PR #19, opened 2024-03-07
CI: add typo check

Young-Flash:
It would be nice to have typos check in CI, `.typos.toml` could be add to allow some intentional typos if needed in the future.


lijunchen (approved):
Thanks!

PR #20, opened 2024-03-07
Try to use `setup-moonbit` action to build on Ubuntu, macOS and Windows

hustcer:
Try to use [`setup-moonbit`](https://github.com/marketplace/actions/setup-moonbit) action to build on Ubuntu, macOS and Windows.
To see an action running detail example, go: https://github.com/hustcer/moonbit-core/actions/runs/8189973593

BTW: 
I'm guessing this PR won't be accepted, but I'll give it a shot 🤣

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -17,17 +17,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: install
-        if: ${{ matrix.os != 'windows-latest' }}
-        run: |
-          curl -fsSL https://cli.moonbitlang.com/install/unix.sh | bash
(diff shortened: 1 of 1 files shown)


hustcer:
I think it's ready now

lijunchen:
Thanks. At this stage, we prefer to write clear scripts directly rather than introducing a black box. If you can help write these CI scripts, we would also appreciate it.

peter-jerry-ye:
Thank you @hustcer for providing this action. 

However `core` is the standard library of the MoonBit and is the place where we make critical changes (for example, adding new backends, changing dependency layouts, etc), thus having an external dependency is not feasible for fast iteration.

With that said, I'm sure that the community can still benefit from it for having a simple way to set up the MoonBit toolchain across different platforms.

PR #29, opened 2024-03-09
feat(string,array): string

Lampese:
## String
- iter
- iteri
- iter_rev
- iteri_rev
- push
- map
- concat
- reserve
- to_list
- to_array
- from_array
- from_list
## Notes
If possible, I strongly recommend using squash merge to merge this pull request to prevent submitting too many useless commits.


array/array.mbt
```diff
@@ -81,6 +81,45 @@ test "iteri" {
   }
 }
 
+pub fn iter_rev[T](self : Array[T], f : (T) -> Unit) {
+  let mut i = self.length() - 1
+  while i >= 0 {
+    f(self[i])
+    i = i - 1
+  }
```
bobzhang, line 89: we can write it in a function style:
```moonbit
for i =  self.length() = 1; i >= 0 ; i = i - 1{
   f(self(i)
}
```

Lampese: I found that the `map_with_index` in the array library uses Functional loop, so I think I should apply this feature to modifiable positions in both the array and string libraries.

Lampese: I have already adopted functional loops for both array and string, and the name of this pull request will be changed to be more accurate.

array/array.mbt
```diff
@@ -25,10 +25,8 @@
 /// [1, 2, 3, 4, 5].iter(fn(x){ print("\(x) ") }) //output: 1 2 3 4 5
 /// ```
 pub fn iter[T](self : Array[T], f : (T) -> Unit) {
-  let mut i = 0
-  while i < self.length() {
-    f(self[i])
-    i = i + 1
+  loop 0 {
+    i => if i < self.length() { f(self[i]); continue i + 1 }
```
bobzhang, line 29: the more idiomatic way:
```
for i = 0 ; I < self.length () ; i = i + 1 {
   f (self[i])
}
```

array/array.mbt
```diff
@@ -58,10 +54,8 @@ test "iter" {
 /// }) //output: (0,1) (1,2) (2,3) (3,4) (4,5)
 /// ```
 pub fn iteri[T](self : Array[T], f : (Int, T) -> Unit) {
-  let mut i = 0
-  while i < self.length() {
-    f(i, self[i])
-    i = i + 1
+  loop 0 {
+    i => if i < self.length() { f(i, self[i]); continue i + 1 }
```
bobzhang, line 58: same here using `for`, note our for is still functional, `i` is immutable

array/array.mbt
```diff
@@ -76,11 +70,44 @@ test "iteri" {
...
     },
   )
-  if failed {
-    return Err("iteri test failed")
+  @assertion.assert_false(failed)?
+}
+
+pub fn iter_rev[T](self : Array[T], f : (T) -> Unit) {
+  loop self.length() - 1 {
+    i => if i > 0 { f(self[i]); continue i - 1 }
```
bobzhang, line 78: same here using for :  `for i = self.length() - 1; i >= 0 ; i = i - 1 {...}` check the boundary to see if `i=0` is applied

array/array.mbt
```diff
@@ -92,10 +119,8 @@ test "iteri" {
 /// ```
 pub fn map[T, U](self : Array[T], f : (T) -> U) -> Array[U] {
   let res = Array::make(self.length(), f(self[0]))
-  let mut i = 0
```
bobzhang, line 95: you need check if the array is empty or not, otherwise this would throw an exception

string/string.mbt
```diff
@@ -11,3 +11,184 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+
+pub fn iter(self : String, f : (Char) -> Unit) {
```
bobzhang, line 15: it would be nice to split contributions to string in a separate package, so that review is a bit easier

bobzhang:
the array part looks good to me, the changes to string need some more work, I would be happy to take the former as is

Lampese:
> the array part looks good to me, the changes to string need some more work, I would be happy to take the former as is

Now the array part is splited to #52, and I will resolve the array modification in this PR when merging branches.

string/string.mbt
```diff
@@ -11,3 +11,190 @@
...
+  for i = 0; i < self.length(); i = i + 1 {
+    f(self[i])
+  }
+}
+
+test "iter" {
```
bobzhang, line 21: we are planning to support multi-platform, so we are a bit cautious adding lots of String/Char functions

Lampese: Okay, this pull request will return to the draft.

string/string.mbt
```diff
@@ -11,3 +11,190 @@
...
+  self + ch.to_string()
+}
+
+test "push" {
+  let str = push("1234", '5')
+  @assertion.assert_eq(str, "12345")?
+}
+
+/// Reserve a string
+pub fn reserve(self : String) -> String {
```
zxch3n, line 132: This might turn a valid UTF-8 string into an invalid one

```
test "reverse" {
  let a = "你好"
  let b = a.reserve()
  let c = b.reserve()
  @assertion.assert_eq(a, c)? // fail `"你好" == "¤Ã½Â Â¥Ã¥Â½Â"`
}
```

PR #33, opened 2024-03-09
Fix error message of assert_ne() to use `!=`

fantix:
Unfortunately, it's not yet easy to write a test to cover this due to the lack of `@string` capabilities.


Yoorkin (approved):
Thanks!

PR #34, opened 2024-03-09
Add assert_is() and assert_is_not()

fantix:
We should probably introduce a `!==` operator ... ?

### DIFF
--- a/assertion/assertion.mbt
+++ b/assertion/assertion.mbt
@@ -95,6 +95,21 @@ test "assert_true.false" {
   assert_ne(assert_true(false), Ok(()))?
 }
 
+/// Assert referential equality of two values.
+///
+/// Returns Ok if the two arguments are the same object by reference, using the
+/// `===` operator; raises an Error otherwise. Certain objects may be equal by
+/// value, but they are different objects in the memory. This function checks
+/// the latter.
+///
+/// # Examples
+///
+/// ```
+/// let a = "4" + "2"
+/// let b = "4" + "2"
+/// assert_is(a, a)?  // this is okay
+/// assert_is(a, b)?  // yields an error
+/// ```
 pub fn assert_is[T : Debug](a : T, b : T) -> Result[Unit, String] {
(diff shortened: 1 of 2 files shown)


assertion/assertion.mbt
```diff
@@ -69,3 +69,51 @@ test "assert_eq.eq" {
...
 }
+
+pub fn assert_is[T : Debug](a : T, b : T) -> Result[Unit, String] {
+  if a === b {
+    Ok(())
+  } else {
+    let a = debug_string(a)
+    let b = debug_string(b)
+    Err("assertion failed for `\(a) === \(b)`")
+  }
```
fantix, line 80: This is preferably `@option.unless(a === b, fn() { ... })`, but crippled by the lack of `test_import` in `moon.pkg.json`, and/or a better handling of cyclic dependencies.

assertion/assertion.mbt
```diff
@@ -69,3 +69,51 @@ test "assert_eq.eq" {
...
+
+test "assert_is.is.not" {
+  let x = 1
+  let s1 = "x:\(x)"
+  let s2 = "x:\(x)"
+  assert_eq(assert_eq(s1, s2), Ok(()))?
+  assert_ne(assert_is(s1, s2), Ok(()))?
+}
+
+pub fn assert_is_not[T : Debug](a : T, b : T) -> Result[Unit, String] {
```
Yoorkin, line 97: Could you add some documentation comments for `assert_is` and `assert_is_not`, since their functionality isn't clear from their names?

fantix: Yeah sure, one sec

fantix: Now added. Thanks!

Yoorkin (approved):
LGTM, thanks for your contribution!

assertion/assertion.mbt
lijunchen: `test_import` will be added as soon as next week

PR #35, opened 2024-03-09
Add `@math.maximum()` and `@math.minimum()`

fantix:
- [x] Added a `@math` package (Rust put min and max in `std::cmp` though)
- [x] The tests in this PR depend on #34
- [x] I'm not sure about the license statements, refs #30

Ported from [muts](https://gitee.com/fantix/muts/blob/v0.5.2/utils/math.mbt).

### DIFF
--- a/math/algebraic.mbt
+++ b/math/algebraic.mbt
@@ -1,0 +1,93 @@
+// Copyright 2024 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// Compares and returns the maximum of two values.
+///
+/// Returns the second argument if the comparison determines them to be equal.
+///
+/// # Examples
+///
+/// ```
+/// print(@math.maximum(1, 2))  // output: 2
+/// print(@math.maximum(2, 2))  // output: 2
+/// ```
+pub fn maximum[T : Compare](x : T, y : T) -> T {
+  if x > y {
+    x
+  } else {
+    y
+  }
+}
+
+test "maximum.value" {
+  @assertion.assert_eq(maximum(1, 2), 2)?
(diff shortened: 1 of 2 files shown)


math/algebraic.mbt
```diff
@@ -0,0 +1,96 @@
...
+///
+/// Returns the second argument if the comparison determines them to be equal.
+///
+/// # Examples
+///
+/// ```
+/// print(@math.maximum(1, 2))  // output: 2
+/// print(@math.maximum(2, 2))  // output: 2
+/// ```
+pub fn maximum[T : Compare](x : T, y : T) -> T {
```
fantix, line 28: The full name `maximum` follows the convention in `list/list.mbt`:

https://github.com/moonbitlang/core/blob/ad0d7c8b12eb3062089a1ccffdb4f3aec462f362/list/list.mbt#L497

However, I personally prefer `max` and `min`, thoughts?

math/algebraic.mbt
```diff
@@ -0,0 +1,96 @@
...
+
+test "minimum.ref" {
+  let v1 = 1
+  let v2 = 2
+  let x1 = "v\(v1)"
+  let x2 = "v\(v2)"
+
+  // We need another value that `== x2` but not `=== x2`
+  let x2t = "v\(v2)"
+  @assertion.assert_is_not(x2, x2t).unwrap()
```
fantix, line 90: Do we have a way to reuse code for testing, without introducing footprints in release builds?

math/algebraic.mbt
```diff
@@ -0,0 +1,96 @@
...
+/// Compares and returns the maximum of two values.
+///
+/// Returns the second argument if the comparison determines them to be equal.
+///
+/// # Examples
+///
+/// ```
+/// print(@math.maximum(1, 2))  // output: 2
+/// print(@math.maximum(2, 2))  // output: 2
+/// ```
```
fantix, line 27: About "code of conduct" for doc strings:

1. I used "Compares" instead of "Compare", while both forms can be found in the code base. I believe it implies "(this function) compares and returns ...".
2. I used "Examples" instead of "Example" because I do have multiple examples. I think we should stick with "Examples" (like Rust) across the repo even if it's just a single example.
3. I used `@math.maximum` instead of just `maximum` so that users could directly copy it from the documentation and run without errors.
4. I'm not sure we like the `print(...)  // output: ...` style in doc strings, it's in compatible with potential future doc tests. Though, `@assertion.assert_eq(@math.maximum(1, 2), 2)` is just too long to read, so I ended up following what the rest of this repo has been doing.

bobzhang: Thanks for the suggestions, about the last one, we will consider generate example based docs directly from `test "example"`

PR #38, opened 2024-03-10
add vector module and test

DawnMagnet:
add vector module and test

support functions
- from_array
- new_default
- push_back
- pop_back
- iter
- iteri
- ...


bobzhang:
we are working on a vector without Default limitation, stay tuned

bobzhang:
@DawnMagnet you can checkout #44 for more details, in that case, Default restriction is no longer needed

PR #44, opened 2024-03-11
add basic setup for vector

Yu-zh:


vec/vec.mbt
```diff
@@ -0,0 +1,81 @@
...
+    self.len = old_cap + 1
+  } else {
+    self.buf[self.len] = value
+    self.len += 1
+  }
+}
+
+/// Retrieves the element at the specified index from the vector.
+pub fn op_get[T](self : Vec[T], index : Int) -> T {
+  self.buf[index]
```
bobzhang, line 63: bounds checking needed

vec/vec.mbt
```diff
@@ -0,0 +1,81 @@
...
+}
+
+/// Sets the value of the element at the specified index.
+pub fn op_set[T](self : Vec[T], index : Int, value : T) {
+  self.buf[index] = value
```
bobzhang, line 68: same here

vec/vec.mbt
Yu-zh: x[i] will panic if i is out of bounds in rust. https://doc.rust-lang.org/std/vec/struct.Vec.html#indexing

PR #45, opened 2024-03-11
fix: tuple with some

pd4d10:
`Some(a, b)` seems not to be a valid syntax with the latest moonbit version.

```sh
moon 0.1.0 (73a0162 2024-03-11)
```


Yu-zh:
Nice catch. Thank you!

PR #52, opened 2024-03-13
refactor: array functions

Lampese:


Lampese:
This PR is the separate array part of the original #29.

PR #53, opened 2024-03-14
add Map[K,V], based on size balanced binary tree.

Yoorkin:
This PR provides some basic operations for map, including `from_array`, `insert`, `remove`, `lookup`, and `member`.

### DIFF
--- a/map/moon.pkg.json
+++ b/map/moon.pkg.json
@@ -1,6 +1,6 @@
 {
     "import": [
-        { "path": "moonbitlang/core/assertion", "alias": "assertion" },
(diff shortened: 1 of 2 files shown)


map/moon.pkg.json
```diff
@@ -0,0 +1,6 @@
+{
```
lijunchen, line 1: can be simplify to

```
{
    "import": [
        "moonbitlang/core/assertion",
        "moonbitlang/core/list"
    ]
}
```

lijunchen (approved):
LGTM

PR #54, opened 2024-03-14
Add balanced binary tree based immutable_set implementation

muqiuhan:
> When creating this PR, I found that the work of immutable_set seems to partially overlap with #53 

This PR implements a set data structure without side effects based on a balanced binary tree, this means the add and find value take time logarithmic in the size of the set.


muqiuhan:
I think this implementation is a bit confusing and may need to be more concise...

muqiuhan:
Now that the tests have passed, we can start to add documentation and tests based on add more functions !

Yoorkin:
Is this PR ready?

muqiuhan:
> Is this PR ready now?

Yes, I've done everything I planned and the local tests are passing.

lijunchen (approved):
LGTM

PR #58, opened 2024-03-14
feat: `ref.incr()` and `ref.decr()`

Yoorkin:
Add two function in package `@moonbitlang/core/ref`,  based on `@moonbitlang/core/num`:

- incr
- decr

## Example

See the test.


lijunchen (approved):
LGTM

PR #61, opened 2024-03-15
feat: add `join` to List

yj-qin:
Add a `join` function to List


Yoorkin:
It's duplicated with [separate_by](https://github.com/moonbitlang/core/blob/6c41d13aff0db30a72c055226116799028d6e911/list/list.mbt#L492)

PR #62, opened 2024-03-15
Add coverage library

lynzrand:
This PR is to facilitate PR 2784 in the compilers repo to add code coverage support


Lampese:
Can you complete the CLA first? Don't forget to log in to it after signing, and cla-check will work fine the next time you commit to rerunning the action.

lynzrand:
Already did, will trigger the pipeline by force pushing some time later. BTW I'm that intern working on coverage the last few days.

coverage/coverage.mbt
```diff
@@ -0,0 +1,100 @@
+pub struct CoverageCounter {
+  priv counter : Array[Int]
+}
+
```
bobzhang, line 4: type defaults to abstract, so you don't need annotate anything here

coverage/coverage.mbt
```diff
@@ -0,0 +1,100 @@
...
+}
+
+fn CoverageCounter::incr(self : CoverageCounter, idx : Int) -> Unit {
+  let counter = self.counter[idx];
+  if counter < 0x7fffffff { // prevent overflow
+    self.counter[idx] = counter + 1
+  }
+}
+
+fn CoverageCounter::to_string(self : CoverageCounter) -> String {
```
bobzhang, line 16: is this for debugging? if so, derive would be enough

coverage/coverage.mbt
```diff
@@ -0,0 +1,100 @@
...
+    abort("error")
+  }
+}
+
+test "incr_counter" {
+  let counter = CoverageCounter::new(10)
+  counter.incr(0)
+  counter.incr(9)
+  let result = counter.to_string()
+  if result != "[1, 0, 0, 0, 0, 0, 0, 0, 0, 1]" {
```
bobzhang, line 44: using assertion package

coverage/coverage.mbt
```diff
@@ -0,0 +1,100 @@
...
+    println(result)
+    abort("error")
+  }
+}
+
+let counters : Ref[List[(String, CoverageCounter)]] = { val: Nil }
+
+fn escape_string(s : String) -> String {
+  let buf = Buffer::make(s.length())
+  let mut ix = 0
```
bobzhang, line 54: where is this used, this could be moved into String package?

coverage/coverage.mbt
```diff
@@ -0,0 +1,100 @@
...
+      }
+    }
+    ix += 1
+  }
+  buf.to_string()
+}
+
+pub fn track(name : String, counter : CoverageCounter) -> Unit {
+  counters.val = Cons((name, counter), counters.val)
+}
```
bobzhang, line 77: CoverageCounter.track -- export as method?

coverage/coverage.mbt
lynzrand: It's for exporting the coverage data in `@coverage.end()`, so it must be in JSON format. I don't know if deriving has such guarantees.

coverage/coverage.mbt
lynzrand: It's for escaping the coverage data's key (filename). I'll do an overhaul and put it into Strings

coverage/coverage.mbt
lynzrand: Update: converted it to a custom implementation of `Debug` and derived `Show`

string/string.mbt
```diff
@@ -11,3 +11,69 @@
...
+      _ =>
+        if ch.to_int() < 0x20 {
+          buf.write_char('\\')
+          buf.write_char('x')
+          let ich = ch.to_int()
+          buf.write_char(to_hex_digit(ich / 16))
+          buf.write_char(to_hex_digit(ich % 16))
+        } else {
+          buf.write_char(ch)
+        }
```
lynzrand, line 52: Do we need to factor this part out to `char` as `Char::escape_to`? Also, do we need to additionally provide escaping for non-ascii chars?

coverage/coverage.mbt
```diff
@@ -0,0 +1,92 @@
...
+/// }
+/// ----- END MOONBIT COVERAGE -----
+/// ```
+pub fn end() -> Unit {
+  println("----- BEGIN MOONBIT COVERAGE -----")
+  println("{")
+  let mut ix = 0
+  loop counters.val {
+    Cons((name, counter), xs) => {
+      if ix != 0 {
```
bobzhang, line 77: rename this to a separate function and write a simple test?

bobzhang (commented):
after second thoughts, the added String::escape is only useful for coverage, we can keep it internal to such package.
The PR is good to go modulo my last two comments

coverage/coverage.mbt
lynzrand: Good idea, will do

bobzhang: note we can use a trait object so that it can be output to a string or console

lynzrand:
> after second thoughts, the added String::escape is only useful for coverage, we can keep it internal to such package. The PR is good to go modulo my last two comments

Maybe it's useful for debug outputting strings?

coverage/coverage.mbt
lynzrand: > a trait object

AFAIK we don't have such infrastructure (like `trait Write`) yet, should I add one in another PR?

bobzhang: I added a new commit, check the TODO in the comment

bobzhang (approved):

PR #67, opened 2024-03-15
Refactor map functions to use loop instead of match

bobzhang:


map/map.mbt
```diff
@@ -189,28 +201,30 @@ fn balance[K : Compare, V](
...
+    //      1                 2
+    //     / \              /   \
+    //    x   3            1     3
+    //       / \    -->   / \   / \
+    //      2   z        x  y1 y2  z
+    //     / \
+    //    y1 y2
   }
 
-  //      1                 2
```
bobzhang, line 194: here is a format bug?

map/map.mbt
```diff
@@ -189,28 +201,30 @@ fn balance[K : Compare, V](
...
-  //       / \    -->   / \   / \
-  //      2   z        x  y1 y2  z
-  //     / \
-  //    y1 y2
   fn double_l {
     k1, v1, x, Map::Tree(k3, v3, _, Tree(k2, v2, _, y1, y2), z) =>
       new(k2, v2, new(k1, v1, x, y1), new(k3, v3, y2, z))
     _, _, _, _ => abort("double_l error")
+
+    //      3                 2
```
bobzhang, line 219: same

map/map.mbt
Yoorkin: It's a known bug in the formatter. Comments cannot be placed between local functions. It will be fixed soon.

PR #69, opened 2024-03-16
feat: add deque

Lampese:


deque/deque.mbt
```diff
@@ -0,0 +1,472 @@
+// Copyright 2024 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
```
bobzhang, line 5: can you add a README.md in this dir to help people get familiar with what's deque and its use case?

Lampese: Allright.

bobzhang (approved):

PR #79, opened 2024-03-18
feat: fold and iters for array/vector

Lampese:


bobzhang:
I resolved conflicts in #80, one minor change I made is adding a label to init, so user can be more flexible with the application order, the type inference is also more influence

PR #90, opened 2024-03-20
Add hash set (mutable set) implementation based on robin hood hash

muqiuhan:
The redundant immutable_set related commits were caused by my rebase in #54


waterlens:
How about implementing it with open addressing? Like robin_hood or swisstable things.

muqiuhan:
> How about implementing it with open addressing? Like robin_hood or swisstable things.

Thank you for your suggestion! I have rethought this problem. Although the current implementation is simpler and easier to maintain, as a standard library, a MutableSet with better performance is needed, so I will use robin_hood or swisstable to reimplement MutableSet in subsequent commits.

muqiuhan:
While trying to implement it, I noticed that the MoonBit standard library seems to be able to add some modules:

- math/random
- bit operations

muqiuhan:
Now, I reimplemented hashset (or mutable_set) using robin hood hash in [4547880](https://github.com/moonbitlang/core/pull/90/commits/45478801d3be911ddbf4275ff7db1fba21de5ff6).

Robin Hood hashing is a technique for implementing hash tables. It is based on open addressing with a simple but clever twist: As new keys are inserted, old keys are shifted around in a way such that all keys stay reasonably close to the slot they originally hash to.

The original implementation of closed addressing based on buckets was [e4826a8](https://github.com/moonbitlang/core/pull/90/commits/e4826a87c8f710c0dc09228c098343be42251b1f)

bobzhang:
@muqiuhan is this good for review? can you do a rebase?

muqiuhan:
> can you do a rebase?

image

It seems that it is not possible to directly update with rebase. This pr does not modify the files of other modules. Why does rebase conflict?

muqiuhan:
> @muqiuhan is this good for review? can you do a rebase?

Yes, I've merged it into one commit.

hash_set/hash_set.mbt
```diff
@@ -0,0 +1,562 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// This module implements mutable set as hashtable.
+
+/// MutableSet is an open-addressing hashset based on robin hood hash.
+pub struct MutableSet[T] {
+  priv mut directory : @vec.Vec[Option[Item[T]]]
```
peter-jerry-ye, line 19: Also it may not be a good idea to use `Vec` here. The `Vec` automatically grows, but you will need to reallocate and rehash anyway, so using `Vec` will cause an extra boxing.

hash_set/hash_set.mbt
```diff
@@ -0,0 +1,562 @@
...
+  priv mut real_size : Int
+  priv mut effective_size : Int
+}
+
+priv struct Item[T] {
+  mut key : T
+  mut probe_length : Int
+} derive(Default, Debug, Show)
```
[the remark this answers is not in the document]
muqiuhan: Sorry, this was an issue left in development and I didn't remember to delete it.

hash_set/hash_set.mbt
muqiuhan: Should I use Array directly to implement it (and then implement expansion-related operations myself)? The performance of List in terms of indexing is not very good.

hash_set/hash_set.mbt
```diff
@@ -0,0 +1,562 @@
...
+
+fn insert_item[T : Hash + Eq + Default](self : MutableSet[T], key : T) -> Int {
+  match self.index_of(key) {
+    Some(index) => {
+      self.directory[index].unwrap().key = key
+      index
+    }
+    None => {
+      if self.effective_size == self.directory.length() {
+        let set = MutableSet::with_capacity(self.real_size * 2)
```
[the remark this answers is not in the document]
muqiuhan: Referring to #93, this is not necessarily ideal. Maybe I should use 0.9 ~ 0.99. Might need some testing to determine it.

hash_set/hash_set.mbt
muqiuhan: Yes! the expansion operation of directory will only appear in with_capacity. It is completely reasonable to use Array. I have changed it to Array (fe3b263).

peter-jerry-ye:
I think the strategy for determining the load factor of #93 looks good to me and maybe you can also adapt that

muqiuhan:
I'm sorry that some natural factors have delayed this small modification for so long, I have modified the load factor to about 0.8 !

mutable_set/mutable_set.mbt
```diff
@@ -0,0 +1,565 @@
...
+        }
+      _ => break None
+    }
+  } else {
+    None
+  }
+}
+
+/// Return true if key exists in the MutableSet.
+pub fn exists[T : Hash + Eq](self : MutableSet[T], key : T) -> Bool {
```
peter-jerry-ye, line 211:
```suggestion
pub fn exist[T : Hash + Eq](self : MutableSet[T], key : T) -> Bool {
```

mutable_set/mutable_set.mbt
```diff
@@ -0,0 +1,565 @@
...
+pub fn size[T](self : MutableSet[T]) -> Int {
+  self.real_size
+}
+
+/// Returns the intersection of self with other.
+/// 
+/// # Example
+/// 
+/// ```
+/// println(MutableSet::[3, 4, 5].inter(MutableSet::[4, 5, 6]).to_array())
```
peter-jerry-ye, line 281:
```suggestion
/// println(MutableSet::[3, 4, 5].intersection(MutableSet::[4, 5, 6]).to_array())
```

mutable_set/mutable_set.mbt
```diff
@@ -0,0 +1,565 @@
...
+
+/// Returns the difference between self and other.
+/// 
+/// # Example
+/// 
+/// ```
+/// println(MutableSet::[1, 2, 3].diff(MutableSet::[4, 5, 1]).to_array())
+/// // output: [3, 2]
+/// ```
+pub fn diff[T : Hash + Eq](
```
peter-jerry-ye, line 355:
```suggestion
pub fn difference[T : Hash + Eq](
```

mutable_set/mutable_set.mbt
```diff
@@ -0,0 +1,565 @@
...
+}
+
+fn insert_item[T : Hash + Eq](self : MutableSet[T], key : T) -> Int {
+  match self.index_of(key) {
+    Some(index) => {
+      self.directory[index].unwrap().key = key
+      index
+    }
+    None => {
+      if self.effective_size == self.directory.length() {
```
peter-jerry-ye, line 394: The load factor is not **how** you resize the hashset; it's about **when** you resize the hashset

For example, if I understand correctly (correct me if I'm wrong), you decide to resize it when `self.effective_size == self.directory.length()`, then it means the load (number of items/length of array) will become 100% before you resize it. It may have some poor performances (when it is 90% full or 95% full or 99% full)

peter-jerry-ye (changes requested):
No worry. It is us that should thank you for your contribution.

With that said, I think you may have misunderstood the meaning of load factor.

mutable_set/mutable_set.mbt
muqiuhan: Oh! I was wrong all along, that is, when the load factor is 0.8, the HashSet should be expanded when it reaches 80% capacity, not when it is full.

muqiuhan: So is it possible to simply modify it to:
```
if self.directory.length() * 80 / 100 == self.effective_size {
  ...
```

mutable_set/mutable_set.mbt
```diff
@@ -0,0 +1,571 @@
...
+      self.effective_size = self.effective_size + 1
+      index
+    }
+  }
+}
+
+fn index_of[T : Hash + Eq](self : MutableSet[T], key : T) -> Option[Int] {
+  let hash = key.hash().abs()
+  let num_slots = self.directory.length()
+  let slot = hash % num_slots
```
peter-jerry-ye, line 443: hash may be negative
```moonbit
test "insert" {
  let set = MutableSet::[-1, 1, 2, 2, 3, 3]
  set.insert(4)
  @assertion.assert_eq(set.to_array(), [4, 3, 2, 1])?
}
```

muqiuhan: This error can be fixed using the same strategy as #93:
```
key.hash().abs().land(self.directory.length() - 1)
```

Although the `hash().abs()` calculation results of -1 and 1 are the same, they cannot be inserted at the same time, so the difference in `directory.length()` can avoid the same index.

mutable_set/mutable_set.mbt
```diff
@@ -0,0 +1,568 @@
...
+    index = index + 1
+  }
+  self.effective_size = 0
+  self.real_size = 0
+}
+
+/// Returns a copy of MutableSet.
+pub fn copy[T](self : MutableSet[T]) -> MutableSet[T] {
+  {
+    directory: self.directory,
```
peter-jerry-ye, line 119:
```moonbit
test "union" {
  let a = MutableSet::[3, 4, 5]
  let b = MutableSet::[3, 4, 5]
  let c = a.union(b)
  a.clear()
  @assertion.assert_true(c.exist(3))?
}
```

mutable_set/mutable_set.mbt
```diff
@@ -0,0 +1,568 @@
...
+
+priv struct Item[T] {
+  mut key : T
+  mut probe_length : Int
+} derive(Debug, Show)
+
+/// Initialize MutableSet by specifying initial capacity.
+pub fn MutableSet::with_capacity[T](initial_capacity : Int) -> MutableSet[T] {
+  {
+    directory: if initial_capacity <= 0 || initial_capacity > 2147483647 {
```
peter-jerry-ye, line 30: is the check `initial_capacity > 2147483647` really necessary?

mutable_set/mutable_set.mbt
peter-jerry-ye:
```moonbit
test "" {
  let a = MutableSet::[0]
  a.insert(1)
}
```

mutable_set/mutable_set.mbt
muqiuhan: In MoonBit, literals that are too long cannot be compiled. In addition, when it is larger than this value, it will overflow to a value < 0, so this is unnecessary

mutable_set/mutable_set.mbt
muqiuhan: I recklessly thought that MoonBit would copy direcory here. After testing, the basic type will perform copying, so I need to copy the directory manually.

muqiuhan:
I have learned a lot here, but I cannot waste too much of your time for my own selfish interests. I am not a good Contributor. In order to save your time, I request to close this PR.

mutable_set/mutable_set.mbt
peter-jerry-ye: We may need `Array::copy` @Yu-zh

peter-jerry-ye:
Feel free to reopen the PR when you feel confident. There are no 'good' or 'bad' contributor. I think you are very close to finish the implementation.

Maybe you can check out the references that inspired the hashmap implementation, i.e.:
- <https://programming.guide/robin-hood-hashing.html>
- <https://cs.uwaterloo.ca/research/tr/1986/CS-86-14.pdf>

There's also a video about hashmap implementation here: <https://www.bilibili.com/video/BV1g64y1P7nq> though it doesn't use robin hood hashing.

PR #91, opened 2024-03-20
feat: new functions to array and comment completion

Lampese:
Basically synchronizes the functions that Array and Vec can share.

### DIFF
--- a/array/array.mbt
+++ b/array/array.mbt
@@ -78,6 +78,18 @@ test "iteri" {
   @assertion.assert_false(failed)?
 }
 
+/// Iterates over each element in reversed turn.
+///
+/// # Arguments
+///
+/// - `self`: The array to iterate over.
+/// - `f`: The function to apply to each element.
+///
+/// # Example
+///
+/// ```
+/// [1, 2, 3, 4, 5].iter(fn(x){ print("\(x) ") }) //output: 5 4 3 2 1
+/// ```
 pub fn iter_rev[T](self : Array[T], f : (T) -> Unit) -> Unit {
   for i = self.length() - 1; i >= 0; i = i - 1 {
     f(self[i])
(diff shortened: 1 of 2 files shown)


array/array.mbt
```diff
@@ -250,7 +297,13 @@ test "fold_rev" {
   @assertion.assert_eq(sum, 15)?
 }
 
-/// Reverse the array in place.
+/// Fold out values from an array according to certain rules in reversed turn.
+/// 
+/// # Example
+/// ```
+/// let arr = [1, 2, 3, 4, 5]
+/// arr // [5, 4, 3, 2, 1]
```
Yu-zh, line 305: should be `reverse(arr)` here?

array/array.mbt
```diff
@@ -289,3 +342,175 @@ pub fn swap[T](self : Array[T], i : Int, j : Int) -> Unit {
...
+/// # Example 
+/// 
+/// ```
+/// let arr = [1, 2, 3, 4, 5]
+/// arr.all(fn(ele) { ele < 6 }) // true
+/// arr.all(fn(ele) { ele < 5 }) // false
+/// ```
+pub fn all[T](self : Array[T], f : (T) -> Bool) -> Bool {
+  for i = 0; i < self.length(); i = i + 1 {
+    if f(self[i]) == false {
```
Yu-zh, line 363: You can use `not` function. See the [doc](https://www.moonbitlang.com/docs/syntax#boolean) for reference

array/array.mbt
```diff
@@ -289,3 +342,175 @@ pub fn swap[T](self : Array[T], i : Int, j : Int) -> Unit {
...
+/// # Example 
+/// 
+/// ```
+/// let arr = [1, 2, 3, 4, 5]
+/// arr.any(fn(ele) { ele < 6 }) // true
+/// arr.any(fn(ele) { ele < 5 }) // true
+/// ```
+pub fn any[T](self : Array[T], f : (T) -> Bool) -> Bool {
+  for i = 0; i < self.length(); i = i + 1 {
+    if f(self[i]) == true {
```
Yu-zh, line 387: `if f(self[i]) { .. }`

PR #102, opened 2024-03-23
fix: readme for deque

Lampese:


deque/README.md
```diff
@@ -11,7 +11,7 @@ Deque is a double-ended queue implemented as a round-robin queue, supporting O(1
 You can create a deque manually via the `new()` or construct it using the `from_array()`.
 
    ```moonbit
-   let dq = Deque::new()
+   let dq : Deque[Int] = Deque::new()
    let dq = Deque::[1, 2, 3, 4, 5]
    ```
```
CAIMEOX, line 16: unexpected indent?

PR #106, opened 2024-03-24
tweak priority queue

bobzhang:
- tweak from_array
- split types into a separate file and export less
- tweak


priority_queue/priority_queue.mbt
```diff
@@ -66,11 +51,15 @@ pub fn PriorityQueue::from_array[T : Compare](
   arr : Array[T],
   ~min : Bool = false
 ) -> PriorityQueue[T] {
-  for i = 0, acc = (Nil : Node[T]); i < arr.length(); {
-    continue i + 1,
-      meld(acc, Cons({ content: arr[i], sibling: Nil, child: Nil }), min)
-  } else {
-    { min, len: arr.length(), top: acc }
+  // CR: bad formatting  
+  let len = arr.length()
```
bobzhang, line 55: @Yoorkin the formatter shows `; ;` here

priority_queue/priority_queue.mbt
```diff
@@ -84,11 +73,12 @@ fn meld[T : Compare](x : Node[T], y : Node[T], min : Bool) -> Node[T] {
...
     (_, Nil) => x
     (Cons(x_top), Cons(y_top)) =>
-      if if min {
+      // CR: bad formatting
+      if (if min {
         x_top.content < y_top.content
       } else {
         x_top.content > y_top.content
-      } {
+      }) {
```
bobzhang, line 81: @Yoorkin for the formatter a paren is needed here to make it less confusing

priority_queue/types.mbt
```diff
@@ -0,0 +1,26 @@
...
+}
+
+priv enum Node[T] {
+  Nil
+  Cons(Cont[T])
+}
+
+struct PriorityQueue[T] {
+  min : Bool
+  mut len : Int
```
bobzhang, line 24: @Lampese for the API design, I feel `min` is not needed here, it slows down the algorithm a little bit and user can always create a new type with different comparison behavior

Lampese: Allright, but my initial consideration was that it might not be user-friendly to define Num for structures that already have a Compare relationship. But it doesn't really hurt to provide only one heap though, as many implementations do.

Lampese: Do I need to re-initiate a PR after this one is over to resolve this issue?

Lampese (approved):

priority_queue/priority_queue.mbt
bobzhang: fixed in upstream

PR #114, opened 2024-03-25
Tweak the hash map

yj-qin:
- Caching the hash value in the entry to speeds up the lookup operation for non-integer keys.
- Tweak the hash calculation to let the higher 16 bits of the hash value participate in the calculation.

### DIFF
--- a/hashmap/hashmap.mbt
+++ b/hashmap/hashmap.mbt
@@ -51,6 +51,9 @@ pub fn HashMap::new[K, V](
   ~capacity : Int = default_init_capacity,
   ~hasher : Option[(K) -> Int] = None
 ) -> HashMap[K, V] {
+  if capacity != 0 && not(is_power_of_two(capacity)) {
+    abort("The capacity is not a zero or a power of 2.")
+  }
   {
     size: 0,
     capacity,
@@ -58,6 +61,10 @@ pub fn HashMap::new[K, V](
     hasher,
     entries: @array.new(capacity, fn() { Empty }),
   }
+}
+
+fn is_power_of_two(n : Int) -> Bool {
+  n > 0 && n.land(n - 1) == 0
 }
 
 /// Create new hash map from array.
--- a/hashmap/hashmap.mbt
+++ b/hashmap/hashmap.mbt
@@ -18,9 +18,36 @@ let default_init_capacity = 8
 enum Entry[K, V] {
   Empty
   // (Probe Sequence Length, Hash, Key, Value)
-  Valid((Int, Int, K, V))
+  Valid(Int, Int, K, V)
 } derive(Debug)
 
+fn psl[K, V](self : Entry[K, V]) -> Option[Int] {
+  match self {
+    Valid(psl, _, _, _) => Some(psl)
+    Empty => None
+  }
+}
+
+fn hash[K, V](self : Entry[K, V]) -> Option[Int] {
+  match self {
+    Valid(_, hash, _, _) => Some(hash)
+    Empty => None
+  }
+}
+
+fn key[K, V](self : Entry[K, V]) -> Option[K] {
+  match self {
+    Valid(_, _, k, _) => Some(k)
+    Empty => None
+  }
+}
+
+fn value[K, V](self : Entry[K, V]) -> Option[V] {
+  match self {
+    Valid(_, _, _, v) => Some(v)
+    Empty => None
+  }
+}
 /// A mutable hash map implements with Robin Hood hashing.
 /// 
 /// reference:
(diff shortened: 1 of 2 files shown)


hashmap/hashmap.mbt
```diff
@@ -17,7 +17,8 @@ let default_init_capacity = 8
 
 enum Entry[K, V] {
   Empty
-  Valid((Int, K, V))
+  // (Probe Sequence Length, Hash, Key, Value)
+  Valid((Int, Int, K, V))
 } derive(Debug)
```
bobzhang, line 22: use `Valid(Int,Int,K,V)`? it is more memory compact

hashmap/hashmap.mbt
```diff
@@ -73,30 +101,36 @@ pub fn set[K : Hash + Eq, V](self : HashMap[K, V], key : K, value : V) -> Unit {
...
-        self.entries[idx] = Valid(entry)
+        self.entries[idx] = entry
         self.size += 1
         break
       }
-      Valid((d, k, v)) => {
-        if k == entry.1 {
-          self.entries[idx] = Valid((d, k, value))
+      Valid(d, h, k, v) => {
+        if h == entry.hash().unwrap() && k == entry.key().unwrap() {
```
peter-jerry-ye, line 116: Don't you already have the hash and key at line 106?

hashmap/hashmap.mbt
```diff
@@ -73,30 +101,36 @@ pub fn set[K : Hash + Eq, V](self : HashMap[K, V], key : K, value : V) -> Unit {
   if self.capacity == 0 || self.size >= self.growAt {
     self.grow()
   }
-  let mut idx = self.hash(key)
-  let mut entry = (0, key, value) // (distance, key, value)
+  let hash = self.make_hash(key)
+  let mut idx = self.index(hash)
```
peter-jerry-ye, line 105:
```suggestion
  loop 0, self.index(hash), 0, hash, key, value {
    i, idx, psl, hash, key, value => {
      if i == self.capacity {
        abort("HashMap is full")
      }
      match self.entries[idx] {
        Empty => {
          self.entries[idx] = Entry::Valid(psl, hash, key, value)
          self.size += 1
          break
        }
        Valid(d, h, k, v) => {
          if h == hash && k == key {
            self.entries[idx] = Valid(d, h, k, value)
            break
          }
          if psl > d {
            self.entries[idx] = Entry::Valid(psl, hash, key, value)
            continue i + 1, self.next_index(idx), d + 1, h, k, v
          }
          continue i + 1, self.next_index(idx), psl + 1, hash, key, value
        }
      }
    }
  }
```

Do you think it is better to write with functional loop?

peter-jerry-ye:
I think you should check the capacity passed in instead of setting it directly. I feel that much of your code assumes that the capacity is 2^n, which does not always hold.

```moonbit
test "remove_unexist_key" {
  let m : HashMap[String, Int] = HashMap::new(
    ~hasher=Some(fn(k) { k.length() }),
    ~capacity=7
  )
  m.set("a", 1)
  m.set("ab", 2)
  m.set("abc", 3)
  m.remove("d")
  @assertion.assert_eq(m.size(), 3)?
  @assertion.assert_eq(
    m.debug_entries(),
    "_,(0,a,1),(0,ab,2),(0,abc,3),_,_,_,_",
  )?
}
```

hashmap/hashmap.mbt
yj-qin: Yes, it is more concise and less error-prone. I will modify other loops.

hashmap/hashmap.mbt
```diff
@@ -114,12 +148,12 @@ pub fn op_set[K : Hash + Eq, V](
 
 /// Get the value associated with a key.
 pub fn get[K : Hash + Eq, V](self : HashMap[K, V], key : K) -> Option[V] {
-  let mut idx = self.hash(key)
-  let mut distance = 0
-  for i = 0; i < self.capacity; i = i + 1 {
+  let hash = self.make_hash(key)
+  let mut idx = self.index(hash)
+  for distance = 0; distance < self.capacity; distance = distance + 1 {
```
peter-jerry-ye, line 153:
```suggestion
  for distance = 0, idx = self.index(hash); distance < self.capacity; distance = distance + 1, idx = self.next_index(idx) {
```

hashmap/hashmap.mbt
```diff
@@ -50,6 +51,9 @@ pub fn HashMap::new[K, V](
   ~capacity : Int = default_init_capacity,
   ~hasher : Option[(K) -> Int] = None
 ) -> HashMap[K, V] {
+  if capacity != 0 && not(is_power_of_two(capacity)) {
+    abort("The capacity is not a zero or a power of 2.")
```
peter-jerry-ye, line 55: I think it's too much to abort directly. You can either
- do not expose capacity at all
- resize the capacity to a power of 2 that is larger than the input (which may have problem with input such as 2^31 - 1)

yj-qin: Due to the lack of unsigned type and some util functions (like Rust 's `checked_next_power_of_two`). Maybe not exposing the capacity is better.

peter-jerry-ye:
LGTM, thanks for the PR

Issue #124, opened 2024-03-26
Queue::from_array looks buggy


bobzhang:
by looking at the code queue.first = queue.last, it seems wrong

Lampese:
@CAIMEOX

PR #125, opened 2024-03-26
show the bug of Queue::from_array

bobzhang:
#124


bobzhang:
@bzy-debug I found another issue, there seems no LSP support for this file

bzy-debug:
`*_test` files are handled differently by build system, haven't supported it on LSP.

Will support this in this week

PR #135, opened 2024-03-27
perf: improve accuracy and add tests

Lampese:


bobzhang:
what are major changes to this PR?
are you aware that we support hex notation for float so it is more precise and reliable, e.g,
```mbt
println(0x1.91eb851eb851fp+1) // 3.14
```

Lampese:
> what are major changes to this PR? are you aware that we support hex notation for float so it is more precise and reliable, e.g,
> 
> ```
> println(0x1.91eb851eb851fp+1) // 3.14
> ```

Okey, Let me explain.
- Since the number of bits in Double is currently tentative, I consider the behavior of the trunc angle to be unsafe and remove it for now.
- I found after adding a new set of large angle tests that I couldn't get within 1e-12 on javascript backends, so I extended the error range to 1e-11.
- Regarding the hex form of storing decimals, I've switched to it and got a tiny improvement in precision, thanks!

bobzhang:
> Since the number of bits in Double is currently tentative, I consider the behavior of the trunc angle to be unsafe and remove it for now.

Can you elaborate a little bit? Note Double is IEEE-754 on all platforms.

> I found after adding a new set of large angle tests that I couldn't get within 1e-12 on javascript backends, so I extended the error range to 1e-11.

Note the JS backend may not be heavily tested, if you can narrow down a little bit, we can have a look cc @hackwaly

PR #138, opened 2024-03-28
Add coverage support for core libraries

lynzrand:
This PR allows for coverage testing for all core libraries, except for the coverage library itself.

This PR removes external dependencies for the coverage library, in order to add it as a dependency of all other dependencies.

Remaining questions:

- Can we insert the dependency to `moonbitlang/core/coverage` from the build system iff coverage is enabled, instead of right here?
- Add coverage output into CI in this PR?

Blocking issues:

- [x] Bug in builtin library that makes string escaping fail in output
- [x] `moon test` does not return error when the test runner fails

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,7 @@
 target/
 .mooncakes/
+
+# Coverage-related files
+moonbit-coverage-*.txt
+_coverage
+bisect.coverage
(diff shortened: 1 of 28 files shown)


bobzhang:
as discussed offline, this is a temporary solution, the `coverage` dependency will be implicit in the future

PR #139, opened 2024-03-28
test core on mac bleeding, allow fail

bobzhang:
@bzy-debug can we make bleeding more forgiving, allow it to fail?


bzy-debug:
`continue-on-error: true` can allow a job to fail

PR #150, opened 2024-03-29
double: add constants and utils

yj-qin:


double/double.mbt
```diff
@@ -12,6 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+let not_a_number : Double = 0x7FF8000000000001L.reinterpret_as_double()
```
bobzhang, line 15: nan?

yj-qin: When the exponent bits are all 1, it is a NaN, and I just reference Go's NaN constants value, so it is 0x7FF8000000000001. 
Should we set up a convention like qNaN and sNaN?
https://stackoverflow.com/questions/18118408/what-is-the-difference-between-quiet-nan-and-signaling-nan

PR #155, opened 2024-03-29
Add coverage test support for core lib

lynzrand:


lynzrand:
For some reason the CI's github token is readonly (probably because this PR comes from a fork?), so it cannot add comments. Otherwise it's usable now.

Issue #128, opened 2024-03-27
Proposal: String conversions package


yj-qin:
Currently, we cannot convert from String to basic types. Maybe we need a package like Go's [strconv](https://pkg.go.dev/strconv)

- [x] String to Int
- [x] String to Int64
- [x] String to Bool
- [x] String to Double

PR #161, opened 2024-03-29
strconv: add Double parser

yj-qin:
This PR implements a string to double conversion algorithm, described in this [post](https://nigeltao.github.io/blog/2020/parse-number-f64-simple.html). It is a slower but comprehensive fallback for the faster Eisel-Lemire algorithm.
#128 

Progress:
- [x] Implements high-precision decimal ds
- [x] Implements the conversion algorithm
- [x] Add more test


strconv/decimal.mbt
```diff
@@ -0,0 +1,625 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// High Precision Decimal structure for "Simple Decimal Conversion" algorithm.
+/// Developed by Ken. Thompson, Russ Cox, Robert Griesemer, Nigel Tao.
+/// 
+/// reference:
+/// - <https://nigeltao.github.io/blog/2020/parse-number-f64-simple.html>
+/// - <https://nigeltao.github.io/blog/2020/eisel-lemire.html>
```
bobzhang, line 20: Could decimal itself be extracted as a useful package?

yj-qin: I think it's feasible to move `Decimal` to a new package and add more functionalities to it, much like Java's BigDecimal.

yj-qin: I found a discussion about the `Decimal` package at [here](https://github.com/golang/go/issues/12127), maybe useful.

yj-qin: @bobzhang IMO, It would be better to keep this `Decimal` within strconv. We can reimplement a more mature `Decimal` that complies with the [decimal arithmetic specification](https://speleotrove.com/decimal/decarith.html)

PR #168, opened 2024-03-30
feat: add sort, sort_by, sort_by_key to Vec

zxch3n:
This PR implements the sort, sort_by, sort_by_key methods on Vec as Array does.

It also adopts the suggestions from https://github.com/moonbitlang/core/pull/72#discussion_r1537002612 to use `T:Compare` directly to enhance the performance by 20%.

### DIFF
--- a/array/sort_by.mbt
+++ b/array/sort_by.mbt
@@ -12,27 +12,49 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-/// Sorts the array
+/// Sorts the array with a key extraction function.
 /// 
 /// It's an in-place, unstable sort(it will reorder equal elements). The time complexity is O(n log n) in the worst case.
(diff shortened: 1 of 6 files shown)


array/sort_by.mbt
```diff
@@ -0,0 +1,298 @@
...
+fn sift_down_by[T](
+  arr : ArraySlice[T],
+  index : Int,
+  cmp : (T, T) -> Int
+) -> Unit {
+  let mut index = index
+  let len = arr.length()
+  let mut child = index * 2 + 1
+  while child < len {
+    if child + 1 < len && cmp(arr[child], arr[child + 1]) < 0 {
```
bobzhang, line 256: note this function could be refactored using functional loop without mutation

bobzhang (approved):

PR #170, opened 2024-03-30
relax  permissions for coverage

bobzhang:


bobzhang:
@lynzrand I relaxed permissions, it seems to work. We can improve the user experience and security later

Issue #169, opened 2024-03-30
Coverage test action cannot add result comment to the corresponding PR


lynzrand:
Permission settings caused the coverage test action added in #155 being unable to comment the result in the corresponding PR. The token issued from the action is readonly, although the action requested write permissions to PRs.

Possibly related documentation: <https://docs.github.com/en/actions/security-guides/automatic-token-authentication#how-the-permissions-are-calculated-for-a-workflow-job>

> Finally, if the workflow was triggered by a pull request from a forked repository, **and the Send write tokens to workflows from pull requests setting is not selected**, the permissions are adjusted to change any write permissions to read only.

Unfortunately, I can't find the highlighted setting in the settings page when looking for it in my repository, so I have no idea how to turn it on.

cc @bobzhang

lynzrand:
Fixed by #170

PR #172, opened 2024-03-30
Comment out coverage-check-bleeding workflow

bobzhang:
enable it after we fix the permission issues cc @lynzrand


Lampese:
#169

PR #187, opened 2024-04-02
strconv: add `FromStr` trait

yj-qin:
## Summary by CodeRabbit

- **New Features**
    - Introduced `from_string` functions for `Bool`, `Double`, `Int`, and `Int64` modules, enabling string to type conversion.
    - Added explicit declarations for `not_a_number`, `positive_infinity`, and `negative_infinity` in the `Double` module, enhancing the handling of special numerical values.
    - Implemented a `FromStr` trait in the `String` module, facilitating parsing of strings into different types.
    - Added a generic `parse` function in the `traits.mbt` file for parsing different types from strings.

- **Refactor**
    - Updated import statements in package files to support new functionalities and removed unnecessary imports.


yj-qin:
Seems like the bool package cannot import the assertion package now

yj-qin:
Seems like the `Boolean` trait needs to be moved to a new package; otherwise, the `bool` package won't be able to import the package that depends on the `assertion` package.

peter-jerry-ye:
Add this in `strconv` only:

```moonbit
pub trait FromStr {
  from_string(String) -> Result[Self, String]
}

pub fn FromStr::from_string(str: String) -> Result[Bool, String] {
  Ok(true)
}

pub fn FromStr::from_string(str: String) -> Result[Int, String] {
  Ok(0)
}

pub fn from_string[A : FromStr](str: String) -> Result[A, String] {
  A::from_string(str)
}
```

And use it outside `strconv` as:

`(@strconv.from_string("bool") : Result[Bool, String])`

PR #193, opened 2024-04-03
fix escape of string in debug_write

Guest0x0:
`debug_write` of `String` will wrongly escape single quote `'`, which is unnecessary, and will make the result invalid in json/MoonBit syntax. This PR fixes this problem

## Summary by CodeRabbit

- **Refactor**
	- Improved character escaping in debug messages for better handling of special characters.
- **Tests**
	- Added tests to ensure special characters are correctly handled in debug messages.


lijunchen (approved):
LGTM

PR #205, opened 2024-04-07
Try fixing coverage CI's API error 

lynzrand:
Uses the two-workflow workaround for coverage issue.

Related: #207 (but not closing just yet!)


lynzrand:
@bobzhang we might have to merge this before checking if it works

bobzhang (approved):

PR #210, opened 2024-04-07
vec: `reserve_capacity` and `shrink_to_fit`

yj-qin:


peter-jerry-ye:
Shrink to fit looks fine to me. The reserve capacity, however, usually means *reserving* the given capacity, instead of *growing* the given capacity, as in [Java](https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html#ensureCapacity-int-) or [C++](https://en.cppreference.com/w/cpp/container/vector/reserve)

yj-qin:
> Shrink to fit looks fine to me. The reserve capacity, however, usually means _reserving_ the given capacity, instead of _growing_ the given capacity, as in [Java](https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html#ensureCapacity-int-) or [C++](https://en.cppreference.com/w/cpp/container/vector/reserve)

I misunderstood, will fix it later.

PR #215, opened 2024-04-08
fix physic_eq related tests

hackwaly:
The semantic of physic eq is not cross platform consistent. 

This PR address this issue: use struct with mutable field to force consistent semantic across platforms.

### DIFF
--- a/assertion/assertion.mbt
+++ b/assertion/assertion.mbt
@@ -142,16 +142,16 @@ pub fn assert_is[T : Debug](
 
 test "assert_is.is" {
   let x = 1
-  let s = "x:\(x)"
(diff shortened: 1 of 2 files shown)


assertion/assertion.mbt
```diff
@@ -140,20 +140,6 @@ pub fn assert_is[T : Debug](
   }
 }
 
-test "assert_is.is" {
-  let x = 1
-  let s = "x:\(x)"
```
bobzhang, line 145: we should fix `assert_is` here instead

hackwaly: My concern is that will put testing only type definition `T` in the `assertion.mbt`, which decrease source code readability.

PR #225, opened 2024-04-09
feat: Array stable sort

He1pa:
Use an improved timsort algorithm to implement Array stable sort. refers from rustc std lib

https://github.com/rust-lang/rust/blob/bd12986fd6659a3091cff7694b8225374f4a26fe/library/core/src/slice/sort.rs#L1034

## Summary by CodeRabbit

- **New Features**
	- Introduced a new stable sorting functionality in arrays using the TimSort algorithm, which preserves the order of equal elements.

- **Chores**
	- Updated package dependencies to include necessary modules for new features.

### DIFF
--- a/array/sort.mbt
+++ b/array/sort.mbt
@@ -48,9 +48,7 @@ fn timsort[T : Compare](arr : ArraySlice[T]) -> Unit {
   // Short arrays get sorted in-place via insertion sort to avoid allocations.
   let len = arr.length()
   if len <= max_insertion {
-    if len > 20 {
       insertion_sort(arr)
-    }
   }
   let mut end = 0
   let mut start = 0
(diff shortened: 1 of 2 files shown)


array/sort.mbt
```diff
@@ -12,6 +12,188 @@
...
+  let len = arr.length()
+  if len <= max_insertion {
+      insertion_sort(arr)
+  }
+  let mut end = 0
+  let mut start = 0
+  let runs : @vec.Vec[TimSortRun] = @vec.Vec::new()
+  while end < len {
+    let (streak_end, was_reversed) = find_streak(
+      arr.slice(start, arr.end - start),
```
Yu-zh, line 58: Should just run till the end.
```suggestion
      arr.slice(start, arr.end),
```

PR #228, opened 2024-04-09
Implements Eisel-Lemire algorithm to speed up floating point parsing.

jialunzhang-psu:
Ported from [fast-float-rust](https://github.com/aldanor/fast-float-rust).
Work in progress.

### DIFF
--- a/strconv/UInt64.mbt
+++ b/strconv/UInt64.mbt
@@ -1,0 +1,90 @@
+// Copyright 2024 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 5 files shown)


yj-qin:
Is it necessary to refactor the decimal based fallback algorithm? Maybe just add a fast path Eisel-Lemire algorithm.

jialunzhang-psu:
I learned that a new version of Eisel-Lemire without fallback has been proposed in [Fast Number Parsing Without Fallback](https://arxiv.org/abs/2212.06644) and its source code is available at [fast_float](https://github.com/fastfloat/fast_float). Implementing this newer version requires a lot more work. But anyway, it still uses Clinger's fast path as the first attack. So I just implement this fast path and use the existing fallback, leaving the full newest Eisel-Lemire for later.

yj-qin (approved):

bobzhang:
hi @jialunzhang-psu thanks for your contributions, I can not push on this branch directly, I rebased and solved the conflicts in #251

PR #230, opened 2024-04-10
CI add moon test --target js

hackwaly:


lijunchen:
Since `moonrun` does not support WasmGC, we directly test the JS backend in CI instead of adding `moon test --target all`. cc @bobzhang

lijunchen (approved):

PR #231, opened 2024-04-10
Fix ryu tests

hackwaly:
and port tests from compiler


lijunchen (approved):

lijunchen:
Once we have fixed data section issue, we can use `inspect` again

PR #234, opened 2024-04-10
rename debug to trace and use %trace intrinsic

hackwaly:
Our expectation is for users to use `debug` more frequently, rather than `println`. However, presently, the situation is the opposite, and I believe the issue lies in the name 'debug'.

This PR renames 'debug' to the more common 'trace', and introduces the `%trace` intrinsic to allow compiling it to `console.log` on the JavaScript backend.

### DIFF
--- a/builtin/debug.mbt
+++ b/builtin/debug.mbt
@@ -29,6 +29,12 @@ pub fn trace[X : Debug](x : X) -> Unit {
   trace_string(buf.to_string())
 }
 
+/// print a value to the console using [Debug], with a tailing newline
+/// @alert deprecated "Use trace instead"
+pub fn debug[X : Debug](x : X) -> Unit {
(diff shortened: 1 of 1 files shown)


builtin/debug.mbt
```diff
@@ -17,11 +17,22 @@ pub trait Debug {
...
+pub fn trace[X : Debug](x : X) -> Unit {
   let buf = Buffer::make(10)
   x.debug_write(buf)
-  println(buf.to_string())
+  trace_string(buf.to_string())
+}
+
+/// print a value to the console using [Debug], with a tailing newline
+/// @deprecated Use trace instead
+pub fn debug[X : Debug](x : X) -> Unit {
```
bobzhang, line 34: `@alert deprecated "Use trace instead"`

PR #240, opened 2024-04-11
Fix endless loop and memleak after reset in buffer

hackwaly:
While it's endless loop, so can't add test for it lol.

## Summary by CodeRabbit

- **Refactor**
	- Updated the buffer initialization and management strategy to enhance performance and reliability.

### DIFF
--- a/builtin/buffer.mbt
+++ b/builtin/buffer.mbt
@@ -28,6 +28,7 @@
 struct Buffer {
   mut bytes : Bytes
   mut len : Int
+  initial_bytes : Bytes
 }
 
 pub fn expect(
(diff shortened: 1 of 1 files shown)


builtin/buffer.mbt
```diff
@@ -28,6 +28,7 @@
 struct Buffer {
   mut bytes : Bytes
   mut len : Int
+  mut initial_capacity : Int
```
bobzhang, line 31: can we keep initial_buffer instead, so that for small buffers it never allocates. 
It is also immutable

bobzhang (changes requested):
see the comments

PR #243, opened 2024-04-11
Expose Decimal struct

hackwaly:
This PR expose Decimal struct detail. So json/json5 can use it to parse double.


bobzhang (commented):
only types needed here?
If that's the case, I would rather duplicate its definitions to avoid inter-dependencies

hackwaly:
> only types needed here? If that's the case, I would rather duplicate its definitions to avoid inter-dependencies

No, the `to_double` needed also. It pulls nearly whole strconv pkg.

PR #254, opened 2024-04-13
Add Bytes::from_array()

fantix:


bobzhang:
shall we add bytes literal syntax instead?

fantix:
> shall we add bytes literal syntax instead?

Yeah that would be better

hackwaly:
> shall we add bytes literal syntax instead?

I think both are worth to have.

PR #256, opened 2024-04-13
String to array

fantix:
Yields an array of `Char`s

## Summary by CodeRabbit

- **New Features**
	- Enabled diffing for specific file types to improve version control.
	- Updated package dependencies to include necessary modules for enhanced functionality.


string/string.mbt
fantix, line 1: I think it is wrong that `moon fmt` replaces hex literals in strings with Unicode characters. It makes Git think such files are binary.

bobzhang:
fixed blob in #259

PR #255, opened 2024-04-13
Make `Result` a `Show`

fantix:
... so that `Result` can be `inspect`ed.

## Summary by CodeRabbit

- **New Features**
	- Added a function to convert result types to strings for improved user readability.
- **Tests**
	- Included new test cases to ensure reliability of the display functionality.

PR #257, opened 2024-04-13
Add `@uuid` package

fantix:
This PR adds RFC 4122 UUID with basic functions.

Reference: https://github.com/python/cpython/blob/3.12/Lib/uuid.py

Depends on:

- [x] #254 
- [x] #255 
- [x] #259

### DIFF
--- a/uuid/moon.pkg.json
+++ b/uuid/moon.pkg.json
@@ -1,0 +1,11 @@
+{
+    "import": [
+        "moonbitlang/core/array",
+        "moonbitlang/core/builtin",
+        "moonbitlang/core/bytes",
+        "moonbitlang/core/coverage",
+        "moonbitlang/core/int64",
+        "moonbitlang/core/result",
+        "moonbitlang/core/string"
+    ]
+}
--- a/uuid/uuid.mbt
+++ b/uuid/uuid.mbt
@@ -1,0 +1,190 @@
+// Copyright 2024 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// UUID that complies with RFC 4122.
+///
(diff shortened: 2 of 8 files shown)


uuid/moon.pkg.json
```diff
@@ -0,0 +1,11 @@
+{
+    "import": [
+        "moonbitlang/core/array",
+        "moonbitlang/core/builtin",
+        "moonbitlang/core/bytes",
+        "moonbitlang/core/coverage",
+        "moonbitlang/core/int64",
+        "moonbitlang/core/result",
+        "moonbitlang/core/string"
```
fantix, line 9: Many are test imports.

uuid/uuid.mbt
```diff
@@ -0,0 +1,185 @@
...
+    rv[i] = self.lo.lsr((15 - i) * 8).to_int()
+  }
+  rv
+}
+
+test "bytes" {
+  let buf = Bytes::[
+    201, 130, 251, 104, 223, 80, 76, 161, 145, 248, 186, 162, 4, 9, 229, 185,
+  ]
+  inspect(from_bytes(buf)?.to_bytes(), ~content="苉棻僟ꅌꊺऄ맥")? // XXX: this is a terrible Show
```
fantix, line 83: `mbt test -u` generated this crap - again, maybe using `Show` in `inspect()` is not the best option, we might need a new trait or something?

uuid/uuid.mbt
```diff
@@ -0,0 +1,185 @@
...
+  fn { c => c.to_int() },
+)
+
+let dash : Int = '-'.to_int()
+
+let segments = [8, 4, 4, 0, 4, 12]
+
+// XXX: use sum() instead; weird `moon fmt`
+let size : Int = (segments.fold_left(fn { acc, x => acc + x }, ~init=0) + segments.length() -
+  2) * 2
```
fantix, line 154: weird `moon fmt`

uuid/uuid.mbt
bobzhang: cc @lijunchen is still the case?

uuid/uuid.mbt
bobzhang: cc @Yoorkin

uuid/uuid.mbt
lijunchen: Yes. This is because the behavior of `Bytes::to_string` is to interpret the content as UTF-16 string.

lijunchen: > Yes. This is because the behavior of `Bytes::to_string` is to interpret the content as UTF-16 string.

Maybe we need to change it to print bytes literal.

bobzhang (approved):

PR #267, opened 2024-04-14
feat: add immutable_stack

Lampese:
## Summary by CodeRabbit

- **New Features**
	- Introduced an immutable stack data structure with functionalities such as push, pop, peek, and more.
	- Added support for Moonbitlang core functionalities in the application.
	- Implemented a struct `ImmutableStack` for managing immutable stack data.
	- Updated README with details on the ImmutableStack data structure.


Lampese:
@CAIMEOX  Help me write the documentation.

Lampese:
@coderabbitai pause

Yoorkin:
Are there any real-world examples of using `immutable_stack`? In my experience, a `List` is sufficient because it's convenient for pattern matching and easy to use with functional code.

Lampese:
> Are there any real-world examples of using `immutable_stack`? In my experience, a `List` is sufficient because it's convenient for pattern matching and easy to use with functional code.

This is indeed an issue that needs to be considered. However, I think ImmutableStack can be replaced with List but semantically it still can't be labeled with a method to use like Stack. But it's true that there's not much need to rewrite a structure, and I'll close this PR for now, and possibly reopen it when it comes up for discussion again.

PR #292, opened 2024-04-19
Add `@time` package

yj-qin:
This PR adds a `time` package, including types and methods for date and time manipulation. We currently need to use FFI to acquire the timestamp and time zone.


time/util.mbt
```diff
@@ -0,0 +1,280 @@
...
+    checked_mul_int64(Int64::max_value(), 2L),
+    content="Err(Int64 overflow)",
+  )?
+  inspect(
+    checked_mul_int64(Int64::min_value(), 2L),
+    content="Err(Int64 overflow)",
+  )?
+}
+
+fn floor(x : Double) -> Double {
```
peter-jerry-ye, line 209: This should be made an intrinsic @Yu-zh

time/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+Package time provides functionality for measuring and manipulating time.
+
+The calendrical calculations always assume a ISO 8601 calendar, with no leap seconds.
+
+To create a datatime that represents the current time, you need to obtain the unix second and time zone offset from the [wasi](https://mooncakes.io/docs/#/peter-jerry-ye/wasi/) package (wasm-gc backend) or other FFI functions, and manually create a datetime.
+
+```
+// creates a UTC+8 fixed time zone.
+let zone = @time.fixed_zone("Asia/Shanghai", 8*60*60)
```
peter-jerry-ye, line 13: There is a database containing all the standard code and offset: https://www.iana.org/time-zones

I start to hesitate whether we should make it part of the core or maintain it as a separate package

time/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+// creates a ZonedDateTime from unix second and time zone.
+let date_time = @time.unix(1714227729L, nanosecond=1000, ~zone) // Ok(2024-04-27T22:22:09.000001+08:00[Asia/Shanghai])
+```
+
+## TODOs
+
+- Convert from/to RFC format string.
```
peter-jerry-ye, line 21: I assume it is [rc9557](https://datatracker.ietf.org/doc/html/rfc9557). Corret me if I'm wrong.

yj-qin: I believe that [rfc3339](https://datatracker.ietf.org/doc/html/rfc3339) is more widely used compared to rfc9557, which is a newer standard. We should also consider supporting [rfc2822](https://datatracker.ietf.org/doc/html/rfc2822) that is used in emails.

time/README.md
yj-qin: The time-zone implementation only supports fixed offset now. Maybe we can refer to how Python integrates time zone data into the standard library ([PEP 615](https://peps.python.org/pep-0615/)).

peter-jerry-ye: Considering we are a compiled language with multiple backends, and one of the backend (wasm) has multiple runtime (web or wasi), I'm not sure if we can really refer to Python's solution.

yj-qin: IMO, we can add public `TimeZone` and `TimeZoneOffset` traits and make zone-related codes depend on these traits. We have two options for the time zone database:
- **Maintaining an independent `tzdata` package.** Only retain a `FixedZone` in the core library. If users need an unfixed time zone, they can use the `tzdata` package to construct a `TimeZone` instance with offset transition rules from the IANA database.
- **Embedding `tzdata` in the core library.** We need to generate moonbit codes from the IANA database to hold the time zone data and implement a lookup mechanism for constructing the `TimeZone` instance. We need to bump the version of the core library when the IANA database updates.

time/plain_time.mbt
```diff
@@ -0,0 +1,337 @@
...
+  }
+  let mut hour = 0
+  let mut minute = 0
+  let mut second = 0
+  let mut nanosecond = 0
+  hour = @strconv.parse(str.substring(end=2))?
+  minute = @strconv.parse(str.substring(start=3, end=5))?
+  if str.length() > 7 {
+    if str[5] != ':' {
+      return Err(invalid_time_err)
```
Lampese, line 51: If you are using Result, should this line be able to be run and added to coverage?

time/plain_time.mbt
```diff
@@ -0,0 +1,337 @@
...
+/// Returns the nanosecond of this time.
+pub fn nanosecond(self : PlainTime) -> Int {
+  self.nanosecond
+}
+
+/// Returns the total seconds of this time.
+pub fn second_of_day(self : PlainTime) -> Int {
+  let sec = self.hour * seconds_per_hour.to_int() + self.minute * seconds_per_minute.to_int() +
+    self.second
+  sec
```
Lampese, line 104: Is the definition of this sec necessary?

time/plain_time.mbt
```diff
@@ -0,0 +1,337 @@
...
+
+/// Returns a new time with the specified nanosecond.
+pub fn with_nanosecond(
+  self : PlainTime,
+  nanosecond : Int
+) -> Result[PlainTime, String] {
+  if nanosecond == self.nanosecond {
+    return Ok(self)
+  }
+  PlainTime::of(self.hour, self.minute, self.second, nanosecond)
```
Lampese, line 260: I think it would be beneficial for the compiler to handle branches more effectively in this way.
```suggestion
if nanosecond == self.nanosecond {
    return Ok(self)
} else {
    PlainTime::of(self.hour, self.minute, self.second, nanosecond)
}
```

time/util.mbt
```diff
@@ -0,0 +1,280 @@
...
+  inspect(floor_div_int(0, 10), content="Ok(0)")?
+  inspect(floor_div_int(10, 0), content="Err(division by zero)")?
+}
+
+fn floor_div_int64(x : Int64, y : Int64) -> Result[Int64, String] {
+  if y == 0L {
+    return Err("division by zero")
+  }
+  let r = x.to_double() / y.to_double()
+  Ok(floor(r).to_int64())
```
Lampese, line 244: same

time/duration.mbt
```diff
@@ -0,0 +1,259 @@
...
+          let mut nanos_str = splits[1]
+          nanos_str = add_suffix_zero(nanos_str, 9)
+          if seconds < 0L {
+            nanos_str = "-" + nanos_str
+          }
+          nanoseconds = @strconv.parse(nanos_str)?
+        }
+        break
+      }
+      _ => return Err(invalid_duration_err)
```
Lampese, line 94: same

time/period.mbt
```diff
@@ -0,0 +1,219 @@
...
+pub fn days(self : Period) -> Int {
+  self.days
+}
+
+/// Adds specified years to this period, and returns a new period.
+pub fn add_years(self : Period, years : Int) -> Result[Period, String] {
+  if years == 0 {
+    return Ok(self)
+  }
+  let new_years = checked_add_int(self.years, years)?
```
Lampese, line 111: same

time/period.mbt
```diff
@@ -0,0 +1,219 @@
...
+  Ok(Period::of(years=new_years, months=self.months, days=self.days))
+}
+
+/// Adds specified weeks to this period, and returns a new period.
+pub fn add_weeks(self : Period, weeks : Int) -> Result[Period, String] {
+  if weeks == 0 {
+    return Ok(self)
+  }
```
Lampese, line 119: same

time/period.mbt
```diff
@@ -0,0 +1,219 @@
...
+  let new_days = checked_add_int(self.days, checked_mul_int(weeks, 7)?)?
+  Ok(Period::of(years=self.years, months=self.months, days=new_days))
+}
+
+/// Adds specified months to this period, and returns a new period.
+pub fn add_months(self : Period, months : Int) -> Result[Period, String] {
+  if months == 0 {
+    return Ok(self)
+  }
```
Lampese, line 128: same

time/period.mbt
```diff
@@ -0,0 +1,219 @@
...
+  let new_months = checked_add_int(self.months, months)?
+  Ok(Period::of(years=self.years, months=new_months, days=self.days))
+}
+
+/// Adds specified days to this period, and returns a new period.
+pub fn add_days(self : Period, days : Int) -> Result[Period, String] {
+  if days == 0 {
+    return Ok(self)
+  }
```
Lampese, line 137: same

time/period.mbt
```diff
@@ -0,0 +1,219 @@
...
+  let new_days = checked_add_int(self.days, days)?
+  Ok(Period::of(years=self.years, months=self.months, days=new_days))
+}
+
+/// Adds other period to this period, and returns a new period.
+pub fn add_period(self : Period, other : Period) -> Result[Period, String] {
+  if other.is_zero() {
+    return Ok(self)
+  }
```
Lampese, line 146: same

Lampese (commented):
There are some subtle code style modification suggestions.

time/plain_time.mbt
yj-qin: This should be covered by the test case "123:00:00"

peter-jerry-ye:
LGTM. I think we can discuss how to include the timezone database in the future.

PR #295, opened 2024-04-20
Add Char and String utils about UTF-16 encoding

yj-qin:


string/string.mbt
```diff
@@ -111,3 +111,42 @@ test "to_array" {
...
+    None
+  } else {
+    Some(self[index].to_int())
+  }
+}
+
+test "char_code_at" {
+  inspect("ABC".char_code_at(0), content="Some(65)")?
+  inspect("🤣".char_code_at(0), content="Some(55358)")?
+  inspect("🤣".char_code_at(1), content="Some(56611)")?
```
bobzhang, line 127: char_code_at and code_point_at seems to be inconsistent with String.prototype.charCodeAt and String.prototype.codePointAt, is this intentional?

yj-qin: In javascript, if the index out of range, charCodeAt will return a `NaN`, and codePointAt will return an `undefined`. How can we approach this behavior in MoonBit?

bobzhang: https://github.com/moonbitlang/core/issues/299
I wrote some of my thoughts here

PR #301, opened 2024-04-22
Mutable map reusing mutable set's red-black tree

jialunzhang-psu:
This PR should wait till the decision on the reorganization of collections. 
See #342 #324.


bobzhang:
@jialunzhang-psu you are welcome to revive this when you have time, close for now since it is inactive for a while.

PR #302, opened 2024-04-22
Json utils

peter-jerry-ye:
Allow easy access to nested item without pattern matching around


json/json.mbti
```diff
@@ -12,6 +12,26 @@ pub enum JsonValue {
   Object(@moonbitlang/core/map.Map[String, JsonValue])
 }
 fn JsonValue::debug_write(JsonValue, @moonbitlang/core/builtin.Buffer) -> Unit
+fn JsonValue::get_as_array(JsonValue) -> Option[@moonbitlang/core/vec.Vec[JsonValue]]
+fn JsonValue::get_as_bool(JsonValue) -> Option[Bool]
```
bobzhang, line 16: the API is too lengthy.
how about
```
inspect(x.array().unwrap()[3].object().unwrap()["xx"])
```

bobzhang (commented):
shall we make sure json/json5 share the types so that we need a package like
jsontypes

hackwaly:
> shall we make sure json/json5 share the types so that we need a package like jsontypes

Currently json5 has no types, it uses types in json directly.

json/json.mbti
peter-jerry-ye: How about just remove `get_` but keeping `as_array`?

With quotation mark, we can have: `x.item(3)?.value_as_string("key")?` instead of `x.as_array().unwrap().get(3).unwrap().as_object().unwrap().lookup("key").unwrap().as_string().unwrap()`

bobzhang: remove `get_` looks good

hackwaly:
I agree to remove functions like `item_as_null(i: Int)` and keep only `as_vec(i: Int)`, so users can use `json.as_vec().unwrap().get()`

peter-jerry-ye:
I don't see the problem with more helper functions.

hackwaly:
> I don't see the problem with more helper functions.

Ok, just my opinion.

bobzhang:
@peter-jerry-ye  I am okay with adding more utils, but the name is not very intuitive: `item_as_number`, `value_as_number` is  hard to figure out the meaning without looking into implementations. Can we split it into two commits so that APIs without controversials can get in first?

We support optional chaining so that the API composition is not that bad, e.g, `self.as_array().?item(3).as_obj().?(key)`

PR #300, opened 2024-04-22
iter: mvp version

hackwaly:
This PR add Iter package to core. Closes #294 

also added a Array::to_iter method.

Other `to_iter`s will be added later in separated PRs

PR #303, opened 2024-04-22
Update moon info output to new version with name shortening

lynzrand:


hackwaly:
This blocks https://github.com/moonbitlang/core/pull/300. So I merge this first.

PR #305, opened 2024-04-22
Bigint

jialunzhang-psu:


bigint/bigint.mbti
```diff
@@ -0,0 +1,41 @@
...
+
+// Values
+fn from_decimal_string(String) -> BigInt
+
+fn from_int(Int) -> BigInt
+
+fn from_int64(Int64) -> BigInt
+
+fn make_zero() -> BigInt
+
```
bobzhang, line 11: TODO: we will add literal syntax support, e.g, `0n`

bigint/bigint.mbti
```diff
@@ -0,0 +1,41 @@
...
+  op_equal(Self, Self) -> Bool
+  op_mod(Self, Self) -> Self
+  op_mul(Self, Self) -> Self
+  op_neg(Self) -> Self
+  op_sub(Self, Self) -> Self
+  to_decimal_string(Self) -> String
+}
+
+type Sign
+impl Sign {
```
bobzhang, line 32: Should this type be publicly exposed?

bigint/bigint.mbti
```diff
@@ -0,0 +1,41 @@
...
+fn from_int64(Int64) -> BigInt
+
+fn make_zero() -> BigInt
+
+// Types and methods
+type BigInt
+impl BigInt {
+  compare(Self, Self) -> Int
+  debug_write(Self, Buffer) -> Unit
+  deep_clone(Self) -> Self
```
bobzhang, line 17: when `deep_clone` is needed given BigInt is pure?

bobzhang (commented):
looks good to me in general.
It would be interesting to see the performance compared with implementation using externref bindings through V8

bigint/bigint.mbti
```diff
@@ -0,0 +1,41 @@
...
+  lsl(Self, Int) -> Self
+  lsr(Self, Int) -> Self
+  op_add(Self, Self) -> Self
+  op_div(Self, Self) -> Self
+  op_equal(Self, Self) -> Bool
+  op_mod(Self, Self) -> Self
+  op_mul(Self, Self) -> Self
+  op_neg(Self) -> Self
+  op_sub(Self, Self) -> Self
+  to_decimal_string(Self) -> String
```
bobzhang, line 28: should be called `to_string` directly?

bigint/bigint.mbti
jialunzhang-psu: annotated

bigint/bigint.mbti
jialunzhang-psu: fixed

bigint/bigint.mbti
jialunzhang-psu: fixed

bigint/bigint.mbti
jialunzhang-psu: Later we might have to_hex_string, etc. But given that decimal is the most common case, we can name it just `to_string`.

PR #308, opened 2024-04-23
fix some operators in Iter and add more test

Yu-zh:


hackwaly:
Could you please split the commit to two commits which first add tests? So we can see which methods are wrong.

I  doubt there  are simpler replacement of some of your proposed changes.

hackwaly:
If this protocol are so hard to implement. I think we should  reconsider it.

bobzhang:
@hackwaly It is indeed confusing for complex cases, but it is one time effort, the change looks good to me

Issue #299, opened 2024-04-21
clean up the design of Char


bobzhang:
Char is a unicode code point in MoonBit, it is encoding agnostics, it is not tied to utf16, so we should not provide APIs like this:

    fn Char::is_surrogate(Char) -> Bool

We should provide String.as_iter which returns Iter[Char]

```mbt
“A😊𠮷B"
   .as_iter()
   .reduce(init=Vec::[], fn {acc, x -> acc.push(x)})
// expected output
[65, 128522, 134071, 66]
```

cc @yj-qin

PR #319, opened 2024-04-25
add as_iter method for string

Yu-zh:
This PR implements #299

### DIFF
--- a/string/string.mbt
+++ b/string/string.mbt
@@ -173,16 +173,26 @@ fn next_code_point(self : String, index : Int) -> Option[(Char, Int)] {
 pub fn as_iter(self : String) -> @iter.Iter[Char] {
   @iter._unstable_internal_make(
     fn(yield) {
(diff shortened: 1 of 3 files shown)


string/string.mbt
```diff
@@ -149,3 +149,53 @@ test "substring" {
...
+  return Some((Char::from_int(c), index + 1))
+}
+
+pub fn as_iter(self : String) -> @iter.Iter[Char] {
+  @iter._unstable_internal_make(
+    fn(yield) {
+      for i = 0 {
+        match self.next_code_point(i) {
+          None => break true
+          Some((c, next_i)) =>
```
hackwaly, line 179: `next_i` can be derived from `c > 0xffff` and `i`, so we can avoid a tuple box cost.

hackwaly:
The typo check is stupid. Can't figure out a way to ignore words in string literal.

PR #321, opened 2024-04-26
ci: ignore typos in string literals

Young-Flash:
cc @hackwaly

### DIFF
--- a/_typos.toml
+++ b/_typos.toml
@@ -4,3 +4,9 @@ extend-exclude = ["**/*_gen.mbt"]
 [default.extend-identifiers]
 IDentifier = "IDentifier"
 
+[default]
+extend-ignore-re = [
+    # ignore string literals
+    "\"(.*?)\"",
+    ".*#|.*",
+]


lijunchen (approved):
bleeding CI failed as expected

_typos.toml
```diff
@@ -4,3 +4,8 @@ extend-exclude = ["**/*_gen.mbt"]
 [default.extend-identifiers]
 IDentifier = "IDentifier"
 
+[default]
+extend-ignore-re = [
+    # ignore string literals
+    "\"(.*?)\"",
+]
```
bobzhang, line 11: also need support `#|` style string literals

hackwaly:
Need a a test to verify it.

Young-Flash:
I have test locally, should we need to add some intentional typos in the codebase to verify?

if so, which file would be appropriate to add these typos to?

hackwaly:
> I have test locally, should we need to add some intentional typos in the codebase to verify?
> 
> if so, which file would be appropriate to add these typos to?

If locally tested, I thinks it's fine to without test here. The coming MR has few cases of that.

PR #324, opened 2024-04-27
Reorganize collections

jialunzhang-psu:
This PR reorganizes containers into a new `collections` folder. There are two motivations:
- The list of packages in core is getting lengthy.
- After putting different containers into a separate `collections` folder, it becomes more convenient to share implementation between containers. For example, `hashmap` and `hashset` share the same underlying implementation. I split it out to be a separate package (`robinhood_hash`) and they both become wrappers over it.

What I did:
- Created a new folder `collections`. Moved relevant packages into it.
- Corrected dependency paths in other packages.
- Create a new `robinhood_hash` package under `collections`. Rewrite `hashset` and `hashmap` to be just wrappers.

Unresolved issue: 
- `core/collections/robinhood_hash` is a package visible from the outside world. No idea how to make it local to `collections`.

Note:
- I left `vec` and `list` outside of `collections`, since they are more commonly used.


bobzhang:
The orgnanization layout is mostly settled, thanks for your contributions

PR #326, opened 2024-04-27
Add Iter::sum() and Iter::product()

fantix:
Replaces #36 

Depends on:

- [ ] #325


bobzhang:
we are in favor of simple code e.g,  `x.to_iter().fold(init=0, fn {x,y=> x+y})`, thanks for your contributions

PR #327, opened 2024-04-27
Add Iter::ascending() and Iter::descending()

fantix:
Depends on:

- [ ] #326


bobzhang:
the diff is so large that it needs significant amount of work to make it alive. note we are going to provide some sugars like this ` for i in 0..= 10 { .. }`

PR #328, opened 2024-04-27
Add BytesView

fantix:
This is a 1:1 clone of `VecView`


bobzhang:
We are going to make Bytes immutable and introduce BytesView at some time. This PR is out of dated and relies on the fact that Bytes is mutable, so I will close it for now

PR #329, opened 2024-04-27
Add `@mem` package

fantix:
This PR adds int/int64 load/store with little/big endian, as well as `@mem.copy()` to memory-like objects.

Ideally, such operations should use appropriate intrinsics.

Depends on:

- [ ] #328

### DIFF
--- a/mem/mem_test.mbt
+++ b/mem/mem_test.mbt
@@ -1,0 +1,94 @@
+// Copyright 2024 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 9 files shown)


mem/mem_test.mbt
```diff
@@ -0,0 +1,94 @@
...
+
+test "load bytes" {
+  test_load(Bytes::[1, 2, 3, 4, 5, 6, 7, 8, 9])?
+}
+
+test "load bytes view" {
+  test_load(
+    {
+      Bytes::[0, 1, 2, 3, 4, 5, 6, 7, 8, 9][1..]
+    },
```
fantix, line 31: `moon fmt` added unnecessary `{}` here.

peter-jerry-ye: cc @Yoorkin

Yoorkin: Fixed. The fix will apply in the next release.

bobzhang:
you are welcome to revive this when you have time, close for now since it is inactive for a while.

PR #338, opened 2024-04-28
rename Char::from_int to Char::from_int_unsafe

lijunchen:
Part of https://github.com/moonbitlang/core/issues/299

This PR rename current `Char::from_int` to `Char::from_int_safe`, and add a safe version of `Char::from_int(Int) -> Option[Char]`

### DIFF
--- a/builtin/console.mbt
+++ b/builtin/console.mbt
@@ -50,7 +50,7 @@ pub fn to_string(self : Int64) -> String {
     if num2 != 0L {
       write_digits(num2)
     }
-    buf.write_char(Char::from_int(abs(num % 10L).to_int() + 48))
+    buf.write_char(Char::from_int_unsafe(abs(num % 10L).to_int() + 48))
   }
 
   write_digits(self)
@@ -78,7 +78,7 @@ pub fn to_string(self : Int) -> String {
     if num2 != 0 {
       write_digits(num2)
     }
-    buf.write_char(Char::from_int(abs(num % 10) + 48))
+    buf.write_char(Char::from_int_unsafe(abs(num % 10) + 48))
   }
 
   write_digits(self)
--- a/builtin/intrinsics.mbt
+++ b/builtin/intrinsics.mbt
@@ -184,7 +184,8 @@ pub fn Double::convert_i64_u(val : Int64) -> Double = "%i64_to_f64_u"
 
 pub fn Char::to_int(self : Char) -> Int = "%char_to_int"
 
-pub fn Char::from_int(val : Int) -> Char = "%char_from_int"
+// Converts a Int to a Char, ignoring validity.
+pub fn Char::from_int_unsafe(val : Int) -> Char = "%char_from_int"
 
 pub fn Char::op_equal(self : Char, other : Char) -> Bool = "%char_eq"
 
--- a/char/char.mbt
+++ b/char/char.mbt
@@ -28,9 +28,9 @@ test "to_string" {
 pub fn debug_write(self : Char, buf : Buffer) -> Unit {
   fn to_hex_digit(i : Int) -> Char {
     if i < 10 {
-      Char::from_int('0'.to_int() + i)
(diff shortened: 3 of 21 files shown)


char/char.mbt
```diff
@@ -74,9 +74,43 @@ test "debug_write" {
...
+  inspect(Char::from_int_unchecked(0xD800).is_valid(), content="false")?
+}
+
+test {
+  inspect(Char::from_int_unchecked(97), content="a")?
+  inspect('中'.to_int(), content="20013")?
+  inspect(Char::from_int(20013), content="Some(中)")?
+  inspect(Char::from_int_unchecked(20013), content="中")?
+  inspect('🤣'.to_int(), content="129315")?
+  inspect(Char::from_int(129315), content="Some(🤣)")?
```
lijunchen, line 115: @Guest0x0 `Some(🤣)` should be `Some('🤣')`

lijunchen:
Fix coverage

builtin/console.mbt
```diff
@@ -50,7 +50,7 @@ pub fn to_string(self : Int64) -> String {
     if num2 != 0L {
       write_digits(num2)
     }
-    buf.write_char(Char::from_int(abs(num % 10L).to_int() + 48))
+    buf.write_char(Char::from_int_unchecked(abs(num % 10L).to_int() + 48))
   }
```
bobzhang, line 54: is this in perf critical place? if not, we would prefer safe API

builtin/intrinsics.mbt
```diff
@@ -184,7 +184,7 @@ pub fn Double::convert_i64_u(val : Int64) -> Double = "%i64_to_f64_u"
 
 pub fn Char::to_int(self : Char) -> Int = "%char_to_int"
 
-pub fn Char::from_int(val : Int) -> Char = "%char_from_int"
+pub fn Char::from_int_unchecked(val : Int) -> Char = "%char_from_int"
```
bobzhang, line 188: this unsafe API does not be exposed, it can be pulled in on demand, so that it is only local to its used package

char/char.mbt
```diff
@@ -74,9 +74,45 @@ test "debug_write" {
...
+  @assertion.assert_eq(repr('\x0f'), "'\\x0f'")?
+  @assertion.assert_eq(repr(Char::from_int_unchecked(0)), "'\\x00'")?
 }
 
 pub fn hash(self : Char) -> Int {
   self.to_int()
 }
+
+pub let min : Char = Char::from_int_unchecked(0x0)
+
```
bobzhang, line 86: char ileral

builtin/console.mbt
lijunchen: no, but we cannot use Option::unwrap in `builtin`

coverage/coverage.mbt
```diff
@@ -102,9 +102,9 @@ fn escape_to(s : String, buf : Buffer) -> Unit {
 
   fn to_hex_digit(i : Int) -> Char {
     if i < 10 {
-      Char::from_int('0'.to_int() + i)
+      Char::from_int_unsafe('0'.to_int() + i)
```
lijunchen, line 105: cannot use Option::unwrap here, because of cyclic dependencies

json5/lex_misc.mbt
```diff
@@ -23,7 +23,7 @@ fn read_char(ctx : ParseContext) -> Option[Char] {
         if c2 >= 0xDC00 && c2 <= 0xDFFF {
           ctx.offset += 1
           let c3 = c1.lsl(10) + c2 - 0x35fdc00
-          return Some(Char::from_int(c3))
+          return Some(Char::from_int_unsafe(c3))
```
lijunchen, line 26: maybe performance critical in `json5` package

builtin/intrinsics.mbt
lijunchen: In my current implementation, it is used by `coverage` and `json5`

PR #340, opened 2024-04-28
fix license header check too slow

lijunchen:
Fix https://github.com/moonbitlang/core/issues/339


lijunchen:
execution time of license header check from >1min to ~4s

Issue #341, opened 2024-04-28
[Feature request] a unified trait is needed for various `View`s


shamiao:
As I saw in the current `core`, a concrete `VecView` can be obtained by `op_as_view(start_including, end_excluding)` operator. Although not yet implemented, similar viewing logic may be applied to other linear containers like `Array` or `List`. 

I suggest adding `View[T]` built-in trait as an abstraction of a minimum subset of methods for any `View`. A suggested definition is:

```
/// View[T] is a logical contiguous sequence of T with random access by index. (Not guaranteed to be contiguous in memory)
pub trait[T] View[T] {    // <- not supported by compiler yet
    length[T](Self) -> Int
    op_get[T](Self, Int) -> T
    as_iter[T](Self) -> Iter[T]
}
```

This allows `Char`, `String` and `Bytes` related methods work for various data sources, without bloating the numbers of methods at the same time.

```
// before
let a = Buffer::make(256);
a.write_string("DEV")
a.write_sub_string("ELOPEZZZ", 0, 5)
a.write_char('R')
println(a.to_string())  // -> DEVELOPER

// after - assuming exists: 
//   fn write[Char](self: Buffer, view: impl View[Char]) -> Unit
//   fn view(self: Char) -> impl View[Char]
//   fn view(self: String) -> impl View[Char]
//   fn subview(self: String, start: Int, end: Int) -> impl View[Char]
let a = Buffer::make(256);
a.write("DEV".view())
a.write("ELOPEZZZ".subview(0, 1))
a.write('R'.view())
println(a.to_string())  // -> DEVELOPER
```

```
// before
let moon = Bytes::of_string("MOON")
let mo = Bytes::[0x6d, 0x00, 0x6f, 0x00, 0x7a, 0x00]
moon.blit(0, mo, 0, 4)
moon.blit_from_string(4, "on", 0, 2)
println(moon.to_string()) // -> moon

// after - assuming exists:
//   fn blit(self: Bytes, source: impl View[Byte]) -> Unit
//   fn op_as_view(self: Bytes, start: Int, end: Int) -> impl View[Byte]
//   fn view_bytes(self: String) -> impl View[Byte]
let moon = Bytes::of_string("MOON")
let mo = Bytes::[0x6d, 0x00, 0x6f, 0x00, 0x7a, 0x00]
moon.blit(0, mo[0..4])
moon.blit(4, "on".view_bytes())
println(moon.to_string()) // -> moon
```

[...]

PR #346, opened 2024-04-29
add: byte view traits & binary representation methods.

shamiao:
This is a direct port of practice-tested `loneoutpost/core-extra/representation` mooncake.

# contents

This PR includes 2 main additions:

 - Abstractions `trait ByteView` & `trait ByteMutView` to indicate byte-slice-like objects, and type `struct BytesView` which is a concrete slice type matches above traits. 
 - Binary representations of `Int`, `Int64` and `Double` supporting user specified endianness. 

Solid and complete tests and documentations provided.

# notes

This PR somewhats implements #341.

This PR may be competing to #328 #329. In my opinion, `ByteView`(can only get Byte) and `ByteMutView`(can get and set) trait abstractions are critical, but `Loadable` or `Storable` abstractions is not needed in `core`. Referencing the practices from Rust, these abstractions is more appropriate to appear as `Serializable`/`Deserializable` in a `serde` 3rd party mooncake.


shamiao:
revoked due to:

 - partly duplicate with #328 (complete duplicate, not competing)
 - wrong commiter email

PR #347, opened 2024-04-29
add: struct BytesView, trait ByteView & binary representation methods.

shamiao:
This is a direct port of practice-tested `loneoutpost/core-extra/representation` mooncake.

# Contents

This PR includes 3 main parts

 - Abstractions `trait ByteView` & `trait ByteMutView` to indicate `Byte` sequences from any data source, read-only or writable.
 - Concrete type `struct BytesView` which indicates a read-and-writable `Byte` sequence sliced from a `Bytes`. 
 - Binary representations of `Int`, `Int64` and `Double` supporting any specified endianness. 

Solid and complete tests and documentations provided.

# Competing PRs

This PR is competing to #328 and #329.

## Concrete `struct BytesView`...

is similar but different from #328. This PR makes the output type `Byte` instead of `Int`. 

## Abstractions `trait ByteView` & `trait ByteMutView`

... are new. Since `Byte` sequence can come from various sources (memory, disk, network...), I think adding such abstractions to `core` is critical for supporting various programs.

## The whole `@mem` module_ appeared in #329 

... which includes abstractions `Loadable`(read from bytes) or `Storable`(write into bytes) and implementations for primitive types, are removed. 

This PR made these R/W operations (between primitive types and `Byte` views) simple methods in the primitive types' package, for the following reasons:

1. `Loadable`/`Storable` is actually an abstraction of serialization/deserialization(serde). As the practices from Rust community, serde is more appropriate to appear as a 3rd party package instead of a core feature. 
2. Additionally, endianness conversions are not always considered enough to be a serde operation. It's called binary representation more often.   
3. We cannot assume source and target of serde operations is memory. Binary representation operations may bypass memory entirely. `@mem` is more suggested to be reserved for usages dealing with real memory. 

# Notes

This PR somewhats implements #341.


shamiao:
note: actions `check/cla-check` failure might be a bug:

 - my commit is using CLA email <shamiao@loneoutpost.cn>
 - my github primary email is <shamiaoc@gmail.com>
 - both are verified by github

I think `check/cla-check` action should verify emails appeared in git log history, instead of primary email of github user?

PR #373, opened 2024-05-08
add `from_hex`, `to_hex` for BigInt

jialunzhang-psu:
Reopen #372


MingLLuo:
thks

jialunzhang-psu:
@MingLLuo Sorry for spoiling your fork, I pushed a wrong branch to it and github automatically closed that PR and prevented me from updating it. I left some TODO and will fix them later.

jialunzhang-psu:
@MingLLuo Forgot to mention that in the development of `core`, we follow a principle of least dependency, i.e., avoid importing other packages as much as possible to accelerate the building process. In this PR, `moonbitlang/char` in the import list is avoidable. I will also update that later.

MingLLuo:
Maybe to_string for Char should be in builtin. I first try just use bytes, set_char, sub_string. But it make me feel more complex.😂

PR #376, opened 2024-05-10
annotate array types

Yu-zh:
This PR annotates the type for all the array literals to pave the way for changing the default type of array literals to the vector.


array/sort.mbt
```diff
@@ -419,13 +419,13 @@ fn sift_down[T : Compare](arr : ArraySlice[T], index : Int) -> Unit {
 }
 
 fn test_sort(f : (Array[Int]) -> Unit) -> Result[Unit, String] {
-  let arr = [5, 4, 3, 2, 1]
+  let arr = Array::[5, 4, 3, 2, 1]
   f(arr)
```
bobzhang, line 423: let arr : Array[_] = [5,4,4,]

PR #377, opened 2024-05-10
CI: show version after install

bzy-debug:


bzy-debug:
close this in favor of #380

PR #405, opened 2024-05-16
[break change] tweak Bytes API, migrate to Byte type

Yoorkin:
MoonBit has introduced a `Byte` type, which represents an 8-bit unsigned integer. Currently we lacks the necessary primitives to convert between `Byte` and other basic types,  which limits its usability.

---

In this PR:

- add some intrinsics and arithmetic operations for byte: 
    - `Int::to_byte` `Int64::to_byte` `Byte::to_int64` 
- refine Bytes APIs (`Bytes::make` `Bytes::op_set` `Bytes::op_get`) to use `Byte` type, update codebase to adapt this changes
- add new method `Buffer::write_byte` for `moonbitlang/x/encoding`
---

NOTE: This PR requires collaboration with the compiler as it involves intrinsics. For the sake of easy testing and review, the implementations of `Int::to_byte` and `Int64::to_byte` in `intrinsics.mbt` temporarily use `%identity`. **These implementations will be replaced by intrinsics provided by the compiler later.**

### DIFF
--- a/builtin/builtin.mbti
+++ b/builtin/builtin.mbti
@@ -213,7 +213,7 @@ impl Bytes {
   blit_from_string(Bytes, Int, String, Int, Int) -> Unit
   copy(Bytes) -> Bytes
   length(Bytes) -> Int
-  make(Int, Byte) -> Bytes
+  make(Int, ~init : Byte = ..) -> Bytes
   of_string(String) -> Bytes
   op_get(Bytes, Int) -> Byte
   op_set(Bytes, Int, Byte) -> Unit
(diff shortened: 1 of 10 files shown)


builtin/builtin.mbti
```diff
@@ -209,10 +213,10 @@ impl Bytes {
   blit_from_string(Bytes, Int, String, Int, Int) -> Unit
   copy(Bytes) -> Bytes
   length(Bytes) -> Int
-  make(Int, Int) -> Bytes
+  make(Int, ~init : Byte = ..) -> Bytes
```
bobzhang, line 216: make sure `moon test` is not broken, I have an impression that the test driver used this API cc @lijunchen

Yoorkin: The test driver depends on Buffer, which hides the implementation detail of using bytes. So `moon test` still working.

Issue #409, opened 2024-05-17
Int64::hash cause the time complexity for hash map insertion to degrade


tonyfettes:
bench:
```rust
fn main {
  let map = @hashmap.HashMap::new()
  for i = 0; i < 1000; i = i + 1 {
    for j = 0; j < 1000; j = j + 1 {
      let k = i.to_int64().lsl(32).lor(j.to_int64())
      map.set(k, (i + j, i * j))
    }
  }
}
```

I modify the output of the JavaScript output of the MoonBit program above:

```diff
 (function () {
   const map = moonbitlang$core$hashmap$$HashMap$new$1$(moonbitlang$core$hashmap$$HashMap$new$46$hasher$46$default$1$());
   let _tmp$14 = 0;
+  let start = Date.now();
   while (true) {
     const i = _tmp$14;
     if (i < 1000) {
       let _tmp$15 = 0;
+      console.log(`elapsed: ${Date.now() - start} ms`);
+      start = Date.now();
       while (true) {
         const j = _tmp$15;
         if (j < 1000) {
           const k = BigInt.asIntN(64, $i64_extend_i32_s(i) << (BigInt(32) & 63n)) | $i64_extend_i32_s(j);
           moonbitlang$core$hashmap$$HashMap$set$1$(map, k, { _0: i + j | 0, _1: Math.imul(i, j) | 0 });
           _tmp$15 = j + 1 | 0;
           continue;
         } else {
           break;
         }
       }
       _tmp$14 = i + 1 | 0;
       continue;
     } else {
       return;
     }
   }
 }());
```

<details>
<summary>Output</summary>

```plaintext
elapsed: 0 ms
elapsed: 1 ms
elapsed: 15 ms
elapsed: 15 ms
elapsed: 42 ms
elapsed: 12 ms
elapsed: 18 ms
elapsed: 127 ms
elapsed: 20 ms
elapsed: 23 ms
elapsed: 27 ms
elapsed: 30 ms
elapsed: 35 ms
elapsed: 37 ms
elapsed: 492 ms
elapsed: 48 ms
elapsed: 51 ms
elapsed: 57 ms
elapsed: 58 ms
elapsed: 61 ms
elapsed: 65 ms
elapsed: 67 ms
elapsed: 71 ms
elapsed: 75 ms
elapsed: 77 ms
elapsed: 81 ms
elapsed: 85 ms
elapsed: 2034 ms
elapsed: 96 ms
elapsed: 97 ms
[...]

PR #414, opened 2024-05-20
use xxhash for int64

tonyfettes:
see #409.

The old version of int64 hash will create a lot of collision when the hi and lo part of the int64 ranges over a similar range, and hence the performance degradation seen in #409.

One drawback of this approach is, it allocates a `Bytes` object every time the `hash` function is called, but as per my micro-benchmarks they doesn't seems to be affect much.

This patch copied the `Int64` to a `Bytes` then call xxhash internally.

### DIFF
--- a/int64/int64.mbt
+++ b/int64/int64.mbt
@@ -60,3 +60,12 @@ pub fn hash(self : Int64) -> Int {
   b[7] = self.land(0xFFL).to_int()
   b.hash()
 }
+
+test "int64 hash" {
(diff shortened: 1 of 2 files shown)


Guest0x0:
If we view a `Int64` as a pair of two `Int`, then in the old version, `(a, b)` and `(b, a)` always have the same full-length hash value, since `xor` is commutative.

Note that in the future we may wish to expose some internal state API of `xxhash` to avoid allocating an intermediate byte.

Guest0x0 (approved):

PR #416, opened 2024-05-20
rename array and vector

Yu-zh:
Change list:
* move type definition from `@vec.Vec[T]` to `@builtin.Array[T]`, and `@vec.VecView[T]` to `@builtin.ArrayView[T]`
* rename `@builtin.Array[T]` to `@builtin.FixedArray[T]`
* add missing methods for `Array`, including `Array::make(len: Int, elem: T) -> Array[T]`, `Array::blit_to`, `Array::copy`, etc
* move functions and methods of `Array` from `@vec` package to `@builtin` package. This needs to be cleaned up because we'd like to keep builtin small.
* fix the `Debug` trait implementation for `Array`, and implement `Debug` trait for `Bytes`


immutable_set/immutable_set.mbti
```diff
@@ -37,9 +36,9 @@ impl ImmutableSet {
   split[T : Compare + Eq](Self[T], T) -> Tuple[Self[T], Bool, Self[T]]
   subset[T : Compare + Eq](Self[T], Self[T]) -> Bool
   to_array[T : Compare + Eq](Self[T]) -> Array[T]
+  to_fixed_array[T : Compare + Eq](Self[T]) -> Array[T]
```
Guest0x0, line 39: wrong name

PR #420, opened 2024-05-21
no warn on core

bobzhang:
@Young-Flash can you help finish the integration of --deny-warn


Young-Flash:
will be fixed at next moon release

Young-Flash (approved):

PR #422, opened 2024-05-21
feat: add int64 xxhash implementation

qazxcdswe123:
follow up of: https://github.com/moonbitlang/core/pull/414

### DIFF
--- a/int64/int64.mbt
+++ b/int64/int64.mbt
@@ -47,25 +47,3 @@ pub fn Int64::max_value() -> Int64 {
 pub fn Int64::min_value() -> Int64 {
   min_val
 }
-
-pub fn hash(self : Int64) -> Int {
-  let b = Bytes::make(8)
-  b[0] = self.lsr(56).to_byte()
-  b[1] = self.lsr(48).land(0xFFL).to_byte()
-  b[2] = self.lsr(40).land(0xFFL).to_byte()
-  b[3] = self.lsr(32).land(0xFFL).to_byte()
-  b[4] = self.lsr(24).land(0xFFL).to_byte()
-  b[5] = self.lsr(16).land(0xFFL).to_byte()
-  b[6] = self.lsr(8).land(0xFFL).to_byte()
-  b[7] = self.land(0xFFL).to_byte()
-  b.hash()
-}
-
-test "int64 hash" {
-  @assertion.assert_eq(0L.hash(), 0L.hash())?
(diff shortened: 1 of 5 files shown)


int64/int64.mbt
```diff
@@ -47,25 +47,3 @@ pub fn Int64::max_value() -> Int64 {
 pub fn Int64::min_value() -> Int64 {
   min_val
 }
-
```
bobzhang, line 50: int64_test.mbt

`slow_hash`

int64/int64.mbt
```diff
@@ -47,25 +47,3 @@ pub fn Int64::max_value() -> Int64 {
...
-  let b = Bytes::make(8)
-  b[0] = self.lsr(56).to_byte()
-  b[1] = self.lsr(48).land(0xFFL).to_byte()
-  b[2] = self.lsr(40).land(0xFFL).to_byte()
-  b[3] = self.lsr(32).land(0xFFL).to_byte()
-  b[4] = self.lsr(24).land(0xFFL).to_byte()
-  b[5] = self.lsr(16).land(0xFFL).to_byte()
-  b[6] = self.lsr(8).land(0xFFL).to_byte()
-  b[7] = self.land(0xFFL).to_byte()
-  b.hash()
```
bobzhang, line 61: `Bytes::[a,b,c]`

int64/int64.mbt
qazxcdswe123: done

int64/int64.mbt
qazxcdswe123: I think the current implementation is more intuitive so I left it as is.

int64/xxhash.mbt
```diff
@@ -0,0 +1,76 @@
...
+  }
+  acc = acc.lxor(acc.lsr(15))
+  acc *= gPRIME2
+  acc = acc.lxor(acc.lsr(13))
+  acc *= gPRIME3
+  acc = acc.lxor(acc.lsr(16))
+  acc
+}
+
+test "int64 hash" {
```
bobzhang, line 61: move this and slow_hash into xxhash_test.mbt so that they will not be compiled in non-test mode.

bobzhang (approved):
lgtm modulo some comments

int64/xxhash.mbt
qazxcdswe123: Done
I thought we have tree shaking in compiler

PR #424, opened 2024-05-22
merge fixedarray into Array package

Yoorkin:
This pr needs the cooperation of compiler, as FixedArray is a built-in type.


Yu-zh:
Why did you remove the Timsort?

PR #473, opened 2024-05-30
hongbo/simplify op set in hashmap

bobzhang:
- tweak
- simplify op_get in hashmap
- clean up

cc @yj-qin 

some questions:
1. does the wrapper of `make_hash` make sense?
2. for the index/next_index, the `land` is a bit confusing, should it be `mod


yj-qin (approved):

yj-qin:
1. If the hash function is good enough, the `make_hash` wrapping can be removed.
2. If  `mod` and `land` have the same performance, it could change to `mod` for better readability.

PR #475, opened 2024-05-31
Refactor int hash and hash in container

qazxcdswe123:
Changed the int hash function to a three round hash functions for 32 bit int, so we don't need additional hash function in hash container.
Hash function credit: https://github.com/skeeto/hash-prospector

### DIFF
--- a/int/int.mbt
+++ b/int/int.mbt
@@ -48,6 +48,15 @@ pub fn Int::min_value() -> Int {
   min_val
 }
 
+// https://github.com/skeeto/hash-prospector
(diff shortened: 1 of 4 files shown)


int/int.mbt
```diff
@@ -48,6 +48,15 @@ pub fn Int::min_value() -> Int {
   min_val
 }
```
bobzhang, line 50: 1. Can you add some more documentation here and clarify how you show the correctness of this piece of code?
2. I had another look at xxhash in int64, the while loop is not needed, since it can be unrolled (twice)

PR #483, opened 2024-06-01
add todo notes and remove abs which is not needed

bobzhang:
TODO:

new(size_hint:Int=..)

make sure the size is always positive and to the power of 2, so the logic can be simpliied

### DIFF
--- a/hashmap/hashmap.mbt
+++ b/hashmap/hashmap.mbt
@@ -25,6 +25,8 @@ pub fn HashMap::new[K, V]() -> HashMap[K, V] {
   }
 }
 
+/// TODO: calculate the size of the array
+/// and pre-allocate the array with the size.
 /// Create new hash map from array.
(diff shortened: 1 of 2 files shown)


hashmap/hashmap.mbt
```diff
@@ -187,8 +189,9 @@ fn grow[K : Hash + Eq, V](self : HashMap[K, V]) -> Unit {
   }
 }
 
+/// TODO: inline this function
 fn index[K : Hash, V](self : HashMap[K, V], hash : Int) -> Int {
-  hash.abs().land(self.capacity - 1)
+  hash.land(self.capacity - 1)
```
bobzhang, line 194: cc @yj-qin the abs operation is not needed?

yj-qin: yes, it's useless.

PR #487, opened 2024-06-02
Optimize LinkedHashMap

yj-qin:
- Add a ListNode array to eliminate circular references
- Remove abs()


bobzhang:
@coderabbitai review

FlyCloudC:
Why is the circular reference here bad？

PR #490, opened 2024-06-02
simplify hashmap

bobzhang:
1. add a labeld argument for capacity_hint, make sure the size is always 2^n and positive so the boundary can be simplified

### DIFF
--- a/hashmap/hashmap.mbt
+++ b/hashmap/hashmap.mbt
@@ -12,26 +12,35 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Default initial capacity
-let default_init_capacity = 8
-
-/// Create new hash map.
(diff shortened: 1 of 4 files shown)


hashmap/hashmap.mbt
```diff
@@ -82,22 +88,21 @@ pub fn op_set[K : Hash + Eq, V](
...
-        Some(entry) => {
-          if entry.hash == hash && entry.key == key {
-            break Some(entry.value)
-          }
-          if i > entry.psl {
-            break None
-          }
-          continue i + 1, self.next_index(idx)
+  for i = 0, idx = hash.land(self.capacity - 1) {
+    // if i == self.capacity {
```
bobzhang, line 92: @yj-qin this seems never going to happen?

yj-qin: Yes, it can be removed.

PR #491, opened 2024-06-02
futher simplify remove and shift_back operations in hashmap

bobzhang:
- simplify the remove operation
- simplify shift_back
- tweak

### DIFF
--- a/hashmap/hashmap.mbt
+++ b/hashmap/hashmap.mbt
@@ -155,11 +155,7 @@ pub fn remove[K : Hash + Eq, V](self : HashMap[K, V], key : K) -> Unit {
 }
 
 fn shift_back[K : Hash, V](self : HashMap[K, V], start_index : Int) -> Unit {
-  for i = 0, prev = start_index, curr = (start_index + 1).land(
-          self.capacity - 1,
-        )
-      i < self.entries.length()
-      i = i + 1, prev = curr, curr = (curr + 1).land(self.capacity - 1) {
+  for prev = start_index, curr = (start_index + 1).land(self.capacity - 1) {
     match self.entries[curr] {
       Some({ psl, hash, key, value, .. }) => {
         if psl == 0 {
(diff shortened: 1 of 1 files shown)


hashmap/hashmap.mbt
```diff
@@ -135,9 +135,7 @@ pub fn contains[K : Hash + Eq, V](self : HashMap[K, V], key : K) -> Bool {
 /// Remove a key-value pair from hash map.
 pub fn remove[K : Hash + Eq, V](self : HashMap[K, V], key : K) -> Unit {
   let hash = key.hash()
-  for i = 0, idx = hash.land(self.capacity - 1)
-      i < self.capacity
```
bobzhang, line 139: `i < self.capacity` check is not needed?

hashmap/hashmap.mbt
```diff
@@ -147,37 +145,37 @@ pub fn remove[K : Hash + Eq, V](self : HashMap[K, V], key : K) -> Unit {
...
       }
-      None => ()
+      None => break
     }
   }
 }
 
 fn shift_back[K : Hash, V](self : HashMap[K, V], start_index : Int) -> Unit {
-  for i = 0, prev = start_index, curr = (start_index + 1).land(
-          self.capacity - 1,
```
bobzhang, line 160: the loop state variable `i` is not needed

bobzhang:
@coderabbitai review

yj-qin (approved):

yj-qin:
The 50% load factor will result in half of the space being wasted, maybe a higher load factor is better?

bobzhang:
> The 50% load factor will result in half of the space being wasted, maybe a higher load factor is better?

I am open to change the load factor later, rationale why I prefer 0.5 instead of higher values:
1. when people use hashtbl, they want the optimal performance, so it is okay to exchange space for speed
2. the wasted the space is not that bad, since the array only holds a pointer(not the full content)

PR #492, opened 2024-06-02
Simplify LinkedHashMap

yj-qin:
- Provide a capacity hint parameter
- Simplify some methods

### DIFF
--- a/builtin/linkedhashmap.mbt
+++ b/builtin/linkedhashmap.mbt
@@ -58,17 +58,37 @@ struct LinkedHashMap[K, V] {
 
 // Implementations
 
-// Default initial capacity
-let default_init_capacity = 8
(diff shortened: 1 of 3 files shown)


builtin/linkedhashmap.mbt
```diff
@@ -58,17 +58,37 @@ struct LinkedHashMap[K, V] {
 
 // Implementations
 
-// Default initial capacity
-let default_init_capacity = 8
+fn power_2_above(x : Int, n : Int) -> Int {
+  let n = n.to_int64()
+  for i = x.to_int64() {
+    if i > 2147483647L {
+      break 2147483647
```
bobzhang, line 65:
```
if i  > max_int_32 {
   break i 
}
```
we want to keep the invariant that the returned value is 2^n

Issue #488, opened 2024-06-02
provide Array::shuffle_in_place


bobzhang:
Its signature could be similar to 
```
val shuffle :
  rand: (int -> int) -> 'a array -> unit
```
so that array is not dependent on Random

PR #506, opened 2024-06-04
feat: Add shuffle function to Array

qazxcdswe123:
close: #488 

Not sure if an API like this is what you want, and I also think we should provide a default rand function for the user so I added the rand package and make the `rand` argument optional.

### DIFF
--- a/array/array.mbt
+++ b/array/array.mbt
@@ -79,3 +79,18 @@ pub fn Array::new_with_index[T](length : Int, value : (Int) -> T) -> Array[T] {
     array
   }
 }
+
+fn rand(upper : Int) -> Int {
+  let summoner = @random.RandomInt::new()
+  summoner.summon() % upper
+}
+
+pub fn Array::shuffle[T](arr : Array[T], ~rand : (Int) -> Int = rand) -> Unit {
+  let n = arr.length()
+  let mut i = n - 1
+  while i > 0 {
+    let j = rand(i + 1)
+    arr.swap(i, j)
+    i = i - 1
+  }
+}
--- a/array/array.mbt
+++ b/array/array.mbt
@@ -97,7 +97,7 @@ pub fn Array::new_with_index[T](length : Int, value : (Int) -> T) -> Array[T] {
 pub fn Array::shuffle_in_place[T](arr : Array[T], ~rand : (Int) -> Int) -> Unit {
   let n = arr.length()
   for i = n - 1; i > 0; i = i - 1 {
-    let j = rand(i + 1) % (i + 1) 
+    let j = rand(i + 1) % (i + 1)
     // for safety, perf is not a concern here
     // TODO: maybe return an error later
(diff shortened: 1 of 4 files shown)


array/array.mbt
```diff
@@ -79,3 +79,15 @@ pub fn Array::new_with_index[T](length : Int, value : (Int) -> T) -> Array[T] {
...
+
+pub fn Array::shuffle[T](arr : Array[T], ~rand : (Int, Int) -> Int) -> Unit {
+  let n = arr.length()
+  let mut i = n - 1
+  while i > 0 {
+    let j = rand(0, i + 1)
+    let tmp = arr[i]
+    arr[i] = arr[j]
+    arr[j] = tmp
+    i = i - 1
```
Frank-III, line 91: I think there is a `swap` method?

qazxcdswe123: Yes, thank you for the reminder

array/array.mbt
```diff
@@ -79,3 +79,18 @@ pub fn Array::new_with_index[T](length : Int, value : (Int) -> T) -> Array[T] {
...
 }
+
+fn rand(upper : Int) -> Int {
+  let summoner = @random.RandomInt::new()
+  summoner.summon() % upper
+}
+
+pub fn Array::shuffle[T](arr : Array[T], ~rand : (Int) -> Int = rand) -> Unit {
+  let n = arr.length()
+  let mut i = n - 1
```
bobzhang, line 90: we tend to avoid mut if possible.
e.g,
```
for i = n - 1; i >0 ; i = i - 1 {}
```

array/moon.pkg.json
```diff
@@ -2,6 +2,7 @@
   "import": [
     "moonbitlang/core/builtin",
     "moonbitlang/core/coverage",
-    "moonbitlang/core/assertion"
+    "moonbitlang/core/assertion",
+    "moonbitlang/core/random"
   ]
```
bobzhang, line 7: we don't want array depend on random

array/array.mbti
```diff
@@ -40,6 +40,7 @@ impl Array {
   blit_to[A](Self[A], Self[A], ~len : Int, ~src_offset : Int = .., ~dst_offset : Int = ..) -> Unit
   copy[T](Self[T]) -> Self[T]
   new_with_index[T](Int, (Int) -> T) -> Self[T]
+  shuffle[T](Self[T], ~rand : (Int) -> Int = ..) -> Unit
   sort[T : Compare + Eq](Self[T]) -> Unit
   sort_by[T](Self[T], (T, T) -> Int) -> Unit
```
bobzhang, line 45: let's provide both `shuffle` and `shuffle_in_place`

bobzhang:
@qazxcdswe123 lgtm, I added two doc strings and a safety check

bobzhang (approved):

PR #511, opened 2024-06-04
Refactor map

qazxcdswe123:
If two keys are the same, we can safely assume that they hash to the same hash, so the `hash` field is not necessary.


yj-qin:
IMO, storing the hash value in an entry can provide a fast equality check. By first comparing hash values before checking key equality, we can prevent perf degradation caused by a poorly implemented key equality check. The downside is a higher memory footprint.

qazxcdswe123:
> By first comparing hash values before checking key equality

For `a && b`, we will need to evaluate **both** a and b (short circuit), unless what you mean is `same_hash || same_key`, but what if two different key hashed to the same hash?

As for poorly implemented key equality check, I guess it is kinda inevitable if we are comparing them?

what do you think?

yj-qin:
If a evaluate to false then b will not be evaluated and the whole expression is false. Please correct me if I'm wrong.

qazxcdswe123:
I see, that is a consideration. You are right.

Closing for now

PR #522, opened 2024-06-05
impl Eq,Debug,Show for Map #501

qazxcdswe123:
close: #501 

Changed: 
- API changed, moved `debug_linked_list` from `linkedhashmap.mbt` to `debug.mbt` and renamed. The result are also changed (more in test) to make the type checking happy. This implement Show and Debug trait.
```
// removed
// fn debug_linked_list[K : Show, V : Show](self : Map[K, V]) -> String {

// added
pub fn to_string[K : Debug, V : Debug](self : Map[K, V]) -> String {

// to debug.mbt
pub fn debug_write[K : Debug, V : Debug](
  self : Map[K, V],
  buf : Buffer
) -> Unit {
``` 
- Eq:
1. check size
2. from head to tail, both the idx(not sure about this), and the K, V should be the same, then we call them equal.


yj-qin:
I think the `to_string` method should return something like `{key1: value1, key2: value2}`

builtin/debug.mbt
```diff
@@ -159,3 +159,36 @@ pub fn debug_write(self : String, buf : Buffer) -> Unit {
...
+pub fn debug_write[K : Debug, V : Debug](
+  self : Map[K, V],
+  buf : Buffer
+) -> Unit {
+  loop self.head {
+    Some({ idx, key, value, hash, .. }) => {
+      buf.write_char('(')
+      // buf.write_string("\(hash),\(key),\(value)")
+      hash.debug_write(buf)
+      buf.write_string(",")
```
yj-qin, line 172: `write_char(',')`

builtin/debug.mbt
```diff
@@ -159,3 +159,36 @@ pub fn debug_write(self : String, buf : Buffer) -> Unit {
...
+      key.debug_write(buf)
+      buf.write_string(",")
```
yj-qin, line 174: same

builtin/linkedhashmap.mbt
```diff
@@ -429,3 +409,20 @@ pub fn to_array[K, V](self : Map[K, V]) -> Array[(K, V)] {
   }
   arr
 }
+
+pub fn op_equal[K : Eq, V : Eq](self : Map[K, V], that : Map[K, V]) -> Bool {
```
yj-qin, line 413: should add a test for `op_equal`

qazxcdswe123:
Thank you for your review! The necessary updates have been made.

PR #529, opened 2024-06-06
string comparation: use proper default strategy & add case-insensitive

shamiao:
Current strategy of `String.compare(self, other)` violates common senses of users too much. With all due respect, this can even be described as untolerable. Considering the following address book app example, if implemented by MoonBit:

example

I understand that comparing by length first may accelerate certain usages, but it simply doesn't worth to sacrifice simple common senses. It's the root of evil called 'premature optimizing'. 

I changed the 2 details in the `moonbitlang/core/string`:

1. Changed default compare strategy to char-by-char (unicode-code-point-by-unicode-code-point). The UTF-16 surrogate pairs are also handled properly, fixing the weird 'Emoticons[U+1F600] < BMP[U+FE11]' mistake in JavaScript, causing by naive 16bit-word-by-16bit-word comparison. 
2. Added a case insensitive variant to string comparisons. It only changes comparision behaviors to ASCII letters. Simple but useful for common needs.

### DIFF
--- a/char/char.mbt
+++ b/char/char.mbt
@@ -81,16 +81,22 @@ pub fn hash(self : Char) -> Int {
   self.to_int()
 }
 
-fn compare_case_insensitive(self: Char, b: Char) -> Int {
-  fn is_ascii_letter(c: Char) -> Bool {
-    ('A' <= c && c <= 'Z') || ('a' <= c && c <= 'z')
+fn compare_case_insensitive(self : Char, b : Char) -> Int {
+  fn is_ascii_letter(c : Char) -> Bool {
+    'A' <= c && c <= 'Z' || 'a' <= c && c <= 'z'
   }
-  fn to_uppercase(c: Char) -> Char {
+
+  fn to_uppercase(c : Char) -> Char {
     // NOTE: 'a' - 'A' == 0x20
-    if 'a' <= c && c <= 'z' { Char::from_int(c.to_int() - 0x20) } else { c }
+    if 'a' <= c && c <= 'z' {
+      Char::from_int(c.to_int() - 0x20)
+    } else {
+      c
+    }
   }
-  if is_ascii_letter(self) && is_ascii_letter(b) { 
-    to_uppercase(self).compare(to_uppercase(b)) 
+
+  if is_ascii_letter(self) && is_ascii_letter(b) {
+    to_uppercase(self).compare(to_uppercase(b))
   } else {
     self.compare(b)
   }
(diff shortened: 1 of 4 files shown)


shamiao:
note: the failed tests in `json` blocked some checks from passing.

char/char.mbt
```diff
@@ -80,3 +80,33 @@ test "debug_write" {
...
+  }
+  fn to_uppercase(c: Char) -> Char {
+    // NOTE: 'a' - 'A' == 0x20
+    if 'a' <= c && c <= 'z' { Char::from_int(c.to_int() - 0x20) } else { c }
+  }
+  if is_ascii_letter(self) && is_ascii_letter(b) { 
+    to_uppercase(self).compare(to_uppercase(b)) 
+  } else {
+    self.compare(b)
+  }
```
peter-jerry-ye, line 96:
```suggestion
  to_uppercase(self).compare(to_uppercase(b)) 
```
There's too many checks here

string/string.mbt
```diff
@@ -37,8 +55,62 @@ test "compare" {
...
+  @assertion.assert_eq(-1, "\u{AC00}".compare("\u{1F600}"))?
+}
+
+/// Compare two strings case-insensitive.
+///
+/// String are compared char-by-char by unicode code point value, but ASCII letters 
+/// are compared case-insensitive. 
+pub fn compare_case_insensitive(self : String, other : String) -> Int {
+  // NOTE: use traditional while loops, since we have no way to 
+  // poll 2 iterators (Iter[Char]) simultaneously.
```
peter-jerry-ye, line 84: I think you can implement a `zip` for `Iter`

peter-jerry-ye (commented):
You need to run `moon info` to update the `mbti` declaration

char/char.mbt
shamiao: @peter-jerry-ye  It's a trap here🤣, it causes this case to fail: 

`'_'(0x5F) < 'n'(0x6E) == false` because `'_'(0x5F) > 'N'(0x4E)`. 

Convert to uppercase before comparing only applies to both chars are ASCII letters. This has already been covered in test cases.

string/string.mbt
shamiao: @peter-jerry-ye  I once considered `zip`, but I think I'm not capable to decide how to design `zip` to deal with iterators having different lengths. (rustlang panics, python trimmes to shortest, how do moonbit?)

This may be postponed for further optimizations after the design of `zip`?

char/char.mbt
peter-jerry-ye: Then your implementation is incorrect as:
- 'N' < '_' < 'n'
- 'N' == 'n'

It is no longer a totally ordered set

string/string.mbt
peter-jerry-ye: I think you may give two versions of `zip` and `zip_exn` where the latter fails and it follows the naming convention that we have right now.

char/char.mbt
shamiao: Yes, it's theologically a partial order set, likes the float set with NaN. But the results of comparison is guaranteed to be deterministic.

peter-jerry-ye: Sorry, I mean it is no longer a strict partial order set

shamiao:
Please forgive me for providing an EXTREMELY UGLY quick fix to `test "stringify"` of `moonbitlang/core/json` just to pass `moon test`. This quick fix is expected be changed in the future, of course.

shamiao:
@peter-jerry-ye I'll close this PR and split my changes into multiple PRs.

PR #531, opened 2024-06-06
feat: Add to_array method to Bytes

qazxcdswe123:
Used in Unicode encoding (https://github.com/moonbitlang/x/commit/daa53ced0ca9ef3538064de41e80dae8bb730fee)
But Bytes itself should also mean byte array, so this may be confusing

### DIFF
--- a/bytes/bytes.mbt
+++ b/bytes/bytes.mbt
@@ -31,9 +31,9 @@ pub fn Bytes::from_array(arr : Array[Int]) -> Bytes {
 }
 
 pub fn to_array(self : Bytes) -> Array[Int] {
-  let rv = Array::with_capacity(self.length())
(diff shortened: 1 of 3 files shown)


bytes/bytes.mbt
```diff
@@ -29,3 +29,11 @@ pub fn Bytes::from_array(arr : Array[Int]) -> Bytes {
   }
   rv
 }
+
+pub fn to_array(self : Bytes) -> Array[Int] {
+  let rv = Array::with_capacity(self.length())
```
bobzhang, line 34: 1. it is more efficient to 
`Array::make(self.length(),0)` and `rv[i]=.. ` directly
2. Array::with_capacity API is a bit weird, can it be unified with 
 Array::init(~capcaity:Int)

PR #538, opened 2024-06-11
change bool to IterResult

bobzhang:
- change join
- explicit iter result


qazxcdswe123:
Accidentally ruined the commit history, but the diff should work just fine and the history should be cleaned after squash and merge.

Issue #514, opened 2024-06-04
A single Random state


bobzhang:
currently the random package has two type 

```
struct RandomInt {
  state : FixedArray[Int]
  mut index : Int
}

struct RandomInt64 {
  state : FixedArray[Int64]
  mut index : Int
}
```
It would be nice to use only one type `RandomState`, so that we have 

```
let default : RandomState  =  ...
pub fn init_state(...) {
}
```
In this case, it is convenient to use the API without passing around the seed why keep it flexible to 
seed the random package

bobzhang:
cc @Lampese

Lampese:
LGTM, and in fact, mt19937 is too slow, and most requirements don't need to be so strict. We have discussed before that the random provided by the compiler or using simpler random algorithms are necessary.

Unfortunately, I have been busy preparing for the NEMT recently and currently do not have time to complete this part of the writing.

PR #543, opened 2024-06-12
RandState based on xorshift*

Lampese:
- As mentioned in #514, the Random module was rewritten using RandomState and the faster xorshift * algorithm was adopted.

- The random number generator that originally used mt19937 will be migrated to [moonbitlang/x](https://github.com/moonbitlang/x) later on.


Lampese:
The next implementation transition will be Lagged Fibnoacci, which will be completed later by @CAIMEOX.

PR #550, opened 2024-06-13
use StringBuilder instead of Buffer in builtin and json package

hackwaly:
This PR introduce StringBuilder in builtin package. Thus avoid expensive Buffer operations when use `Show` trait in js/wasm-gc target.


builtin/builtin.mbti
```diff
@@ -186,6 +186,13 @@ impl SourceLoc {
   to_string(Self) -> String
 }
 
+type StringBuilder
+impl StringBuilder {
+  add_string(Self, String) -> Unit
+  make() -> Self
```
Yoorkin, line 192: rename `make()` to `new()`? related #593

builtin/moon.pkg.json
```diff
@@ -18,6 +18,8 @@
     "int64_nonjs.mbt": ["not", "js"],
     "arraycore_js.mbt": ["js"],
     "arraycore_nonjs.mbt": ["not", "js"],
-    "array_nonjs_test.mbt": ["not", "js"]
+    "array_nonjs_test.mbt": ["not", "js"],
+    "stringbuilder_buffer.mbt": ["not", ["js", "wasm-gc"]],
+    "stringbuilder_concat.mbt": ["js", "wasm-gc"]
```
hackwaly, line 23: I added `wasm-gc` to this list, since we are going to support `jsstring` in near future. 

But it still a question that can we assume every `wasm-gc` impies `jsstring` proposal?

builtin/moon.pkg.json
```diff
@@ -18,6 +18,8 @@
     "int64_nonjs.mbt": ["not", "js"],
     "arraycore_js.mbt": ["js"],
     "arraycore_nonjs.mbt": ["not", "js"],
-    "array_nonjs_test.mbt": ["not", "js"]
+    "array_nonjs_test.mbt": ["not", "js"],
+    "stringbuilder_buffer.mbt": ["not", ["js"]],
+    "stringbuilder_concat.mbt": ["js"]
```
hackwaly, line 23: Mark todo: add `wasm-gc` here once we have support for jsstring proposal on wasm-gc

bobzhang (commented):
do we have some concrete data points to justify the change?

Yoorkin:
I was supposed the StringBuilder is a faster (in js backend) and safer abstraction than Buffer. The Byte/Buffer is more suitable to writing binary stuff, programmers use it at their own risk: the `Buffer::to_string` do not check the output string. While StringBuilder will ensure the result is valid, and the `String` and `Char` being writing into are also already checked.

But in this PR `StringBuilder::sub_string` will write unpaired surrogate into result string, [which is not valid in UTF16](https://unicode.org/faq/utf_bom.html#utf16-7). Furthermore the StringBuilder is specialized for js backend and in others backends it just a wrapper of Buffer. It seems possible to do the same thing for Buffer's implementation, which leads the StringBuilder becoming a subset of Buffer.

I wonder why we still want to introduce a new abstraction. What functionalities should StringBuilder and Buffer provide separately?

hackwaly:
`write_sub_string` will be removed in `StringBuilder`. For now, to keep it just to make the diff smaller.

> I wonder why we still want to introduce a new abstraction. What functionalities should StringBuilder and Buffer provide separately?

As you said, Buffer is more suitable to writing binary stuff, and StringBuilder only cares string concatenation. Buffer is not always the best approach to concat strings on each targets. It's too slow compared to concat based approach on js platform even wasm-gc platform if we support JsStringBuiltins proposal.

I'll present some data result later to show that.

hackwaly:
The test result use this code: https://gist.github.com/hackwaly/24c2c16e25ef660772f3b06625ddb7a4

On my machine. Under js target: buffer based cost 5.7s. concat based cost 1.25s.

Yoorkin:
> and the String and Char being writing into are also already checked.

I was wrong because the invalid UTF16 string (contains unpaired surrogate) can still produced by `String::sub_string`, or just take a char by `String::op_get` and write it into StringBuilder. Will StringBuilder validate the input and result string in future?

hackwaly:
Probably not. It will align with the javascript's behavior. I think validation should be opt-in for users. It's best placed in unicode package?

hackwaly:
By replace the internal StringBuilder previously used in json package. The performance `@json.parse` also be improved after this MR.

bobzhang:
Per discussions offline, we agree to make a reasonably fast StringBuilder package on all platform.
1. StringBuilder should not depend on Buffer
2. we will move Buffer out of builtin package

hackwaly:
Now `StringBuilder` doesn't depend on `Buffer`.

builtin/stringbuilder_buffer.mbt
```diff
@@ -0,0 +1,70 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+struct StringBuilder {
+  mut bytes : Bytes
+  mut len : Int
+}
+
+pub fn StringBuilder::new(~size_hint : Int = 0) -> StringBuilder {
```
gmlewis, line 20: @bobzhang - shouldn't the "size_hint" refer to the number of characters and shouldn't the `size_hint = size_hint << 1` be performed within this function and not require users of the `size_hint` to perform the doubling operation themselves since that is an implementation detail?

I ask because of line 76 here:
https://github.com/moonbitlang/core/pull/1731/files#diff-183f011b464e6065c3eef786743e1ba9d2969609afa597570565cf15cd2e9958R76

Either way, it would be nice if there were documentation for `StringBuilder::new` that described whether `size_hint` means the number of `Byte`s or the number of `Char`s.

bobzhang: >  shouldn't the "size_hint" refer to the number of characters and shouldn't the size_hint = size_hint << 1 be performed within this function and not require users of the size_hint to perform the doubling operation themselves since that is an implementation detail?

It would mean `Byte`s instead of `Char`s, since the latter is hard to be precise (due to various code ranges)

I will add a docstring later

PR #557, opened 2024-06-14
feat: string add index_of api

schizobulia:


bobzhang:
the API lgmt, but the performance could be improved later.

PR #560, opened 2024-06-16
tests for rational

bobzhang:


bobzhang (commented):
test coding from codespaces

PR #562, opened 2024-06-17
chore: move immut/sparse_array to internal lib

Young-Flash:
cc @bobzhang, @lijunchen


bobzhang:
@Young-Flash internal/sparse_array/... instead of internal

PR #578, opened 2024-06-20
rename `contain`/`member` to `contains`

Yoorkin:
Rename the function `contain`/`member` in `immut/list`, `sorted_set`, `sorted_map` to `contains` to align with other collections.


bobzhang:
`contains` -> `has`?

PR #580, opened 2024-06-20
rename `forall/exists` to `all/any` in sorted_set

Yoorkin:
Both `forall/exists` and `all/any` make sense to me, but `all/any` are shorter. So in this PR, the `forall/exists` in sorted_set will be renamed to align with other collections.


hackwaly:
Another candidate: every/some

PR #596, opened 2024-06-24
feat: add Map::set_default

### LINKED ISSUE #588: [proposal]: Add `DefaultMap[K,V]` or `Map[K,V]::set_default` like in Python
Currently, there is no concise way to set a default value for a key in `Map[K,V]`. The `Map::get_or_default` method returns the default value without updating the map. This makes tasks like word counting more verbose.

I propose adding a `Map::set_default` method that updates the map with the default value and returns it if the key does not exist.

Alternatively, a `DefaultMap[K,V]` could be introduced to automatically create entries for missing keys. This would be particularly useful if it follows a standard `Map` trait.

KKKIIO:
Close #588 .

### DIFF
--- a/builtin/linked_hash_map.mbt
+++ b/builtin/linked_hash_map.mbt
@@ -193,6 +193,22 @@ pub fn get_or_default[K : Hash + Eq, V](
   }
 }
 
+/// Returns the value for the given key, or sets and returns a default value if the key does not exist.
+pub fn set_default[K : Hash + Eq, V](
+  self : Map[K, V],
+  key : K,
+  default : () -> V
+) -> V {
+  match self.get(key) {
+    Some(v) => v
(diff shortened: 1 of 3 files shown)


builtin/linked_hash_map.mbt
```diff
@@ -193,6 +193,22 @@ pub fn get_or_default[K : Hash + Eq, V](
   }
 }
 
+/// Returns the value for the given key, or sets and returns a default value if the key does not exist.
+pub fn set_default[K : Hash + Eq, V](
+  self : Map[K, V],
```
bobzhang, line 198: should it be called `get_or_init`?

KKKIIO: sure

PR #604, opened 2024-06-25
CI: add more test target

bzy-debug:


bzy-debug:
disable --deny-warn temporarily
cc @Yu-zh

PR #616, opened 2024-06-28
replace user defined UInt64 with builtin ones

bobzhang:


strconv/number.mbt
```diff
@@ -206,3 +208,30 @@ fn parse_inf_nan(s : String) -> (Double, Int)? {
   }
   None
 }
+
+fn lsl(a : UInt64, b : Int) -> UInt64 {
+  a.to_int64().lsl(b).to_uint64()
+}
```
bobzhang, line 214: @Yoorkin would UInt64 provide its own lsl?

Yoorkin: Yes, the bitwise operations for unsigned integers will be available on today's release.

strconv/number.mbt
```diff
@@ -206,3 +208,30 @@ fn parse_inf_nan(s : String) -> (Double, Int)? {
...
+  }
+  if a == 1UL {
+    return Some(b)
+  }
+  if b == 1UL {
+    return Some(a)
+  }
+  if b < 0UL || a < 0UL {
+    return None
+  }
```
peter-jerry-ye, line 230: This one here introduced a bug. As they used to be represented by `Int64`, comparing with 0 was effectively checking the highest bit. But now it became a useless check.

PR #625, opened 2024-06-29
feat: make Buffer::grow_if_necessary public

KKKIIO:
Making `Buffer::grow_if_necessary` public allows users to explicitly grow or pre-allocate memory, enabling various optimizations that can improve performance.

### DIFF
--- a/builtin/buffer.mbt
+++ b/builtin/buffer.mbt
@@ -75,6 +75,11 @@ pub fn grow_if_necessary(self : Buffer, required : Int) -> Unit {
   }
 }
 
+/// Returns the total number of bytes the Buffer can hold without reallocating.
+pub fn capacity(self : Buffer) -> Int {
+  self.bytes.length()
+}
+
 /// Return a new String contains the data in buffer.
 pub fn to_string(self : Buffer) -> String {
   self.bytes.sub_string(0, self.len)
--- a/builtin/buffer.mbt
+++ b/builtin/buffer.mbt
@@ -58,7 +58,7 @@ pub fn expect(
 }
 
 /// Expand the buffer size if capacity smaller than required space.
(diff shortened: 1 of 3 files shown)


bobzhang:
do you have a concrete use case for this API?

KKKIIO:
This API can be useful for implementing efficient encoding algorithms. For example, it can help encode a `String` in UTF-8 or write a list of `Int` values without the need for multiple memory reallocations.

qazxcdswe123:
> This API can be useful for implementing efficient encoding algorithms. For example, it can help encode a `String` in UTF-8 or write a list of `Int` values without the need for multiple memory reallocations.

If you already know the size in advance, why not set the `size_hint` in `Buffer::new`?

KKKIIO:
> > This API can be useful for implementing efficient encoding algorithms. For example, it can help encode a `String` in UTF-8 or write a list of `Int` values without the need for multiple memory reallocations.
> 
> If you already know the size in advance, why not set the `size_hint` in `Buffer::new`?

I may have misunderstood Moonbit's `Buffer`.

Initially, I thought it was similar to Golang's `bytes.Buffer`, which often reuses the same instance during a connection and calls `Reset` to reset the read/write position while retaining the storage before writing a new packet.

However, Moonbit's `Buffer::reset` discards any grown `Bytes` and reuses the `initial_bytes`, suggesting that a sufficiently large `size_hint` should be provided during initialization rather than relying on lazy growth. In this context, exposing `grow_if_necessary` seems unnecessary.

I'm curious about the benefits of this implementation. While it may reduce extra memory usage, it also incurs a penalty with each grow operation. If the maximum packet size can be limited, the entire 'grow' algorithm might be unnecessary.

Issue #632, opened 2024-06-30
Inappropriate parentheses removed


Lampese:
Before format:
```moonbit
test "panic op_get" {
  @deque.Deque::[1, 2, 3][-1] |> ignore
  (@deque.Deque::[1, 2, 3])[-1] |> ignore
}
```
then:
```moonbit
test "panic op_get" {
  @deque.Deque::[1, 2, 3][-1] |> ignore
  @deque.Deque::[1, 2, 3][-1] |> ignore
}
```
Based on the feedback I received on the coverage tool, the results obtained from these two writing methods are different, but the formatting tool removed parentheses for them.

qazxcdswe123:
Thank you for the report. This is an oversight, but we are deprecating the syntax `T::[]` so this shouldn't be an issue.

PR #633, opened 2024-06-30
tidy up deque

Lampese:
- Adding more tests has improved coverage
- Added the to_string function to implement Show trait
- Changed names that were not completely changed in the implementation
- Specifically distinguishing the location of black and white box testing
- Found a formatting issue #632


deque/test/test.mbt
```diff
@@ -197,10 +199,107 @@ test "iter2" {
...
+}
+
+test "panic pop_back_exn" {
+  @deque.from_array(([] : Array[Int])).pop_back_exn()
+}
+
+//Don't format heer!!! See Issue #632
+test "panic op_get" {
+  @deque.Deque::[1, 2, 3][-1] |> ignore
+}
```
bobzhang, line 226: This bug needs to be investigated (cc @Yoorkin )
For this specific use case, we are planning to deprecate the syntactic sugar to encourage user only use `@deque.of([1,2,3])`

deque/test/test.mbt
```diff
@@ -197,10 +199,107 @@ test "iter2" {
...
+  @assertion.assert_eq(dv.back(), Some(7))?
+  let dv = @deque.of([1])
+  @assertion.assert_eq(dv.pop_front(), Some(1))?
+  @assertion.assert_eq(dv.pop_front(), None)?
+  @assertion.assert_eq(dv.pop_back(), None)?
+}
+
+test "is_empty" {
+  let dv = @deque.Deque::new()
+  @assertion.assert_true(dv.is_empty())?
```
bobzhang, line 295: here we would encourage user just use `@deque.new()` instead.

bobzhang (approved):

PR #646, opened 2024-07-03
hongbo/one int max

bobzhang:
- use Int::min, Int::max, avoid duplication
- add Double::max
- tweak


qazxcdswe123:
I'd it's kinda confusing if we provide a MIN MAX for every type since they may also mean MAX_VALUE for each type. It's also not ergonomic given that we need to find the appropriate function in the first place

Issue #639, opened 2024-07-02
add more String method


lijunchen:
- [ ] trim
- [ ] to_upper
- [ ] split
- [ ] c style strcmp

qazxcdswe123:
What should we do with the current `compare` in `string.mbt` though.
It seems that it is not used anywhere else and this should probably align with other languages.

lijunchen:
> What should we do with the current `compare` in `string.mbt` though. It seems that it is not used anywhere else and this should probably align with other languages.

@bobzhang

Lampese:
- #993

PR #647, opened 2024-07-04
add string utils

yj-qin:
#639


string/string.mbt
```diff
@@ -118,36 +118,248 @@ pub fn iter(self : String) -> Iter[Char] {
...
+      } else {
+        buf.write_char(c)
+      }
+    },
+  )
+  buf.to_string()
+}
+
+/// Joins an array of string by a seperator.
+pub fn join(strs : Array[String], seperator : String) -> String {
```
qazxcdswe123, line 354: FYI: `join` had beed added in https://github.com/moonbitlang/core/pull/597/

qazxcdswe123 (changes requested):

qazxcdswe123:
I think it's better to use KMP search for `index_of`

I've implemented it myself and could you help verify it?

```
fn compute_prefix(pattern : String) -> Array[Int] {
  let m = pattern.length()
  let prefix = Array::make(m, 0)
  let mut k = 0
  for q = 1; q < m; q = q + 1 {
    while k > 0 && pattern[k] != pattern[q] {
      k = prefix[k - 1]
    }
    if pattern[k] == pattern[q] {
      k = k + 1
    }
    prefix[q] = k
  }
  return prefix
}

fn kmp_search(text : String, pattern : String) -> Int {
  let n = text.length()
  let m = pattern.length()
  if m == 0 {
    return 0
  } else if m > n {
    return -1
  }
  let prefix = compute_prefix(pattern)
  let mut q = 0
  for i = 0; i < n; i = i + 1 {
    while q > 0 && pattern[q] != text[i] {
      q = prefix[q - 1]
    }
    if pattern[q] == text[i] {
      q = q + 1
    }
    if q == m {
      return i - m + 1
    }
  }
  return -1
}
```

Thanks!

qazxcdswe123 (dismissed):

yj-qin:
I have implemented a basic brute force matching algorithm, which should be fast when the length of strings and substrings does not exceed a threshold. When the string is longer, it can be switched to KMP or Rabin-Karp algorithm. 
You can do an optimize in another PR.

FYR, Go's string index search:
https://github.com/golang/go/blob/master/src/internal/stringslite/strings.go#L24

@qazxcdswe123

string/string.mbt
```diff
@@ -118,36 +118,234 @@ pub fn iter(self : String) -> Iter[Char] {
   )
 }
 
+/// Removes the around whitespaces of this string.
+pub fn trim(self : String) -> String {
+  let len = self.length()
+  let mut start = 0
+  let mut end = len
+  while start < end && self[start] <= ' ' {
+    start += 1
```
bobzhang, line 127: can you add commment why `self[start] <= '  ' `works, does other language follow the same semantics

string/string.mbt
```diff
@@ -118,36 +118,234 @@ pub fn iter(self : String) -> Iter[Char] {
...
+  }
+  while start < end && self[end - 1] <= ' ' {
+    end -= 1
```
bobzhang, line 130: same here

string/string.mbt
```diff
@@ -118,36 +118,234 @@ pub fn iter(self : String) -> Iter[Char] {
...
+    if found {
+      return start + 1
+    }
+    i -= 1
+  }
+}
+
+/// Returns true if this string contains a sub string.
+pub fn contains(self : String, str : String) -> Bool {
+  self.index_of(str) > -1
```
bobzhang, line 236: ` >= 0`?

string/string.mbt
```diff
@@ -118,36 +118,234 @@ pub fn iter(self : String) -> Iter[Char] {
...
+    }
+    subs.push(self.substring(~start, ~end))
+    start = end + sep_len
+  }
+  subs
+}
+
+/// Replace the first old string in this string to new string.
+pub fn replace(self : String, old : String, new : String) -> String {
+  let old_idx = self.index_of(old)
```
bobzhang, line 282: use the label here? ` (self : String,  ~old: String, ~new : String)`

string/string.mbt
```diff
@@ -118,36 +118,234 @@ pub fn iter(self : String) -> Iter[Char] {
...
-        return i
+    self.substring(end=old_idx) + new + self.substring(
+      start=old_idx + old.length(),
+    )
+  }
+}
+
+/// Replace all old string in this string to new string.
+pub fn replace_all(self : String, old : String, new : String) -> String {
+  let buf = Buffer::new()
```
bobzhang, line 294: same here

string/string.mbt
```diff
@@ -118,36 +118,234 @@ pub fn iter(self : String) -> Iter[Char] {
...
+    }
+    -1
+  }
+}
+
+/// Returns the last index of the sub string.
+pub fn last_index_of(
+  self : String,
+  str : String,
+  ~from : Int = self.length()
```
bobzhang, line 189: should `last_index_of` and `index_of` be fused to a single API, e.g,
```
index_of(self : String, str : String, ~rev : Bool = true, ~from = if rev { self.length()} else {0})
```

cc @Guest0x0

bobzhang:
@qazxcdswe123 kmp could be provided via a new API

string/string.mbt
yj-qin: The Java JDK was implemented like this. I rethought it, maybe provide `trim`, `trim_left` and `trim_right` with a `trim_chars` parameter, and build `trim_space` on top of them, like Go's api.

PR #649, opened 2024-07-04
Adapt to abstract builtin array

hackwaly:


hackwaly:
This plan has dropped.

PR #653, opened 2024-07-05
add `push_iter` method to array

iningz:
This PR introduces the `push_iter` method to the `array`, allowing users to append all elements from an iterator to the end of an array.

### Changes
- Implemented the `push_iter` method in `array/array.mbt`.
- The method iterates over each element provided by the iterator and appends it to the array using the built-in `push` method.

### Example
```moonbit
let u = [1, 2, 3]
let v = [4, 5, 6]
u.push_iter(v.iter())
```

### DIFF
--- a/array/array.mbt
+++ b/array/array.mbt
@@ -26,6 +26,21 @@ pub fn iter[T](self : Array[T]) -> Iter[T] {
   )
 }
 
+/// Adds all elements from an iterator to the end of the array.
+///
+/// This function iterates over each element in the provided iterator
+/// and adds them to the array using the `push` method.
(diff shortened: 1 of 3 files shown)


hackwaly:
I think `push_all` feels better than `push_iter`

iningz:
> I think `push_all` feels better than `push_iter`

Yeah `push_all` is more in line with its goal

iningz:
`*_then` methods are omitted as we are going to support dot chaining. `push_iter` is decided to be the method name after disscussion.

bobzhang (approved):

PR #660, opened 2024-07-09
feat: debug for arrayview

qazxcdswe123:
close: https://github.com/moonbitlang/core/issues/395


builtin/debug.mbt
```diff
@@ -106,6 +106,17 @@ pub fn debug_write[X : Debug](self : Array[X], buf : Buffer) -> Unit {
   buf.write_char(']')
 }
 
+pub fn debug_write[X : Debug](self : ArrayView[X], buf : Buffer) -> Unit {
```
Yoorkin, line 109: also need to provide test

PR #664, opened 2024-07-09
Debug_write test

qazxcdswe123:


builtin/debug_test.mbt
```diff
@@ -12,71 +12,152 @@
...
+test "debug function with Map[Int, String]" {
+  let map = Map::new()
+  map.set(1, "one")
+  map.set(2, "two")
+  let buf = Buffer::new()
+  map.debug_write(buf)
+  inspect(
+    buf,
+    content=
+      #|(69681622,1,"one"),prev:_,next:-236984087)
```
qazxcdswe123, line 157: Should we change it to `(key: 69.., index: 1, content:"xx"),...`?

@Yoorkin

Yoorkin: we can make it align with the map literal: `{ key1: value1, key2: value2 }`

PR #671, opened 2024-07-10
tuple: add hash method

### LINKED ISSUE #551: add tuple hash support

yj-qin:
This PR adds some hash methods to the tuple using the xxHash32 algorithm.
Closes #551


qazxcdswe123:
I think we need a "source of truth" to verify the implementation, something like https://github.com/moonbitlang/core/blob/41826be11fd09bb63c983eb23f63b2a12fbbaf90/int64/xxhash_test.mbt#L15

We should probably also change the hash result to `uint`, for all hash functions, to avoid unnecessary `to_uint`

tuple/xxhash.mbt
```diff
@@ -0,0 +1,192 @@
...
+
+let gPRIME2 = 0x85EBCA77U
+
+let gPRIME3 = 0xC2B2AE3DU
+
+let gPRIME4 = 0x27D4EB2FU
+
+let gPRIME5 = 0x165667B1U
+
+pub fn hash[T0 : Hash, T1 : Hash](self : (T0, T1)) -> Int {
```
bobzhang, line 27: we are thinking of change the Hash trait into something like
```
trait Hash {
   hash_fold(Self, UInt) -> UInt
   hash(Self) -> Int
}
```
thoughts?

bobzhang: or maybe 
```
trait Hash {
  hash(Self, ~st : UInt = 0) -> UInt
}
```

yj-qin: So calculating the combination of hashes is making a recursive call to the hash function? That makes sense.

yj-qin: I think it makes sense. Or we could provide a hash combinator, like the [.NET HashCode class](https://github.com/dotnet/runtime/blob/1d1bf92fcf43aa6981804dc53c5174445069c9e4/src/libraries/System.Private.CoreLib/src/System/HashCode.cs)

bobzhang:
https://github.com/swiftlang/swift-evolution/blob/main/proposals/0206-hashable-enhancements.md

I am thinking of follow the same approach as Swift 4.2
```
trait Hash {
   hash_iter(Self, Hasher) -> Unit
   hash(Self) -> Int
}

impl Hash with hash(self) {
   let hasher = Hasher::new()
   hasher.hash_iter(self)
   hasher.finalize()
}
```

PR #674, opened 2024-07-11
internal: adapt /test to bbtest

Young-Flash:


bytes/moon.pkg.json
```diff
@@ -1,4 +1,13 @@
 {
-  "import": ["moonbitlang/core/builtin", "moonbitlang/core/coverage"],
-  "test-import": ["moonbitlang/core/test"]
-}
+  "import": [
+    "moonbitlang/core/builtin",
+    "moonbitlang/core/coverage"
+  ],
+  "test-import": [
```
bobzhang, line 6: if whitebox test is not used, shall we remove the field?

PR #686, opened 2024-07-14
hongbo/fix overflow

### LINKED ISSUE #672: `until` (number range) in `iter` have edge case
Currently the implementation looks like this

```
  Iter(
    fn(yield) {
      for i = self; i < end; i = i + step {
        if yield(i) == IterEnd {
          break IterEnd
        }
      } else {
        IterContinue
      }
    },
  )
```

And we cannot generate value like `2^31 - 1`, even if we change the `<` to `<=` since it will overflow and the condition holds true.

bobzhang:
- fix #672
- add test
@qazxcdswe123 can you check if this fix is correct and apply it to other functions, add tests for other cases too

### DIFF
--- a/builtin/iter.mbt
+++ b/builtin/iter.mbt
@@ -200,10 +200,15 @@ pub fn until(self : Int, end : Int, ~step : Int = 1) -> Iter[Int] {
   if step < 0 {
     return Iter(
       fn(yield) {
-        for i = self; i > end; i = i + step {
+        for i = self; i > end; {
+          let next_i = i + step
+          if next_i > i {
+            break IterEnd
+          }
           if yield(i) == IterEnd {
             break IterEnd
           }
+          continue next_i
         } else {
           IterContinue
         }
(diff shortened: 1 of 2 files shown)


qazxcdswe123:
Unless we make breaking changes to the interface (inclusive on both end), it won't generate `INT_MAX` since now it's exclusive on the `end` side.

something like (inclusive on both end)

```cpp
uint32_t bounded_rand(rng_t& rng, uint32_t range) {
    std::uniform_int_distribution<uint32_t> dist(0, range-1); // inclusive right end.

    return dist(rng);
}
```

qazxcdswe123:
Unless we make breaking changes to the interface (inclusive on both end), it won't generate `INT_MAX` since now it's exclusive on the `end` side.

something like (inclusive on both end)

```cpp
uint32_t bounded_rand(rng_t& rng, uint32_t range) {
    std::uniform_int_distribution<uint32_t> dist(0, range-1); // inclusive right end.

    return dist(rng);
}
```

bobzhang:
> Unless we make breaking changes to the interface (inclusive on both end), it won't generate INT_MAX since current implementation is exclusive on the end

Can you write a test and document your expected behavior?

qazxcdswe123:
> > Unless we make breaking changes to the interface (inclusive on both end), it won't generate INT_MAX since current implementation is exclusive on the end
> 
> Can you write a test and document your expected behavior?

```
// intention: generate [2147483646, 2147483647]
test "range overflow" {
  let v = Int::max_value() - 1
  inspect(v.until(Int::max_value()), content="[2147483646]")!
}
```

bobzhang:
`upto`, `downto`  (both inclusive []), `step = 0`

Issue #687, opened 2024-07-14
ambiguous debug output for `List`


skylee03:
The current dubug output cannot accurately display the nested structure of lists.

For example:

```moonbit
fn main {
  let ls = @immut/list.of([
    @immut/list.of([1, 2, 3]),
    @immut/list.of([4, 5, 6]),
    @immut/list.of([7, 8, 9]),
  ])
  debug(ls) // 1, 2, 3, 4, 5, 6, 7, 8, 9
  println(ls) // List::[List::[1, 2, 3], List::[4, 5, 6], List::[7, 8, 9]]
}
```

It is better to change the output of `debug_write` to a format similar to the one of `to_string`.

---

Related PRs:
- #716 
- #719

Yoorkin:
the output should be `[1,2,3,...,N]`

skylee03:
> The result of `List::debug_write` should be `[1,2,3,...,N]`

I'll try to update it, then I can update #688 accordingly.

skylee03:
I noticed that for other data structures, such as `Queue`, we used the result of `to_string` as the result of `debug_write`. Should we do the same for `List` to ensure consistency?

```moonbit
fn main {
  let q = @queue.of([1, 2, 3])
  println(q) // Queue::[1, 2, 3]
  debug(q) // Queue::[1, 2, 3]
}
```

qazxcdswe123:
Shouldn't the `debug_write` also show the type?

> The result of List::debug_write should be [1,2,3,...,N]

or `List::[1,2,3,...,N]`

@Yoorkin

Yoorkin:
> Should the `debug_write` also show the type?
> 
> > The result of List::debug_write should be [1,2,3,...,N]
> 
> or `List::[1,2,3,...,N]`
> 
> @Yoorkin

There will be some changes to the built-in `Debug` and `Show` traits, so we could fix these behaviors later.

skylee03:
The `Debug` trait has been removed and the output of `println` looks fine. Therefore, this issue can be considered fixed.

PR #688, opened 2024-07-14
docs: correct usage of `rev()` and sample output format of `List`

### LINKED ISSUE #510: about arr.reverse return value && doc in CONTRIBUTING.md Usage.
image

the code not work
```
$ moon new hello
$ cd hello
$ echo """fn main {
  println([1, 2, 3].reverse())
}""" > main/main.mbt
$ moon run main
# Output: [3, 2, 1]
```

the rignt way

```
$ moon new hello
$ cd hello
$ echo """fn main {
  let v = [1, 2, 3]
  v.reverse()
  println(v)
}""" > main/main.mbt
$ moon run main
# Output: [3, 2, 1]
```

skylee03:
In #656, the reverse functions for `Array`, `FixedArray` and `List` have been renamed, but the docs and comments have not been updated accordingly.

In #607, occurrences of `T::[_]` have been replaced with `T::of`. However, this should not include some sample outputs in the comments, since the format is determined by `debug_write` and `to_string`.

This PR aims to resolve both issues above. It also fixes #510.


immut/list/README.md
```diff
@@ -23,7 +23,7 @@ let list = Cons(1, Cons(2, Cons(3, Nil)))
 
 Build a repeated list by using the `repeat()` method:
 ```moonbit
-let list = repeat(3, 1) // of([1, 1, 1])
```
qazxcdswe123, line 26: This is actually correct since we've removed the sugar `T::[]` in favor of `of([])`

skylee03: We are not using any print or debug statements here. So should the format in the comment match `to_string`, `debug_write` or the actual construction?

As I mentioned in #687, we'd better unify the format.

skylee03:
Fixed in #724.

The adjustment of the sample output format in the comments is outside the scope of this PR. We need to wait until #687 is resolved.

Issue #689, opened 2024-07-15
Hash and deriving support


bobzhang:
relevant issue https://github.com/moonbitlang/core/pull/671

- [x] We provide a Hasher type in the builtin package
        It could be xxhasher, but it is behavior is implementation detail, subject to change for security upgrades later on
        the mbti would be look like this:

        type Hasher
        impl Hasher {
            new() -> Self
            combine_primitive( String|Int|Int64|Double..) -> Unit
            finalize() -> Int
        } 
        triat Hash {
          combine(Self, Hasher) -> Unit
          hash(Self) -> Int // default methods
          // default methods is needed so that we can provide optimal implementation for a specific type
          // when needed
        }
        
- [x] After above is implemented, add deriving support

PR #706, opened 2024-07-17
add xxhash32 hasher

yj-qin:
This PR implements a xxhash32 hasher follow the first step of #689 .


bobzhang:
@Young-Flash the failure is due to we are in the transition from whitebox test to blackbox test, can you help fix it?

PR #714, opened 2024-07-18
move string hasher into builtin

bobzhang:
I provided a default implementation for Hash trait and moved String:Hash into builtin, but found some errors
@yj-qin do you have any idea why?

### DIFF
--- a/builtin/builtin.mbti
+++ b/builtin/builtin.mbti
@@ -451,4 +451,7 @@ pub trait Show {
 }
 
 // Extension Methods
+impl Hash for String
+
+impl Hash for Hash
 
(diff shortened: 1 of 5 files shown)


yj-qin:
This branch has no error on my current toolchain, but some test results changed due to the change in the hash value of the string.
My toolchain version:
```
moon 0.1.20240717 (a22c580 2024-07-17) ~\.moon\bin\moon.exe
moonc v0.1.20240717+d22c317b1 ~\.moon\bin\moonc.exe
```

bobzhang:
@yj-qin you are right, since default method has higher precedence compared with method, so this diff may change the hash behavior for some type

PR #715, opened 2024-07-18
small diff to show the weird semantic diff

bobzhang:


Yoorkin:
This pr is ready now.

---

By the way, the behavior of the default implementation is confusing to me. For example:

```
trait MyTrait {
  f(Self) -> Int
  g(Self) -> Int
}

impl MyTrait with g(self) { 100 }

enum Foo { Foo }
impl MyTrait for Foo with f(self) { 1 }
impl MyTrait for Foo with g(self) { 2 }

enum Bar { Bar }
fn f(self : Bar) -> Int { 1 }
fn g(self : Bar) -> Int { 2 }

fn test_g[A : MyTrait](a : A) -> Int {
  a.g()
}

test {
  let foo = Foo 
  inspect!(test_g(foo), content="2")
  let bar = Bar  
  inspect!(test_g(bar), content="100") // expecting 2
}

```

@Guest0x0

PR #716, opened 2024-07-18
refine design of the [Show] trait

Guest0x0:
This PR refine the design of the `Show` trait.

## Problem with the previous design
Consider the problem of `Show` implementation for `String`: should it add quotes and escapes?

- For interpolation `"... \(str) ..."` and `println(str)` where `str : String`, we want `str` itself to be interpolated/printed as-is
- Whenever a `String` occur in nested structure, such as `Ref[String]`, it should be quoted for a sensible result

Previously, the `Show` trait has only a `to_string` method, which is not enough to produce the correct behavior above. Furthermore, combining `to_string` with string concatenation is inefficient.

## The proposed design here
The new `Show` trait in this PR is:
```rust
trait Show {
  output(Self, Logger) -> Unit
  to_string(Self) -> String
}
```
Here `output` is for composition, it write a string representation of `Self` to a `Logger` (which is a trait object for logging stuffs). This allows efficient composition without intermediate allocation. End users of the `Show` trait can still use the familiar `to_string` method.

The method `to_string` has a default implementation using `output`, so library authors only need to implement the `output` method to implement the `Show` trait.

`String` overrides the default `to_string` in its implementation of `Show` and returns the string itself. In the mean time, `String`'s implementation for `output`, which is used for composition, will escape and quote the string. So when a `String` occurs inside nested structure, it is properly quoted and escaped. When a `String` is used directly for `println` and interpolation, the identity `to_string` will be used.

### DIFF
--- a/array/array_wbtest.mbt
+++ b/array/array_wbtest.mbt
@@ -143,10 +143,20 @@ test "map" {
   let v = [3, 4, 5]
   let e : Array[Int] = []
   inspect!(v.map(fn(x) { x + 1 }), content="[4, 5, 6]")
-  inspect!(v.map(fn(x) { x.to_string() }), content="[3, 4, 5]")
+  inspect!(
+    v.map(fn(x) { x.to_string() }),
+    content=
+      #|["3", "4", "5"]
+    ,
+  )
   inspect!(e.map(fn(x) { x + 1 }), content="[]")
   // Is it correct?
-  inspect!(["a", "b", "c"], content="[a, b, c]")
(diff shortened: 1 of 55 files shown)


Yoorkin:
@qazxcdswe123 We need to fix the implementation of `to_string` and `output` after this PR is merged. As described in the description of the PR, the result of `debug_write` should be the same as `to_string` for all types except `String` and `Char`, but currently, it's misused to output the internal data of collections.

Issue #669, opened 2024-07-10
iterator/loop utils function for ArrayView


Yoorkin:
- ArrayView::iter
- Do we need `fold_left` `fold_right` and `each` for ArrayView?

A real world example:

```
pub fn separate(sep : Document, docs : Array[Document]) -> Document {
  let cat_by_sep = fn(l,r) { concat(l,concat(sep, r)) }
  match docs {
    [] => empty()
    [x,.. as xs] => xs.fold_left(cat_by_sep, init=x) // fail, the type of xs is ArrayView[_] 
  }
}
```

PR #726, opened 2024-07-21
feat: add each for arrayview

QizhengMo:
I do find each useful for ArrayView as memtioned in https://github.com/moonbitlang/core/issues/669

And I noticed there was a rename for FixedArray::iter related functions, but the tests were not renamed. Fixed in the PR as well

### DIFF
--- a/array/array.mbti
+++ b/array/array.mbti
@@ -49,6 +49,7 @@ impl Array {
 }
 impl ArrayView {
   rev_inplace[T](Self[T]) -> Unit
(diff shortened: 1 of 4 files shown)


peter-jerry-ye:
Thank you for your contribution. Please use `moon info` command to regenerate the `mbti` file.

QizhengMo:
> Thank you for your contribution. Please use `moon info` command to regenerate the `mbti` file.

Done! Thanks for your quick review.

PR #749, opened 2024-07-24
get rid of array spread in builtin

Guest0x0:
Desugaring of array spread `[ a, ..xs, b ]` requires `iter` method and `push_iter` method of `Array`, which are in the `array` package, not `builtin`. So using array spread in `builtin` is not stable w.r.t. location of API & desugaring of array spread.

This PR removes unnecessary usage of array spread in builtin package.


bobzhang:
@Guest0x0 this is ok, since `test` package is not belonging to builtin

PR #763, opened 2024-07-26
various misc improvements on the [Iter] data type:

Guest0x0:
- avoid some unncessary closure in [Iter::new]
- reduce the indentation level of code by removing some unnecessary use of matrix function
- deprecate the [Iter::tap] method: we now have native cascade operator


Yoorkin (approved):
LGTM

PR #765, opened 2024-07-26
add [each_err] & [fold_err] API to [Iter]

Guest0x0:
Add two APIs, `each_err` and `fold_err` to the `Iter` type. These two APIs are similar to `each` and `fold`, except that the callback may raise an error. If the callback raises any error, the iteration stops immediately and raises the same error.

### DIFF
--- a/builtin/iter.mbt
+++ b/builtin/iter.mbt
@@ -75,6 +75,40 @@ pub fn eachi[T](self : Iter[T], f : (Int, T) -> Unit) -> Unit {
   |> ignore
 }
 
+/// Iterates over each element in the iterator, applying the function `f` to each element.
+/// If `f` raises any error, `each_err` will terminate immediately and raise the same error.
+///
+/// # Type Parameters
+///
+/// - `T`: The type of the elements in the iterator.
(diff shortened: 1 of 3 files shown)


builtin/iter.mbt
```diff
@@ -75,6 +75,40 @@ pub fn eachi[T](self : Iter[T], f : (Int, T) -> Unit) -> Unit {
...
+/// - `T`: The type of the elements in the iterator.
+/// - `Err`: The type of the error that `f` may raise.
+///
+/// # Arguments
+///
+/// - `self`: The iterator to consume.
+/// - `f`: A function that takes an element of type `T` and returns `Unit`.
+///    This function is applied to each element of the iterator and may raise error of type `Err`.
+pub fn each_err[T, Err](self : Iter[T], f : (T) -> Unit!Err) -> Unit!Err {
+  let mut err = None
```
bobzhang, line 92: I think eventually, we should just have only one `each` function which accept the error callback, the optimizer should be able to generate as efficient code for callbacks without errors
cc @Yu-zh

PR #768, opened 2024-07-29
feat: add u64 to f64 intrinsic

qazxcdswe123:
~~In the intrinsic test, wasm/wasmgc and js yields different result. And the correct one should be 1.0~~

~~Coule this be a issue in wasm or ryu?~~

### DIFF
--- a/builtin/int64.js.mbt
+++ b/builtin/int64.js.mbt
@@ -335,6 +335,9 @@ extern "js" fn MyInt64::trunc_double_u(value : Double) -> MyInt64 =
 
 extern "js" fn MyInt64::convert_to_double_u(self : MyInt64) -> Double =
   #|(a) => (a.hi >>> 0) * 4294967296.0 + (a.lo >>> 0)
+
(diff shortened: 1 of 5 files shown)


builtin/int64.js.mbt
```diff
@@ -497,6 +497,10 @@ pub fn UInt64::to_int(self : UInt64) -> Int {
   MyInt64::from_uint64(self).to_int()
 }
 
+pub fn UInt64::to_double(self : UInt64) -> Double {
+  MyInt64::from_uint64(self).to_double()
```
Yoorkin, line 501: This implementation is not correct. Like the implementations of `UInt64::compare` and `Int64::compare`, the compare operations are different for signed and unsigned integer. In webassembly, there also exists two instructions `i64.trunc_f64_s`/`i64.trunc_f64_u` (which the wasm backend currently uses) to convert double to i64

Yoorkin:
LGTM, except the `MyInt64::convert_to_double` part

hackwaly (approved):

PR #794, opened 2024-08-01
[POC, DON'T MERGE] use error to implement early return of `Iter`

Guest0x0:
This PR change the internal representation of `Iter` from `(() -> IterResult) -> IterResult` to `(() -> Unit!Error) -> Unit!Error`. This brings several benefits:

-  APIs such as `each_err` and `fold_err` are easier to implement
-  It is possible to distinguish different sources of early return in an `Iter`, via different error constructors. For example the implementation of `take` and `takeWhile` can be simplified
- `!` makes writing `Iter`s a lot simpler

### DIFF
--- a/builtin/iter.mbt
+++ b/builtin/iter.mbt
@@ -13,17 +13,14 @@
 // limitations under the License.
 
 // Types
-type Iter[T] ((T) -> IterResult) -> IterResult
+type Iter[T] ((T) -> Unit!Error) -> Unit!Error
 
 //TODO: Add intrinsic for Iter::run
-pub fn run[T](self : Iter[T], f : (T) -> IterResult) -> IterResult {
-  (self.0)(f)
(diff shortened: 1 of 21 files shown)


builtin/iter.mbt
```diff
@@ -107,13 +118,13 @@ pub fn eachi[T](self : Iter[T], f : (Int, T) -> Unit) -> Unit {
...
-      IterContinue
-    },
-  )
-  acc
+  try {
+    self.run!(fn { x => acc = f(acc, x) })
+  } catch {
+    _ => abort("unhandled error in iter")
+  } else {
+    _ => acc
```
bobzhang, line 126: this looks hard to optimize

bobzhang (commented):
the readability is significantly improved. Note if that's the main benefit,
IterResult ~~ Unit!IterBreak may also work

PR #802, opened 2024-08-02
add Iter2 to prepare for foreach sugar

bobzhang:


bobzhang:
cc @Guest0x0

bobzhang:
cc @lynzrand do you have any idea why iter2.mbt is red in the coverage report, I am pretty sure I added some test for that file

lynzrand:
> cc @lynzrand do you have any idea why iter2.mbt is red in the coverage report, I am pretty sure I added some test for that file

Investigating 👀

lynzrand:
> cc @lynzrand do you have any idea why iter2.mbt is red in the coverage report, I am pretty sure I added some test for that file

The issue was caused by black box tests of `builtin` not considered to be part of the project itself, thus not using `builtin` itself as coverage provider when compiling `builtin`. A fix is under way.

PR #817, opened 2024-08-06
fix `json.stringify` escape

### LINKED ISSUE #812: json.stringify does not handle the escaped characters
details #811

yj-qin:
fix #812

### DIFF
--- a/json/json.mbt
+++ b/json/json.mbt
@@ -122,19 +122,35 @@ pub fn stringify(self : JsonValue) -> String {
 }
 
 fn escape(str : String) -> String {
-  let buf = Buffer::new()
+  fn to_hex_digit(i : Int) -> Char {
+    if i < 10 {
+      Char::from_int('0'.to_int() + i)
+    } else {
+      Char::from_int('a'.to_int() + (i - 10))
+    }
+  }
+
+  let buf = Buffer::new(size_hint=str.length())
   for i = 0; i < str.length(); i = i + 1 {
     let c = str[i]
     match c {
-      '"' | '\\' | '/' => {
-        buf.write_char('\\')
-        buf.write_char(c)
-      }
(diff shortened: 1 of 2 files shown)


json/json.mbt
```diff
@@ -121,6 +121,25 @@ pub fn stringify(self : JsonValue) -> String {
   buf.to_string()
 }
 
+fn escape(str : String) -> String {
+  let buf = Buffer::new()
```
qazxcdswe123, line 125: set `size_hint` to `str.length()` to avoid some reallocation?

json/json.mbt
```diff
@@ -121,6 +121,25 @@ pub fn stringify(self : JsonValue) -> String {
...
+  for i = 0; i < str.length(); i = i + 1 {
+    let c = str[i]
+    match c {
+      '"' | '\\' | '/' => {
+        buf.write_char('\\')
```
qazxcdswe123, line 130: Maybe use string interpolation `\\\{c}`

```
'"' | '\\' | '/' => buf.write_string("\\\{c}")
```

json/json.mbt
```diff
@@ -121,6 +121,25 @@ pub fn stringify(self : JsonValue) -> String {
...
+        buf.write_char(c)
+      }
+      '\n' => buf.write_string("\\n")
+      '\r' => buf.write_string("\\r")
+      '\b' => buf.write_string("\\b")
+      '\t' => buf.write_string("\\t")
```
qazxcdswe123, line 136: https://stackoverflow.com/questions/19176024/how-to-escape-special-characters-in-building-a-json-string

According to this.
Probably missing `\f` and `\u`?

yj-qin: Thanks!

CAIMEOX:
Still fails at the following case:
```moonbit
let json : @json.JsonValue = { "\x0c": True }
```

PR #818, opened 2024-08-06
fix sorted_set iter

bobzhang:


sorted_set/set.mbt
```diff
@@ -337,13 +337,31 @@ pub fn to_array[V](self : T[V]) -> Array[V] {
...
+
+fn iter[A](self : Node[A]) -> Iter[A] {
+  Iter::new(
+    fn(yield) {
+      let { left, value, right, .. } = self
+      let res = match left {
+        None => IterContinue
+        Some(l) => l.iter().run(yield)
+      }
+      if res == IterEnd {
```
bobzhang, line 356: use bool we could write it as 
```
res && yield(value) && (
  match right {
     None => true
     Some(r) => r.iter().run(yield) 
  } 
) 
```

PR #820, opened 2024-08-07
move json type to builtin

Guest0x0:
Move the `@json.JsonValue` type to builtin package, so that:

- it can be referenced more easily with `Json` directly
- builtin types can have a `.to_json` method

Note that we don't want to pull all the json parsing and processing stuff into builtin, just the type definition. To make this work some compiler special treatment is needed, so that `@json` can define methods for the `Json` type. So this PR need compiler adoption before merging.

A type alias is placed in the `@json` package, so this PR is backward compatible. All existing code that make use of `@json.JsonValue` will continue to work.


Guest0x0:
The most-recent compiler release also come with some formatter behavior changes, and those changes must be shipped in this PR, otherwise CI will break

PR #828, opened 2024-08-08
feat: add inclusive for iter until

qazxcdswe123:
Added `inclusive` function for later `..=` and `..<`

### DIFF
--- a/builtin/builtin.mbti
+++ b/builtin/builtin.mbti
@@ -365,6 +365,7 @@ impl Int {
   to_string(Int) -> String
   to_uint(Int) -> UInt
   until(Int, Int, ~step : Int = ..) -> Iter[Int]
+  until_inclusive(Int, Int, ~step : Int = ..) -> Iter[Int]
 }
 impl Int64 {
   asr(Int64, Int) -> Int64
@@ -399,6 +400,7 @@ impl Int64 {
   to_string(Int64) -> String
   to_uint64(Int64) -> UInt64
   until(Int64, Int64, ~step : Int64 = ..) -> Iter[Int64]
+  until_inclusive(Int64, Int64, ~step : Int64 = ..) -> Iter[Int64]
 }
 impl UInt {
   compare(UInt, UInt) -> Int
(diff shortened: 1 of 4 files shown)


builtin/iter_test.mbt
```diff
@@ -259,6 +259,38 @@ test "range" {
...
+  )
+  inspect!(
+    (@int64.max_value() - 2).until(@int64.max_value() - 1L, inclusive=true),
+    content="[9223372036854775805, 9223372036854775806]",
+  )
+  inspect!(
+    (@int64.max_value() - 1).until(@int64.max_value(), inclusive=true),
+    content="[9223372036854775806, 9223372036854775807]",
+  )
+  inspect!(
```
qazxcdswe123, line 278: Since the accuracy is not enough, the inclusive param doesn't affect the result here.

builtin/iter.mbt
```diff
@@ -226,6 +226,40 @@ pub fn until(self : Int, end : Int, ~step : Int = 1) -> Iter[Int] {
...
+///
+/// # Returns
+///
+/// Returns an iterator that iterates over the range of Int from `start` to `end`.
+pub fn until_inclusive(self : Int, end : Int, ~step : Int = 1) -> Iter[Int] {
+  if step == 0 {
+    return Iter::empty()
+  }
+  fn(yield) {
+    let mut i = self
```
bobzhang, line 246:
```
for i in a ..< b {
}
==>
    for i a.upto(b)
for i in a ..= b {
} 
==>
    for i in a.upto(b, inclusive=true)
```

builtin/builtin.mbti
```diff
@@ -360,7 +360,8 @@ impl Int {
   to_json(Int) -> Json
   to_string(Int) -> String
   to_uint(Int) -> UInt
-  until(Int, Int, ~step : Int = ..) -> Iter[Int]
+  until(Int, Int, ~step : Int = .., ~inclusive : Bool = ..) -> Iter[Int]
```
qazxcdswe123, line 363: This is too repetitive though, they serve the same purpose (default `step` for `until` is 1, behave the same as `upto`). I think we can keep only `until` and desugar to it.

qazxcdswe123: The compiler should work just fine BTW

PR #829, opened 2024-08-08
add [ToJson] trait & [to_json] method for builtin types

Guest0x0:
- add `to_json` method to builtin types
- add `ToJson` trait, so that the compiler can automatically `derive(ToJson)` in the future

### DIFF
--- a/builtin/json.mbt
+++ b/builtin/json.mbt
@@ -21,3 +21,46 @@ pub enum Json {
   Array(Array[Json])
   Object(Map[String, Json])
 } derive(Eq)
+
+/// Trait for types that can be converted to `Json`
+pub trait ToJson {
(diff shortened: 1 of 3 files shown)


builtin/json.mbt
```diff
@@ -21,3 +21,46 @@ pub enum Json {
...
 } derive(Eq)
+
+/// Trait for types that can be converted to `Json`
+pub trait ToJson {
+  to_json(Self) -> Json
+}
+
+pub fn Bool::to_json(self : Bool) -> Json {
+  if self {
+    True
```
bobzhang, line 32: I am thinking of we should rename True into JsonTrue to avoid name pollution, note JsonTrue is used much less after we introduce first class json support

PR #834, opened 2024-08-09
add flat_map to list, deprecate concat_map

Yoorkin:
- add `immut/list.flat_map` to replace `concat_map`


hackwaly:
Is keep concat_map and deprecate it in favor of flat_map a better option? It could serve as a documentation as a guide to use flat_map.

As my experience, sometimes I can't recall flat_map and just got concat_map in my mind.

PR #835, opened 2024-08-09
add assert_not_eq, assert_false, assert_true to built-in

Yoorkin:
- Move some frequently used functions from the `test` package to built-in.
- Mark old API deprecated
- Fix tests


builtin/builtin.mbti
```diff
@@ -5,6 +5,12 @@ fn abort[T](String) -> T
 
 fn assert_eq[T : Eq + Show](T, T, ~loc : SourceLoc = _) -> Unit!Failure
 
+fn assert_false(Bool, ~loc : SourceLoc = _) -> Unit!Failure
+
+fn assert_ne[T : Eq + Show](T, T, ~loc : SourceLoc = _) -> Unit!Failure
+
```
bobzhang, line 11: assert_not_eq?

bobzhang: we don't care about the error type, can we make it a bit more simpler to use, i.e,

fn assert_ne[T : Eq + Show](T, T, ~loc : SourceLoc = _) -> Unit!

bobzhang (approved):

PR #863, opened 2024-08-16
refactor: use existing string escape

qazxcdswe123:
Does the updated output make sense? @CAIMEOX
It's technically more "correct" but not as concise as before.

### DIFF
--- a/builtin/json.mbt
+++ b/builtin/json.mbt
@@ -47,42 +47,8 @@ pub fn Double::to_json(self : Double) -> Json {
   Number(self)
 }
 
-fn escape_json_string(str : String) -> String {
-  fn to_hex_digit(i : Int) -> Char {
-    if i < 10 {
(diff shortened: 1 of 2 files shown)


builtin/json.mbt
```diff
@@ -47,42 +47,8 @@ pub fn Double::to_json(self : Double) -> Json {
...
-        }
-      }
-    }
-  }
-  buf.to_string()
-}
-
 pub fn String::to_json(self : String) -> Json {
-  String(escape_json_string(self))
+  String(self.escape())
```
bobzhang, line 51: cc @hackwaly I remember this is not the same?

hackwaly: Yes, Json escape is standardized, and our String::escape isn't. I suggest that keep the original version.

hackwaly: I'm not sure if the original version follow the standard. Maybe we should add more tests here.

PR #865, opened 2024-08-16
feat: better prng

### LINKED ISSUE #879: Merge core/random and x/rand
I would like to merge x/rand with core/random, but these two packages doesn't seem to share the same underlying implementation of PRNG.

I would also like to see more support for random in core, e.g.

1. Generate `Int/Int64/Float/Double` with range.
2. More random operation on array:
   1. `sample[T](Array[T], ~k: Int) -> Array[T]`
   2. `choice[T](Array[T]) -> T`

qazxcdswe123:
This PR introduced new PRNG, with output exactly the same as go's `math/rand/v2`.

It is also cryptographically secure, but comes with a more strict interface for seeding.

close #879


random/rand.mbt
```diff
@@ -0,0 +1,234 @@
...
+
+/// Currently we only support [chacha8] as the source of randomness.
+struct Rand {
+  src : ChaCha8
+}
+
+/// Create a new random number generator with [seed].
+/// @alert unsafe "Panic if seed is not 32 bytes long"
+pub fn Rand::new(
+  ~seed : Bytes = string_to_bytes("ABCDEFGHIJKLMNOPQRSTUVWXYZ123456")
```
qazxcdswe123, line 23: Go's rand is automatically seeded by its runtime, which is not supported by moonbit yet. This interface is somewhat awkward.

random/rand.mbti
```diff
@@ -0,0 +1,28 @@
+package moonbitlang/x/rand
+
+// Values
+
+// Types and methods
+type ChaCha8
+
+type Rand
```
bobzhang, line 8: please remove this file

random/random.mbti
```diff
@@ -3,13 +3,21 @@ package moonbitlang/core/random
...
 
 // Types and methods
-type RandomState
-impl RandomState {
-  gen_double(Self) -> Double
-  gen_int(Self) -> Int
-  gen_int64(Self) -> Int64
-  init_state(~seed : Int = ..) -> Self
-  new() -> Self
+type ChaCha8
```
bobzhang, line 6: why this is exposed

qazxcdswe123: It was removed in x , but out of sync here. Sorry

random/rand_source_chacha.mbt
```diff
@@ -0,0 +1,469 @@
+// Copyright 2024 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
```
bobzhang, line 3: is it easy to move it into random/internal/source_chacha?

random/rand.mbt
```diff
@@ -0,0 +1,232 @@
...
+  src : ChaCha8
+}
+
+/// Create a new random number generator with [seed].
+/// @alert unsafe "Panic if seed is not 32 bytes long"
+pub fn Rand::new(~seed : Bytes = b"ABCDEFGHIJKLMNOPQRSTUVWXYZ123456") -> Rand {
+  if seed.length() != 32 {
+    abort("seed must be 32 bytes long")
+  }
+  let src = ChaCha8::new(seed)
```
bobzhang, line 26: no need panic? you can always make it length 32,right?

random/rand.mbt
```diff
@@ -0,0 +1,232 @@
...
+  assert_eq!(n, exp)
+}
+
+/// [int] Return a non-negative pseudo-random 31-bit integer as an Int.
+pub fn Rand::int(self : Rand) -> Int {
+  self.next().lsr(33).to_int()
+}
+
+/// [int64] returns a non-negative pseudo-random 63-bit integer as an Int64
+pub fn Rand::int64(self : Rand) -> Int64 {
```
bobzhang, line 53: can you document its range precisely?

random/rand.mbt
```diff
@@ -0,0 +1,232 @@
...
+}
+
+test "next" {
+  let r = Rand::new()
+  let n = r.next()
+  let exp = 13219109469176600229UL
+  assert_eq!(n, exp)
+}
+
+/// [int] Return a non-negative pseudo-random 31-bit integer as an Int.
```
bobzhang, line 47: precise range

random/random.mbti
```diff
@@ -3,13 +3,19 @@ package moonbitlang/core/random
...
-type RandomState
-impl RandomState {
-  gen_double(Self) -> Double
-  gen_int(Self) -> Int
-  gen_int64(Self) -> Int64
-  init_state(~seed : Int = ..) -> Self
-  new() -> Self
+type Rand
+impl Rand {
+  double64(Self) -> Double
```
bobzhang, line 8: just double?

random/random.mbti
```diff
@@ -3,13 +3,19 @@ package moonbitlang/core/random
...
+  int(Self) -> Int
+  int64(Self) -> Int64
+  int64N(Self, Int64) -> Int64
+  intN(Self, Int) -> Int
+  new(~seed : Bytes = ..) -> Self
```
bobzhang, line 13: what does `intN` `N` mean here, can you give it a more meaningful name?

random/rand.mbt
qazxcdswe123: Yes it is possible, but another benefits for using chacha as random source is that it is cryptographically secure.

If we just pad it to 32, it no longer hold this property.

random/random.mbti
qazxcdswe123: I tried to add float but float show is not implemented and thus cannot be tested, so I git stashed it and forgot..

random/random.mbti
qazxcdswe123: `N` means range, thats what go's API looks like.

Probably rename to `int_range`?

random/rand_source_chacha.mbt
qazxcdswe123: Done

random/random.mbti
```diff
@@ -3,13 +3,20 @@ package moonbitlang/core/random
...
-  init_state(~seed : Int = ..) -> Self
-  new() -> Self
+type Rand
+impl Rand {
+  double64(Self) -> Double
+  float32(Self) -> Float
+  int(Self) -> Int
+  int64(Self) -> Int64
+  int64N(Self, Int64) -> Int64
+  intN(Self, Int) -> Int
```
tonyfettes, line 13: Use optional argument for `intN`? for example `int(Self, ~limit?:Int)`

tonyfettes (changes requested):

random/random.mbti
bobzhang: uuml avoid allocation -- imperative API

tonyfettes (approved):
LGMT

PR #866, opened 2024-08-16
add: a new method String::rev()

iamzhz:
I have added a `reverse()` method to `String`. 

I think the reverse() method has some good points. First, it makes the String class more useful and easier to work with. It makes reversing a string simple, which is needed often in programming.   
I have tested it in my computer. I can run correctly.
And you can using it like this:
```
fn main {
  let x: String = "hello world";
  println(x.reverse())
}
```
It will display `dlrow olleh`

### DIFF
--- a/string/string.mbt
+++ b/string/string.mbt
@@ -441,3 +441,14 @@ pub fn to_upper(self : String) -> String {
   }
   buf.to_string()
 }
+
+/// Reverse string
+pub fn reverse(self : String) -> String {
+  // TODO: deal with non-ascii characters
+  let mut result : String = ""
+  let len = self.length()
+  for i = len - 1; i >= 0; i = i - 1 {
+    result = result + self[i].to_string()
+  }
+  result
+}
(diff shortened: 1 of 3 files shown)


iamzhz:
> LGTM. Could you rename `String::reverse()` to `String::rev()`, to make it look consistent with other collections? (`Array::rev()`, `List::rev()`)

I did it just now.

string/string.mbt
```diff
@@ -441,3 +441,14 @@ pub fn to_upper(self : String) -> String {
...
+/// Reverse string
+pub fn rev(self : String) -> String {
+  // TODO: deal with non-ascii characters
+  let mut result : String = ""
+  let len = self.length()
+  for i = len - 1; i >= 0; i = i - 1 {
+    result = result + self[i].to_string()
+  }
+  result
+}
```
bobzhang, line 454: note the function is very in-efficient

string/string_test.mbt
```diff
@@ -348,6 +348,12 @@ test "replace_all" {
   )
 }
 
+test "reverse" {
+  inspect!("hello".rev(), content="olleh")
+  inspect!("Nothing".rev(), content="gnihtoN")
+  inspect!("< >".rev(), content="> <")
```
bobzhang, line 354: we should at least add some unicode/emoji test

PR #882, opened 2024-08-21
fix(json): correct integer stringify

### LINKED ISSUE #878: Feature request: Int.to_json().stringify() should not add trailing ".0"
When using this version of moon:

```bash
$ moon version --all
moon 0.1.20240819 (284058b 2024-08-19) ~/.moon/bin/moon
moonc v0.1.20240820+85e9a0dc8 ~/.moon/bin/moonc
moonrun 0.1.20240820 (ecf5abc 2024-08-20) ~/.moon/bin/moonrun
```

if I have a struct like this:

```rust
/// `VowelReport` represents the JSON struct returned to the host.
pub struct VowelReport {
  count : Int
  total : Int
  vowels : String
} derive(Show, Eq)

pub fn to_json(self : VowelReport) -> Json {
  {
    "count": self.count.to_json(),
    "total": self.total.to_json(),
    "vowels": self.vowels.to_json(),
  }
}
```

and I want to `stringify` it, the `Int`s have trailing `.0` which is confusing and misleading.
It would be nice to render `Int`s as integers in the resulting JSON.

Here is the test that currently fails (but should pass) as a demonstration:

```rust
test "to_json - this test fails but should pass" {
  let got = { count: 1, total: 2, vowels: "some string" }.to_json().stringify()
  let want =
    #|{"count":1,"total":2,"vowels":"some string"}
  assert_eq!(got, want)
}
```

Here is the test that currently passes (but should fail) as a demonstration:

```rust
test "to_json - this test passes but should fail" {
  let got = { count: 1, total: 2, vowels: "some string" }.to_json().stringify()
  let want =
    #|{"count":1.0,"total":2.0,"vowels":"some string"}
  assert_eq!(got, want)
}
```

yj-qin:
Fix #878

### DIFF
--- a/json/json.mbt
+++ b/json/json.mbt
@@ -113,12 +113,14 @@ pub fn stringify(self : JsonValue) -> String {
       buf.write_string(escape(s))
       buf.write_char('\"')
     }
-    Number(n) =>
-      return if n == n.to_int64().to_double() {
-        n.to_int64().to_string()
+    Number(n) => {
+      let int_n = n.to_int64()
+      return if n == int_n.to_double() {
+        int_n.to_string()
       } else {
         n.to_string()
       }
+    }
     True => return "true"
     False => return "false"
     Null => return "null"
--- a/json/json.mbt
+++ b/json/json.mbt
@@ -113,7 +113,12 @@ pub fn stringify(self : JsonValue) -> String {
       buf.write_string(escape(s))
       buf.write_char('\"')
     }
-    Number(n) => return n.to_string()
+    Number(n) =>
+      return if n == n.to_int64().to_double() {
+        n.to_int64().to_string()
+      } else {
+        n.to_string()
+      }
     True => return "true"
     False => return "false"
     Null => return "null"
--- a/json/json.mbt
+++ b/json/json.mbt
@@ -114,11 +114,14 @@ pub fn stringify(self : JsonValue) -> String {
       buf.write_char('\"')
     }
     Number(n) => {
-      let int_n = n.to_int64()
-      return if n == int_n.to_double() {
-        int_n.to_string()
+      if n == 0 {
+        return "0"
+      }
+      let s = n.to_string()
+      return if n.trunc() == n && s.ends_with(".0") {
+        s.substring(end=s.length() - 2)
       } else {
-        n.to_string()
+        s
       }
     }
     True => return "true"
--- a/json/json_test.mbt
+++ b/json/json_test.mbt
@@ -317,6 +317,7 @@ test "stringify number" {
   let nums : Array[Json] = [
     0.0,
     -0.0,
+    1.0,
     12345,
     123.45,
     9223372036854775807,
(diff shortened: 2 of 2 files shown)


Yoorkin:
LGTM

qazxcdswe123:
"edge double": 1.9999999999999999,
    
```moonbit
  let json : Json = {
    "key": [1.0, true, Null, [], { "key": "value", "value": 100.0 }],
    "int": 12345,
    "double": 123.45,
    "edge double": 1.9999999999999999,
    "null": Null,
    "bool": false,
    "obj": {  },
  }
  inspect!(
    json.stringify(),
    content=
      #|{"key":[1,true,null,[],{"key":"value","value":100}],"int":12345,"double":123.45,"edge double":2,"null":null,"bool":false,"obj":{}}
    ,
  )
```

yj-qin:
I think the first case is normal, and it behaves like JS's `JSON.stringify`. Double::to_string has the same result.
```
(1.9999999999999999).to_string() // "2.0"
```

Yoorkin:
> ```
> "edge double": 1.9999999999999999,
> ```
> 
> ```
>   let json : Json = {
>     "key": [1.0, true, Null, [], { "key": "value", "value": 100.0 }],
>     "int": 12345,
>     "double": 123.45,
>     "edge double": 1.9999999999999999,
>     "null": Null,
>     "bool": false,
>     "obj": {  },
>   }
>   inspect!(
>     json.stringify(),
>     content=
>       #|{"key":[1,true,null,[],{"key":"value","value":100}],"int":12345,"double":123.45,"edge double":2,"null":null,"bool":false,"obj":{}}
>     ,
>   )
> ```
> 
> Note that the original implementation yield fatal error without stacktrace. But should be handled anyway?
> 
> ```
> "edge double2": @double.inf(1).to_json(),
> "edge double2": @double.nan().to_json(),
> ```
> 
> ```
> RuntimeError: float unrepresentable in integer range
>     at Json::stringify.fn/435 (wasm://wasm/001cb4f2:wasm-function[603]:0xfdd3)
>     at Json::stringify.fn/604 (wasm://wasm/001cb4f2:wasm-function[604]:0xfe4c)
>     at Map::eachi|String+@moonbitlang/core/builtin.Json|.fn/216 (wasm://wasm/001cb4f2:wasm-function[367]:0x6ecb)
>     at Json::stringify.fn/435 (wasm://wasm/001cb4f2:wasm-function[603]:0xfd7a)
>     at $moonbitlang/core/json_blackbox_test.__test_6a736f6e5f746573742e6d6274_7.fn/483 (wasm://wasm/001cb4f2:wasm-function[658]:0x120bc)
>     at $moonbitlang/core/json_blackbox_test.__test_6a736f6e5f746573742e6d6274_7.clo/765 (wasm://wasm/001cb4f2:wasm-function[14]:0x25a5)
>     at *main*/770 (wasm://wasm/001cb4f2:wasm-function[774]:0x19780)
> ```
> 
> ```
> 0.0
> -0.0
> 9223372036854775807 (and larger value)
> ```

Maybe `if Double::trunc(x)==x { x.to_string().split('.').head() } else { x.to_string() }` is a better solution?

json/json.mbt
```diff
@@ -113,7 +113,17 @@ pub fn stringify(self : JsonValue) -> String {
       buf.write_string(escape(s))
       buf.write_char('\"')
     }
-    Number(n) => return n.to_string()
+    Number(n) => {
+      if n == 0 {
+        return "0"
+      }
+      let s = n.to_string()
```
bobzhang, line 120: cc @lijunchen should we fix s.to_string instead to align with JS semantics

qazxcdswe123 (approved):
The json parse error message could be better but thats beyond the scope of this PR

json/json_test.mbt
```diff
@@ -310,3 +312,47 @@ test "stringify escape" {
     Ok(newjson) => assert_eq!(newjson, json)
   }
 }
+
+test "stringify number" {
+  let nums : Array[Json] = [
+    0.0,
+    -0.0,
+    1.0,
```
qazxcdswe123, line 320: I think this should still be `1.0`, same as `0.0` and `-0.0`?

PR #886, opened 2024-08-22
hongbo/string rev iter

bobzhang:
- **disable gutte**
- **add String::rev_iter**

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "testing.gutterEnabled": false
+}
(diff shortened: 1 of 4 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
qazxcdswe123, line 1: unexpected?

bobzhang: this is intentional to disable guttering which is duplicate of `Run tests| Update tests`

PR #888, opened 2024-08-22
add double constants

### LINKED ISSUE #885: `min_value` for Double is not idiomatic
Currently, the min value for double is

```
let min_val : Double = 0x1L.reinterpret_as_double()
  inspect!(@double.min_value(), content="5.0e-324")
```

For me, the more idiomatic value would be

```
let min_val : Double = 4503599627370496L.reinterpret_as_double()
0 00000000001 0000000000000000000000000000000000000000000000000000 // bit value
  inspect!(@double.min_value(), content="2.2250738585072014e-308")
```

For example: https://doc.rust-lang.org/std/primitive.f64.html#associatedconstant.MIN_POSITIVE

Also probably separate MIN and MIN_POSITIVE?

Probably related: https://github.com/moonbitlang/core/pull/882

yj-qin:
Fix #885
- add double constants, rename `min_normal` to `min_positive`, and let `min` equal to `-max`.
- deprecate some functions that return constants.

### DIFF
--- a/double/double.mbt
+++ b/double/double.mbt
@@ -12,17 +12,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-let not_a_number : Double = 0x7FF8000000000001L.reinterpret_as_double()
+pub let not_a_number : Double = 0.0 / 0.0
 
-let positive_infinity : Double = 0x7FF0000000000000L.reinterpret_as_double()
+pub let infinity : Double = 1.0 / 0.0
 
-let negative_infinity : Double = 0xFFF0000000000000L.reinterpret_as_double()
+pub let neg_infinity : Double = -1.0 / 0.0
 
-let max_val : Double = 0x7FEFFFFFFFFFFFFFL.reinterpret_as_double()
+pub let max : Double = 1.7976931348623157e+308
 
-let min_val : Double = 0x1L.reinterpret_as_double()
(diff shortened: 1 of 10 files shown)


double/double.mbt
```diff
@@ -12,17 +12,17 @@
...
+pub let not_a_number : Double = 0.0 / 0.0
 
-let positive_infinity : Double = 0x7FF0000000000000L.reinterpret_as_double()
+pub let infinity : Double = 1.0 / 0.0
 
-let negative_infinity : Double = 0xFFF0000000000000L.reinterpret_as_double()
+pub let neg_infinity : Double = -1.0 / 0.0
 
-let max_val : Double = 0x7FEFFFFFFFFFFFFFL.reinterpret_as_double()
+pub let max : Double = 1.7976931348623157e+308
```
qazxcdswe123, line 21: I think that using literal to represent floating point number is not precise enough. 

And rust doc and MS doc disagree on the precision `1.7976931348623157E+308f64`[(rust)](https://doc.rust-lang.org/std/primitive.f64.html#associatedconstant.MIN_POSITIVE) vs 
`1.7976931348623158e+308`[(Microsoft)](https://learn.microsoft.com/en-us/cpp/cpp/floating-limits?view=msvc-170)

The old approach `reinterpret_as_double()` is better IMO

PR #891, opened 2024-08-22
add binary_search functionality for Array

guuzaa:
The functionality added:
- binary_search
- binary_search_by

See code for more.

### DIFF
--- a/builtin/array.mbt
+++ b/builtin/array.mbt
@@ -577,19 +577,28 @@ pub fn search[T : Eq](self : Array[T], value : T) -> Int? {
   }
 }
 
-/// Binary search the array index for a given element in a sorted array.
+/// Performs a binary search on a sorted array to find the index of a given element.
 /// 
-/// If the element is found then Ok is returned, containing the matching array index. 
-/// If there are multiple matches, then the leftmost match could be returned. 
-/// If the element is not found then Err is returned, containing the array index where a matching element cloud be inserted while maintaining sorted order.
-/// 
-/// The array must be sorted in increaing order. If not, the returned result is meaningless.
-///
 /// # Example
 /// ```
 /// let v = [3, 4, 5]
-/// v.binary_search(3)
+/// let result = v.binary_search(3)
+/// assert_eq!(result, Ok(0)) // The element 3 is found at index 0
 /// ```
+/// 
+/// # Arguments
+/// - `self`: The array in which to perform the search.
+/// - `value`: The element to search for in the array.
+/// 
+/// # Returns
+/// - `Result<Int, Int>`:
+/// If the element is found, an `Ok` variant is returned, containing the index of the matching element in the array. 
+/// If there are multiple matches, the leftmost match will be returned. 
+/// If the element is not found, an `Err` variant is returned, containing the index where the element could be inserted to maintain the sorted order.
+/// 
+/// # Notes
+/// - Ensure that the array is sorted in increasing order before calling this function. 
+/// - If the array is not sorted, the returned result is undefined and should not be relied on.
 pub fn binary_search[T : Compare](
   self : Array[T],
   value : T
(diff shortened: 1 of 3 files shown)


builtin/array.mbt
```diff
@@ -577,6 +577,66 @@ pub fn search[T : Eq](self : Array[T], value : T) -> Int? {
...
+pub fn binary_search[T : Compare](
+  self : Array[T],
+  value : T
+) -> Result[Int, Int] {
+  return self.binary_search_by(
+    fn(v : T) { if v < value { -1 } else if v == value { 0 } else { 1 } },
+  )
+}
+
+/// Binary search the array index for a given element in a sorted array with a custom comparator function.
```
qazxcdswe123, line 602: The wording here probably needs some improvement, here is an example from GPT4

```
/// Performs a binary search on a sorted array using a custom comparator function, and returns the index of the given element if found.
/// 
/// The comparator function should return an `Int` that indicates whether the argument is less than `(-1)`, equal to `(0)`, or greater than `(1)` 
/// the target element.
/// 
/// If the element is found, an `Ok` variant is returned containing the index of the matching element in the array. 
/// If there are multiple matches, the leftmost match will be returned. 
/// 
/// If the element is not found, an `Err` variant is returned containing the index where the element could be inserted while maintaining the sorted order of the array.
/// 
/// # Important Considerations
/// - The array must be sorted, and the comparator function should be consistent with the sorted order of the array.
/// - If the array is not sorted or if the comparator function is not consistent with the sorted order, the result of this function will be meaningless.
/// 
/// # Example
/// ```
/// let v = [3, 4, 5];
/// let result = v.binary_search_by(|v| {
///     if v < 3 { -1 }
///     else if v == 3 { 0 }
///     else { 1 }
/// });
/// assert_eq!(result, Ok(0)); // The element 3 is found at index 0
/// ```
/// 
/// # Type Parameters
/// - `T`: The type of elements in the array. It must implement the `Compare` trait.
/// 
/// # Arguments
/// - `self`: The array in which to perform the search.
/// - `cmp`: A closure that defines the comparator function, taking an element of type `T` and returning an `Int`.
/// 
/// # Returns
/// - `Result[Int, Int]`: If the element is found, `Ok` is returned with the index of the element.
/// If the element is not found, `Err` is returned with the index where the element could be inserted to maintain sorted order.
```

builtin/array.mbt
```diff
@@ -577,6 +577,66 @@ pub fn search[T : Eq](self : Array[T], value : T) -> Int? {
   }
 }
 
+/// Binary search the array index for a given element in a sorted array.
```
qazxcdswe123, line 580: Same wording issue here

```
/// Performs a binary search on a sorted array to find the index of a given element.
/// 
/// If the element is found, an `Ok` variant is returned, containing the index of the matching element in the array. 
/// If there are multiple matches, the leftmost match will be returned. 
/// 
/// If the element is not found, an `Err` variant is returned, containing the index where the element could be inserted to maintain the sorted order.
/// 
/// The array must be sorted in increasing order. If the array is not sorted, the result of this function is undefined and should not be relied upon.
/// 
/// # Example
/// ```
/// let v = [3, 4, 5];
/// let result = v.binary_search(&3);
/// assert_eq!(result, Ok(0)); // The element 3 is found at index 0
/// ```
/// 
/// # Arguments
/// - `self`: The sorted array in which to perform the search.
/// - `target`: The element to search for in the array.
/// 
/// # Returns
/// - `Result<usize, usize>`: If the element is found, `Ok` is returned with the index of the element.
/// If the element is not found, `Err` is returned with the index where the element could be inserted to maintain sorted order.
/// 
/// # Notes
/// - Ensure that the array is sorted in increasing order before calling this function.
/// - The function performs a logarithmic search, making it efficient for large arrays.
```

builtin/array.mbt
```diff
@@ -577,6 +577,66 @@ pub fn search[T : Eq](self : Array[T], value : T) -> Int? {
...
+/// )
+/// ```
+pub fn binary_search_by[T : Compare](
+  self : Array[T],
+  cmp : (T) -> Int
+) -> Result[Int, Int] {
+  let mut i = 0
+  let mut j = self.length()
+  while i < j {
+    let h = (i + j) / 2
```
qazxcdswe123, line 626: Potential overflow issue here.

Probably use rust approach?

        let mut size = self.len();
            let mid = left + size / 2;

            size = right - left;

builtin/array.mbt
```diff
@@ -577,6 +577,66 @@ pub fn search[T : Eq](self : Array[T], value : T) -> Int? {
...
+/// If the array is not sorted or if the comparator function is not consistent with the sorted order of array, the returned result is meaningless.
+/// 
+/// # Example
+/// ```
+/// let v = [3, 4, 5]
+/// v.binary_search_by(
+///   fn(v : T) { if v < 3 { -1 } else if v == 3 { 0 } else { 1 } },
+/// )
+/// ```
+pub fn binary_search_by[T : Compare](
```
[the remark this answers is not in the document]
guuzaa: You're right. It doesn't need Compare trait. The Eq trait is fine here.

builtin/array.mbt
guuzaa: I've no clue what the standard style for the document is. Your suggestions are inspiring. I'll adopt to this style.

builtin/array.mbt
guuzaa: The rust approach is good. But I think `let h = i + (j - i) / 2` is better here, where i is always less than j.

bobzhang:
@guuzaa thanks for you first contribution.
I made two tweaks:
- inline the callback in binary_search
- use our functional loop to make the code easier to reason about

PR #897, opened 2024-08-23
correct the type of `queue::fold`

Yoorkin:
The type of `fold` should be `(T[A], B, (B, A) -> B) -> B`


queue/queue.mbt
```diff
@@ -237,7 +237,7 @@ pub fn eachi[A](self : T[A], f : (Int, A) -> Unit) -> Unit {
 /// let queue : Queue[Int] = of([1, 2, 3, 4])
 /// let sum = queue.fold(0, fn(acc, x) { acc + x })
 /// ```
-pub fn fold[A](self : T[A], ~init : A, f : (A, A) -> A) -> A {
+pub fn fold[A, B](self : T[A], ~init : B, f : (B, A) -> B) -> B {
   loop self.first, init {
     cell, acc =>
       match cell {
```
Yu-zh, line 243: loop self.first, init {
  Nil, acc => ..
  Cons, acc => ..
}

PR #921, opened 2024-08-26
Refactor ryu to match ECMAScript standard

qazxcdswe123:
cc @bobzhang


double/internal/ryu/ryu.mbt
```diff
@@ -604,6 +576,9 @@ fn d2d_small_int(
 }
 
 pub fn ryu_to_string(val : Double) -> String {
+  if val == 0.0 {
```
lijunchen, line 579: consider `-0.0`, it equals `0.0` but should be printed as `-0`

bobzhang: I tried console, the JS engine does print it as "0"

PR #924, opened 2024-08-27
factor out error helper

bobzhang:


json/from_json.mbt
```diff
@@ -26,53 +26,55 @@ pub fn from_json[T : FromJson](
   FromJson::from_json!(json, path)
 }
```
bobzhang, line 28: cc @Yu-zh introduce never type may still be useful

PR #927, opened 2024-08-27
weird error message

bobzhang:


bobzhang:
cc @Young-Flash the confusing part seems to be that
```
Error: The toplevel identifier debug_string is declared twice: it was previously defined at /home/runner/work/core/core/builtin/bigint_wbtest.mbt:15:4.
```
the error message is printed twice

PR #928, opened 2024-08-27
hongbo/debug my bigint

bobzhang:
- **weird error message**
- **remove unneeded mutation**
- **more internal tests**


bobzhang:
cc @peter-jerry-ye make bigint immutable

PR #931, opened 2024-08-28
fix: bigint bugs

peter-jerry-ye:
- `neg` didn't take into account of zero
- `from_int` broke invariant that `limbs[i]` < `radix`


qazxcdswe123:
lgtm

PR #934, opened 2024-08-29
hongbo/new iter features

bobzhang:
- **add Iter::op_as_view**
- **add List::add**
- **add String::fold & String::rev_fold**
- **more tests**

### DIFF
--- a/immut/list/list.mbt
+++ b/immut/list/list.mbt
@@ -17,6 +17,10 @@ pub enum T[A] {
   Cons(A, T[A])
 } derive(Eq)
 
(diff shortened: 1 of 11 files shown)


immut/list/list.mbt
```diff
@@ -17,6 +17,10 @@ pub enum T[A] {
   Cons(A, T[A])
 } derive(Eq)
 
+pub fn add[A](self : T[A], head : A) -> T[A] {
```
hackwaly, line 20: addFront, addFirst or insert?

PR #935, opened 2024-08-29
feat: change radix of bigint to 2^32

peter-jerry-ye:
Based on #925

### DIFF
--- a/builtin/bigint.mbt
+++ b/builtin/bigint.mbt
@@ -241,7 +241,18 @@ pub fn op_div(self : BigInt, other : BigInt) -> BigInt {
   if other == zero {
     abort("division by zero")
   }
-  grade_school_div(self, other).0
+  // Handle negative numbers
+  if self.sign == Negative {
+    if other.sign == Negative {
+      grade_school_div(-self, -other).0
+    } else {
+      -grade_school_div(-self, other).0
+    }
+  } else if other.sign == Negative {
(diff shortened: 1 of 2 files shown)


builtin/bigint.mbt
```diff
@@ -290,110 +285,126 @@ fn grade_school_div(self : BigInt, other : BigInt) -> (BigInt, BigInt) {
...
-  let mut back = divisor.limbs[divisor.len - 1].to_int64()
-  while back < radix / 2L {
-    back = back.lsl(1)
-    lshift += 1
-  }
+  // D1. normalize
+  // m = divident.limbs.length() - divisor.limbs.length()
+  // left shift divident & divisor such that 
+  // - divisor.limbs.last() >= radix / 2
+  // - divident.limbs.length() = self.limbs.length() + 1
```
jialunzhang-psu, line 321: For test case `self = 18446744073709551615, other = 4294967296`, this condition doesn't hold, where `dividend.limbs.length() == 3, self.limbs.length() == 4`, and `a_len == 3`. In general, I think all invariants or pre/post conditions should not talk about `x.limbs.length()`, which is not part of the representation invariant (or at least it is not constrained by any upper bound).

peter-jerry-ye: Yes, you are right. It should've been `len` instead of `limbs.length`

PR #936, opened 2024-08-29
rewrite in functional style

bobzhang:
unclear why changes are lost 
https://github.com/moonbitlang/core/pull/891#issuecomment-2308864108
cc @guuzaa


guuzaa:
Thank you! I appreciate your feedback on the style. You know, I'm new to moonbit.  But why doesn't binary_search call binary_search_by? In my opinion, doing so could reduce duplication.

bobzhang:
@guuzaa yes, you are right, but for core library, the *predictable* preformance matters, even though the optimizer could inline it.

PR #939, opened 2024-08-30
feat: add utils for bigint

peter-jerry-ye:
Based on #935 

- Add `pow` which may perform `mod` at the same time
- Add `from_octets` and `to_octets` to convert from/to octet string (`Bytes`), which can be useful in many scenarios


bobzhang:
cc @jialunzhang-psu

PR #943, opened 2024-08-31
Refactoring in Array

guuzaa:
# What I've done
- Rename the `find_index` function to `search_by` for consistency with the `search` function.
- Fix some errors in documents and tests introduced by my last [PR](https://github.com/moonbitlang/core/pull/891).
- Enhance the `retain` and `append` functions.
- Add more tests.


bobzhang:
thanks for your contributions, it is merged in #944 
btw it would be nice to keep the scope of a PR small, so it can be iterated quickly

PR #944, opened 2024-08-31
Refactoring Array

bobzhang:
cc @guuzaa it seems that I can not do modifications directly in your branch, I appended a commit to keep the old API for a while and some minor clean up


bobzhang:
@peter-jerry-ye the AI review looks great, it would be nice to output checkbox format, so I can click which suggestions are handled, which can be ignored

guuzaa:
Yes, we shouldn't remove `find_index` right away. Meanwhile we cannot replace its all tests with search_by.

PR #958, opened 2024-09-04
hongbo/test new features2

bobzhang:
- **tweak**
- **tweak**

### DIFF
--- a/builtin/array.mbt
+++ b/builtin/array.mbt
@@ -253,8 +253,9 @@ pub fn each[T](self : Array[T], f : (T) -> Unit) -> Unit {
 }
 
 pub fn rev_each[T](self : Array[T], f : (T) -> Unit) -> Unit {
-  for i = self.length() - 1; i >= 0; i = i - 1 {
-    f(self[i])
+  let len = self.length()
+  for i in 0..<len {
+    f(self[len - i - 1])
   }
 }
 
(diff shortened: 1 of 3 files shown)


builtin/array.mbt
```diff
@@ -283,8 +284,8 @@ pub fn eachi[T](self : Array[T], f : (Int, T) -> Unit) -> Unit {
 /// ```
 pub fn rev_eachi[T](self : Array[T], f : (Int, T) -> Unit) -> Unit {
   let len = self.length()
-  for i = 0; i < len; i = i + 1 {
-    f(i, self[len - i - 1])
+  for i in 0..<len {
+    f(len - i - 1, self[i])
```
gmlewis, line 288: Didn't this change the order that the array elements are processed?

Say there are 3 elements...
This code used to call:

```
f(0, self[2]) // sum += 0 + 5 + 0 = 5
f(1, self[1]) // sum += 5 + 4 + 1 = 10
f(2, self[0]) // sum += 10 + 3 + 2 = 15
```

Now it calls:

```
f(2, self[0]) // sum += 0 + 3 + 2 = 5
f(1, self[1]) // sum += 5 + 4 + 1 = 10
f(0, self[2]) // sum += 10 + 5 + 0 = 15
```

So the provided example isn't affected, but the behavior is different.

But from the description of the function:

`Iterates over the elements of the array with index in reversed order.`

I would have expected this behavior:

```
f(2, self[2]) // sum += 0 + 5 + 2 = 7
f(1, self[1]) // sum += 7 + 4 + 1 = 12
f(0, self[0]) // sum += 12 + 3 + 0 = 15
```

bobzhang: fixed in #960

PR #959, opened 2024-09-05
Add to_array and to_string funtions for Iter2

guuzaa:
I've implemented `to_string` and `to_array` functions and the `Show` trait for `Iter2`.

### DIFF
--- a/builtin/builtin.mbti
+++ b/builtin/builtin.mbti
@@ -203,6 +203,7 @@ impl Iter {
   each[T](Self[T], (T) -> Unit) -> Unit
   eachi[T](Self[T], (Int, T) -> Unit) -> Unit
   empty[T]() -> Self[T]
+  enumerate[T](Self[T]) -> Iter2[Int, T]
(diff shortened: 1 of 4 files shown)


qazxcdswe123:
```
  let array = [5, 6, 7]
  for k, v in array {
    println("\{k}, \{v}")
  }
  ```

This will do

guuzaa:
In other words, if you'd like to use `for k, v in array`, you need to implement another `iter2` function for your type. However, in many scenarios, it could be tedious to do this for each new type.

bobzhang:
can you split changes to Iter/Iter2 into two PRs?
the interface changes in Iter2 looks good to me, for Iter::enumerate, do you want to return Iter2[Int,A] or Iter[(Int,A)], if performance is not a concern, you may want to return the second for convenience

guuzaa:
This PR only includes changes to Iter2. As for the enumerate function, I'll address it on another PR later.

PR #964, opened 2024-09-06
chore: maintain map/set API consistency

Lampese:


immut/hashmap/hashmap.mbti
```diff
@@ -5,11 +5,11 @@ package moonbitlang/core/immut/hashmap
 // Types and methods
 type T
 impl T {
-  add[K : Eq + Hash, V](Self[K, V], K, V) -> Self[K, V]
   each[K, V](Self[K, V], (K, V) -> Unit) -> Unit
-  find[K : Eq + Hash, V](Self[K, V], K) -> V?
```
bobzhang, line 10: the old name makes more sense to me, also we would prefer keep the old APIs for two weeks(with deprecations) and remove them later

Lampese: It's a question worth discussing, and I'll talk more about later on what's the appropriate name for these apis.

PR #970, opened 2024-09-10
format code

Yoorkin:


string/string.mbt
```diff
@@ -343,16 +343,14 @@ pub fn last_index_of(
...
-    let found = {
-      while j > start {
-        if self[j] != str[k] {
-          break false
-        }
-        j -= 1
-        k -= 1
-      } else {
-        true
+    let found = while j > start {
```
bobzhang, line 346: @Yoorkin we prefer it indented as:

```
let found = 
  while j > start {
  }
if found ...   
```

Yoorkin: ok, I will adjust it later

PR #973, opened 2024-09-10
test guard let

bobzhang:


array/array.mbt
```diff
@@ -16,9 +16,8 @@ pub fn iter[T](self : Array[T]) -> Iter[T] {
   Iter::new(
     fn(yield) {
       for v in self {
-        if yield(v) == IterEnd {
-          break IterEnd
-        }
+        guard let IterContinue = yield(v) else { x => break x }
```
bobzhang, line 19: cc @Yoorkin  @lynzrand we need check if the coverage works with the new guard let

array/array.mbt
```diff
@@ -30,9 +29,8 @@ pub fn iter2[A](self : Array[A]) -> Iter2[Int, A] {
   Iter2::new(
     fn(yield) {
       for i, v in self {
-        if yield(i, v) == IterEnd {
-          break IterEnd
-        }
+        guard let IterContinue = yield(i, v) else { x => break x }
+
```
bobzhang, line 33: cc @Yoorkin the new line seems not needed

PR #978, opened 2024-09-11
test derive Arbitrary

bobzhang:

PR #984, opened 2024-09-12
use int64's popcnt

lijunchen:
the following code in js backend will cause an infinity loop in https://github.com/moonbitlang/core/pull/978 

It is a compiler bug. This PR fixes it by using Int64's popcnt method. The compiler bug should be fixed later.

```
fn pop_count(c : UInt64) -> Int {
  loop 0, c {
    x, 0 => x
    c, w => continue c + 1, w & (w - 1)
  }
}

test {
  let _ = pop_count(1UL)
}
```


bobzhang:
merged in #978

PR #986, opened 2024-09-12
feat: add float show

qazxcdswe123:
WeChatWorkScreenshot_966be902-1444-4716-a301-239a910aa38c

As JS does not have native float type, it first convert it to double and then use double show.

If we are going to stick with JS's behavior, we dont need an `ryu` for `float`. This is enough.

This can also be used as a temporary solution even if we want a dedicated `ryu` for `float`

JS test code
```
// Create a Float32Array with one element
let floatArray = new Float32Array(1);

// Assign a 32-bit float value
floatArray[0] = 3.14159;  // Example float value

// Print the value
console.log(floatArray[0]); // This will print 3.141590118408203
```


bobzhang:
cc @qazxcdswe123 I added some todo notes

PR #988, opened 2024-09-12
hongbo/tweak tojson/fromjson and add derive fromjson test

bobzhang:
- **tweak tojson/fromjson**
- **more tests**
- **fmt**

### DIFF
--- a/json/from_json.mbt
+++ b/json/from_json.mbt
@@ -32,8 +32,8 @@ fn decode_error[T](path : JsonPath, msg : String) -> T!JsonDecodeError {
 
 pub impl FromJson for Bool with from_json(json, path) {
   match json {
-    True => true
-    False => false
+    true => true
+    false => false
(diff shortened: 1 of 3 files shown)


json/from_json.mbt
```diff
@@ -122,6 +123,11 @@ pub impl[X : FromJson] FromJson for FixedArray[X] with from_json(json, path) {
 }
 
 pub impl[V : FromJson] FromJson for Map[String, V] with from_json(json, path) {
+  // Note in theory, we can make it more general like
+  // Map[K : FromString, V : FromJson]
+  // but currently we don't have FromString 
+  // Here ToJSON we have 
+  // Map[K : Show, V : ToJson]
```
qazxcdswe123, line 130: The comment `Map[K : Show, V : ToJson]` doesn't match the signature `Map[String, V : ToJson]`?

bobzhang: it does, can you double check?

PR #993, opened 2024-09-13
add `concat`, `from_array` methods for String

### LINKED ISSUE #989: Perferred way to concatenate an array of characters
`Show::to_string` hides `Char::to_string` and joining an array of `Char` produces a messy result unlike Python (tho there's no builtin "char" in Python). Is this intended? If so, what is the perferred way to concatenate an array of characters? There seems to be no such method to collect an array or iterator into a string like Rust and folding it is quite tedious.

```moonbit
println(['1', '2', '3'].to_string()) // ['1', '2', '3']
println(['1', '2', '3'].join("")) // '1''2''3'
println(['1', '2', '3'].fold(init="", fn(s, c) { s + c.to_string() })) // 123
```

Yoorkin:
- add `concat`, `from_array` methods for String
- remove `Array::join`
    The behavior of Array::join is not the same as javascript, because the `Show` trait is used to logging or debugging, the output result is always remains "as is".  e.g.  in moonbit,`["abc","def"].join() == "\"abc\"\"def\""`. So I prefer to remove this API and use `String::join` instead.

fix #989

### DIFF
--- a/array/array.mbt
+++ b/array/array.mbt
@@ -68,7 +68,7 @@ pub fn Array::new_with_index[T](length : Int, value : (Int) -> T) -> Array[T] {
 ///
 /// The joined string.
 /// 
-/// @alert deprecated "Use `String::join` instead"
+/// @alert deprecated "Use `String::concat` instead"
 pub fn join[T : Show](self : Array[T], sep : String) -> String {
   let len = self.length()
   if len == 0 {
--- a/array/array.mbt
+++ b/array/array.mbt
@@ -113,33 +113,6 @@ pub fn Array::shuffle[T](arr : Array[T], ~rand : (Int) -> Int) -> Array[T] {
   new_arr
 }
 
-/// Joins the elements of an array into a single string, using the specified separator.
-///
-/// # Arguments
-///
-/// * `self` - The array of strings to join.
-/// * `sep` - The separator string to insert between each element.
-///
-/// # Returns
-///
-/// The joined string.
-pub fn join[T : Show](self : Array[T], sep : String) -> String {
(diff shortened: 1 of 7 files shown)


Lampese (approved):

array/array.mbt
```diff
@@ -139,33 +139,6 @@ pub fn Array::shuffle[T](arr : Array[T], ~rand : (Int) -> Int) -> Array[T] {
...
-///
-/// # Arguments
-///
-/// * `self` - The array of strings to join.
-/// * `sep` - The separator string to insert between each element.
-///
-/// # Returns
-///
-/// The joined string.
-pub fn join[T : Show](self : Array[T], sep : String) -> String {
```
qazxcdswe123, line 152: deprecate before removing it?

string/string.mbti
```diff
@@ -5,16 +5,19 @@ package moonbitlang/core/string
...
   default() -> String
   ends_with(String, String) -> Bool
   fold[A](String, ~init : A, (A, Char) -> A) -> A
+  from_array(Array[Char]) -> String
   index_of(String, String, ~from : Int = ..) -> Int
   is_blank(String) -> Bool
   is_empty(String) -> Bool
   iter(String) -> Iter[Char]
   iter2(String) -> Iter2[Int, Char]
+  join(Array[String], ~separator : String) -> String
```
bobzhang, line 20: should `concat` and `join` be merged as a single function

gmlewis:
It's quite disappointing that `x.join(...)` is deprecated, as this is such a common operation and was so easy to type.

PR #997, opened 2024-09-14
clean up some deprecated function

Yoorkin:
remove some deprecated function:
- fold_left, fold_right, fold_lefti, fold_righti, ...
- `List::concat_map`, `Ref::ref`, `Iter::tap`
- pop_exn, head_exn, nth_exn, ...
- some util function in test package


bobzhang:
instead of removing them, shall we use the convention e.g, create a file called deprecated.mbt and put them all together to avoid some breakage. Keep them for a bit longer, e.g, a couple of months?

Yoorkin:
> instead of removing them, shall we use the convention e.g, create a file called deprecated.mbt and put them all together to avoid some breakage. Keep them for a bit longer, e.g, a couple of months?

putting them together in deprecated.mbt looks good to me

qazxcdswe123:
Some may use `@pkg.func`, will this be affected?

bobzhang:
> Some may use @pkg.func, will this be affected if we move them to separated file?

No, it is backwards compatible

PR #1002, opened 2024-09-15
Add String.pad_left and String.pad_right

bikallem:
This PR adds 2 string methods that I have been copy pasting in a few projects. It would be nice if moonbit core library supports them out of the box. 
1. pad_left - given a total_width and a padding_char, pads the left of the string.
2. pad_right - given a total_width and a padding_char, pads the right of the string.

With regards to implementation details, I have used the buffer approach rather than the string concatenation approach. However, I can be convinced to use the string concatenation approach since string allocations are minimal for our use case.

### DIFF
--- a/string/string.mbt
+++ b/string/string.mbt
@@ -538,8 +538,8 @@ pub fn rev(self : String) -> String {
   result
 }
 
-/// Returns a new string with length `total_width` and `padding_char` prefixed 
-/// if `self.length() < total_width`.
+/// `pad_left` is a new string with `padding_char`s prefixed to `self` if `self.length() < total_width`. The length of the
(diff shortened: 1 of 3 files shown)


CryZe:
It should be called `pad_start` and `pad_end`, not left / right. Not every language is written from left to right.

bikallem:
@CryZe As you have noticed, the function names are consistent with the current convention. However, I am open to renaming if the repo. owner deems it suitable.

bobzhang:
hi @bikallem since I can not make changes to your contribution directly, I will have a follow up pr to tweak the name to align with JS conventions later, thank you.

Issue #938, opened 2024-08-29
Is it necessary to implement hash/compare trait and for Array?


Lampese:
- [x] Compare (done in #1003) (but to be discussed)
- [ ] Hash (to be discussed)

bobzhang:
compare is needed, not sure about hash though

tonyfettes:
use case for hash: BPE tokenizer. In BPE tokenizer I need to maintain a `Map[(Array[Int], Array[Int]), Array[Int]]` when merging multi-byte pairs.

tonyfettes:
I have another use case: hash cons when compiling a computation graph.

The const in a computation graph is a `FixedArray[Float]` (or `Array[Float]`) and I need it to be hashable so that I can derive Hash for my graph data type.

bobzhang:
I agree we should add hash for Array, user can always misuse HashMap if they want

PR #1003, opened 2024-09-16
feat: compare for Array

Lampese:
- #938


bobzhang:
@Lampese like the string comparison, I tweaked the algorithm to compare by length first

PR #1011, opened 2024-09-18
move splitmix from random into quickcheck

bobzhang:
cc @CAIMEOX can you help document functions inside splitmix


CAIMEOX (approved):
We can merge this first and I will open another PR for documentation.

PR #1027, opened 2024-09-20
fix: proper option fromjson

qazxcdswe123:
mainly for `Option[Option[T]]`

> for T? inside record, e.g,
struct Hello[T] {
x : T?
y : Int
}
when x is None, it does not show up in JSON, when x is Some(val), only val shows up

This needs to be changed in compiler

### DIFF
--- a/builtin/json.mbt
+++ b/builtin/json.mbt
@@ -125,7 +125,7 @@ pub fn Map::to_json[K : Show, V : ToJson](self : Map[K, V]) -> Json {
 pub fn Option::to_json[T : ToJson](self : T?) -> Json {
   match self {
     None => Null
-    Some(value) => value.to_json()
+    Some(value) => { "Some": value.to_json() }
   }
 }
 
--- a/builtin/json.mbt
+++ b/builtin/json.mbt
@@ -125,7 +125,7 @@ pub fn Map::to_json[K : Show, V : ToJson](self : Map[K, V]) -> Json {
 pub fn Option::to_json[T : ToJson](self : T?) -> Json {
   match self {
     None => Null
-    Some(value) => { "Some": value.to_json() }
+    Some(value) => [value.to_json()]
   }
 }
 
--- a/json/from_json.mbt
+++ b/json/from_json.mbt
@@ -144,8 +144,24 @@ pub impl[V : FromJson] FromJson for Map[String, V] with from_json(json, path) {
 
 pub impl[T : FromJson] FromJson for T? with from_json(json, path) {
   match json {
+    Object(obj) => {
+      if obj.size() != 1 {
+        decode_error!(path, "Option::from_json: expected object with one field")
+      }
+      match obj {
+        { "Some": value } =>
+          Some(FromJson::from_json!(value, path.add_key("Some")))
+        _ =>
+          decode_error!(
+            path, "Option::from_json: expected object with None or Some field",
+          )
+      }
+    }
     Null => None
-    _ => Some(FromJson::from_json!(json, path))
+    _ =>
+      decode_error!(
+        path, "Option::from_json: expected object with None or Some field",
+      )
   }
 }
 
--- a/json/from_json.mbt
+++ b/json/from_json.mbt
@@ -144,24 +144,9 @@ pub impl[V : FromJson] FromJson for Map[String, V] with from_json(json, path) {
 
 pub impl[T : FromJson] FromJson for T? with from_json(json, path) {
   match json {
-    Object(obj) => {
-      if obj.size() != 1 {
-        decode_error!(path, "Option::from_json: expected object with one field")
-      }
-      match obj {
-        { "Some": value } =>
-          Some(FromJson::from_json!(value, path.add_key("Some")))
-        _ =>
-          decode_error!(
-            path, "Option::from_json: expected object with None or Some field",
-          )
-      }
-    }
     Null => None
-    _ =>
-      decode_error!(
-        path, "Option::from_json: expected object with None or Some field",
-      )
+    Array([value]) => Some(FromJson::from_json!(value, path.add_index(0)))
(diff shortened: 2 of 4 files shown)


builtin/json.mbt
```diff
@@ -124,8 +124,8 @@ pub fn Map::to_json[K : Show, V : ToJson](self : Map[K, V]) -> Json {
 
 pub fn Option::to_json[T : ToJson](self : T?) -> Json {
   match self {
-    None => Null
-    Some(value) => value.to_json()
```
Yoorkin, line 128: if we follow the json format generated by enum type, the mapping between Option[T] and json should be

```
None   <->  { "$tag": "None" }
Some(x)  <->  { "$tag": "Some", 0: x }
```

It seems that using a special json format like this saves more memory. It looks good to me, but I am not sure what problems it might cause...

Anyway, the `"None"` can be omitted here:

```
None       <->   nulll
Some(x)  <->  { "Some": x }
```

builtin/json.mbt
```diff
@@ -124,8 +124,8 @@ pub fn Map::to_json[K : Show, V : ToJson](self : Map[K, V]) -> Json {
...
+    None => { "None": Null }
+    Some(value) => { "Some": value.to_json() }
```
Yoorkin, line 128: if we follow the json format generated by enum type, the mapping between Option[T] and json should be

```
None   <->  { "$tag": "None" }
Some(x)  <->  { "$tag": "Some", 0: x }
```

It seems that using a special json format like this saves more memory. It looks good to me, but I am not sure what problems it might cause...

Anyway, the `"None"` can be omitted here:

```
None       <->   null
Some(x)  <->  { "Some": x }
```

gmlewis:
This looks really unusual to me. 

Over the last 14 years working with the Go and JavaScript programming languages both on frontend and backend, I have never once come across a "Some" in JSON. 

Go has the concept of `omitempty` combined with the use of a pointer for the underlying type such that if the field is `nil` and there is `omitempty` in the tag, the field will simply be omitted (not even listed) in the JSON. This is how I would expect an optional field to be handled in MoonBit - if `None` then it could be omitted or `null` in the JSON, but if it is `Some(x)` then the value `x` would be listed in the JSON _WITHOUT_ the `Some`.

qazxcdswe123:
> This looks really unusual to me.
> 
> Over the last 14 years working with the Go and JavaScript programming languages both on frontend and backend, I have never once come across a "Some" in JSON.
> 
> Go has the concept of `omitempty` combined with the use of a pointer for the underlying type such that if the field is `nil` and there is `omitempty` in the tag, the field will simply be omitted (not even listed) in the JSON. This is how I would expect an optional field to be handled in MoonBit - if `None` then it could be omitted or `null` in the JSON, but if it is `Some(x)` then the value `x` would be listed in the JSON _WITHOUT_ the `Some`.

For `Option[Option]`, eg. `Some(None)` It would be `None.to_json()` -> `null`. Which would be indistinguishable from `None`

bobzhang:
@qazxcdswe123  we can do things like this:
for `T?` itself :
 None -> null
 Some(val) -> [val]
 
 for T? inside record, e.g,
  struct Hello[T] {
     x : T? 
     y : Int
  } 
  when x is None, it does not show up in JSON, when x is Some(val), only `val` shows up

So we need do some special handling for json encode/decode in record

qazxcdswe123:
Note that old behaviour match `serde`'s behaviour:

```
Serialized None: {"field":null}
Serialized Some(None): {"field":null}
Serialized Some(Some): {"field":"Hello"}
Deserialized None: MyStruct { field: None }
Deserialized Some(None): MyStruct { field: None }
Deserialized Some(Some): MyStruct { field: Some(Some("Hello")) }
```

<details><summary>Test Code</summary>
```rust
use serde::{Serialize, Deserialize};

#[derive(Debug, Serialize, Deserialize)]
struct MyStruct {
    field: Option<Option<String>>,
}

fn main() {
    let example_with_none: MyStruct = MyStruct {
        field: None,
    };

    let example_with_some_none: MyStruct = MyStruct {
        field: Some(None),
    };

    let example_with_some_value: MyStruct = MyStruct {
        field: Some(Some("Hello".to_string())),
    };

    // Serialize examples to JSON
    let serialized_none = serde_json::to_string(&example_with_none).unwrap();
    let serialized_some_none = serde_json::to_string(&example_with_some_none).unwrap();
    let serialized_some_value = serde_json::to_string(&example_with_some_value).unwrap();

    println!("Serialized None: {}", serialized_none); // Output: {"field":null}
    println!("Serialized Some(None): {}", serialized_some_none); // Output: {"field":null}
    println!("Serialized Some(Some): {}", serialized_some_value); // Output: {"field":"Hello"}

    // Deserialize examples from JSON
    let deserialized_none: MyStruct = serde_json::from_str(&serialized_none).unwrap();
    let deserialized_some_none: MyStruct = serde_json::from_str(&serialized_some_none).unwrap();
    let deserialized_some_value: MyStruct = serde_json::from_str(&serialized_some_value).unwrap();

    println!("Deserialized None: {:?}", deserialized_none);
    println!("Deserialized Some(None): {:?}", deserialized_some_none);
    println!("Deserialized Some(Some): {:?}", deserialized_some_value);
}
```
</details>

lynzrand:
Revisiting this PR: 

Special-casing `T?` in objects might not be a good idea, as it causes major inconsistency. People use `T?` both inside and outside object all the time, and people's natural expectation is `serialize(Some(T)) == serialize(T)`, and this consistent behavior benefits manual implementation of serialize/deserialize functions.

On the other hand, `T??` is very rarely used when compared to `T?`, and one can always manually implement a workaround function for this, which occurs in far fewer places than `T?` in manual implementaitons.

I suggest we revert the behavior to the original one, which matches Serde.

PR #1034, opened 2024-09-23
Enable nightly and bleeding release

bzy-debug:


bobzhang:
we can enable both nightly and stable

PR #1044, opened 2024-09-26
add `length()` and `is_empty()` to `Buffer`

rami3l:
As required in the Markdown parser implementation.

### DIFF
--- a/buffer/buffer.mbt
+++ b/buffer/buffer.mbt
@@ -52,6 +52,18 @@ fn grow_if_necessary(self : Buffer, required : Int) -> Unit {
 }
 
 ///|
+/// Return the given buffer's length in bytes.
+pub fn length(self : Buffer) -> Int {
+  self.len
+}
+
+///|
+/// Return whether the given buffer is empty.
+pub fn is_empty(self : Buffer) -> Bool {
+  self.len == 0
+}
+
+///|
 /// Return a new String contains the data in buffer.
 pub fn to_string(self : Buffer) -> String {
   self.bytes.sub_string(0, self.len)
--- a/buffer/buffer.mbti
+++ b/buffer/buffer.mbti
@@ -5,6 +5,8 @@ package moonbitlang/core/buffer
 // Types and methods
 type Buffer
 impl Buffer {
+  is_empty(Self) -> Bool
+  length(Self) -> Int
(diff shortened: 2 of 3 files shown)


builtin/buffer.mbt
```diff
@@ -72,6 +72,16 @@ fn grow_if_necessary(self : Buffer, required : Int) -> Unit {
   }
 }
 
+/// Return the given buffer's length in bytes.
```
bobzhang, line 75: this is encoding agnostics? How is it useful

rami3l: @bobzhang This API mirrors OCaml's [`Buffer.length`](https://ocaml.org/manual/5.2/api/Buffer.html#VALlength) which counts in bytes (i.e. `char`s in OCaml), so it is indeed encoding agnostic. Currently it's feasible via `b.to_bytes().length()`, but `.to_bytes()` introduces an extra copy.

Although in Cmarkit this API is only used to check whether the buffer is empty (namely `length b <> 0`), other use cases [do exist](https://github.com/search?q=language%3AOCaml+Buffer.length&type=code), notably getting a copy of the underlying `Bytes` towards its end via [`Buffer.sub`](https://ocaml.org/manual/5.2/api/Buffer.html#VALsub) (e.g. `sub b 42 (length b - 1 - 42)`).

Thus, I think it would still be beneficial to expose the buffer length. Of course, I can still withdraw the `.length()` API if you consider this to be unnecessary.

bobzhang: we are moving buffer outside builtin to provide a dedicated buffer package mimicking ocaml's behavior cc @hackwaly

hackwaly:
related https://github.com/moonbitlang/core/pull/1084

bobzhang:
@rami3l can you cherry pick these APIs into the new buffer package

rami3l:
> @rami3l can you cherry pick these APIs into the new buffer package

Rebased.

rami3l:
May I ask what is blocking this PR from being merged? Many thanks in advance :)

bobzhang (approved):
lgtm @rami3l sorry for the delay

rami3l:
> lgtm @rami3l sorry for the delay

@bobzhang Thanks again! The PR is now rebased.

PR #1049, opened 2024-09-29
fix some unused mutability & make some type private

Guest0x0:
While improving MoonBit's unused warning, I found several struct field with unused mutability & some types that should be made private on core. This PR fixes these unused things

### DIFF
--- a/math/algebraic_test.mbt
+++ b/math/algebraic_test.mbt
@@ -56,5 +56,5 @@ test "minimum.ref" {
 
 // For testing purposes
 priv struct T {
-  mut x : Int
+  x : Int
 } derive(Show, Eq, Compare)
--- a/builtin/builtin.mbti
+++ b/builtin/builtin.mbti
@@ -697,7 +697,7 @@ impl Hash for Option
 
 impl Hash for Result
 
-impl Hash for $default_impl
+impl Hash::hash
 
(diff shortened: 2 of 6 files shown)


math/algebraic_test.mbt
```diff
@@ -56,5 +56,5 @@ test "minimum.ref" {
 
 // For testing purposes
 priv struct T {
-  mut x : Int
```
hackwaly, line 59: The `mut` is indented to make compiler unable to do unbox optimization on type `T`.

hackwaly: While there's no something likes `//@warning "-unused-mut"`. I think we need a manual assignment statement to eliminate the warning?

Guest0x0: Currently `T` will not be unboxed globally even if its only field is immutable. If some value of `T` is used locally by accessing field only then that particular value may be unboxed, but that's not the case here, and mutable structs also get unboxed in that case.

This test itself may be controversial but let's discuss and fix that in a separate PR

Guest0x0:
There are some bug fixes in yesterday's compiler release on the format of default trait `impl` in `.mbti`, and `.mbti` files on core should be updated. Since it breaks CI, I'll update the changes in this PR.

PR #1062, opened 2024-10-06
add `Array::dedup_consecutive` for removing consecutive repeated elements in array

### LINKED ISSUE #1061: `Array::dedup` removes duplications ACROSS consecutive groups.
The doc-comment of the `Array::dedup` function says that it should remove consecutive repeated elements.

```moonbit
/// Removes consecutive repeated elements in the vector according to the Eq trait.
/// 
/// # Example
/// ```
/// let v = Vec::[3, 4, 4, 5, 5, 5]
/// v.dedup() // v = Vec::[3, 4, 5]
/// ```
```

However, as for array `[3, 4, 5, 6, 6, 4]`, this method will return `[3, 4, 5, 6]`. It removes `4`, which is not repeated (i.e., occurred only once in its own consecutive group).

```moonbit
test "dedup" {
  let array = [3, 4, 5, 6, 6, 4]
  array.dedup()
  inspect!(array, content="[3, 4, 5, 6]") // should be [3, 4, 5, 6, 4]
}
```

Thanks @Demonmasterlqx for reporting this.

tonyfettes:
For backward compatibility, this PR updates the documentation for `Array::dedup`, so that it shall not break any existing code that use this function. This PR also add another function (`Array::unique`) to implement the originally proposed functionality, i.e. de-duplication of consecutive repeated elements.

cc @Demonmasterlqx

close #1061 

## Implementation in other programming language

| Language | Consecutive | Non-consecutive |
|----------|-------------|-----------------|
| C++      | [`unique`](https://en.cppreference.com/w/cpp/algorithm/unique) | N/A |
| Rust     | [`dedup`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.dedup) | [`unique`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.unique) |
| NumPy    | N/A | [`unique`](https://numpy.org/doc/2.0/reference/generated/numpy.unique.html) |
| JavaScript | N/A | N/A |
| Python   | N/A | N/A |
| Scala    | N/A | [`distinct`](https://www.scala-lang.org/api/2.13.9/scala/collection/immutable/List.html#distinct:C) |
| *nix     | [`uniq`](https://man7.org/linux/man-pages/man1/uniq.1.html) | N/A |
| Haskell  | N/A | [`Data.List.Unique`](https://hackage.haskell.org/package/Unique-0.4.7.9/docs/Data-List-Unique.html) |

See more on https://rosettacode.org/wiki/Remove_duplicate_elements

## Algorithm Complexity

For the time-complexity of removing non-consecutive duplications:

> There are basically three approaches seen here:
> - Put the elements into a hash table which does not allow duplicates. The complexity is O(n) on average, and O(n^2) worst case. This approach requires a hash function for your type (which is compatible with equality), either built-in to your language, or provided by the user.
> - Sort the elements and remove consecutive duplicate elements. The complexity of the best sorting algorithms is O(n * log(n)). This approach requires that your type be "comparable", i.e., have an ordering. Putting the elements into a self-balancing binary search tree is a special case of sorting.
> - Go through the list, and for each element, check the rest of the list to see if it appears again, and discard it if it does. The complexity is O(n^2). The up-shot is that this always works on any type (provided that you can test for equality).

quoted from https://rosettacode.org/wiki/Remove_duplicate_elements

Removing consecutive duplications should only require `Eq` trait and of `O(n)` time complexity.

### DIFF
--- a/builtin/array.mbt
+++ b/builtin/array.mbt
@@ -843,11 +843,17 @@ pub fn dedup[T : Eq](self : Array[T]) -> Unit {
 /// v.dedup_consecutive() // v = [3, 4, 5, 6, 4]
 /// ```
 pub fn dedup_consecutive[T : Eq](self : Array[T]) -> Unit {
-  for i = 0; i < self.length(); i = i + 1 {
-    while i + 1 < self.length() && self[i] == self[i + 1] {
-      self.remove(i + 1) |> ignore
+  if self.is_empty() {
+    return
+  }
+  let mut w = 1
+  for i = 1; i < self.length(); i = i + 1 {
+    if self[i] != self[w - 1] {
+      self[w] = self[i]
+      w = w + 1
     }
   }
+  self.unsafe_truncate_to_length(w)
 }
 
 /// Extract elements from the array according to the given function.
--- a/builtin/array.mbt
+++ b/builtin/array.mbt
@@ -843,13 +843,9 @@ pub fn dedup[T : Eq](self : Array[T]) -> Unit {
 /// v.unique() // v = [3, 4, 5, 6, 4]
 /// ```
 pub fn unique[T : Eq](self : Array[T]) -> Unit {
-  for i in 0..<self.length() {
-    for j = i + 1; j < self.length(); j = j + 1 {
-      if self[i] == self[j] {
-        self.remove(j) |> ignore
-      } else {
-        break
-      }
+  for i = 0; i < self.length(); i = i + 1 {
+    while i + 1 < self.length() && self[i] == self[i + 1] {
+      self.remove(i + 1) |> ignore
     }
   }
 }
--- a/builtin/array.mbt
+++ b/builtin/array.mbt
@@ -835,6 +835,25 @@ pub fn dedup[T : Eq](self : Array[T]) -> Unit {
   }
 }
 
+/// Removes consecutive repeated elements in the array according to the Eq trait.
+/// 
+/// # Example
+/// ```
+/// let v = [3, 4, 5, 6, 6, 4]
+/// v.unique() // v = [3, 4, 5, 6, 4]
+/// ```
+pub fn unique[T: Eq](self : Array[T]) -> Unit {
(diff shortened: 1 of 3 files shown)


CAIMEOX (approved):

hackwaly:
Just my opinion: `unique` is not a good name for "removing consecutive repeated elements in array". How about keep `dedup` for `consecutive=true` and `unique` for `consecutive=false`.

bobzhang:
note the original `dedup` is terribly inefficient (called too many remove), but it is qutie common API. (good first issue)
The unique API here is not very common, what are the use cases? if we want to call unique, we generally do such things:
```
[...@hashset.from_iter()]
```

tonyfettes:
@bobzhang using

```
[...@hashset.from_iter()]
```

has roughtly the same behavior as the `Array::dedup`, not `Array::unique`, as it will only keep one of the repeated elements.

For example, for array `[3, 4, 5, 6, 6, 4, 4]`,

- `Array::dedup`

  ```moonbit
  [3, 4, 5, 6]
  ```

- `[...@hashset.from_iter()]`

  ```moonbit
  [6, 3, 4, 5]
  ```

- `Array::unique`

  ```moonbit
  [3, 4, 5, 6, 4]
  ```

Additionally, using `[...@hashset.from_iter()]` discards the order that these elments has in the array, which might not be what the user would expect.

builtin/array.mbt
```diff
@@ -835,6 +835,27 @@ pub fn dedup[T : Eq](self : Array[T]) -> Unit {
   }
 }
 
+/// Removes consecutive repeated elements in the array according to the Eq trait.
+/// 
+/// # Example
+/// ```
+/// let v = [3, 4, 5, 6, 6, 4]
```
bobzhang, line 842: 1. investigate [... @hash_set.from_iter(...)] should preserve the order
2. dedup() (breaking changes, document assuming the input is already sorted

PR #1064, opened 2024-10-08
add `Float::exp`

tonyfettes:
This PR adds `Float::exp`. The code is basically transcribed from the musl libc.

https://git.musl-libc.org/cgit/musl/tree/src/math/expf.c

### DIFF
--- a/float/exp.mbt
+++ b/float/exp.mbt
@@ -32,7 +32,6 @@ let exp2f_table_bits = 5
 
 priv struct Exp2fData {
   tab : Array[UInt64]
-  poly : Array[Double]
   shift : Double
   invln2_scaled : Double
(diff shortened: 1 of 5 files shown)


float/exp.mbt
```diff
@@ -0,0 +1,96 @@
...
+}
+
+fn __math_uflowf(sign : UInt) -> Float {
+  return __math_xflowf(sign, 0x1.0p-95)
+}
+
+let exp2f_table_bits = 5
+
+priv struct Exp2fData {
+  tab : Array[UInt64]
```
qazxcdswe123, line 34: They could be fixedarray?

qazxcdswe123 (approved):

PR #1074, opened 2024-10-09
inline field access on [Array]

Guest0x0:
A while ago field access on `Array` (`buf` and `len`) are migrated to helper methods. The intention of that migration is to compile `Array` to native array in JS backend. However, that goal is achieved in a different way via conditional compilation now. So it is no longer necessary to use helper methods for `buf` and `len` now.

This PR inlines `Array::buffer` and `Array::length` in `builtin`, and remove the private `Array::buffer` helper, to improve performance of various array operations.


hackwaly:
The js/nonjs versions share lots of impl based on `.buffer()`. You should make tests pass first.

PR #1084, opened 2024-10-10
create a dedicated buffer package and depercate @builtin.Buffer

### LINKED ISSUE #1081: create a dedicated buffer package and depercate @builtin.Buffer

hackwaly:
Closes #1081

Since we can't make builtin package depends on buffer package. So to make deprecation work: I simply left a Buffer impl in builtin package, and copy it's impl to buffer package.

Other notable changes:

- update docs to guide users to use StringBuilder instead of Buffer for string concatenation

In CI we can see the test driver still rely on `@builtin.Buffer`, this usage should migrate to `@buffer.Buffer` at moonbuild side. @Young-Flash


hackwaly:
The commit https://github.com/moonbitlang/core/pull/1084/commits/f466c5fdc93e9dd204e9c9b340c3d7ebaf1913b1 is out of scope. I'll file a new PR for it.

buffer/moon.pkg.json
```diff
@@ -1,4 +1,3 @@
 {
-  "import": ["moonbitlang/core/builtin"],
-  "test-import": ["moonbitlang/core/coverage"]
```
hackwaly: I can't understand why coverage need to listed in "import" but not "test-import". @lynzrand

hackwaly:
```
failed to read file `/Users/runner/work/core/core/coverage/__generated_driver_for_internal_test.mbt`, [2000] warning: Warning (Alert deprecated): Buffer is moved to buffer package, use @buffer.new instead
```

This error in CI is even weird. /cc @Young-Flash

buffer/moon.pkg.json
lynzrand: My best guess is that we didn't have `test-import` at all at the time when coverage was added.

You can see from `hashset/moon.pkg.json`: When coverage was added https://github.com/moonbitlang/core/blob/b5f6010e093b40c88d6dee36510f2979adcfbbad/hashset/moon.pkg.json; When `test-import` was first added (it was snake-case instead of kebab-case at that time): https://github.com/moonbitlang/core/blob/dbf949f10f95802d69f23150d26fc2bcb5f4b6c1/hashset/moon.pkg.json

Young-Flash:
> ```
> failed to read file `/Users/runner/work/core/core/coverage/__generated_driver_for_internal_test.mbt`, [2000] warning: Warning (Alert deprecated): Buffer is moved to buffer package, use @buffer.new instead
> ```
> 
> This error in CI is even weird. /cc @Young-Flash

nightly hadn't been release

buffer/moon.pkg.json
hackwaly: /cc @Young-Flash @lijunchen

bobzhang (approved):

PR #1085, opened 2024-10-10
refactor: use StringBuilder instead of Buffer

hackwaly:
This PR refactor all the usages of Buffer in core in favor of StringBuilder.

This PR is extracted from https://github.com/moonbitlang/core/pull/1084


hackwaly:
related https://github.com/moonbitlang/core/pull/1086

PR #1091, opened 2024-10-11
chore: change to_uint64 to reinterpret_as_uint64 to match the behavior of Int

qazxcdswe123:
Same as https://github.com/moonbitlang/core/pull/948, but for Uin64 for consistency


qazxcdswe123:
The current naming convention for 64bit int seems to be

```
  reinterpret_as_i64(Double) -> Int64
  reinterpret_as_u64(Double) -> UInt64
```

Is this intentional? `reinterpret_as_uint64` is indeed too long...

Yoorkin:
> Is this intentional? 

No, `reinterpret_as_i64` and `reinterpret_as_u64` were introduced at a very early stage.

qazxcdswe123:
> > Is this intentional?
> 
> No, `reinterpret_as_i64` and `reinterpret_as_u64` were introduced at a very early stage.

Should we also rename the i64 to `Int64` u64 to UInt64 for consistency?

bobzhang (approved):

Yoorkin:
> Should we also rename the i64 to `Int64` u64 to UInt64 for consistency?

rename to `reinterpret_as_int64` and `reinterpret_as_uint64` make sense to me

Issue #1098, opened 2024-10-11
`List::last` should return Option


Yoorkin:
https://github.com/moonbitlang/core/blob/85ab5cfbd5e93be37e601ad79d18594383741852/immut/list/list.mbt#L227

qazxcdswe123:
This will be changed on the next release.

In the current release I just deprecated it with guidance to migrate

bobzhang:
landed on master

PR #1103, opened 2024-10-12
fix: last return option

qazxcdswe123:
close: https://github.com/moonbitlang/core/issues/1098

### DIFF
--- a/immut/list/list.mbt
+++ b/immut/list/list.mbt
@@ -224,10 +224,10 @@ pub fn unsafe_last[A](self : T[A]) -> A {
 /// println(of([1, 2, 3, 4, 5]).last())
 /// // output: Some(5)
 /// ```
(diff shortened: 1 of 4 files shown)


lijunchen:
This PR breaks 

https://github.com/moonbitlang/quickcheck/blob/9071c4e394fae20bae31ce248320bdb6388bca6d/src/lib/axiom_wbtest.mbt#L33-L35

qazxcdswe123:
> This PR breaks
> 
> https://github.com/moonbitlang/quickcheck/blob/9071c4e394fae20bae31ce248320bdb6388bca6d/src/lib/axiom_wbtest.mbt#L33-L35

It's kinda unavoidable unless we deprecate it for one release period and change this in the next release. Maybe thats a better approach

PR #1106, opened 2024-10-12
add [Array::unsafe_get] and [Array::unsafe_set] & array intrinsic

Guest0x0:
This PR utilizes several new compiler intrinsics, `%array.{length,unsafe_get,get,unsafe_set,set}`. The unsafe variants have no bound check at all and may lead to arbitrary memory access, so they are kept private in `builtin`. The safe variants performs bound check on runtime and crashes on out of bound access. They will be inlined by the compiler in release mode for performance.

Since intrinsics are only utilized in release mode, so the use of unsafe operations won't make debug harder.

Previously many `Array` APIs in the core uses `arr.buffer()[index]` for unsafe get/set, but the `FixedArray` indexing still performs bound check at runtime. So these uses are replaced by `unsafe_get`/`unsafe_set` to maximize performance


bobzhang (approved):

bobzhang:
cc @hackwaly

PR #1117, opened 2024-10-12
Add `FixedArray::unsafe_get`, `FixedArray::unsafe_set`

tonyfettes:


builtin/builtin.mbti
```diff
@@ -618,6 +618,8 @@ impl FixedArray {
   to_json[X : ToJson](Self[X]) -> Json
   to_string[X : Show](Self[X]) -> String
   unsafe_blit[A](Self[A], Int, Self[A], Int, Int) -> Unit
+  unsafe_get[T](Self[T], Int) -> T
```
bobzhang, line 621: the unsafe API is not encouraged to be public, what is the motivation here?

tonyfettes: For performance critical code that use `FixedArray`, it might be better to eliminate all the boundary checks when access the element.

PR #1122, opened 2024-10-14
deprecate Js_string

Yu-zh:
We will support the Js-string-builtin proposal, so the legacy Js-string is removed.


bobzhang:
we can make a heads up warning first before removing it

Yu-zh:
> we can make a heads up warning first before removing it

Done. But now `moon info` will always complain the following because it is not smart enough:
```
/Users/whitepie/playground/ideas/core/builtin/console.mbt:156:5-156:17 [E2000] Warning (Alert deprecate): Use js-string-builtins api instead
/Users/whitepie/playground/ideas/core/builtin/console.mbt:156:5-156:17 [E2000] Warning (Alert deprecate): Use js-string-builtins api instead
```

PR #1136, opened 2024-10-17
feat: implement `(From|To)Json` for more `@immut/*.T` types

rami3l:
This pull request introduces JSON (de)serialization functionality to various immutable data structures (`list`, `sorted_map`, and `sorted_set`). The changes include implementing the `(From|To)Json` traits for these structures, adding corresponding methods, and updating the test cases to validate the new functionality.

As previously discussed with @bobzhang.

### DIFF
--- a/immut/list/list.mbt
+++ b/immut/list/list.mbt
@@ -29,6 +29,18 @@ pub fn T::to_string[A : Show](xs : T[A]) -> String {
   Show::to_string(xs)
 }
 
+pub impl[A : ToJson] ToJson for T[A] with to_json(self) {
+  let capacity = self.length()
+  guard capacity != 0 else { return Array([]) }
+  let jsons = Array::new(~capacity)
+  self.each(fn(a) { jsons.push(a.to_json()) })
+  Array(jsons)
+}
+
+pub fn T::to_json[A : ToJson](self : T[A]) -> Json {
+  ToJson::to_json(self)
+}
+
 /// Convert array to list.
 ///
 /// # Example
--- a/immut/list/list.mbt
+++ b/immut/list/list.mbt
@@ -41,6 +41,27 @@ pub fn T::to_json[A : ToJson](self : T[A]) -> Json {
   ToJson::to_json(self)
 }
 
+pub impl[A : @json.FromJson] @json.FromJson for T[A] with from_json(json, path) {
+  match json {
+    Array(arr) =>
+      for i = arr.length() - 1, list = Nil; i >= 0; {
+        continue i - 1, list.add(A::from_json!(arr[i], path))
+      } else {
+        list
+      }
+    _ =>
+      raise @json.JsonDecodeError(
+        (path, "@immut/list.from_json: expected array"),
+      )
+  }
+}
+
+pub fn T::from_json[A : @json.FromJson](
+  json : Json
+) -> T[A]!@json.JsonDecodeError {
+  @json.from_json!(json)
+}
+
 /// Convert array to list.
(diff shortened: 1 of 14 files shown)


immut/sorted_map/utils.mbt
```diff
@@ -257,3 +257,13 @@ pub impl[K : Show, V : Show] Show for T[K, V] with output(self, logger) {
 pub fn to_string[K : Show, V : Show](self : T[K, V]) -> String {
   Show::to_string(self)
 }
+
+pub impl[K : Show, V : ToJson] ToJson for T[K, V] with to_json(self) {
```
qazxcdswe123, line 261: maybe pre-alloc memory with capacity specified? As we've already know the size of the `jsons`. 

Same applies for other impl

rami3l: @qazxcdswe123 Addressed for `sorted_map`. However, lengths of `list`s and `sorted_set`s are (currently) not pre-calculated AFAIK, so it's probably out of the scope of this PR to make it so.

rami3l: @qazxcdswe123 Addressed for `list`s and `sorted_set`s by calculating the capacity required ad hoc.

immut/list/list.mbt
```diff
@@ -29,6 +29,18 @@ pub fn T::to_string[A : Show](xs : T[A]) -> String {
   Show::to_string(xs)
 }
 
+pub impl[A : ToJson] ToJson for T[A] with to_json(self) {
```
Yoorkin, line 32: FromJson is also needed

rami3l: @Yoorkin Can I implement this in another PR?

rami3l: @Yoorkin Unfortunately, I've found a bug in `@immut/sorted_set.T::op_equal()` that has broken my quicktests. I'll fix that first.

rami3l: @Yoorkin All fixed.

immut/list/moon.pkg.json
```diff
@@ -4,7 +4,8 @@
     "moonbitlang/core/test",
     "moonbitlang/core/array",
     "moonbitlang/core/coverage",
-    "moonbitlang/core/quickcheck"
+    "moonbitlang/core/quickcheck",
```
Yoorkin, line 7: It seems the list itself doesn't depend on QuickCheck, it should be moved to `test-import`

rami3l: @Yoorkin Do you mean this is an existing error?

rami3l: @Yoorkin Unfortunately, this will cause a build error like so:

```rust
      ╭─[[..]/moonbit-core/immut/list/list.mbt:1007:14]
      │
 1007 │ pub impl[X : @quickcheck.Arbitrary] @quickcheck.Arbitrary for T[X] with arbitrary(
      │              ──────────┬──────────  
      │                        ╰──────────── Package "quickcheck" not found in the loaded packages.
──────╯
```

Yoorkin: Yes. We have two import configuration: `import` for the package itself, and `test-import` for the dependencies used only in black box test.

rami3l: > Yes. We have two import configurations: `import` for the package itself, and `test-import` for the dependencies used only in black box test.

@Yoorkin Thanks. However, I don't think this can be removed that easily, as per my previous comment.

PR #1137, opened 2024-10-17
feat: add formjson for jsonvalue

qazxcdswe123:


qazxcdswe123:
@Yoorkin dce can eliminate the guard true.

PR #1141, opened 2024-10-18
refactor(to_string): change the behaviour of bytes to string

qazxcdswe123:
BREAKING CHANGE for bytes to string


builtin/bytes.mbt
```diff
@@ -14,7 +14,26 @@
...
-  self.sub_string(0, self.length())
+  let sb = StringBuilder::new(size_hint=self.length())
+  fn to_hex_digit(i : Int) -> Char {
+    if i < 10 {
+      Char::from_int('0'.to_int() + i)
+    } else {
+      Char::from_int('a'.to_int() + (i - 10))
+    }
+  }
+
```
bobzhang, line 25: can we print ascii as is?
e.g,    b"a\0b\0"

qazxcdswe123: It's kind hard to find the ascii inside a bunch of `\00`. Also encounter: `Fatal error: exception File "inlined_snapshot/_build/Core_format.ml", line 200, characters 8-14: Assertion failed`. Investigating

string/string_test.mbt
```diff
@@ -40,10 +40,42 @@ test "compare" {
...
+  inspect!(
+    "中文".to_bytes(),
+    content=
+      #|b"-N\x87e"
+    ,
+  )
+  inspect!(
+    "\n\t\"\\\r\b".to_bytes(),
+    content=
+      #|b"\n\x00\t\x00\"\x00\\\x00\r\x00\b\x00"
```
qazxcdswe123, line 52: I'm still not quite sure whether `\n` should yield `\\n` or `\n` like this

builtin/bytes.mbt
Yoorkin: > can we print ascii as is? e.g, b"a\0b\0"

displaying some byte as ASCII may not be a good idea. `Bytes` are not always converted from a String, they may come from data of an image or some pieces of memory. ASCII is meaningless in such cases.

qazxcdswe123: Second this as UTF-16 encoded string has lots of `\x00` and printing them is very ugly.

string/string_test.mbt
gmlewis: `\\n` should never be used unless the user absolutely wants to see the backslash, which is typically _not_ the case.

Issue #1155, opened 2024-10-22
Missing `Default` trait


qazxcdswe123:
- [x] UInt
- [x] Bytes
- [x] Float
- [x] Array
- [x] Json

PR #1157, opened 2024-10-22
impl Default for some types

qazxcdswe123:
close: #1155

### DIFF
--- a/builtin/json.mbt
+++ b/builtin/json.mbt
@@ -162,3 +162,7 @@ pub fn Result::to_json[Ok : ToJson, Err : ToJson](
 pub fn Unit::to_json(_self : Unit) -> Json {
   Null
 }
+
+pub fn Json::default() -> Json {
+  Null
+}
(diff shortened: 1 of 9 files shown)


builtin/json.mbt
```diff
@@ -162,3 +162,7 @@ pub fn Result::to_json[Ok : ToJson, Err : ToJson](
 pub fn Unit::to_json(_self : Unit) -> Json {
   Null
 }
+
+pub fn Json::default() -> Json {
+  {}
```
bobzhang, line 167: false? -- no allocations

bytes/bytes.mbt
```diff
@@ -69,3 +69,7 @@ pub fn iter(self : Bytes) -> Iter[Byte] {
     },
   )
 }
+
+pub fn Bytes::default() -> Bytes {
+  Bytes::new(1)
```
bobzhang, line 74: empty bytes? b""

PR #1160, opened 2024-10-23
Remove coverage import altogether

lynzrand:
Remove import of `coverage` from all modules.

Should be added after https://github.com/moonbitlang/moon/pull/415 is merged, so coverage imports are added automatically.


bobzhang:
@lynzrand I tested it on #1306, it does work, can you revive and finish this PR?

Young-Flash (approved):

PR #1162, opened 2024-10-23
feat: add `@cmp.(max|min)imum_by_key`

rami3l:
Mirrors Rust's [max_by_key](https://doc.rust-lang.org/std/cmp/fn.max_by_key.html#) and [min_by_key](https://doc.rust-lang.org/std/cmp/fn.min_by_key.html#) functions.

### DIFF
--- a/cmp/cmp.mbt
+++ b/cmp/cmp.mbt
@@ -1,0 +1,49 @@
+// Copyright 2025 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+///| Returns the element that gives the maximum value from the specified function.
+///
+/// Returns the second argument if the comparison determines them to be equal.
+///
+/// # Examples
+///
+/// ```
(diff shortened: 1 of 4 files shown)


bobzhang:
Do we need these tiny functions? I think they are not particularly useful

rami3l:
> Do we need these tiny functions? I think they are not particularly useful

I think it makes some business logic more readable, like the example below:

#### Before
```ocaml
  let span l0 l1 =
    let first_byte, first_line =
      if l0.first_byte < l1.first_byte
      then l0.first_byte, l0.first_line
      else l1.first_byte, l1.first_line
    in
    let last_byte, last_line, file =
      if l0.last_byte < l1.last_byte
      then l1.last_byte, l1.last_line, l1.file
      else l0.last_byte, l0.last_line, l0.file
    in
    v ~file ~first_byte ~first_line ~last_byte ~last_line
```

#### After
```rs
pub fn TextLoc::span(self : TextLoc, other : TextLoc) -> TextLoc {
  let { first_byte, first_line, .. } = minimum_by_key(self, other, fn(it) { it.first_byte })
  let { file, last_byte, last_line, .. } = maximum_by_key(self, other, fn(it) { it.last_byte })
  { file, first_byte, last_byte, first_line, last_line }
}
```

rami3l:
Python does the same thing with the `key` parameter in [`max(arg1, arg2, *args, key=None)`](https://docs.python.org/3/library/functions.html#max).

bobzhang:
@rami3l I think it is okay to have this util, but it does not belong to `math` package, can you create a `cmp` package like rust or `Fun` module like OCaml?

rami3l:
> @rami3l I think it is okay to have this util, but it does not belong to `math` package, can you create a `cmp` package like rust or `Fun` module like OCaml?

@bobzhang Thanks for the feedback! In that case, however, should `@math.(max|min)` be moved elsewhere as well?

rami3l:
@bobzhang Moved to `@cmp` as previously discussed.

PR #1176, opened 2024-10-30
fix some unused/never supplied optional argument

Guest0x0:
The bleeding compiler recently adds feature for detecting optional arguments that are never/always supplied. Several warnings are reported on core, this PR fixes the unused optional argument issues.

One notable case is a `inspect_double_array` function in `rational/rational_test.mbt`. This function is a wrapper over `inspect`, and has an optional argument `~content : String = ""` just like inspect. The problem is, this optional argument will be automatically filled by `moon test -u`, so its default value will only be used when the test is initially added. After the test result is promoted, the default value become unused.

This PR removes the `inspect_double_array` wrapper to fix the warning. `inspect` itself is public so no warning will be reported on it.

### DIFF
--- a/array/fixedarray_sort.mbt
+++ b/array/fixedarray_sort.mbt
@@ -293,10 +293,8 @@ fn fixed_get_limit(len : Int) -> Int {
 /// It will only tolerate at most 8 unsorted elements. The time complexity is O(n).
 /// 
 /// Returns whether the array is sorted.
-fn fixed_try_bubble_sort[T : Compare](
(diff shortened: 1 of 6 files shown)


bobzhang:
@Guest0x0 This PR itself looks good to me, but it is a good indicator that 

> optional arguments that are never/always supplied.

such warning should be opt in (off by default), e.g, only turned on for package authors who have a very high standard

bobzhang (approved):

Issue #1188, opened 2024-11-02
`Json::op_equal` returns false when the keys of 2 maps have different order


FlammeShadow:
[The commit](https://github.com/moonbitlang/core/commit/e5b54cb2bfe3806f5cd7b56b5adfa1c39640404b) in #522 implemented `op_equal` for `Map`. But no note is left in the documentation that its comparison (by pattern matching) is regarding to the idx order. As the contributor mentiond:
> from head to tail, both the idx(not sure about this), and the K, V should be the same, then we call them equal.

Due to this, the first assertion below will fail, but the second one will succeed.
```moonbit
test {
  let map1 = {
    "a": 1, "b": 2
  }
  let map2 = {
    "b": 2, "a": 1
  }
  let map3 = {
    "a": 1, "b": 2
  }
  assert_eq!(map1, map2)
  assert_eq!(map1, map3)
}
```
I'm not sure if map equality really requires the order of keys to be equal. However, though the comparison between `Json` obviously does not require this, according to its definition:
```moonbit
pub enum Json {
  Null
  True
  False
  Number(Double)
  String(String)
  Array(Array[Json])
  Object(Map[String, Json])
} derive(Eq)
```
Its equal operator is derived from `Map`, whose comparison is regarding to the order. So `assert_eq!({"1": 2, "2": 2}.to_json(), {"2": 2, "1": 1})` will fail. I think this behavior should be corrected.

PR #1191, opened 2024-11-03
fix: Json::op_equal returns false when the keys of 2 maps have different order

Lampese:
- #1188


bobzhang (approved):
In Dart, two maps are considered equal if they contain the same keys and values, and the values for each key are equal. However, order matters for Map equality in Dart; this is because Dart Maps, by default, preserve insertion order.

For example:

```dart
void main() {
  var map1 = {'a': 1, 'b': 2};
  var map2 = {'b': 2, 'a': 1};

  print(map1 == map2); // false, because the order is different
}
```

PR #1193, opened 2024-11-04
add constant PI to math module and update trigonometric functions to use it

bobzhang:


math/trigonometric.mbt
```diff
@@ -18,9 +18,12 @@ let pi4b = 0x0.000000A2216800000988
 
 let pi4c = 0x0.000000000000C234BE07
 
+pub const PI = 0x3.243F6A8885A308CA8A54
+
 pub let pi = 0x3.243F6A8885A308CA8A54
 
-let pi4 : Double = 4.0 / pi
+// todo: allow const here
+let pi4 : Double = 4.0 / PI
```
bobzhang, line 27: we  need support this as well:
```
const Pi4 = 4.0 / PI
```

Lampese: It means we need to add a new keyword `const`?

Lampese (commented):
In Nightly's CI, Moon Check seems to believe that PI is a type rather than a variable (because it starts with uppercase). I don't know if the problem lies here, but that's why I didn't use uppercase when I first wrote this constant.

math/trigonometric.mbt
bobzhang: we already made const a keyword

bobzhang:
> In Nightly's CI, Moon Check seems to believe that PI is a type rather than a variable (because it starts with uppercase). I don't know if the problem lies here, but that's why I didn't use uppercase when I first wrote this constant.

Might be that nightly CI is not the latest one, cc @Young-Flash we came across this issue before

PR #1203, opened 2024-11-08
feat: implement Exp trait for generic exponentiation

bobzhang:


math/exp.mbt
```diff
@@ -12,6 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-pub fn exp(input : Double) -> Double {
+trait Exp {
+  exp(Self) -> Double
+}
```
bobzhang, line 17: what's the expected return type of exp function?
should it be `Self`, `Double` or `Float`?

hackwaly: I think `Self` is a proper choose.

If `Int.exp(x)` need a `Double` result, users always can turn it to `Double` first.

peter-jerry-ye:
Closing this PR as per previous discussion, we will have `exp` for `(Double) -> Double` and `expf` for `(Float) -> Float` and we will not providing `Floating` or similar traits.

Issue #1069, opened 2024-10-09
Proposal: add func for `float` in `@math` pkg with postfix `f`


qazxcdswe123:
> Unfortunately unless we bring back Num trait in X and make it generic, it's not possible to overload the function with different argument and same name.

IMO adding `f` to every `Float` function looks good enough, like:

- `lnf`
- `logf`
- `sinf`
- `cosf`
- ...

_Originally posted by @tonyfettes in https://github.com/moonbitlang/core/issues/1050#issuecomment-2398611224_

@bobzhang

bobzhang:
we are planning to allow some overriding so that `ln ( self : Float | Double)` could both work

Lampese:
> we are planning to allow some overriding so that `ln ( self : Float | Double)` could both work

I have two worries about this solution:

- I know we can use self as Float/Double, and then how to infer type if `let x = self` and use the funcitons of self?
- There is a problem that the algorithm of sin/cos/tan.... is based on the IEEE754 and double's bit. If we implement same algorithm for float, then this algorithm will be error. I think sin(self: Float) -> self.to_double().sin().to_float() is even a good solution.

peter-jerry-ye:
We can implement this with sealed trait now. @Lampese @tonyfettes what do you think of the correctness of the implementation?

peter-jerry-ye:
We will choose the option of implementing functions for `float` using postfix `f` because:
- Sealed trait will not work with literal overloading. `1` can be interpreted as `Double` when called with a function `(Double) -> Double` but can only be interpreted as `Int` when called with a function `[T : Floating](T) -> T`.
- Float version requires special care, and is not often used.

We will also migrate `double.cos` to `@math.cos` to accelerate the compilation.

Kaida-Amethyst:
We'll make math functions could accept both `Double` and `Float` type, like:

```moonbit
let x : Double = 1.0;
println("cos(x) = \{@math.cos(x)}"); \\ ok

let y: Float = 1.0; // <-- Float Type
println("cos(x) = \{@math.cos(x)}"); \\ ok
```

PR #1204, opened 2024-11-08
add abstract Floating trait. todo: impl Floating for Float

bobzhang:


math/float.mbt
```diff
@@ -0,0 +1,53 @@
...
+
+trait Floating {
+  asin(Self) -> Self
+  // ideas: `as`
+  // or use @math.Floating::asin(x)  
+  // how about 
+  // pub fn asin(x : Self) -> Self 
+  acos(Self) -> Self
+}
+
```
bobzhang, line 23: note we may introduce a sugar like this:
```
pub trait Floating::(asin, acos)
```

peter-jerry-ye:
Closing due to the reason stated here: https://github.com/moonbitlang/core/issues/1069#issuecomment-2803930203

PR #1213, opened 2024-11-12
[draft] weird error message, `T::of` vs `of`

bobzhang:


Yu-zh:
Seems that some packages are imported unexpectedly. I will take a look.

bobzhang:
this should be fixed with the new semantics cc @Guest0x0

PR #1218, opened 2024-11-14
 linked_hash_set是否可以将insert函数和remove函数加上返回值

suiyunonghen:
insert函数返回Bool,如果key存在，返回false，表示存在插入失败，不存在返回true，插入成功
remove函数也返回Bool,如果key存在返回true，表示删除成功，key不存在返回false，表示删除失败

### DIFF
--- a/builtin/linked_hash_set.mbt
+++ b/builtin/linked_hash_set.mbt
@@ -70,13 +70,13 @@ pub fn Set::new[K](~capacity : Int = 8) -> Set[K] {
 /// Create a hash set from array.
 pub fn Set::from_array[K : Hash + Eq](arr : Array[K]) -> Set[K] {
   let m = Set::new(capacity=arr.length())
-  arr.each(fn(e) { m.insert(e) })
+  arr.each(fn(e) { let _ = m.insert(e) })
   m
 }
 
 ///|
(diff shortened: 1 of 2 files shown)


hackwaly:
To keep API consistent, it's better to make other Map/Set::add methods returns Bool, too.

To align this change, immut/xx_set needs to return tuple. It may cost unexpected more.

There is a low cost workaroud if we don't introduce this change.

```
let old_size = set.size()
set.add(...)
let success = set.size() > old_size
```

Yoorkin:
Since MoonBit has cascade operator, the method with a `Unit` return type are useful, e.g.

```
let result = 
  @set.of([1,2,3,4])
  ..add(5)
  ..remove(4)
  ..remove(3)
  .to_string()
```

The `remove` method can't be chained by `..` if it return `Bool` or `Option[K]`. Even worse is that you cannot discard return value without `let _ = ...` or `... |> ignore`.  So I agree with @hackwaly. 

I propose to add another method:

```
fn pop[K](self : Set[K], elem : K) -> Option[K]
fn pop[K,V](self : Map[K,V], k : K) -> Option[V]
```

`pop` may not be a suitable name, here are some alternative: `extract`, `move_out`

PR #1231, opened 2024-11-19
refactor buffer to use `FixedArray[Byte]` instead of `Bytes`

Yu-zh:
This refactor is a preparation for making `Bytes` immutable. It basically changes the use of `Bytes` in the type `Buffer` to `FixedArray[Byte]`, and added `Bytes::cast_from_fixedarray` and `cast_to_fixedarray` to convert between `Bytes` and `FixedArray[Byte]`. Currently, these casts need to allocate, and we plan to implement these casts as primitives in the compiler to avoid the allocation eventually.

### DIFF
--- a/buffer/buffer.mbt
+++ b/buffer/buffer.mbt
@@ -33,16 +33,16 @@ pub(all) typealias Buffer = T
 /// println(buf.to_string()) // output: Test
 /// ```
 struct T {
-  mut bytes : FixedArray[Byte]
+  mut data : FixedArray[Byte]
   mut len : Int
-  initial_bytes : FixedArray[Byte]
+  initial_data : FixedArray[Byte]
 }
 
 ///|
 /// Expand the buffer size if capacity smaller than required space.
 fn grow_if_necessary(self : T, required : Int) -> Unit {
   // TODO: get rid of mut
-  let mut enough_space = self.bytes.length()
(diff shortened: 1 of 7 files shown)


buffer/buffer.mbt
```diff
@@ -108,7 +116,7 @@ pub fn write_object(self : T, value : Show) -> Unit {
 pub fn write_bytes(self : T, value : Bytes) -> Unit {
   let val_len = value.length()
   self.grow_if_necessary(self.len + val_len)
-  self.bytes.blit(self.len, value, 0, val_len)
+  self.bytes.unsafe_blit(self.len, value.cast_to_fixedarray(), 0, val_len)
   self.len += val_len
```
bobzhang, line 120: blit_from_bytes

bytes/bytes.mbt
```diff
@@ -29,6 +29,32 @@ pub fn Bytes::from_array(arr : Array[Byte]) -> Bytes {
   rv
 }
 
+///|
+/// Makes a new Bytes from a fixedarray.
+pub fn Bytes::from_fixedarray(
+  arr : FixedArray[Byte],
+  len~ : Int = arr.length()
```
bobzhang, line 36: let's not use this feature, we may remove the labelled dependency to keep the language simple

PR #1232, opened 2024-11-19
Enhance round functions

hiawui:
1. refactor round functions
2. use JS / WASM native round functions
3. add round functions to float


math/round.mbt
```diff
@@ -19,7 +19,7 @@ pub fn round(input : Double) -> Double {
 
 test "round" {
   assert_eq!(round(4503599627370497.5), 4503599627370498.0)
-  assert_eq!(round(2251799813685248.5), 2251799813685249.0)
+  assert_eq!(round(2251799813685248.5), 2251799813685248.0)
```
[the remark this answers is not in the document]
hackwaly: Yes


Number following the rounding digit | Rounding direction | Example
-- | -- | --
0 - 4 | Rounded down | 2.121->2.12
5 (being the last digit) | The rounding digit is an even number. | Rounded down | 2.125->2.12
The rounding digit is an odd number. | Rounded up | 2.135->2.14
5 (not being the last digit) | Rounded up | 2.1351->2.14
6 - 9 | Rounded up | 2.127->2.13

hackwaly:
This is a breaking change, we should consider it seriously. We need investigate other language's choice. @hiawui 

The alternative is to keep banker's round an another method `Double::bankers_round`.

hiawui:
> This is a breaking change, we should consider it seriously. We need investigate other language's choice. @hiawui
> 
> The alternative is to keep banker's round an another method `Double::bankers_round`.

OK. I will wait for your investigation, and I will update according to your decisions.

hackwaly:
@hiawui Could you split non-disputed part (eg. trunc, ceil, floor) into separated PR? So we can merge those changes first.

hiawui:
> @hiawui Could you split non-disputed part (eg. trunc, ceil, floor) into separated PR? So we can merge those changes first.

@hackwaly Can I continue using this PR while only reverting the rounding algorithm to how it was before?

hackwaly:
@hiawui Certainly.

double/round.mbt
```diff
@@ -59,31 +59,13 @@ pub fn floor(self : Double) -> Double {
...
-  }
-  let mask_shift = biased_exp - exp_bias + exp_bits
-  let parity_mask = sign_mask >> mask_shift
-  return if (u32 & parity_mask) > 0 { 1 } else { 0 }
-}
-
-///|
-// Round to nearest, ties to Even
+// Round to nearest, ties to Ceiling
 pub fn round(self : Double) -> Double {
```
hackwaly, line 63: `floor(self + 0.5)`?

hiawui: This is better

hackwaly:
lgtm.  @Lampese

hackwaly (approved):

Lampese (approved):
LGTM.

PR #1237, opened 2024-11-22
Add append_fast and set methods to the array

suiyunonghen:
I have looked at the methods of batch adding arrays to array, and they all directly use push to add them one by one. This method is not a problem when dealing with small amounts of data, but when dealing with large amounts of data, it can be time-consuming. Recently, [when implementing msgpack encoding](https://github.com/suiyunonghen/moonvalue), I have been trying to find out if there is an efficient batch operation function in the core library. Blit is what I need, but I found that it limits the length of the array. Therefore, I adjusted two functions based on the relevant code, both of which can automatically expand and use batch operation functions, thus reducing the instruction count

### DIFF
--- a/builtin/builtin.mbti
+++ b/builtin/builtin.mbti
@@ -47,6 +47,7 @@ type Array
 impl Array {
   append[T](Self[T], Self[T]) -> Unit
   append_fast[T](Self[T], Self[T]) -> Unit
+  appendex[T](Self[T], Self[T], Int, Int) -> Unit
   binary_search[T : Compare + Eq](Self[T], T) -> Result[Int, Int]
   binary_search_by[T](Self[T], (T) -> Int) -> Result[Int, Int]
   blit_to[A](Self[A], Self[A], len~ : Int, src_offset~ : Int = .., dst_offset~ : Int = ..) -> Unit
--- a/builtin/builtin.mbti
+++ b/builtin/builtin.mbti
@@ -46,6 +46,7 @@ impl Show for ArgsLoc
 type Array
 impl Array {
   append[T](Self[T], Self[T]) -> Unit
+  append_fast[T](Self[T], Self[T]) -> Unit
   binary_search[T : Compare + Eq](Self[T], T) -> Result[Int, Int]
   binary_search_by[T](Self[T], (T) -> Int) -> Result[Int, Int]
   blit_to[A](Self[A], Self[A], len~ : Int, src_offset~ : Int = .., dst_offset~ : Int = ..) -> Unit
@@ -106,6 +107,7 @@ impl Array {
   rev_inplace[T](Self[T]) -> Unit
   search[T : Eq](Self[T], T) -> Int?
   search_by[T](Self[T], (T) -> Bool) -> Int?
+  set[T](Self[T], Int, Self[T], Int, Int) -> Unit
   shrink_to_fit[T](Self[T]) -> Unit
   split[T](Self[T], (T) -> Bool) -> Self[Self[T]]
   split_at[T](Self[T], Int) -> (Self[T], Self[T])
--- a/builtin/builtin.mbti
+++ b/builtin/builtin.mbti
@@ -45,9 +45,7 @@ impl Show for ArgsLoc
 
 type Array
 impl Array {
-  append[T](Self[T], Self[T]) -> Unit
-  append_fast[T](Self[T], Self[T]) -> Unit
-  appendex[T](Self[T], Self[T], Int, Int) -> Unit
+  append[T](Self[T], Self[T], from~ : Int = .., len~ : Int = ..) -> Unit
   binary_search[T : Compare + Eq](Self[T], T) -> Result[Int, Int]
   binary_search_by[T](Self[T], (T) -> Int) -> Result[Int, Int]
   blit_to[A](Self[A], Self[A], len~ : Int, src_offset~ : Int = .., dst_offset~ : Int = ..) -> Unit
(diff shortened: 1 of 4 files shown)


suiyunonghen:
图片
Local compilation and running are all successful, I don't know why I submitted the compilation. However, let's take a look at the difference between append_fast and append
```
pub fn fastappend()->Unit{  
  for i in 0..<1000_0000{
    let b: Array[Byte] = [1,2,3,4,5,6,7,8,9,10,11,20,3,5,6,6,6,6,6,6,9]
    let b2: Array[Byte]=[1,2,3,4,5,6,7,8,9,10,11,20,3,5,6,6,6,6,6,6,9]
    b.append_fast(b2)
  }
}

pub fn append()->Unit{  
  for i in 0..<1000_0000{
    let b: Array[Byte] = [1,2,3,4,5,6,7,8,9,10,11,20,3,5,6,6,6,6,6,6,9]
    let b2: Array[Byte]=[1,2,3,4,5,6,7,8,9,10,11,20,3,5,6,6,6,6,6,6,9]
    b.append(b2)
  }
}
```

Lampese:
> 图片 Local compilation and running are all successful, I don't know why I submitted the compilation. However, let's take a look at the difference between append_fast and append
> 
> ```
> pub fn fastappend()->Unit{  
>   for i in 0..<1000_0000{
>     let b: Array[Byte] = [1,2,3,4,5,6,7,8,9,10,11,20,3,5,6,6,6,6,6,6,9]
>     let b2: Array[Byte]=[1,2,3,4,5,6,7,8,9,10,11,20,3,5,6,6,6,6,6,6,9]
>     b.append_fast(b2)
>   }
> }
> 
> pub fn append()->Unit{  
>   for i in 0..<1000_0000{
>     let b: Array[Byte] = [1,2,3,4,5,6,7,8,9,10,11,20,3,5,6,6,6,6,6,6,9]
>     let b2: Array[Byte]=[1,2,3,4,5,6,7,8,9,10,11,20,3,5,6,6,6,6,6,6,9]
>     b.append(b2)
>   }
> }
> ```

I guess that because we have a JavaScript backend and our Array can be directly compile to JavaScript's native Array so it makes some mistake?

cc @hackwaly

hackwaly:
@Lampese Yes. Array is implemented platform specific in `arraycore_js.mbt`, `arraycore_nonjs.mbt`.

bobzhang:
@suiyunonghen the feature request looks reasonable to me.
for append_fast and appendex, we can just change the existing API instead of adding two new functions. e.g
```
append[T](Self[T], Self[T], offset? : Int, len? : Int ) -> Unit
```
and for `set` function, it looks reasonable to me that we just relax blit_to to allow it grow when size is not enough?

suiyunonghen:
@bobzhang yes，I just need to wait for the official response, don't I need to adjust the pull request anymore？

bobzhang:
> I just need to wait for the official response, don't I need to adjust the pull request anymore？

@suiyunonghen Can you adapt the API based on my suggestions, we can check the correctness later

suiyunonghen:
@bobzhang blit_to has implemented automatic expansion, and append_fast has been merged into append. Let's see if it works now

bobzhang:
cc @suiyunonghen can you make this branch editable (not protected) so that we can  push atop

hackwaly:
> cc @suiyunonghen can you make this branch editable (not protected) so that we can push atop

PR branch can be edited by using "Github Desktop".

suiyunonghen:
@bobzhang When forking, I didn't pay much attention to whether there was protection. Now I have created a branch ruleset like this, I don't know if it counts as deprotection
图片

builtin/builtin.mbti
```diff
@@ -106,6 +106,7 @@ impl Array {
   rev_inplace[T](Self[T]) -> Unit
   search[T : Eq](Self[T], T) -> Int?
   search_by[T](Self[T], (T) -> Bool) -> Int?
+  set[T](Self[T], Int, Self[T], Int, Int) -> Unit
```
bobzhang, line 109: why this one is needed?

builtin/builtin.mbti
```diff
@@ -45,7 +45,7 @@ impl Show for ArgsLoc
 
 type Array
 impl Array {
-  append[T](Self[T], Self[T]) -> Unit
+  append[T](Self[T], Self[T], from~ : Int = .., len~ : Int = ..) -> Unit
   binary_search[T : Compare + Eq](Self[T], T) -> Result[Int, Int]
   binary_search_by[T](Self[T], (T) -> Int) -> Result[Int, Int]
   blit_to[A](Self[A], Self[A], len~ : Int, src_offset~ : Int = .., dst_offset~ : Int = ..) -> Unit
```
bobzhang, line 51: I suggest relax `blit_to` to allow grow automatically

bobzhang: @suiyunonghen I have some problem in pushing this branch directly (might be your name is also `main`). Can you clean up your commit message to make it a single commit given the commit is not that large, thanks

builtin/builtin.mbti
suiyunonghen: It's not necessary, it just retains a function similar to JavaScript's array, which can also be removed

builtin/builtin.mbti
suiyunonghen: I deleted the main branch and created a new master as the main branch

bobzhang:
@suiyunonghen we implemented this feature request in #1279, so we allow `Array::blit_to` to grow by need, it should solve your performance issue, let me know if it works

suiyunonghen:
> @suiyunonghen we implemented this feature request in #1279, so we allow `Array::blit_to` to grow by need, it should solve your performance issue, let me know if it works

Performance is no longer a problem. I tested it and found that there was a 7-fold improvement from 5000 milliseconds before to 700 milliseconds now

PR #1243, opened 2024-11-25
Show::to_string is enough

bobzhang:


builtin/fixedarray.mbt
```diff
@@ -42,9 +42,10 @@ pub fn iter2[T](self : FixedArray[T]) -> Iter2[Int, T] {
 }
 
 ///|
-pub fn to_string[X : Show](self : FixedArray[X]) -> String {
-  Show::to_string(self)
-}
+// pub fn to_string[X : Show](self : FixedArray[X]) -> String {
+//   // Show::to_string(self)
+//   self.to_string() // this may cause nonterminated recursion
+// }
```
bobzhang, line 48: this is actually non-terminating, we may detect this cc @Yu-zh

PR #1246, opened 2024-11-26
feat(buffer):  `to_unchecked_string` method optional params

jetjinser:
`offset? : Int` and `length? : Int`


bobzhang:
@jetjinser when do you need this API, @buffer.to_string? instead of @buffer.contents which return bytes?

jetjinser:
> @jetjinser when do you need this API, @buffer.to_string? instead of @buffer.contents which return bytes?

Yes, I used `buffer.T` when I wrote the encoding package before, but it seemed that the usage was not appropriate, so I don’t use it now. :smiling_face_with_tear:

I think this PR can be closed.

Issue #1148, opened 2024-10-20
Make `String::concat` take `ArrayView[T]`


tonyfettes:
I was writing a MoonBit code generate to automate the implementation of some trait for tuple in MoonBit, and when I was trying to build the type of tuple `(A, B, C, ...)`, I found there is no way to concat an `ArrayView[String]` into a `String` with some separator.

Example code:

```moonbit
let type_names = ["A", "B", ...]

fn build_tuple_type(length : Int) -> String {
  String::concat(type_names[:length], separator=", ")
}
```

Therefore, I wish there is a way to concat an `ArrayView[T]` with a separator.

Yu-zh:
You are encouraged to use Iter for this kind of issue.
```
let type_names: Array[String] = ["A", "B"]

fn build_tuple_type(length : Int) -> String {
  type_names[:length].iter().join("")
}
```

PR #1249, opened 2024-11-26
feat: add String::concat_view

Lampese:
- #1148

### DIFF
--- a/string/string.mbt
+++ b/string/string.mbt
@@ -32,16 +32,16 @@ pub fn String::from_array(chars : Array[Char]) -> String {
 }
 
 ///|
-/// Concatenate strings.
+/// Concatenate strings by a string array view.
 /// 
 /// ```
-/// let s = @string.concat(["Hello", ", ", "world!"])
+/// let s = @string.concat_view(["Hello", ", ", "world!"][:])
 /// assert_eq!(s, "Hello, world!")
-/// let s = @string.concat(["a", "b", "c"], separator=",")
+/// let s = @string.concat(["a", "b", "c"][:], separator=",")
 /// assert_eq!(s, "a,b,c")
 /// ```
-pub fn String::concat(
-  strings : Array[String],
+pub fn String::concat_view(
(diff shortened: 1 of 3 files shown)


string/string.mbt
```diff
@@ -64,6 +64,22 @@ pub fn String::concat(
   buf.to_string()
 }
```
bobzhang, line 65: what's the rationale to make it a plain function istead of a method?

Lampese:
> How about:
> 
> ```
> pub(open) trait IterString {
>   iter(Self) -> Iter[String]
> }
> 
> pub fn concat[T : IterString](strings : T, separator~ : String = "") -> String {
>   let buf = StringBuilder::new()
>   let mut str : String? = None
>   for s in strings {
>     if str.is_empty().not() {
>       buf.write_string(str.unwrap())
>       buf.write_string(separator)
>     }
>     str = Some(s)
>   } else {
>     if str.is_empty().not() {
>       buf.write_string(str.unwrap())
>     }
>   }
>   buf.to_string()
> }
> 
> test {
>   inspect!(concat(([] : Array[String])), content="")
>   inspect!(concat(["a", "b", "c"]), content="abc")
>   inspect!(concat(["a", "b", "c"], separator="|"), content="a|b|c")
>   inspect!(concat(["a", "b", "c"][1:], separator="|"), content="b|c")
>   inspect!(
>     concat((["a", "b", "c"] : FixedArray[String]), separator="|"),
>     content="a|b|c",
>   )
> }
> ```

It is a quite feasible plan

peter-jerry-ye:
We are now considering moving it to `Iter[String]::join`

Yoorkin:
`String::concat` was a mistake. In the self parameter, the types in the type constructor can be a concrete type like String. After discussing offline, we will move to `Iter[String]::join` and make `String::concat` a function. The join method will be added later because users can always use other methods instead.

Lampese:
- #1257

PR #1252, opened 2024-11-26
feat(buffer): add `op_set` and `op_get` operator

jetjinser:


buffer/buffer.mbt
```diff
@@ -173,6 +173,20 @@ pub fn to_bytes(self : T) -> Bytes {
   Bytes::from_fixedarray(self.data, len=self.len)
 }
 
+///|
+pub fn op_set(self : T, index : Int, value : Byte) -> Unit {
+  let len = self.length()
+  guard index >= 0 && index < len
+  self.data[index] = value
+}
```
bobzhang, line 181: buffer is an opaque type, when is this API needed?

jetjinser: I made a mistake, `buffer` doesn't need random access, just use `bytes`. I think the opaque nature can be retained and I'll close this pr.

PR #1254, opened 2024-11-26
feat(buffer): new api from_{bytes, array, iter}

jetjinser:


buffer/buffer.mbt
```diff
@@ -99,6 +99,31 @@ pub fn T::new(size_hint~ : Int = 0) -> T {
...
+  }
+  buf
+}
+
+///|
+/// Create a buffer from an array.
+pub fn T::from_iter(iter : Iter[Byte]) -> T {
+  let arr = iter.collect()
+  T::from_array(arr)
+}
```
peter-jerry-ye, line 125: How about
```moonbit
for byte in iter {
  buf.write_byte(byte)
}
```

And let the `T::from_array` just use `T::from_iter`?

jetjinser: Yes, the changes to `from_iter` should be reasonable I think, there is no need to allocating memory to construct an array.

But if use `T::from_iter` to impl `T::from_array`, `from_array` will not be able to use the `array.length()` as `size_hint~` information :thinking:. I think it should be fine just to keep it as it is?

jetjinser: Should let `@buffer.new` use `iter.count()` as size_hint?

I guess `arr.length()` is cheaper than `iter.count()`, so I thought `from_array` should keep it as is.

peter-jerry-ye: `iter.count` means run the `iter` once and count each element, which is not what we want. But I'm not quite sure about collecting first and then use `T::from_array` though. Because when we are collecting, we also have the underlying expansion logic upon the array. Also we would have an extra copy iteration.

So I think it would be reasonable to have two separate implementations instead of one relying on another.

buffer/buffer.mbt
```diff
@@ -99,6 +99,31 @@ pub fn T::new(size_hint~ : Int = 0) -> T {
   { data, len: 0, initial_data: data }
 }
 
+///|
+/// Create a buffer from a bytes.
+pub fn T::from_bytes(bytes : Bytes) -> T {
+  let buf = T::new(size_hint=bytes.length())
+  buf.write_bytes(bytes)
```
peter-jerry-ye, line 106: We should inline `write_bytes`'s logic here, otherwise we would have an uncessary `grow_if_necessary` check inside the `write_bytes`.

peter-jerry-ye:
Sorry for hanging this PR for so long. I adjusted the API according to the current convention, and fixed the issue with `from_iter`.

PR #1259, opened 2024-11-27
feat: new min/max api for `Int(64)`

jetjinser:
new helper api:
- `Int::min`, `Int::max`
- `Int64::min`, `Int64::max`


hackwaly:
Redundant? There is `@math.maximum`

jetjinser:
> Redundant? There is `@math.maximum`

Ah! I missed it!

Issue #1029, opened 2024-09-21
[feature request] add error-awared functions like: `makei_error`, `map_error`


tonyfettes:
It's often the case when I want to construct an `Array[T]` or `FixedArray[T]` from a function that can raise error. For example, I want to construct a `Tensor` from a `Json` value:

```moonbit
fn main {
  match json {
    Json::Number(x) => Tensor::new(x)
    Json::Array(xs) => {
      let value : FixedArray[Tensor] = FixedArray::makei(xs.length(), fn(i) {
        Tensor::from_json!(xs[i], @json.add_index(path, i))
      })
      Tensor::stack(value)
    }
    _ => raise @json.JsonDecodeError((path, "Expected an array"))
  }
}
```

For now it is not possible to achieve this, since `FixedArray::makei` require the callback to be error-free.

htoooth:
@tonyfettes  Maybe the functions of all `core` module have to add the `error` feature. But i wish you do not use error feature in you code. Because this feature influences all related code if you use it. Just like now scene you face.

bobzhang:
```
let value : FixedArray[Tensor] = FixedArray::makei(xs.length(), fn(i) {
        Tensor::from_json!(xs[i], @json.add_index(path, i))
      })
Tensor::stack(value)
```
Note you work around this using `for  .. in `

FlammeShadow:
> ```
> let value : FixedArray[Tensor] = FixedArray::makei(xs.length(), fn(i) {
>         Tensor::from_json!(xs[i], @json.add_index(path, i))
>       })
> Tensor::stack(value)
> ```
> Note you work around this using `for  .. in `, I am hestitant to add such API since eventually want our `map` to propagate callbacks with Error something like this:
> ```
> map[A](self : Array[A], f : (A) ->B!_) -> Array[B]!_ // when call back has error, result has error
> ```

The function type `(A) -> B` is incompatitble with `(A) -> B!Error` at now. Is `!_` here a syntax that imply that the error type can be unused (which will be reported as an error currently)?

FlammeShadow:
> ```
> let value : FixedArray[Tensor] = FixedArray::makei(xs.length(), fn(i) {
>         Tensor::from_json!(xs[i], @json.add_index(path, i))
>       })
> Tensor::stack(value)
> ```
> Note you work around this using `for  .. in `, I am hestitant to add such API since eventually want our `map` to propagate callbacks with Error something like this:
> ```
[...]

PR #1260, opened 2024-11-27
feat: new api `try_collect` for builtin/iter

jetjinser:
Collects the `Ok` elements of the iterator into an array, short circuiting if a `Err` is encountered.


tonyfettes:
See https://github.com/moonbitlang/core/issues/1029

peter-jerry-ye:
We will support effect polymorphism in the near future.

PR #1269, opened 2024-11-29
feat(builtin): add `dump()` for easy print-based debugging

rami3l:
[`dbg!()`](https://doc.rust-lang.org/std/macro.dbg.html) is a very popular Rust `std` macro that can be used like so:

```rust
let a = 2;
let b = dbg!(a * 2) + 1;
//      ^-- prints: [src/main.rs:2:9] a * 2 = 4
assert_eq!(b, 5);
```

... note that `a * 2` is not only **printed for debugging** but also **passed on to the rest of the calculation**, which makes it a perfect fit for inspecting complex evaluations such as those of chained method calls.

The idea of this PR is to mimic the behavior of that macro with what is currently available with MoonBit's compiler magic, i.e. `loc~ : SourceLoc = _` and `"%any.to_string"`.

The signature of this function `dump()` is as follows:

```rust
pub fn dump[T](t : T, name? : String, loc~ : SourceLoc = _) -> T 
```

... where the identifiers `dump` and `name` are taken from [a similar Swift function](https://developer.apple.com/documentation/swift/dump(_:name:indent:maxdepth:maxitems:)).


builtin/console.mbt
```diff
@@ -26,6 +29,18 @@ pub fn print[T : Show](input : T) -> Unit {
...
+/// @alert deprecated "This function is for debugging only and should not be used in production"
+pub fn dump[T](t : T, name? : String, loc~ : SourceLoc = _) -> T {
+  let name = match name {
+    Some(name) => name
+    None => ""
+  }
+  println("dump(\{name}@\{loc}) = \{any_to_string(t)}")
+  t
+}
+
```
bobzhang, line 43: any_to_string is introduced for arbitrary code transformations where `T` is not sure if it satsifies `Show`. How is this different from `dump[T:Show](...)`
cc @Guest0x0 it makes sense to introduce support of ArgsRepr which is essentialy a stringified repr of args. so that we can do something like this:
```mbt
pub fn dump[T:Show](t : T, name~ : ArgsRepr = _, loc~ : SourceLoc = _) -> T
```

rami3l: @bobzhang One thing that distinguishes Swift's `dump()` and Rust's `dbg!()` is that the former doesn't pose any restrictions on `T` whereas the latter depends on the `Debug` trait (which is exclusively used for log-oriented pretty-printing).

I think the `Show` trait in MoonBit corresponds to the `Display` trait rather than `Debug`, so it makes more sense to add no bounds at all.

OTOH `ArgsRepr` will definitely be a nice addition.

PR #1274, opened 2024-12-01
fix(array): quick_sort store wrong pred pivot element

yjl9903:
We should store the pred pivot element before updating arr.

And, there is a test case added which will trigger this bug.


Yu-zh:
lgtm. Thank you for fixing this bug.

Yu-zh (approved):

PR #1278, opened 2024-12-02
deprecate `Bytes::blit_from_string` and `Bytes::set_utf16_char`

Yu-zh:
This PR:
1) deprecated `Bytes::blit_from_string` and `Bytes::set_utf16_char`
2) refactored stringbuilder_buffer.mbt to use `FixedArray[Byte]` instead of `Bytes`
3) Added `FixedArray::reinterpreted_as_bytes` and `Bytes::reinterpreted_as_fixedarray` for non-copy conversion between the two types.

### DIFF
--- a/buffer/buffer.mbt
+++ b/buffer/buffer.mbt
@@ -85,10 +85,7 @@ pub fn to_string(self : T) -> String {
 /// Note this function does not validate the encoding of the byte sequence, 
 /// it simply copy the bytes into a new String.
 pub fn to_unchecked_string(self : T) -> String {
-  Bytes::from_fixedarray(self.data).to_unchecked_string(
-    offset=0,
-    length=self.len,
-  )
+  self.data.to_unchecked_string(offset=0, length=self.len)
 }
 
 ///|
--- a/buffer/buffer.mbt
+++ b/buffer/buffer.mbt
@@ -75,17 +75,25 @@ pub fn is_empty(self : T) -> Bool {
 ///|
 /// Return a new string contains the data in buffer.
 /// 
-/// @alert deprecated "Use `Buffer::to_unchecked_string` instead"
+/// @alert deprecated "Use `Buffer::contents` to read the contents of the buffer"
 pub fn to_string(self : T) -> String {
-  self.to_unchecked_string()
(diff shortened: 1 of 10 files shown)


buffer/buffer.mbt
```diff
@@ -85,10 +85,7 @@ pub fn to_string(self : T) -> String {
 /// Note this function does not validate the encoding of the byte sequence, 
 /// it simply copy the bytes into a new String.
```
bobzhang, line 86: `@buffer.contents() -> Bytes`

builtin/bytes.mbt
```diff
@@ -12,10 +12,20 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+///|
+/// Reinterpret the byte sequence as Bytes.
+pub fn reinterpret_as_bytes(self : FixedArray[Byte]) -> Bytes = "%identity"
+
```
bobzhang, line 18: `pub` removed, private local unsafe function

builtin/builtin.mbti
```diff
@@ -640,19 +640,21 @@ impl FixedArray {
   make[T](Int, T) -> Self[T]
   op_get[T](Self[T], Int) -> T
   op_set[T](Self[T], Int, T) -> Unit
+  reinterpret_as_bytes(Self[Byte]) -> Bytes
```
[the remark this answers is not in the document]
bobzhang: try to not expose it

string/string.mbt
```diff
@@ -101,10 +101,7 @@ pub fn String::default() -> String {
 ///|
 /// `String` holds a sequence of UTF-16 code units encoded in little endian format
 pub fn to_bytes(self : String) -> Bytes {
-  let len = self.length()
-  let bytes = Bytes::new(len * 2)
-  bytes.blit_from_string(0, self, 0, len)
-  bytes
+  Bytes::of_string(self)
 }
```
bobzhang, line 105: FTR, since both String and Bytes are immutable, this can be a nop in the future?

Yu-zh: Yeah that makes sense to me.

PR #1279, opened 2024-12-02
enhance Array::blit_to to allow grow destination array

Yu-zh:
This PR completes #1267 to allow the destination array to grow in the following case:

Before blit:
```
src array: [ 0  1  2  3  4 ]
src range:      ^--------^
dst array: [ 0  0  0  0  0 ]
dst range:               ^--------^
```
after blit:
```
dst array: [ 0  0  0  1  2  3  4]
```

If the start position of the dst range does not overlap with the dst array, it will panic:
```
dst array: [ 0  0  0  0  0 ]
dst range:                               ^--------^
```


hackwaly:
lgtm. @Yu-zh

bobzhang (approved):

PR #1286, opened 2024-12-03
Immutable array based on RRB Vector

jialunzhang-psu:
This PR modifies the tree structure based on *RRB Vector: A Practical General Purpose Immutable Sequence, ICFP'15*. It will support operations including removal and concat.


jialunzhang-psu:
Originally I wanted to add `concat` and `split` and all other derived operations for `immut/array`, but now it turns out to be a large enough PR for just adding `concat`. So I pause here. There is more work to be done on this component, as listed in README.

Lampese:
> @jialunzhang-psu  can you do a rebase? @Lampese are you interested in doing a review? 

OK, I will do it later.

immut/array/array.mbt
```diff
@@ -127,31 +156,45 @@ pub fn set[A](self : T[A], index : Int, value : A) -> T[A] {
...
-/// ```
-/// let v = @array.of([1, 2, 3])
-/// assert_eq!(v, @array.from_array([1, 2, 3]))
-/// ```
-pub fn from_array[A](arr : Array[A]) -> T[A] {
-  makei(arr.length(), fn(i) { arr[i] })
+/// Return an iterator over the array.
+pub fn iter[A](self : T[A]) -> Iter[A] {
+  Iter::new(fn(yield_) {
+    let arr = self.to_array() // TODO: it first converts to an array, which is not efficient
```
Lampese, line 189: Isn't it a good idea to initialize the entire array when defining iterators?

IMHO Is there no problem with structure based iteration?

immut/array/array.mbt
```diff
@@ -12,67 +12,103 @@
...
-    for i = 0; i < self.size; i = i + 1 {
-      if yield_(arr[i]) == IterEnd {
-        break IterEnd
-      }
-    } else {
-      IterContinue
+/// Physically copy the array.
+/// Since it is an immutable data structure,
+/// it is rarely the case that you would need this function.
+pub fn copy[A](self : T[A]) -> T[A] {
```
Lampese, line 47: I'm not sure about the necessity of this API.
cc @peter-jerry-ye

immut/array/array.mbt
```diff
@@ -127,31 +156,45 @@ pub fn set[A](self : T[A], index : Int, value : A) -> T[A] {
...
-      size: self.size + 1,
-      shift: self.shift,
-    }
+  let (tree, shift) = self.tree.push_end(self.shift, value)
+  { tree, size: self.size + 1, shift }
+}
+
+///|
+/// Given two trees, concatenate them into a new tree.
+pub fn T::concat[A](self : T[A], other : T[A]) -> T[A] {
```
Lampese, line 165: According to the current Array API, the concat operation should be implemented as `op_add`.

immut/array/array.mbt
```diff
@@ -100,19 +140,8 @@ pub fn op_get[A](self : T[A], index : Int) -> A {
 /// assert_eq!(v.set(1, 10), @array.of([1, 10, 3, 4, 5]))
 /// ```
 pub fn set[A](self : T[A], index : Int, value : A) -> T[A] {
```
Lampese, line 142: Should `op_set` be added?

immut/array/array.mbt
peter-jerry-ye: Hmm, I don't think it's necessary. Even it were, maybe it can just return itself?

immut/array/array.mbt
peter-jerry-ye: I think it would be fine to leave both.

Notice the recent change upon calling convention though. `T::` might need to be removed.

immut/array/array.mbt
jialunzhang-psu: `a[i] = j` makes no sense if `a` is immutable.

immut/array/array.mbt
jialunzhang-psu: The entire array is a tree and the array content is only on leaves. We can just iterate over the leaves.

immut/array/array.mbt
jialunzhang-psu: The API was already there when I touched it. I have no problem in removing it or making it private.

immut/array/array.mbt
jialunzhang-psu: Added and fixed.

immut/array/array.mbt
Lampese: I'm stupid, sorry.

immut/array/array.mbt
Lampese: OK, So the current plan is acceptable, if there are any issues later on, we can make further modifications.

immut/array/array.mbt
Lampese: IMO we need to remove it.

jialunzhang-psu: I removed it.

bobzhang:
@jialunzhang-psu we can leave `copy` and mark it as deprecated. I tried to rebase and merge it but found lots of conflicts, it would be nice that you can help rebase, thanks

jialunzhang-psu:
@bobzhang I have marked `copy` as deprecated. Now it shows no conflict on my end, and auto-merging is okay. Please check.

Lampese (approved):

PR #1287, opened 2024-12-04
fix `abs()` behavior on negative zero

tonyfettes:


double/double.mbt
```diff
@@ -37,11 +37,7 @@ pub fn Double::from_int(i : Int) -> Double {
 
 ///|
 pub fn abs(self : Double) -> Double {
-  if self < 0.0 {
-    -self
-  } else {
-    self
-  }
+  (self.reinterpret_as_uint64() & 0x7FFFFFFFFFFFFFFFUL).reinterpret_as_double()
```
lijunchen, line 40: use `0x7FFF_FFFF_FFFF_FFFFUL` for readability

lijunchen (approved):
LGTM, but maybe we need to add fabs intrinsic

PR #1289, opened 2024-12-04
local type issues

bobzhang:


bool/bool_test.mbt
```diff
@@ -18,12 +18,14 @@ test "Bool hash function" {
...
 test "Bool hash function with struct" {
+  struct TestHash {
+    flag : Bool
+    x : Int
+  } derive(Hash, Eq, Show)
+  // type MyInt Int derive(Show) this does not work
+  // let m : Map[TestHash, Char] = {}
   let m = {}
   m[{ flag: true, x: 1 }] = '3'
   m[{ flag: false, x: 1 }] = '3'
```
bobzhang, line 31: 1. new type not supported
2. type error when using `let m = { ..}`

cc @Yu-zh

PR #1298, opened 2024-12-05
rename to_unchecked_x

Yoorkin:
rename `unchecked_to_x` to `to_unchecked_x`, make them consistent with the existing `@buffer.to_unchecked_string`


tonyfettes:
Some cents on the naming convention:

1. For type/encoding/format conversion, use `to_unchecked_x`.
2. For data structure operations (`op_get`, `op_set`, etc) that might leads to trapping, use `usafe_x`.

NOTE: `to_unchecked_int` for float-point data types traps for out-of-rand floats on Wasm/WasmGC as well.

Yoorkin:
@bobzhang  we need to merge this before todays release

PR #1301, opened 2024-12-05
add Iter[String]::join

### LINKED ISSUE #1257: add `Iter[String]::join`
`String::concat` was a mistake. In the self parameter, the types in the type constructor can be a concrete type like String. After discussing offline, we will move to `Iter[String]::join` and make `String::concat` a function. The join method will be added later because users can always use other methods instead.

_Originally posted by @Yoorkin in https://github.com/moonbitlang/core/issues/1249#issuecomment-2502569493_

hackwaly:
Closes #1257

### DIFF
--- a/builtin/builtin.mbti
+++ b/builtin/builtin.mbti
@@ -243,7 +243,7 @@ impl Iter {
   head[A](Self[A]) -> A?
   intersperse[A](Self[A], A) -> Self[A]
   iter[T](Self[T]) -> Self[T]
-  join(Self[String], sep~ : String = ..) -> String
+  join(Self[String], String) -> String
   just_run[T](Self[T], (T) -> IterResult) -> Unit
   last[A](Self[A]) -> A?
   map[T, R](Self[T], (T) -> R) -> Self[R]
--- a/builtin/builtin.mbti
+++ b/builtin/builtin.mbti
@@ -243,6 +243,7 @@ impl Iter {
   head[A](Self[A]) -> A?
   intersperse[A](Self[A], A) -> Self[A]
   iter[T](Self[T]) -> Self[T]
+  join(Self[String], separator~ : String = ..) -> String
   just_run[T](Self[T], (T) -> IterResult) -> Unit
   last[A](Self[A]) -> A?
   map[T, R](Self[T], (T) -> R) -> Self[R]
--- a/builtin/builtin.mbti
+++ b/builtin/builtin.mbti
@@ -243,7 +243,7 @@ impl Iter {
   head[A](Self[A]) -> A?
   intersperse[A](Self[A], A) -> Self[A]
   iter[T](Self[T]) -> Self[T]
-  join(Self[String], separator~ : String = ..) -> String
+  join(Self[String], sep~ : String = ..) -> String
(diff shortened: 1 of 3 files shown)


builtin/iter.mbt
```diff
@@ -781,6 +781,28 @@ pub fn collect[T](self : Iter[T]) -> Array[T] {
   result
 }
 
+///|
+/// Collects the elements of the iterator into a string.
+pub fn join(self : Iter[String], separator~ : String = "") -> String {
+  let buf = StringBuilder::new()
```
bobzhang, line 787: 1. there are some duplications with `Logger::write_iter`, can you factor out some common parts
2. `if separator == ""` this rarely happens, such optimization is not worth it

builtin/iter.mbt
```diff
@@ -781,6 +781,14 @@ pub fn collect[T](self : Iter[T]) -> Array[T] {
   result
 }
 
+///|
+/// Collects the elements of the iterator into a string.
+pub fn join(self : Iter[String], separator~ : String = "") -> String {
```
hackwaly, line 786: Shall we change the label "separator" to "sep"

hackwaly:
related work #1512

builtin/iter.mbt
hackwaly: `Logger::write_iter` has unexpected behavior which turn "Test" to "\"Test\"" before write.

gmlewis: I vote for `sep : String`... that way you always see the one argument that is passed to join and it is always extremely clear what the code is doing:

```moonbit
arr.iter().join("")
arr.iter().join(", ")
arr.iter().join(" ")
etc.
```

Additionally, you don't have the jarring noise of `sep=` cluttering the code unnecessarily.

gmlewis: Now, if the MoonBit compiler could simply perform the `.iter()` under the covers whenever an `iter` is needed where it is obvious that the array must be iterated, then that would get us back to the original, beautiful syntax which was one of the reasons early on that I liked MoonBit so much in the first place!!!

```moonbit
arr.join("")
arr.join(", ")
arr.join(" ")
etc.
```

hackwaly: > I vote for `sep : String`... that way you always see the one argument that is passed to join and it is always extremely clear what the code is doing:
> 
> ```moonbit
> arr.iter().join("")
> arr.iter().join(", ")
> arr.iter().join(" ")
> etc.
> ```
> 
> Additionally, you don't have the jarring noise of `sep=` cluttering the code unnecessarily.
> 
> And it is much easier to write, and muscle memory takes over so that you don't have to think about it, because joins happen frequently in code.

lgtm.

builtin/builtin.mbti
```diff
@@ -256,6 +256,7 @@ impl Iter {
   head[A](Self[A]) -> A?
   intersperse[A](Self[A], A) -> Self[A]
   iter[T](Self[T]) -> Self[T]
+  join(Self[String], sep~ : String = ..) -> String
```
gmlewis, line 259: I think this needs to be `sep~`. Maybe `moon check` will update this?

gmlewis (approved):
LGTM

PR #1307, opened 2024-12-08
Add more write_X method and some bytes conversions to buffer

SyoujyoujiNaiki:
Add methods to write UInt64, Int64, UInt, Int, Double and Float into buffer with their conversion to Bytes.


buffer/buffer.mbt
```diff
@@ -148,6 +196,176 @@ pub fn to_bytes(self : T) -> Bytes {
...
+}
+
+///| Convert an UInt into a 4 bytes long Bytes.
+/// 
+/// # Example
+/// 
+/// ```
+/// let uint_bytes = 114514 |> @buffer.uint_to_bytes
+/// ```
+pub fn uint_to_bytes(i : UInt) -> Bytes {
```
Lampese, line 253: For the sake of API consistency, I think the naming and location of these conversion functions need further consideration, such as whether bytes_to_xxx can be made to xxx(self: bytes), and then whether xxx_to_bytes should not be written here.

SyoujyoujiNaiki: Moved xxx_to_bytes into their own type definitions, but for bytes_to_xxx methods I think they should be kept. The bytes_to_xxx methods is only for specific long Bytes rather general Bytes.

Yoorkin:
Thanks for your contribution!

The `write_X` methods for the buffer look good to me, but the bytes conversions seem unrelated to the buffer package. We also need further consideration about the API convention.

Can you separate out the bytes conversions part from this PR? Let's merge the `write_X` part first.

buffer/buffer.mbt
SyoujyoujiNaiki: Removed bytes_to_xxx methods.

Yoorkin:
Some consideration about performance:

The buffer is used to write large data consisting of multiple basic types (int, bool, double, etc.). It will call `write_x` frequently. We should try to eliminate the memory copy overhead in the `write_x` methods. Instead of converting to bytes and writing the bytes into the buffer, consider using `write_byte` directly

If you don't want to duplicate the code, maybe changing the implementation of `X::to_bytes` to use the `x |> @buffer.write_x |> @buffer.to_bytes` is a better choice, because those conversions are not frequently used. What do you think?

buffer/moon.pkg.json
```diff
@@ -2,6 +2,12 @@
...
     "moonbitlang/core/builtin",
     "moonbitlang/core/coverage",
-    "moonbitlang/core/bytes"
+    "moonbitlang/core/bytes",
+    "moonbitlang/core/uint64",
+    "moonbitlang/core/uint",
+    "moonbitlang/core/int64",
+    "moonbitlang/core/int",
+    "moonbitlang/core/double",
+    "moonbitlang/core/float"
```
Yoorkin, line 11: These imports seem unused now

double/double.mbti
```diff
@@ -37,6 +37,7 @@ impl Double {
   pow(Double, Double) -> Double
   round(Double) -> Double
   signum(Double) -> Double
+  to_bytes(Double) -> Bytes
```
Yoorkin, line 40: can you also move changes about `T::to_bytes` to another PR, along with `bytes_to_t`?

Yoorkin (changes requested):

Yoorkin (approved):

buffer/buffer.mbti
```diff
@@ -16,10 +16,16 @@ impl T {
...
+  write_float(Self, Float) -> Unit
+  write_int(Self, Int) -> Unit
+  write_int64(Self, Int64) -> Unit
   write_object(Self, Show) -> Unit
   write_string(Self, String) -> Unit
   write_sub_string(Self, String, Int, Int) -> Unit //deprecated
   write_substring(Self, String, Int, Int) -> Unit
+  write_uint(Self, UInt) -> Unit
+  write_uint64(Self, UInt64) -> Unit
 }
```
bobzhang, line 29: we need mark the endianess, for example, in ocaml
```
val add_uint16_ne : t -> int -> unit
(** [add_uint16_ne b i] appends a binary native-endian unsigned 16-bit
    integer [i] to [b].
    @since 4.08
*)

val add_uint16_be : t -> int -> unit
(** [add_uint16_be b i] appends a binary big-endian unsigned 16-bit
    integer [i] to [b].
    @since 4.08
*)

val add_uint16_le : t -> int -> unit
(** [add_uint16_le b i] appends a binary little-endian unsigned 16-bit
    integer [i] to [b].
    @since 4.08
*)
```

SyoujyoujiNaiki: Added some endian variants methods. But for native endian, does MoonBit have a function to check it?

SyoujyoujiNaiki: I also notice that write_char and write_string does not provide different endian variants. Do I need to complete them?

Yoorkin: > I also notice that write_char and write_string does not provide different endian variants. Do I need to complete them?

Encoding-related APIs like `write_char` and `write_string` will be moved to x/encoding

PR #1313, opened 2024-12-09
ai generated @immut/sorted_map.T::intersection

### LINKED ISSUE #1308: [Feature Request]: provide @immut/sorted_map.T::intersection_map2
```
pub intersection_map2[K, V, M](
  self : T[K, V],
  other : T[K, V],
  f : (V, V) -> M
) -> T[K, M]
```

Use case in moonyacc:

```
fn weak_compat(set1 : EncodedLR1ItemSet, set2 : EncodedLR1ItemSet) -> Bool {
  let common = @immut/sorted_set.intersection(set1.core(), set2.core())
    .to_array()
    .map(fn(key) { (key, set1._[key].unwrap(), set2._[key].unwrap()) })
  for i in 0..<common.length() {
    let (_, la_set1i, la_set2i) = common[i]
    for j in i..<common.length() {
      guard i != j else { continue }
      let (_, la_set1j, la_set2j) = common[j]
      let compat = (la_set1i.disjoint(la_set2j) && la_set2i.disjoint(la_set1j)) ||
        not(la_set1i.disjoint(la_set1j)) ||
        not(la_set2i.disjoint(la_set2j))
      if not(compat) {
        return false
      }
    }
  }
  true
}
```

If we have `intersection_map2`

```
fn weak_compat(set1 : EncodedLR1ItemSet, set2 : EncodedLR1ItemSet) -> Bool {
  let common = @immut/sorted_map.intersection_map2(set1, set2, fn (v1, v2) { (v1, v2) })
    .to_array()
  for i in 0..<common.length() {
    let (_, (la_set1i, la_set2i)) = common[i]
    for j in i..<common.length() {
      guard i != j else { continue }
      let (_, (la_set1j, la_set2j)) = common[j]
      let compat = (la_set1i.disjoint(la_set2j) && la_set2i.disjoint(la_set1j)) ||
        not(la_set1i.disjoint(la_set1j)) ||
        not(la_set2i.disjoint(la_set2j))
      if not(compat) {
        return false
      }
    }
  }
  true
}
```

hackwaly:
Closes #1308

### DIFF
--- a/immut/sorted_map/sorted_map.mbti
+++ b/immut/sorted_map/sorted_map.mbti
@@ -23,6 +23,7 @@ impl T {
   from_iter[K : Compare, V](Iter[(K, V)]) -> Self[K, V]
   from_json[V : @json.FromJson](Json) -> Self[String, V]!@json.JsonDecodeError
   insert[K : Compare, V](Self[K, V], K, V) -> Self[K, V] //deprecated
+  intersection[K : Compare, V1, V2](Self[K, V1], Self[K, V2]) -> Iter[(K, V1, V2)]
   is_empty[K, V](Self[K, V]) -> Bool
   iter[K, V](Self[K, V]) -> Iter[(K, V)]
   iter2[K, V](Self[K, V]) -> Iter2[K, V]
--- a/immut/sorted_map/map.mbt
+++ b/immut/sorted_map/map.mbt
@@ -119,6 +119,58 @@ pub fn filter_with_key[K : Compare, V](
   }
 }
 
+///| Get the intersection of two maps.
+/// This returns an iterator that traverses over the common keys in both maps,
+/// yielding tuples (key, value1, value2).
+/// O(n log m) where n is the size of the smaller map and m is the size of the larger map.
+pub fn intersection[K : Compare, V1, V2](
+  self : T[K, V1],
+  other : T[K, V2]
+) -> Iter[(K, V1, V2)] {
+  Iter::new(
+    fn(yield_) {
+      if self.size() <= other.size() {
+        intersection_visit(self, other, yield_)
+      } else {
+        intersection_visit(
+          other,
+          self,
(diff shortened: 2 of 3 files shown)


yj-qin:
LGTM

yj-qin (approved):

immut/sorted_map/sorted_map.mbti
```diff
@@ -23,6 +23,7 @@ impl T {
   from_iter[K : Compare, V](Iter[(K, V)]) -> Self[K, V]
   from_json[V : @json.FromJson](Json) -> Self[String, V]!@json.JsonDecodeError
   insert[K : Compare, V](Self[K, V], K, V) -> Self[K, V] //deprecated
+  intersection[K : Compare, V1, V2](Self[K, V1], Self[K, V2]) -> Iter[(K, V1, V2)]
```
Yoorkin, line 26: The `intersection` should return another sorted_map, similar to the `intersection` methods of `set`, `sorted_set`, and `immut/sorted_set`

hackwaly: The result `@immut/sorted_map.T[K, (V1, V2)]` is often useless, as users often want to do some merge work of `V1` and `V2`.

Yoorkin: `intersection(Map, Map) -> Map` can be implemented in O(m log(n/m + 1)) time complexity, which is better than the current implementation O(m log n). Of course, returning a map will cause more space overhead, but users who only iterate over the result can always easily optimize it by `map1.iter().filter(existInMap2).each(f)`. 

If it returns an iter, the order information will be lost and it will take O(m log n) + O(k log k) (k < m < n) time to construct an intersection map. Furthermore, you have to break existing set operations to keep API consistent.

hackwaly: Agreed. @Yoorkin Could you fire a new PR for `O(m log(n/m + 1))` implementation that returns `Map`?

PR #1315, opened 2024-12-10
add @hashmap.T::iter2

hackwaly:
This PR added missing `@hashmap.T::iter2`

### DIFF
--- a/hashmap/utils.mbt
+++ b/hashmap/utils.mbt
@@ -55,6 +55,26 @@ pub fn iter[K, V](self : T[K, V]) -> Iter[(K, V)] {
 }
 
 ///|
+pub fn T::iter2[K, V](self : T[K, V]) -> Iter2[K, V] {
+  Iter2::new(
+    fn(yield_) {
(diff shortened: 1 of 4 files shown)


hashmap/utils.mbt
```diff
@@ -54,6 +54,26 @@ pub fn iter[K, V](self : T[K, V]) -> Iter[(K, V)] {
   )
 }
 
+///|
+pub fn T::iter2[K, V](self : T[K, V]) -> Iter2[K, V] {
+  Iter2::new(
+    fn(yield_) {
+      let len = self.entries.length()
+      for i = 0; i < len; i = i + 1 {
+        match self.entries[i] {
```
bobzhang, line 63: is this the efficient way to iterate hashmap?

hackwaly: This is copied from `T::iter`. Since `@hashmap.T` do not record keys in linked list as `@builtin.Map`, there's no efficient way to iterate over keys.

Issue #1013, opened 2024-09-18
Feature request: `ArrayView` should have all the public methods that `Array` has


gmlewis:
Discord discussion: https://discord.com/channels/1135479745207869510/1285600688197341287

Since it is not easy to create an `Array` from an `ArrayView` without copying, it would be nice if `ArrayView` could do everything that an `Array` can do.

Additionally, I often find myself in similar situations with these types that are closely related yet frequently incompatible without copying:
* `Bytes`
* `FixedArray[Byte]`
* `Array[Byte]`
* `ArrayView[Byte]`

Ideally, it should be easy to convert to/from any of these 4 types to any of the other 4 types _WITHOUT_ any copying.

peter-jerry-ye:
The `without any copying` part is impossible to give guarantee.

For example, take the wasm-gc backend. There's possibility that `Bytes` and `FixedArray[Byte]` can be converted back and forth with `array<u8>`, but impossible with `Array[Byte]`.

On the other hand, there's no difference between `FixedArray` and `Array` and JS native array. But still, it's impossible to convert `ArrayView` to `Array` without copying.

gmlewis:
Thank you, @peter-jerry-ye, for the information.

Is it safe to say that anything that takes an `Array` as an argument would be more useful if it instead took an `ArrayView`?

For example, take `Bytes::from_array` for example:

Instead of:

```rust
pub fn Bytes::from_array(arr : Array[Byte]) -> Bytes { ... }
```

wouldn't it be more flexible if it were this?

```rust
pub fn Bytes::from_array(arr : ArrayView[Byte]) -> Bytes { ... }
```

And if that is true, would you want to make sweeping breaking API changes throughout the code base, or would you prefer two versions of every function where the `Array` version would call the (more flexible) `ArrayView` version?
(Or maybe instead deprecate the existing one and add a new one?)

If I were to take on this "good first issue", would it be OK to address it with baby-steps like this first Array/ArrayView step?

peter-jerry-ye:
@gmlewis That's a good question. `ArrayView` is indeed more general than `Array`. We may need to take that into account. cc @bobzhang @Yoorkin

FlammeShadow:
Should we implement methods for `ArrayView` that change `buf`? If so, Should we retain the `start` and `end` unchanged when an `ArrayView` is called with `sort(_by)` or `*_inplace`? The existing `rev_inplace` in `@array` seems not to change it.

PR #1321, opened 2024-12-11
Implement map* and filter for ArrayView

gmlewis:
This PR is for discussion purposes.
It relates to #1013 and #1275.

I believe that it is _MUCH_ more flexible for the `map*` functions to return an `Array` (instead of an `ArrayView`).
Then, code that was performing a map on an entire array could easily insert a `[1:]` and now map all but the first entry without modifying any other code.

Additionally, it is much easier to turn an `Array` into an `ArrayView` than to go the other direction.

### DIFF
--- a/builtin/arrayview.mbt
+++ b/builtin/arrayview.mbt
@@ -216,7 +216,7 @@ pub fn map_inplace[T](self : ArrayView[T], f : (T) -> T) -> Unit {
 /// ```
 /// let v = [3, 4, 5]
 /// let v2 = v[1:].mapi(fn (i, x) {x + i})
-/// assert_eq!(v2, [3, 4, 6])
+/// assert_eq!(v2, [4, 6])
 /// ```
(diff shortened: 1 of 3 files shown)


builtin/arrayview.mbt
```diff
@@ -173,3 +173,92 @@ pub fn rev_foldi[A, B](
...
+}
+
+///|
+/// Maps a function over the elements of the array view with index.
+///
+/// # Example
+/// ```
+/// let v = [3, 4, 5]
+/// let v2 = v[1:].mapi(fn (i, x) {x + i})
+/// assert_eq!(v2, [3, 4, 6])
```
bobzhang, line 219: the CI failed because the test doc is incorrect

gmlewis: Sorry about that, @bobzhang !  
I forgot about `moon test --doc`.
It should be fixed now.

PR #1324, opened 2024-12-13
add `Bool::negate`, avoid usage like `x.not()`

Yoorkin:
We will introduce the `not x` syntax to avoid the illusion of LLM. For any code written like `not(x)` or `not x`, it will not break, except for `x.not()`. This PR remove usage like `x.not()` in core and introduces the `Bool::negate` method.


Yoorkin:
#1330

PR #1333, opened 2024-12-17
AI generated docs and tests

bobzhang:


builtin/string.mbt
```diff
@@ -31,3 +65,45 @@ pub fn substring(
   guard start >= 0 && start <= end && end <= self.length()
   unsafe_substring(self, start, end)
```
bobzhang, line 66: in the future, for patterns like this, `guard` followed by a function call, we may always in the function to eliminate the guard statement cc @Yu-zh

PR #1334, opened 2024-12-17
perf(builtin/string)!: prevent calculating `.length()` twice in `substring()`

rami3l:
cc @bobzhang


gmlewis:
`String.length()` isn't `O(1)` ?!?  Why not?

rami3l:
> Wasn't aware but is `String.length()` O(n)  and this PR is optimizing for that?

It's O(1) but the current implementation delegates to a primitive twice.

PR #1336, opened 2024-12-17
Add land/lor/lxor for BigInt.

SyoujyoujiNaiki:
#1159


SyoujyoujiNaiki:
cc @Lampese

peter-jerry-ye:
Will check the pipeline issue.

However, I'm not quite sure about the implementation:

```bash
% node
Welcome to Node.js v23.4.0.
Type ".help" for more information.
> (-10n) & 11n
2n
> (-10n) | 11n
-1n
> (-10n) ^ 11n
-3n
```

SyoujyoujiNaiki:
> Will check the pipeline issue.
> 
> However, I'm not quite sure about the implementation:
> 
> ```shell
> % node
> Welcome to Node.js v23.4.0.
> Type ".help" for more information.
> > (-10n) & 11n
> 2n
> > (-10n) | 11n
> -1n
> > (-10n) ^ 11n
> -3n
> ```

The implementation of NodeJS BigInt is different from the MoonBit NonJS implementation. I just apply the land/lor/lxor method on the uint array and the sign.

peter-jerry-ye:
> The implementation of NodeJS BigInt is different from the MoonBit NonJS implementation. I just apply the land/lor/lxor method on the uint array and the sign.

I noticed that. But MoonBit needs consistency across backends, i.e. NonJS & JS backends.

SyoujyoujiNaiki:
> > The implementation of NodeJS BigInt is different from the MoonBit NonJS implementation. I just apply the land/lor/lxor method on the uint array and the sign.
> 
> I noticed that. But MoonBit needs consistency across backends, i.e. NonJS & JS backends.

But then the land/lor/lxor would be not a direct and clear bit operation.

peter-jerry-ye:
> But then the land/lor/lxor would be not a direct and clear bit operation.

It has never been: https://262.ecma-international.org/15.0/index.html#sec-bigintbitwiseop

builtin/bigint_nonjs.mbt
```diff
@@ -876,3 +876,41 @@ pub fn BigInt::to_octets(self : BigInt, length? : Int) -> Bytes {
   }
   result
 }
+
+///|
+pub fn land(self : BigInt, other : BigInt) -> BigInt {
+  let new_limbs = FixedArray::make(self.limbs.length(), 0U)
+  for i = 0; i < self.len; i = i + 1 {
+    new_limbs[i] = self.limbs[i] & other.limbs[i]
```
inkydragon, line 884: The implementation does not seem to take into account the length of the two inputs.
With
```
  // large input
  let a = BigInt::from_string("123456789012345678901234567890")
  inspect!(a & 0N, content="0")
```
We have: `RuntimeError: array element access out of bounds`

inkydragon (commented):
Maybe we could learn something from quickjs (MIT).

`quickjs` uses a struct for representing a BigInt, moonbit doesn't have an `expn` field, and a similar implementation of `bf_logic_op` would be more compact. 
```c
/* +/-zero is represented with expn = BF_EXP_ZERO and len = 0,
   +/-infinity is represented with expn = BF_EXP_INF and len = 0,
   NaN is represented with expn = BF_EXP_NAN and len = 0 (sign is ignored)
 */
typedef struct {
    struct bf_context_t *ctx;
    int sign;
    slimb_t expn;
    limb_t len;
    limb_t *tab;
} bf_t;
```
https://github.com/quickjs-ng/quickjs/blob/be2db73000765717f231e32a4110cb72556d0bec/libbf.c#L2345-L2447

peter-jerry-ye:
You might need something like this:

```moonbit
pub fn BigInt::land(self : BigInt, other : BigInt) -> BigInt {
  let mut x = self
  let mut y = other
  let mut result = 0N
  let mut shift = 0
  while not((x == 0 || x == -1) && (y == 0 || y == -1)) {
    let x0 = x.limbs[0] & 1
    let y0 = y.limbs[0] & 1
    if (x0 & y0) != 0 {
      result = result + (1N << shift)
    }
    x = (x - BigInt::from_int(x0.reinterpret_as_int())) / 2
    y = (y - BigInt::from_int(y0.reinterpret_as_int())) / 2
    shift += 1
  }
  if (x.limbs[0] & 1U & (y.limbs[0] & 1U)) != 0 {
    result = result - (1N << shift)
  }
  result
}
```

SyoujyoujiNaiki:
> > The implementation of NodeJS BigInt is different from the MoonBit NonJS implementation. I just apply the land/lor/lxor method on the uint array and the sign.
> 
> I noticed that. But MoonBit needs consistency across backends, i.e. NonJS & JS backends.

Pushed the new implementation which results the same as what NodeJS output.

builtin/bigint_nonjs.mbt
SyoujyoujiNaiki: Pushed the new implementation that makes two BigInt's limbs when calculating same size long. Should fix this.

builtin/bigint_js.mbt
```diff
@@ -212,3 +212,15 @@ extern "js" fn js_shl(self : BigInt, other : Int) -> BigInt =
...
+extern "js" fn js_land(self : BigInt, other : BigInt) -> BigInt =
+  #|(x, y) => x & y
+
+///|
+extern "js" fn js_lor(self : BigInt, other : BigInt) -> BigInt =
+  #|(x, y) => x | y
+
+///|
+extern "js" fn js_lxor(self : BigInt, other : BigInt) -> BigInt =
+  #|(x, y) => x ^ y
```
peter-jerry-ye, line 226:
```suggestion

///|
pub extern "js" fn land(self : BigInt, other : BigInt) -> BigInt =
  #|(x, y) => x & y

///|
pub extern "js" fn lor(self : BigInt, other : BigInt) -> BigInt =
  #|(x, y) => x | y

///|
pub extern "js" fn lxor(self : BigInt, other : BigInt) -> BigInt =
  #|(x, y) => x ^ y
```

peter-jerry-ye (changes requested):
This should make the tests pass

builtin/bigint_nonjs.mbt
```diff
@@ -925,3 +925,264 @@ pub fn BigInt::to_octets(self : BigInt, length? : Int) -> Bytes {
...
+      if x_limbs[i] != 0xFFFFFFFFU {
+        break
+      }
+    }
+    for i = 0; i < x_limbs.length(); i = i + 1 {
+      x_limbs[i] = x_limbs[i] ^ 0xFFFFFFFFU
+    }
+  }
+  
+  // return the result
```
peter-jerry-ye, line 999:
```suggestion
  }

  // return the result
```

builtin/bigint_nonjs.mbt
```diff
@@ -925,3 +925,264 @@ pub fn BigInt::to_octets(self : BigInt, length? : Int) -> Bytes {
...
```
peter-jerry-ye, line 1086:
```suggestion
  }

  // return the result
```

builtin/bigint_nonjs.mbt
```diff
@@ -925,3 +925,264 @@ pub fn BigInt::to_octets(self : BigInt, length? : Int) -> Bytes {
...
```
peter-jerry-ye, line 1173:
```suggestion
  }

  // return the result
```

builtin/bigint_test.mbt
```diff
@@ -557,3 +557,93 @@ test "to_hex" {
...
+  let a = BigInt::from_string("0")
+  let b = BigInt::from_string("-10")
+  inspect!(a ^ b, content="-10")
+}
+
+test "lxor_5" {
+  let a = BigInt::from_string("0")
+  let b = BigInt::from_hex("ffffffffffffffff")
+  inspect!(a ^ b, content="18446744073709551615")
+}
```
peter-jerry-ye, line 649:
```suggestion
}

```

peter-jerry-ye (changes requested):
These should make the format test pass

peter-jerry-ye (approved):

peter-jerry-ye:
Thank you for your contribution

PR #1342, opened 2024-12-19
fix bigint pow

peter-jerry-ye:
BigInt::pow should not have default modulus value.

API change will not break existing code.


hackwaly:
I think it's better to split it to two functions, since we do not have specialize optimization yet.

hackwaly:
FYI https://leetcode.com/problems/super-pow/solutions/1132201/javascript-BigInt-pow-mod-one-line-192ms/

PR #1343, opened 2024-12-19
Use native array map under js target

hackwaly:
This MR make `Array::map` implementation platform specific to use native JavaScript Array::map under js target.

### DIFF
--- a/builtin/array.mbt
+++ b/builtin/array.mbt
@@ -264,26 +264,6 @@ pub fn clear[T](self : Array[T]) -> Unit {
 }
 
 ///|
-/// Maps a function over the elements of the array.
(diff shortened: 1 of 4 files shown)


bobzhang:
what's the motivation here, this might actually be slower

hackwaly:
If `f` cannot be inlined. map is faster than for loop absolutely.

https://www.measurethat.net/Benchmarks/Show/5059/0/for-loop-vs-map

bobzhang:
I tweaked the benchmark a little bit: https://www.measurethat.net/Benchmarks/Show/33085/0/for-loop-vs-map-with-size

PR #1345, opened 2024-12-20
Implement FromJson and ToJson for UInt and UInt64

gmlewis:


gmlewis:
Note that if I remove the `uint` and `uint64` imports from `json/moon.pkg.json`, that I get the error:

```
$ moon test --target all
...
Error: [4037]
    ╭─[/Users/glenn/src/github.com/moonbitlang/core/json/to_json_test.mbt:29:18]
    │
 29 │   inspect!((42U).to_json(), content="Number(42)")
    │                  ───┬───
    │                     ╰───── Cannot call method of type UInt: package moonbitlang/core/uint is not imported.
────╯
Error: [4037]
    ╭─[/Users/glenn/src/github.com/moonbitlang/core/json/to_json_test.mbt:33:17]
    │
 33 │   inspect!(42UL.to_json(), content="Number(42)")
    │                 ───┬───
    │                    ╰───── Cannot call method of type UInt64: package moonbitlang/core/uint64 is not imported.
────╯
error: failed to run test for target Wasm

Caused by:
    failed when testing
```

gmlewis:
So I'm not sure how to get all the checks to pass, as there are currently warnings for some targets.

uint64/uint64.mbt
```diff
@@ -17,3 +17,8 @@ pub let min_value : UInt64 = 0UL
 
 ///|
 pub let max_value : UInt64 = 18446744073709551615UL
+
+///|
+pub fn UInt64::to_json(self : UInt64) -> Json {
+  Number(self.to_double())
```
bobzhang, line 23: this may loose the precision, the conservative way may just store them as string

gmlewis: But wouldn't that add quotes around the number in the resulting JSON?

I would rather lose the precision than to change the actual type to string.
I say this because if the client is communicating with a system that demands a number, it may error if there are quotes around the number.

peter-jerry-ye: I think it might be better to leave it as an explicit conversion if it may lose precision.

With that said, the JSON format does not define the range of the number as per ECMA 404 section 8. Maybe we need to modify the `JSON` enum defined in MoonBit to include BigInt / BigNum or something like that

bobzhang:
```
In Serde, the serialization and deserialization of i64 (a 64-bit signed integer) to and from JSON is handled based on the capabilities of the JSON format, which generally supports numbers that fit within JavaScript’s Number type. Here’s how it works:

1. Serialization of i64 to JSON
	•	By default, Serde will serialize an i64 as a JSON number if the value fits within the range of a JavaScript Number (i.e., -2^53 to 2^53-1).
	•	If the value is outside this range, it will be serialized as a JSON string to prevent precision loss in systems that only support JavaScript Number.

Example:
	•	i64::MAX (9223372036854775807) → "9223372036854775807" (JSON string, because it’s too large for a JavaScript Number)
	•	12345_i64 → 12345 (JSON number, because it’s within the safe range)

You can customize this behavior by using Serde attributes or a custom serializer.

2. Deserialization of i64 from JSON
	•	During deserialization, Serde will attempt to parse the JSON value as an i64:
	•	If the JSON value is a number and fits within the i64 range, it will be deserialized directly.
	•	If the JSON value is a string containing a valid i64, Serde will also parse it.
	•	If the value is out of the i64 range or cannot be parsed as an i64, deserialization will fail.

Example:
	•	"9223372036854775807" → i64::MAX
	•	12345 → 12345_i64
	•	"not_a_number" → Deserialization error

3. Customizing the Behavior

You can customize the behavior using Serde’s attributes. For example:
	•	Use #[serde(with = "serde_with::rust::display_fromstr")] to enforce string-based serialization and deserialization.
	•	Write a custom serialize and deserialize function for more control over how i64 values are handled.

Practical Implication

Since JSON is often consumed by JavaScript-based systems, the string-based fallback for large i64 values is crucial to avoid precision loss. However, this means you need to be careful when handling these numbers on the consuming side, ensuring they are parsed correctly back into integers.

Would you like an example implementation of custom serialization/deserialization for i64 in Serde?
```

peter-jerry-ye: > By default, Serde will serialize an i64 as a JSON number if the value fits within the range of a JavaScript Number (i.e., -2^53 to 2^53-1).
>	•	If the value is outside this range, it will be serialized as a JSON string to prevent precision loss in systems that only support JavaScript Number.

This might not be true

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=657d9dc6cc11f9b8efc8524acfc4e874

lynzrand: As can be seen from `serde_json`'s source code, it just emits the string representation of the number:

https://github.com/serde-rs/json/blob/b2a1415aad0c76684c675d13a5746850d275c178/src/ser.rs#L112-L116

https://github.com/serde-rs/json/blob/b2a1415aad0c76684c675d13a5746850d275c178/src/ser.rs#L1625-L1633

In the meantime, it by default accepts numbers in `[int64.min, uint64.max] U float64` when deserializing numbers with unspecified precision, and has special deserilization paths for `{u,}int128`. 

https://github.com/serde-rs/json/blob/b2a1415aad0c76684c675d13a5746850d275c178/src/de.rs#L350-L437

It has an `arbitrary_precision` flag to enable arbitrarily-long numbers to be stored in intermediate representations or `serde_json::Number` instances, and roundtripped back to JSON if needed.

https://github.com/serde-rs/json/blob/b2a1415aad0c76684c675d13a5746850d275c178/src/de.rs#L931-L948

The behavior of serializing large numbers to strings is indeed present in some other serializers, but as far as I'm aware it's not present in `serde_json`

gmlewis:
@peter-jerry-ye - please feel free to close this if #1366 is the preferred solution.

peter-jerry-ye:
As discussed in #1366 we will make sure that `FromJson` and `ToJson` is roundtrip & does not lose precision so other approaches will be taken.

PR #1346, opened 2024-12-20
Fix Atan output when Atan's input is `inf`

Kaida-Amethyst:
`atan(inf)` should be `1.5707963267948966`, not `Infinity`
`atan(-inf)` should be `-1.5707963267948966`, not `-Infinity`

### DIFF
--- a/math/trigonometric.mbt
+++ b/math/trigonometric.mbt
@@ -244,7 +244,7 @@ fn satan(x : Double) -> Double {
 
 ///|
 pub fn atan(x : Double) -> Double {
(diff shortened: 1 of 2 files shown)


Kaida-Amethyst:
Note: `1.5707963267948966` is `pi/2`, `-1.5707963267948966` is `-pi/2`, here is the python result:

```python
>>> from math import atan, inf
>>> atan(inf)
1.5707963267948966
>>> atan(-inf)
-1.5707963267948966
```

Lampese (approved):

PR #1347, opened 2024-12-20
add get_or_init for hashmap

### LINKED ISSUE #1344: Missing get_or_init for hashmap.T

hackwaly:
Closes #1344 

This PR introduce `@hashmap.T::get_or_init`. It's signature slightly different from `linked_hashmap.T::get_or_init`. Which `init` parameter is a function takes key as argument. `linked_hashmap.T::get_or_init` doesn't.

There's no safe way to migrate `get_or_init` from old signature to new signature. Maybe introduce a new name `get_or_add`?

### DIFF
--- a/hashmap/hashmap.mbt
+++ b/hashmap/hashmap.mbt
@@ -118,6 +118,45 @@ pub fn op_get[K : Hash + Eq, V](self : T[K, V], key : K) -> V? {
 }
 
 ///|
+/// Gets the value associated with the given key. If the key doesn't exist in the
+/// map, initializes it with the given function and returns the new value.
+///
+/// Parameters:
+///
+/// * `map` : The hash map to operate on.
+/// * `key` : The key to look up in the map.
+/// * `init_fn` : A function that takes the key and returns a new value. This
+/// function is called only when the key is not found in the map.
+///
+/// Returns the value associated with the key, either existing or newly
+/// initialized.
+///
+/// Example:
+///
+/// ```moonbit
+/// test "get_or_init" {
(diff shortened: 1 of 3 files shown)


hashmap/hashmap.mbti
```diff
@@ -16,6 +16,7 @@ impl T {
   from_iter[K : Hash + Eq, V](Iter[(K, V)]) -> Self[K, V]
   get[K : Hash + Eq, V](Self[K, V], K) -> V?
   get_or_default[K : Hash + Eq, V](Self[K, V], K, V) -> V
+  get_or_init[K : Hash + Eq, V](Self[K, V], K, (K) -> V) -> V
   is_empty[K, V](Self[K, V]) -> Bool
```
bobzhang, line 20: Can you align with the old one using `() -> V`, note the callback used to ben an expensive computation so capture the key should not matter that much.
We could provide `get_or_add` as an alternative

hashmap/hashmap.mbt
```diff
@@ -101,15 +101,23 @@ pub fn op_set[K : Hash + Eq, V](self : T[K, V], key : K, value : V) -> Unit {
...
 pub fn get[K : Hash + Eq, V](self : T[K, V], key : K) -> V? {
-  let hash = key.hash()
+  match self.get_entry_with_hash(key, key.hash()) {
+    Some(entry) => Some(entry.value)
+    None => None
+  }
+}
+
+///|
+fn get_entry_with_hash[K : Hash + Eq, V](
```
hackwaly, line 111: This is for later optimization on grow. To avoid grow to reuse old entries instead of recreating entries.

Issue #1264, opened 2024-11-27
Add a `.range()` method to sorted_map and sorted_set


zxch3n:
This is a very useful API that allows users to:

- Select entries within the target range
- Find the first entry >, >=, <, or <= a target key

https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.range

zxch3n:
The current API is not flexible enough.

```ts
///| Returns a new array of key-value pairs that are within the specified range [low, high).
pub fn range[K : Compare, V](self : T[K, V], low : K, high : K) -> Iter2[K, V] {
    ...
}
```

For example, E may be an enum type with only three variants: A < B < C. This API doesn't allow you to find all the elements that are either B or C. 

We may add a `Bound` enum to solve this, like Rust.

```
pub enum Bound<T> {
    Included(T),
    Excluded(T),
    Unbounded,
}
```

```ts
pub fn range[K : Compare, V](self : T[K, V], low : Bound[K], high : Bound[K]) -> Iter2[K, V] {
    ...
}
```

PR #1359, opened 2024-12-24
#1264 Add .range() method to sorted_map and sorted_set

E-M-T-F:
Added .range() method to sorted_map and sorted_set, and privode the test cases

### DIFF
--- a/sorted_map/map.mbt
+++ b/sorted_map/map.mbt
@@ -241,6 +241,17 @@ pub impl[K : @quickcheck.Arbitrary + Compare, V : @quickcheck.Arbitrary] @quickc
   @quickcheck.Arbitrary::arbitrary(size, rs) |> T::from_iter
 }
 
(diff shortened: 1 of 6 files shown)


bobzhang:
the implementation looks good to me, but this is not commonly used operation, shall we leave it in the user land?

E-M-T-F:
Related: #1264

E-M-T-F:
This PR is the solution of issue #1264, so I think it is an useful solution for the user land

bobzhang:
@E-M-T-F that looks good to me, can you change the API to return Iter2 instead?
```
range[K : Compare, V](Self[K, V], K, K) -> Iter2[K,V]
```

E-M-T-F:
@bobzhang OK，I will change the API to return Iter2 instead

bobzhang:
@E-M-T-F I added some tweaks, thank you

PR #1362, opened 2024-12-25
fix(deque): fix incorrect `@deque.push*()` behavior when empty

### LINKED ISSUE #1361: `@deque.T`'s implementation is not correct
Minimal reproduction below:

```moonbit
fn main {
  let d = @deque.of(["RP", "NL"])
  ignore(d.pop_front())
  guard let Some(f) = d.pop_front()
  d.push_back(f)
  println("\{d}") // Expected `@deque.of(["NL"])`; got `@deque.of(["RP"])`.
}
```

rami3l:
Closes #1361 by addressing the immediate concern WRT the usage in `cmark`.

Whether `tail` should be a computed state instead of a save state is still an open question.

### DIFF
--- a/deque/deque.mbt
+++ b/deque/deque.mbt
@@ -157,8 +157,10 @@ pub fn push_front[A](self : T[A], value : A) -> Unit {
   if self.len == self.buf.length() {
     self.realloc()
   }
-  if self.len != 0 {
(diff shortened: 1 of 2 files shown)


Lampese (approved):

rami3l:
Converting to draft since @peter-jerry-ye would like to have another look at the invariants in this implementation.

PR #1364, opened 2024-12-25
feat: impl Show for Failure

Lampese:


peter-jerry-ye:
Should we include the name of the constructor?

builtin/show.mbt
```diff
@@ -136,6 +136,18 @@ pub impl Show for String with output(self, logger) {
 ///|
 pub impl Show for String with to_string(self) { self }
 
+///|
+pub impl Show for Failure with to_string(self) {
+  let Failure(msg) = self
```
bobzhang, line 141: to_string has a default implementation, this is not needed

PR #1372, opened 2024-12-26
test(iter): add invariant iter should only run once

peter-jerry-ye:
All the methods provided should make sure that `Iter` is only ran once.

Counter example: `tap` #1371


lynzrand:
My best guess is that the reduction in `result.mbt` comes from a rare name clash in coverage data output, as I can't reproduce it in my local environment. A PR there is under way: https://github.com/moonbitlang/moon/pull/538

PR #1379, opened 2024-12-27
test: improve coverage for `array/sort.mbt`

rami3l:
It looks like this cannot be achieved with black box tests, so the agent has repeatedly abandoned on this task... Anyway, the file coverage is now 100%.

### DIFF
--- a/array/sort.mbt
+++ b/array/sort.mbt
@@ -313,6 +313,12 @@ test "quick_sort limit check" {
 }
 
 test "quick_sort with pred check" {
+  let arr = []
+  for i = 16; i >= 0; i = i - 1 {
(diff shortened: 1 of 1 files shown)


bobzhang:
> It looks like this cannot be achieved with black box tests, so the agent has repeatedly abandoned on this task... Anyway, the file coverage is now 100%.

This test is supplied by *youself* not AI?

rami3l:
> > It looks like this cannot be achieved with black box tests, so the agent has repeatedly abandoned on this task... Anyway, the file coverage is now 100%.
> 
> This test is supplied by _youself_ not AI?

@bobzhang Yes.

Issue #1363, opened 2024-12-25
Add more test for `@deque`


Lampese:
ref: https://github.com/rust-lang/rust/blob/a92c3cf04d945aa2071067ef299bfacec7bfebe4/library/alloc/src/collections/vec_deque/tests.rs

PR #1385, opened 2024-12-29
Add more test for deque.

SyoujyoujiNaiki:
Related issue #1363

### DIFF
--- a/array/sort.mbt
+++ b/array/sort.mbt
@@ -313,6 +313,12 @@ test "quick_sort limit check" {
 }
 
 test "quick_sort with pred check" {
+  let arr = []
+  for i = 16; i >= 0; i = i - 1 {
+    arr.push(i)
+  }
+  quick_sort(arr[:], Some(8), 0)
+  assert_eq!(arr, [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16])
   let arr = [5, 4, 3, 2, 1]
   quick_sort(arr[:], Some(3), 0)
   assert_eq!(arr, [1, 2, 3, 4, 5])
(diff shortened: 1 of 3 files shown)


Lampese:
Is there any problem with the ubuntu-latest CI?

peter-jerry-ye (changes requested):
It would be better if the coverage can be raised. To trigger those `guard`, you can `break` inside a `for .. in` loop.

Also please avoid using meme numbers. If possible, please use 0, 1, 2, 3, 12345, 10000, etc., or min/max of Int instead.

SyoujyoujiNaiki:
> To trigger those `guard`, you can `break` inside a `for .. in` loop.

I don't really know what that means.

peter-jerry-ye:
> > To trigger those `guard`, you can `break` inside a `for .. in` loop.
> 
> I don't really know what that means.

Ok, here's the thing. When you are constructing an `Iter`, there's two state after a value is consumed: `IterContinue` or `IterBreak`.

The `IterBreak` comes from :

```moonbit
for i in [1, 2, 3] {
  if i == 2 {
      break // <- here
  }
}
```

And when we receive the signal, we as the developer of `Iter`, know that the iteration is done, so we break. That's why you'll see:

```moonbit
guard let yield_(value) == IterContinue else { x => return x } // <- here
```

So here's a little example:

```moonbit
///|
fn iter_array[T : Show](array : Array[T]) -> Iter[T] {
  Iter::new(fn(yield_) {
    for i = 0; i < array.length(); i = i + 1 {
      guard let IterContinue = yield_(array[i]) else {
        IterEnd => {
          println(array[i])
          return IterEnd
        }
      }

    } else {
      IterContinue
    }
  })
}

///|
fn main {
  for i in iter_array([1, 2, 3, 4, 5]) {
    if i == 3 {
      break
    }
  }
}
```

The previous code should print "3".

In the case of `deque`, there are a [lot of lines missing coverage](https://coveralls.io/builds/71521186/source?filename=deque%2Fdeque.mbt#L625), so I would like you to try to come up with some data that will allow it to break in the `for .. in` loop so that the `guard let` statements get tested.

peter-jerry-ye (approved):

peter-jerry-ye:
Thanks for the PR

PR #1389, opened 2024-12-31
Use saturating conversion primitives float conversion on wasm/wasm-gc backends

tonyfettes:


builtin/moon.pkg.json
```diff
@@ -26,6 +26,9 @@
     "panic_nonjs_test.mbt": ["wasm", "wasm-gc"],
     "array_nonjs_test.mbt": ["not", "js"],
     "stringbuilder_buffer.mbt": ["not", "js"],
-    "stringbuilder_concat.mbt": ["js"]
+    "stringbuilder_concat.mbt": ["js"],
+    "double.mbt": ["not", "wasm", "wasm-gc"],
```
bobzhang, line 30: this combination makes it very hard to reason about if, for example, js platform cover all API

PR #1392, opened 2024-12-31
feat: bigint conversion

peter-jerry-ye:


hackwaly (approved):

builtin/bigint_nonjs.mbt
```diff
@@ -1186,3 +1202,32 @@ pub fn lxor(self : BigInt, other : BigInt) -> BigInt {
     },
   }
 }
+
+///|
```
hackwaly, line 1206: Missing documentation?

builtin/bigint_nonjs.mbt
```diff
@@ -1186,3 +1202,32 @@ pub fn lxor(self : BigInt, other : BigInt) -> BigInt {
...
+pub fn to_int(self : BigInt) -> Int {
+  self.to_uint().reinterpret_as_int()
+}
+
+///|
```
hackwaly, line 1211: And here

tonyfettes (approved):
LGTM. It might be better to add more documentation on publich functions as proposed by @hackwaly .

builtin/bigint_nonjs.mbt
bobzhang: we should document it clear what happens when it overflow

PR #1393, opened 2024-12-31
fix: remove accidentally leaked apis

peter-jerry-ye:


builtin/int64_nonjs.mbt
```diff
@@ -177,7 +177,7 @@ pub fn Double::reinterpret_as_u64(self : Double) -> UInt64 = "%f64_to_i64_reinte
 ///|
 pub fn Double::reinterpret_as_uint64(self : Double) -> UInt64 = "%f64_to_i64_reinterpret"
 
-///|
+///| @alert deprecated "Use `UInt64::to_double` instead"
```
hackwaly, line 180: UInt64 is not safe to represent in Double. I think `UInt64::to_double` is not a good name?

peter-jerry-ye: It is as good / bad as `(int)1.0`.

bobzhang: We should not do two things in a single PR, can you split your changes (as tilte) into a single PR?

peter-jerry-ye:
#1473

PR #1401, opened 2025-01-01
when I run `moon test --doc --update`, it generates invalid output

bobzhang:
cc @Young-Flash when I run moon test --doc --update, it generates invalid output, see the commit for more details


Young-Flash:
will fix at next release

Issue #1366, opened 2024-12-25
Extend JSON to include Int64, UInt64, Arbitrary(String)


peter-jerry-ye:
## Spec

[RFC 8259](https://datatracker.ietf.org/doc/rfc8259/):

> This specification allows implementations to set limits on the range
   and precision of numbers accepted.  Since software that implements
   IEEE 754 binary64 (double precision) numbers [IEEE754] is generally
   available and widely used, good interoperability can be achieved by
   implementations that expect no more precision or range than these
   provide, in the sense that implementations will approximate JSON
   numbers within the expected precision.

[ECMA 404](https://ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf)

> JSON is agnostic about the semantics of numbers. In any programming language, there can be a variety of 
number types of various capacities and complements, fixed or floating, binary or decimal. That can make 
interchange between different programming languages difficult. JSON instead offers only the representation of 
numbers that humans use: a sequence of digits. All programming languages know how to make sense of digit 
sequences even if they disagree on internal representations. That is enough to allow interchange.

## Reference implementation

[serde_json](https://docs.rs/serde_json/1.0.134/src/serde_json/number.rs.html#22-34)
[gson](https://javadoc.io/doc/com.google.code.gson/gson/latest/com.google.gson/com/google/gson/JsonPrimitive.html)
[jackson](https://fasterxml.github.io/jackson-core/javadoc/2.14/com/fasterxml/jackson/core/JsonParser.NumberType.html)
[golang](https://pkg.go.dev/encoding/json#Number)

## Related

#1345

peter-jerry-ye:
One possible solution is:

For `FromJson` and `ToJson` implementation in MoonBit, we guarantee roundtrip & lose no accuracy by:
- Using Double for `UInt` / `Int`
- Using String for `UInt64` / `Int64`

For access custom data source (i.e. JSON shape isn't defined in MoonBit), we support them by adding a field to Number, something like:
```moonbit
enum Json {
  Number(Double, repr? : String)
}
```
And for users who wants to manipulate the number, they can handle the `repr`:
```moonbit
match json {
  { "a" : Number(-100..<100) } => ...
  { "a" : Number(_, repr~) } => ...
}
```

What do you think @gmlewis ?

gmlewis:
That sounds good to me, @peter-jerry-ye - please feel free to close #1345 or let me know if you would like me to do anything to support this.

Thank you!

[...]

PR #1417, opened 2025-01-03
feat: add from/to json for uint/uint64

peter-jerry-ye:
For `FromJson` and `ToJson`, we only guarantee the roundtrip property.

So for `UInt64` and `Int64`, we'll be using `String` as representation, and for `UInt` and `Int`, we'll be using `Number` as representation.

### DIFF
--- a/builtin/builtin.mbti
+++ b/builtin/builtin.mbti
@@ -554,6 +554,7 @@ impl UInt {
   to_byte(UInt) -> Byte
   to_float(UInt) -> Float
   to_int(UInt) -> Int //deprecated
+  to_json(UInt) -> Json
   to_string(UInt) -> String
   to_uint64(UInt) -> UInt64
   trunc_double(Double) -> UInt
@@ -590,6 +591,7 @@ impl UInt64 {
   to_float(UInt64) -> Float
   to_int(UInt64) -> Int
   to_int64(UInt64) -> Int64 //deprecated
+  to_json(UInt64) -> Json
   to_string(UInt64) -> String
   to_uint(UInt64) -> UInt
   trunc_double(Double) -> UInt64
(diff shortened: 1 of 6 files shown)


peter-jerry-ye:
Waiting for #1419

gmlewis:
> So for `UInt64` and `Int64`, we'll be using `String` as representation, and for `UInt` and `Int`, we'll be using `Number` as representation.

I think this decision is a mistake, as it makes MoonBit incompatible with servers written in mainstream languages like Go that serialize `int64` and `uint64` as JSON numbers. I understand that precision can be lost. However, requiring a number to be represented as a string will cause major incompatibility problems in my opinion.
At the very least, if a `Number` is found, it should be parsed and returned.
You can still use `String` as an optional possibility, but to reject a `Number` I think is incorrect.

See: https://github.com/gmlewis/modus/blob/f4e0f1e13d39ca5363c045a8560129b9be56c6ed/sdk/moonbit/examples/auth/main_test.mbt#L33-L36

peter-jerry-ye:
> I think this decision is a mistake, as it makes MoonBit incompatible with servers written in mainstream languages like Go that serialize `int64` and `uint64` as JSON numbers. I understand that precision can be lost. However, requiring a number to be represented as a string will cause major incompatibility problems in my opinion. At the very least, if a `Number` is found, it should be parsed and returned. You can still use `String` as an optional possibility, but to reject a `Number` I think is incorrect.

I think we discussed this in #1366. We will support parsing `int64` and `uint64` as JSON numbers and make the `Number` constructor as `Number(Double, repr~: String)`. The reason why this did not proceed was due to name convention conflict (#1418). The `FromJSON` and `ToJSON` will provide a default behavior which can be overriden by implementing explicitly the `FromJson` and `ToJson` instead of using deriviation.

PR #1418, opened 2025-01-03
feat: smart constructor for Json

peter-jerry-ye:


bobzhang:
I would like to merge this but it is a bit dangerous currently since `boolean`, `array` would pollute the builtin namespace, maybe  we can wait until the new method mechanism landed cc @Guest0x0

bobzhang:
@peter-jerry-ye We should revisit this PR now given the ambiguity is resolved now

peter-jerry-ye:
@bobzhang The code is rebased, but we don't have the attribute to deprecate the constructor yet.

bobzhang (approved):

PR #1429, opened 2025-01-07
add Clone trait

illusory0x0:


illusory0x0:
@peter-jerry-ye 
main branch now execute `moon info` would failed by default. how to solve this problem?

peter-jerry-ye:
Investigating...

FlammeShadow:
Whether or not to "deep" copy seems to be up to the implementor of the trait. Would it be better to be called `Clone`?

Lampese:
> Whether or not to "deep" copy seems to be up to the implementation of this trait. Would it be better to be called `Clone`?

I'm more sympathetic to the idea

peter-jerry-ye:
The problem with `DeepCopy` trait is what should we do with the immutable structure? Do we copy them as well?

illusory0x0:
> Whether or not to "deep" copy seems to be up to the implementation of this trait. Would it be better to be called `Clone`?

Yes. immutable type just return himself reference or value, mutable type should DFS do deep_clone.

immutable container not always immutable type,

```moonbit
trait Mutability {
  is_mutable() -> Bool 
  clone(Self) -> Self 
}

impl Mutability for String with is_mutable() { false }
impl Mutability for String with clone(x) { x }

impl Mutability for Ref[Int] with is_mutable() { true }
impl Mutability for Ref[Int] with clone(x) { x.map(fn (x) {x}) }

fn clone[T : Mutability](xs : @immut/list.T[T]) ->  @immut/list.T[T] {
  if T::is_mutable() {
    xs.map(T::clone)
  } else {
    xs 
  }
}

let xs : @immut/list.T[Ref[Int]] = ([1,2,3] : FixedArray[Int]).map(@ref.new) |> @immut/list.of 

fn main {
  let xs_ = clone(xs)

  let x = xs.head_exn()
  x.val = 30

  println(xs_)
  println(xs)
}
```

```output
@list.of([{val: 1}, {val: 2}, {val: 3}])
@list.of([{val: 30}, {val: 2}, {val: 3}])
```

illusory0x0:
@Lampese @peter-jerry-ye @FlammeShadow  What do you think?

```moonbit
///|
enum Mutability {
  Mutable
  Immutable
}

/// Use clone method intended for cloned object's internal state won't be influenced original object
trait Clone {
  mutability() -> Mutability 
  clone(Self) -> Self 
}
```

peter-jerry-ye:
I think we can give it a try

PR #1431, opened 2025-01-07
internal: add sys package

Young-Flash:
cc @peter-jerry-ye


sys/sys.mbti
```diff
@@ -0,0 +1,19 @@
+package moonbitlang/core/sys
+
+// Values
+fn exit(Int) -> Unit
+
+fn get_cli_args() -> Array[String]
+
+fn get_env_vars() -> Map[String, String]
```
peter-jerry-ye, line 8: Maybe we should have `get_env` for now

bobzhang: @env.args
@env.var

PR #1433, opened 2025-01-07
add other function for Iter2 similar to Iter1

illusory0x0:
Iter2::combine may conflict with Hasher::combine, `zip` will be better possible.

API reference

https://ocaml.org/manual/5.2/api/Array.html#VALcombine

https://hackage.haskell.org/package/base-4.21.0.0/docs/[Prelude](https://hackage.haskell.org/package/base-4.21.0.0/docs/Prelude.html#v:zip).html#v:zip

### DIFF
--- a/builtin/iter.mbt
+++ b/builtin/iter.mbt
@@ -904,3 +904,20 @@ pub fn contains[A : Eq](self : Iter[A], value : A) -> Bool {
     false
   }
 }
+
+///|
+pub fn combine[A, B](self : Iter[A], b : Iter[B]) -> Iter2[A, B] {
+  Iter2::new(fn(k) {
+    let (x, xs) = (self.head(), self.drop(1))
+    let (y, ys) = (b.head(), b.drop(1))
+    match (x, y) {
+      (Some(x), Some(y)) =>
+        match k(x, y) {
+          IterEnd => IterEnd
+          IterContinue => xs.combine(ys).run(k)
+        }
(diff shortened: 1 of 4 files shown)


illusory0x0:
`Iter1[T]` and `Iter2[A,B]` is stream or CoInductive type, we can peek one element with small step, rather than using to_array enumerate all element, so we can `combine` or `zip` `iter1` and `iter2` feel free. It necessary to add zip operation

> In Batteries Included, enumerations are used pervasively, both as a uniform manner of reading and manipulating the contents of a data structure

[ocaml batteries](https://ocaml-batteries-team.github.io/batteries-included/hdoc2/BatEnum.html)

illusory0x0:
- #1435 
use Iter[V] may cause worse performance.

use Iter[V] pervasively is naive thought.

builtin/iter.mbt
```diff
@@ -904,3 +904,20 @@ pub fn contains[A : Eq](self : Iter[A], value : A) -> Bool {
...
+      (Some(x), Some(y)) =>
+        match k(x, y) {
+          IterEnd => IterEnd
+          IterContinue => xs.combine(ys).run(k)
+        }
+      (None, None) => IterEnd
+      _ => panic()
+    }
+  })
+}
```
peter-jerry-ye, line 923: This is basically impractical as this has n^2 complexity.

PR #1435, opened 2025-01-07
use Iter to avoid create array for op_equal

illusory0x0:


hackwaly:
I guess new implementation would perform much worse than old one.

Lampese:
However, there may be areas where performance can be improved, such as early stopping in unequal cases may be better than converting to Array, but the implementation in this PR may be a bit complicated.

PR #1436, opened 2025-01-08
add `StringView`

Yu-zh:
Basic usage:
```
let str = "Hello🤣🤣🤣"
guard let Some(start) = CharOffset::offset_by(str, 1)
guard let Some(end) = CharOffset::offset_by(str, 6)
let view = str[start:end]
inspect!(view, content="ello🤣")
```

To create a `CharsView` from a `String`, we need to use the `CharOffset` as indices. `CharOffset` is an opaque type that represents a position in the utf-16 String. It can be constructed via `CharOffset::offset_by(str: String, offset: int)` where the offset is the number of characters skipped between the start of the string and the returning position. It can also be constructed via `CharOffset::unsafe_from_int(index: Int)`.

This is to avoid confusion with indexing `String` using indices, i.e. indices indicate utf-16 code unit offset when accessing `String` whereas when viewing `String` the indices represents offsets of characters.

Once we hold a `CharsView`, we can use integers directly to access chars or create subviews, and here the index always represents the codepoint offset.

---

Update:
renamed the types and methods:

`CharsView` -> `StringView`
`CharOffset` -> `StringIndex`
`offset_by` -> `index_at`

### DIFF
--- a/builtin/char.mbt
+++ b/builtin/char.mbt
@@ -1,4 +1,4 @@
-// Copyright 2024 International Digital Economy Academy
+// Copyright 2025 International Digital Economy Academy
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/string/string.mbti
+++ b/string/string.mbti
@@ -3,6 +3,18 @@ package moonbitlang/core/string
 // Values
 
 // Types and methods
+type StringIndex
+impl Eq for StringIndex
+impl Show for StringIndex
+
+type StringView
+impl StringView {
+  length(Self) -> Int
+  op_as_view(Self, start~ : Int = .., end? : Int) -> Self
+  op_get(Self, Int) -> Char
+}
+impl Show for StringView
+
 impl String {
   compare(String, String) -> Int
   concat(Array[String], separator~ : String = ..) -> String
@@ -13,12 +25,14 @@ impl String {
   fold[A](String, init~ : A, (A, Char) -> A) -> A
   from_array(Array[Char]) -> String
   from_iter(Iter[Char]) -> String
+  index_at(String, Int, start~ : StringIndex = ..) -> StringIndex?
   index_of(String, String, from~ : Int = ..) -> Int
   is_blank(String) -> Bool
   is_empty(String) -> Bool
   iter(String) -> Iter[Char]
   iter2(String) -> Iter2[Int, Char]
   last_index_of(String, String, from~ : Int = ..) -> Int
+  op_as_view(String, start~ : StringIndex = .., end? : StringIndex) -> StringView
   pad_end(String, Int, Char) -> String
   pad_start(String, Int, Char) -> String
   repeat(String, Int) -> String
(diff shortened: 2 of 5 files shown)


builtin/charsview.mbt
```diff
@@ -0,0 +1,253 @@
...
+  // - `start`: Starting UTF-16 code unit index into the string
+  // - `end`: Ending UTF-16 code unit index into the string (not included)
+  // - `len`: Number of Unicode codepoints (characters) in this view
+  //
+  // Note that `start` and `end` are UTF-16 code unit indices, while `len` 
+  // counts actual Unicode codepoints which may span multiple UTF-16 code units.
+  str : String
+  start : Int
+  end : Int
+  len : Int
```
bobzhang, line 32: remove `len`

builtin/charsview.mbt
```diff
@@ -0,0 +1,253 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+///| 
+/// A `StringView` represents a view of a String that maintains proper Unicode
+/// character boundaries. It allows safe access to a substring while handling 
+/// multi-byte characters correctly.
+struct StringView {
+  // # Fields
+  //
```
bobzhang, line 21: filename rename

bobzhang: support
```
a[-3:-1]
a[-1:]
a[0:-1]
```

builtin/builtin.mbti
```diff
@@ -11,6 +11,8 @@ fn assert_not_eq[T : Eq + Show](T, T, loc~ : SourceLoc = _) -> Unit!
 
 fn assert_true(Bool, loc~ : SourceLoc = _) -> Unit!
 
+fn code_point_of_surrogate_pair(Char, Char) -> Char
+
```
bobzhang, line 15: hide

builtin/builtin.mbti
```diff
@@ -388,6 +394,22 @@ impl StringBuilder {
 }
 impl Show for StringBuilder
 
+type StringIndex
+impl StringIndex {
+  index_at(String, Int) -> Self?
+  unsafe_from_int(Int) -> Self
```
bobzhang, line 400: hide `unsafe_from_int`

builtin/charsview_test.mbt
```diff
@@ -0,0 +1,131 @@
...
+  inspect!(offset, content="Some(StringIndex(4))")
+  let offset = StringIndex::index_at(str, 3)
+  inspect!(offset, content="None")
+}
+
+test "charsview basic" {
+  let str = "Hello🤣🤣🤣"
+  guard let Some(start) = StringIndex::index_at(str, 1)
+  guard let Some(end) = StringIndex::index_at(str, 6)
+  inspect!(
```
bobzhang, line 37: String::(self, startIndex : StringIndex = 0, by : Int) -> StringIndex

bobzhang (commented):
try to move StringIndex/StringView into string package

Issue #1423, opened 2025-01-03
FixedArray::deep_clone is NOT really a deep clone


CAIMEOX:
The implementation of `deep_clone` in `fixedarray.mbt` is not really a deep clone because it won't clone elements.
https://github.com/moonbitlang/core/blob/e68786433d2cce64773cad7eebace0ca4a2f634c/array/fixedarray.mbt#L958
Unexpected behavior happens when there is internal mutability:
```moonbit
test {
  let arr : FixedArray[_] = [@ref.new(1), @ref.new(2), @ref.new(3)]
  let cloned = deep_clone(arr)
  arr[0].val = 0
  inspect!(cloned, content="[{val: 0}, {val: 2}, {val: 3}]")
}
```

illusory0x0:
@CAIMEOX  
cc #1429

peter-jerry-ye:
`deep_clone` no longer exists. Closing the issue.

Issue #527, opened 2024-06-06
Array::deep_clone is not appropriate


hackwaly:
The `Array::deep_clone` currently implemented as https://github.com/moonbitlang/core/blob/b231cb9098d9f98b5dd6e3a195aadfd899edf2b5/builtin/array.mbt#L165-L171

In my opinion. Deep clone should looks like this:

```
fn deep_clone[T : Clonable](self : Array[T]) -> Array[T] {
  // ...
}
```

I suggest to change `deep_clone` to `copy` or `shallow_copy` or `clone`.

PR #1438, opened 2025-01-08
add Clone trait

illusory0x0:
- #1429 

I apologize for turning off the previous pull request by mistake.

### DIFF
--- a/builtin/clone.mbt
+++ b/builtin/clone.mbt
@@ -1,0 +1,261 @@
+// Copyright 2025 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+///|
+impl Clone for Unit with mutability() { Immutable }
+
+///|
+impl Clone for Unit with clone(self) { self }
+
+///|
+impl Clone for Bool with mutability() { Immutable }
+
+///|
+impl Clone for Bool with clone(self) { self }
+
+///|
+impl Clone for Int with mutability() { Immutable }
+
+///|
+impl Clone for Int with clone(self) { self }
+
+///|
+impl Clone for Int64 with mutability() { Immutable }
+
+///|
+impl Clone for Int64 with clone(self) { self }
+
+///|
+impl Clone for UInt with mutability() { Immutable }
+
+///|
+impl Clone for UInt with clone(self) { self }
+
+///|
+impl Clone for Byte with mutability() { Immutable }
+
+///|
+impl Clone for Byte with clone(self) { self }
+
+///|
+impl Clone for String with mutability() { Immutable }
+
+///|
+impl Clone for String with clone(self) { self }
+
+// mutable containers
+
+///|
+fn copy_array[T](self : Array[T]) -> Array[T] {
+  let len = self.length()
+  if len == 0 {
+    []
+  } else {
+    let arr = Array::make(len, self[0])
+    Array::unsafe_blit(arr, 0, self, 0, len)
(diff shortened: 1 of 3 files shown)


peter-jerry-ye (approved):

builtin/clone.mbt
```diff
@@ -0,0 +1,261 @@
...
+test "Ref[Int]::clone" {
+  let x = { val: 0 }
+  let y = x.clone()
+  x.val = 10
+  inspect!(x, content="{val: 10}")
+  inspect!(y, content="{val: 0}")
+}
+
+///|
+impl Clone for Bytes with mutability() { Mutable }
```
peter-jerry-ye, line 159: We will switch to immutable @Yu-zh

builtin/clone.mbt
```diff
@@ -0,0 +1,261 @@
...
+
+///|
+impl[T : Clone, E : Clone] Clone for Result[T, E] with clone(self) {
+  match T::mutability() {
+    Mutable =>
+      match self {
+        Ok(x) => Ok(T::clone(x))
+        Err(x) => Err(E::clone(x))
+      }
+    Immutable => self
```
Yu-zh, line 217: You also need to check E::mutability() here. I'd suggest
```
match self {
  Ok(x) => if T is mutable ...
  Err(x) => if E is mutable ...
}
```

builtin/clone.mbt
```diff
@@ -0,0 +1,261 @@
...
+}
+
+test "mutability" {
+
+  // immutable 
+  let oi = Some(0)
+  inspect!(get_mutability(oi), content="Immutable")
+  let ris : Result[Int, String] = Ok(32)
+  inspect!(get_mutability(ris), content="Immutable")
+  // TODO, we need to test reference equal
```
Yu-zh, line 252: we have a function `physical_equal` but be aware that its behavior is backend-dependent.

bobzhang:
what's the motivation for this API? it looks to me we are adding lots of API for a niche use case?

illusory0x0:
> what's the motivation for this API? it looks to me we are adding lots of API for a niche use case?

- #1448
- #1423
- #527

FlammeShadow:
It seems the type `Mutability` is only used by `mutability`. Why not just use `is_mutable(Self) -> Bool` ?

illusory0x0:
stm32 use RESET SET, ENABLE DISABLE
enum, rather than only true and false.

add new enum for type safe, and more meaningful

> It seems the type `Mutability` is only used by `mutability()`. Why not just use `is_mutable(Self) -> Bool` ?

FlammeShadow:
> this is just compiletime constant, not depend instance, below is enough
> `is_mutable() -> Bool`

My fault. It was my mistake.

illusory0x0:
cc @Guest0x0

Lampese:
> what's the motivation for this API? it looks to me we are adding lots of API for a niche use case?

I don't think this thing is niche. There is indeed a lack of effective methods to deeply replicate a data structure, especially when some mutable data structures (Array [sorted set]) are nested. Using Clone Trap is also much more convenient This is considered good news for some mutable structures.

Lampese:
> > what's the motivation for this API? it looks to me we are adding lots of API for a niche use case?
> 
> * [sorted_set.add is not appropriate #1448](https://github.com/moonbitlang/core/issues/1448)
> * [FixedArray::deep_clone is NOT really a deep clone #1423](https://github.com/moonbitlang/core/issues/1423)
> * [Array::deep_clone is not appropriate #527](https://github.com/moonbitlang/core/issues/527)

cc @bobzhang @illusory0x0 

Is this referring to discussing the necessity of the mutable API rather than the necessity of the Clone trait?

I think the original intention of distinguishing between immutable and mutable in the design was because the Clone behavior of immutable is unreasonable, but there may be some problems with the API design. Users are already explicitly using this API when using Clone, which proves that the cost of deep replication of immutable data structures is visible and affordable for users (we do have such a need, if there is a lot of mutable data stored in an immutable data structure) Sacrificing convenience to distinguish the need for variability may not be ideal

illusory0x0:
> Is this referring to discussing the necessity of the mutable API rather than the necessity of the Clone trait?

- #1448

cc @Lampese

peter-jerry-ye:
Closing this PR for now. The content of this PR can be implemented in a community package, and we can discuss whether we should merge it into core when that package gets widely adopted.

PR #1456, opened 2025-01-09
improve eachi for sorted_set

illusory0x0:
```moonbit
Benchmark Task [fast_eachi] Count = 10
----------------------------------------
Average Time: 0.04309862s/per test
Max Time Per Test: 0.0475998s/per test
Min Time Per Test: 0.0427086s/per test
----------------------------------------
Benchmark Task [eachi] Count = 10
----------------------------------------
Average Time: 0.0559544s/per test
Max Time Per Test: 0.0558127s/per test
Min Time Per Test: 0.0287929s/per test
----------------------------------------
Total tests: 31, passed: 31, failed: 0.
```

```
test "benchmark" {
  let criterion = @bm.Criterion::new()
  let test_data = new()
  let rand = @random.new()
  for i in 0..<100000 {
    test_data.add(rand.int())
  }
  criterion.add(
    @bm.Task::new("fast_eachi", fn() {
      let mut b = false
      for _ in 0..<20 {
        test_data.fast_eachi(fn(_, v) { b = b || v % 2 == 0 })
      }
    }),
  )
  criterion.add(
    @bm.Task::new("eachi", fn() {
      let mut b = false
      for _ in 0..<20 {
        test_data.eachi(fn(_, v) { b = b || v % 2 == 0 })
      }
    }),
  )
  let result = criterion.run()
  println(result["fast_eachi"].unwrap())
  println(result["eachi"].unwrap())
}
```


peter-jerry-ye:
> Potential Stack Overflow:

> The modified code uses a recursive depth-first search (dfs) function to traverse the tree. While this approach is more concise, it could lead to a stack overflow for very large trees due to deep recursion. The original iterative approach using a stack (s) would be more memory-efficient and avoid this risk.

hackwaly:
The depth never be too large since we are using balanced binary tree. I think dfs is acceptable.

illusory0x0:
self.size < 10000,use this impl, otherwise use original impl

> The depth never be too large since we are using balanced binary tree. I think dfs is acceptable.
> 
> Another question: Do we really need `eachi` as we have `iter` already.

PR #1459, opened 2025-01-10
improve to_array for sorted_set

illusory0x0:
I think `Array::make_uninit` should be public , and use it carefully, rename `make_uninit` to `unsafe_make_uninit` or another better name


peter-jerry-ye:
Not available on JS backend.

peter-jerry-ye:
Use it under `JS` backend will cause the `length` method return the wrong number:

```js
let a = Array(5)
console.log(a.length) // <- 5
```

PR #1460, opened 2025-01-11
Chore: update copyright year for all files to 2025 + run `moon fmt`

mindreframer:
Just to make sure the copyright year is consistent. 

Thanks for making MoonBit, it's a joy to use!
Roman


bobzhang:
@mindreframer thank you for nice words!

PR #1468, opened 2025-01-13
improve to_array for sorted_set (use safe way)

illusory0x0:
- #1459

### DIFF
--- a/sorted_set/set.mbt
+++ b/sorted_set/set.mbt
@@ -350,9 +350,28 @@ pub fn eachi[V](self : T[V], f : (Int, V) -> Unit) -> Unit {
 ///|
 /// Converts the set to an array.
 pub fn to_array[V](self : T[V]) -> Array[V] {
-  let arr = []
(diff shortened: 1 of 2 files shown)


sorted_set/set.mbt
```diff
@@ -350,9 +350,14 @@ pub fn eachi[V](self : T[V], f : (Int, V) -> Unit) -> Unit {
...
-  let arr = []
-  self.each(fn(v) { arr.push(v) })
-  arr
+  if self.size == 0 {
+    []
+  } else {
+    let padding = self.root.unwrap().value
+    let arr = Array::make(self.size.to_int(), padding)
+    self.eachi(fn(i, v) { arr[i] = v })
+    arr
```
bobzhang, line 359: since we are squeezing performance here, can you inline `eachi` here?

illusory0x0: so we also need inline `each` into `eachi`?

PR #1470, opened 2025-01-13
add compare for sorted_set

illusory0x0:


bobzhang:
cc @hackwaly I remember you wrote one reasonably efficient version before?

hackwaly:
It's there: 
https://github.com/moonbitlang/core/blob/b45fcd74b4249d22c244afcf2341874a7c392e3f/immut/sorted_set/generic.mbt#L64-L77

peter-jerry-ye:
I think the design principle for comparing containers in core is to compare the size first?

Lampese:
I have one concern: Will structure-based comparisons affect the results of the comparison because of the order in which the nodes are inserted? I'll test that out later.

hackwaly:
> I think the design principle for comparing containers in core is to compare the size first?

I don't think so. String::compare isn't, and Some container don't have `O(1)` length at all.

peter-jerry-ye:
> > I think the design principle for comparing containers in core is to compare the size first?
> 
> I don't think so. Usually, `String::compare` isn't, and Some containers don't have `O(1)` length at all.
> 
> but, here we can compare size first due to performance reason.

I'm pretty sure we compare string with length first. Also, the length can always be traced for any container.

sorted_set/set.mbt
```diff
@@ -299,9 +299,42 @@ pub fn disjoint[V : Compare](self : T[V], src : T[V]) -> Bool {
...
+    }
+    ()
+  }
+
+  dfs(self.root)
+  if result != 0 {
+    result
+  } else {
+    (self.size - other.size).to_int()
+  }
```
peter-jerry-ye, line 332: Here we do dfs first and then compare length, which is weird.

hackwaly:
> Also, the length can always be traced for any container.

The case is `Iter::compare`, and if we gonna to have something like `LazySeq`.

peter-jerry-ye:
> > Also, the length can always be traced for any container.
> 
> The case is `Iter::compare`, and if we gonna to have something like `LazySeq`.

Do you mean comparing two `Iter`? I don't think that's possible with internal iterator, same reason as `Iter::zip`. As of `LazySeq`, yeah, that's the only case where we have to compare the elements first. But do we really need that? You might encounter infinite length Seq.

peter-jerry-ye:
Closing this PR for now.

I think we need stronger arguments for having `compare` on mutable structure and `immut/sorted_set` can always be an alternative. The `op_equal` need to be improved later.

PR #1472, opened 2025-01-13
refine the call syntax & semantic of method

Guest0x0:
We plan to change the call syntax rule of methods to make it simpler & more robust in the near future, this PR migrates API in core.

## Current semantic
- methods can be declared with `fn f(self : T, ..)` or `fn T::f(..)`
- methods whose first parameter is of type `Self` can be called with dot syntax
- when there is no ambiguity, methods call be called with `f(..)` or `@pkg.f(..)` directly

the third rule above is useful (for example, `@hashmap.from_array(..)` v.s. `@hashmap.T::from_array(..)`). But it is implicit and sometimes chaotic (especially in the `builtin` package). We want to make the API/available call syntax of toplevel definitions clear and explicit. So we plan to change the call syntax rules of methods to:

## Proposed new semantic
- methods can be declared with `fn f(self : T, ..)` or `fn T::f(..)`
- methods whose first parameter is of type `Self` can be called with dot syntax
- functions declared with `fn f(..)` can be called via `f(..)` or `@pkg.f(..)`. Functions declared with `fn T::f(..)` must be called via `T::f(..)`

notice the difference in the third rule. The new rule makes valid call syntax of methods clear and explicit: users call functions in the same way you they are declare (`f(..)` or `T::f(..)`).

## Migration & API design
The new rule require library authors to modify their definitions. This PR can serve as an example for how to perform such migration. Under the new rule, the guideline for API design is:

- if a function has no ambiguity in a package, declare it with `fn f(..)`
    - if you want users to call this function with dot syntax, name the first parameter `self`
- if a function is ambiguous in a package, declare it with `fn T::f(..)`, and use the namespace of `T` to do disambiguation

For example, in the `@hashmap` package, previously there is a method `fn T::new()`. But since `new` is unambiguous in `@hashmap`'s public API, and we want users to call it directly via `@hashmap.new()`, so `new` should really be defined as a regular function:
```moonbit
fn new() -> T {
  ...
}
```
In the builtin package, there are a lot of types and methods. So all these methods should be declared with `fn T::f(..)` to avoid name clash and confusion. The previous API is chaotic. For example `from_string(...)` is actually `BigInt::from_string`, because it is the only `from_string` in `@builtin`. In the future such call syntax will be forbidden. Users must write `BigInt::from_string` explicitly.

## Migration schedule
Later this week, we will report a warning on the old behavior (name clash between `fn f(self :T)` and regular function, calling `fn T::f(..)` via `f(..)` directly). A few weeks later we will remove the old behavior completely and migrate to the new behavior.

## The migration here
The migration here in core consist of:

- remove unnecessary `T::` in the unambiguous case
    - for methods whose first parameter is not `self`, a deprecated method definition is added, for compatibility with the fully qualified `@pkg.T::new()` syntax
- add `T::` in the ambiguous case, e.g. `@builtin`
- migrate fully qualified `@pkg.T::method(..)` call to `@pkg.method(..)` if possible


gmlewis:
> We plan to change the call syntax rule of methods to make it simpler & more robust in the near future, this PR migrates API in core.

@Guest0x0 - thank you for the excellent, detailed, and clearly-written heads-up!  It is deeply appreciated!

bobzhang (approved):

Issue #1481, opened 2025-01-14
Deque::truncate run failed when `len` < 0


illusory0x0:
```moonbit
test {
  let d = @deque.from_array([1,2,3,4,5])
  d.truncate(-1)
  println(d)
}
```

we might need handle `len < 0` or use `UInt`

bobzhang:
this is documented to be a panic behavior like out of bound access

PR #1483, opened 2025-01-15
deprecate Deque::truncate, rename it to unsafe_truncate

illusory0x0:
if a API might `panic`, I think rename it to `unsafe_xxx` would be better, or we can handle the `panic` case make it **safe**,

like this.

```moonbit
/// @alert unsafe "Panic if the array is empty."
pub fn Array::unsafe_pop[T](self : Array[T]) -> T
```

related issues.
- #1481

### DIFF
--- a/deque/deque.mbt
+++ b/deque/deque.mbt
@@ -866,19 +866,21 @@ pub fn shrink_to_fit[A](self : T[A]) -> Unit {
 ///
 /// Parameters:
 ///
-/// * `self` : The deque to be truncated.
+/// * `self` : The deque to be unsafe_truncated.
 /// * `len` : The new length of the deque.
 ///
 /// Example:
 ///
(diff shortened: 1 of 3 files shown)


peter-jerry-ye:
All the `op_get` would be unsafe. We need to discuss this later.

Lampese:
> All the `op_get` would be unsafe. We need to discuss this later.

＊ex: Map

peter-jerry-ye:
`Map` was designed exceptionally because of the 'map pattern matching', which can also be discussed whether `op_get` should be used or some other function should be used...

peter-jerry-ye:
We will have no effect if there's an invalid input.

And as of the naming, we will add `unsafe_` when there is a safe equivalent.

peter-jerry-ye:
Replaced by #2003

PR #1485, opened 2025-01-15
neg_infinity.ln() should return nan, not -inf

Kaida-Amethyst:
`neg_infinity.ln()` should return `not_a_number`, not `neg_infinity`.

See the following python code:

```python
>>> import numpy as np
>>> import warnings # set ignore runtimewarning
>>> warnings.filterwarnings("ignore", category=RuntimeWarning)
>>> np.log([-np.inf])
array([nan])
```

In javaScript, run the following code:

```javascript
console.log(Math.log(-Infinity))
```

we could get `NaN`

### DIFF
--- a/double/log.mbt
+++ b/double/log.mbt
@@ -95,10 +95,10 @@ fn frexp(f : Double) -> (Double, Int) {
 /// }
 /// ```
 pub fn ln(self : Double) -> Double {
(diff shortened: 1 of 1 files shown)


Lampese (approved):

tonyfettes (approved):
LGTM

Issue #1487, opened 2025-01-16
[Feature Request] Char missing useful method


illusory0x0:
[Rust Char](https://doc.rust-lang.org/std/primitive.char.html#method.is_ascii_digit)

we also need `is_digit`, `is_lowercase` and anothor predicate for Char.

and other convert function like , `to_digit`,  `to_lowercase`

bungcip:
adding `to_ascii_uppercase` and `to_ascii_lowercase` would be nice

peter-jerry-ye:
Accomplished in #2140

PR #1494, opened 2025-01-16
Add parts of methods for Char.

SyoujyoujiNaiki:
Related issue: #1487

### DIFF
--- a/char/char.mbt
+++ b/char/char.mbt
@@ -108,17 +108,17 @@ pub fn is_control(self : Char) -> Bool {
 }
 
 ///| Checks if a char is a digit in the given radix (range from 2 to 36).
-pub fn is_digit(self : Char, radix : UInt) -> Bool? {
+pub fn is_digit(self : Char, radix : UInt) -> Bool!Error {
+  guard radix >= 2 && radix <= 36 else {
+    fail!("radix must be in the range from 2 to 36")
+  }
   let v = self.to_uint()
   match radix {
-    2..=10 => Some(v >= 48 && v <= radix + 47)
-    11..=36 =>
-      Some(
-        (v >= 48 && v <= 57) ||
-        (v >= 65 && v <= radix + 54) ||
-        (v >= 97 && v <= radix + 86),
-      )
-    _ => None
+    2..=10 => v >= 48 && v <= radix + 47
+    _ =>
+      (v >= 48 && v <= 57) ||
+      (v >= 65 && v <= radix + 54) ||
+      (v >= 97 && v <= radix + 86)
   }
 }
 
--- a/char/char.mbt
+++ b/char/char.mbt
@@ -107,16 +107,16 @@ pub fn is_control(self : Char) -> Bool {
   (self >= '\u007F' && self <= '\u009F')
 }
 
-///| Checks if a char is a digit in the given radix (range from 1 to 36).
+///| Checks if a char is a digit in the given radix (range from 2 to 36).
 pub fn is_digit(self : Char, radix : UInt) -> Bool? {
   let v = self.to_uint()
   match radix {
-    1..=10 => Some(v >= 48 && v <= radix + 48 - 1)
+    2..=10 => Some(v >= 48 && v <= radix + 47)
     11..=36 =>
       Some(
-        (v >= 48 && v <= radix + 48 - 1) ||
-        (v >= 65 && v <= radix + 65 - 11) ||
-        (v >= 97 && v <= radix + 97 - 11),
+        (v >= 48 && v <= 57) ||
+        (v >= 65 && v <= radix + 54) ||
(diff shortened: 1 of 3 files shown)


char/char.mbt
```diff
@@ -19,3 +19,295 @@ pub impl Hash for Char with hash(self) { self.to_int() }
 pub impl Hash for Char with hash_combine(self, hasher) -> Unit {
   hasher.combine_char(self)
 }
+
+///| Checks if the value is within the ASCII range.
+pub fn is_ascii(self : Char) -> Bool {
+  guard self >= '\x00' && self <= '\x7F' else { false }
+  true
```
peter-jerry-ye, line 26:
```suggestion
  self >= '\x00' && self <= '\x7F'
```

char/char.mbt
```diff
@@ -19,3 +19,295 @@ pub impl Hash for Char with hash(self) { self.to_int() }
...
+  let v = self.to_uint()
+  match radix {
+    1..=10 => Some(v >= 48 && v <= radix + 48 - 1)
+    11..=36 =>
+      Some(
+        (v >= 48 && v <= radix + 48 - 1) ||
+        (v >= 65 && v <= radix + 65 - 11) ||
+        (v >= 97 && v <= radix + 97 - 11),
+      )
+    _ => None
```
peter-jerry-ye, line 143: This is false

SyoujyoujiNaiki: What does 'This' refer to?

peter-jerry-ye:
This condition. If it’s not number nor ascii, it’s certainly not a digit.

在 2025年1月16日，19:31，Lim Weiihang ***@***.***> 写道：

﻿

@SyoujyoujiNaiki commented on this pull request.

________________________________

In char/char.mbt<https://github.com/moonbitlang/core/pull/1494#discussion_r1918292897>:

> +  }
+  true
+}
+
+///| Checks if a char is a digit in the given radix (range from 1 to 36).
+pub fn is_digit(self : Char, radix : UInt) -> Bool? {
+  let v = self.to_uint()
+  match radix {
+    1..=10 => Some(v >= 48 && v <= radix + 48 - 1)
+    11..=36 =>
+      Some(
+        (v >= 48 && v <= radix + 48 - 1) ||
+        (v >= 65 && v <= radix + 65 - 11) ||
+        (v >= 97 && v <= radix + 97 - 11),
+      )
+    _ => None


What does 'This' refer to?

char/char.mbt
```diff
@@ -19,3 +19,267 @@ pub impl Hash for Char with hash(self) { self.to_int() }
...
+  (self >= '\x21' && self <= '\x2F') ||
+  (self >= '\x3A' && self <= '\x40') ||
+  (self >= '\x5B' && self <= '\x60') ||
+  (self >= '\x7B' && self <= '\x7E')
+}
+
+///| Checks if the value is an ASCII uppercase character:
+/// U+0041 ‘A’ ..= U+005A ‘Z’
+pub fn is_ascii_uppercase(self : Char) -> Bool {
+  self >= 'A' && self <= 'Z'
```
bobzhang, line 91: you can pattern match over char too.
```
match self {
  'A'..='Z' => true
  _ => false 
}
```

char/char.mbt
bobzhang: he probaly means it should return `Bool`

peter-jerry-ye: `is_digit('1', 100) == None` ==> `is_digit('1', 100) == false`

peter-jerry-ye: or maybe we can simply just panic?

peter-jerry-ye: Or else, you can just follow the Java's API: <https://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#isDigit-int->

char/char.mbt
SyoujyoujiNaiki: How could we benefit from pattern match rather boolean expression?

Lampese: > How could we benefit from pattern match rather boolean expression?

This way, the writing of 'range' is easier to read and more uniform

peter-jerry-ye (approved):

char/char.mbt
```diff
@@ -19,3 +19,300 @@ pub impl Hash for Char with hash(self) { self.to_int() }
...
+  match self {
+    '\u0000'..='\u001F' | '\u007F'..='\u009F' => true
+    _ => false
+  }
+}
+
+///| Checks if a char is a digit in the given radix (range from 2 to 36).
+pub fn is_digit(self : Char, radix : UInt) -> Bool!Error {
+  guard radix >= 2 && radix <= 36 else {
+    fail!("radix must be in the range from 2 to 36")
```
lynzrand, line 140: This should panic instead of failing; the error is rare and should be considered user error. Will fix in another PR.

PR #1495, opened 2025-01-16
Add conversion between Char and Bytes/BytesView.

SyoujyoujiNaiki:


peter-jerry-ye:
The problem with this API is: there's unlikely to be strings with UTF-32 representation (or UCS-4). We should not give people some false idea.

SyoujyoujiNaiki:
> The problem with this API is: there's unlikely to be strings with UTF-32 representation (or UCS-4). We should not give people some false idea.

The conversion is treating char as Unicode point like what [documents present](https://docs.moonbitlang.com/en/latest/language/fundamentals.html#char:~:text=Char%20represents%20a%20Unicode%20code%20point.) .  There is nothing to do with UTF-32.

peter-jerry-ye:
Yes, but when would you need to convert a char from/to string other than use encoding it?

peter-jerry-ye:
When we use the APIs that converts data from/to Bytes representation, we are performing serialization/deserialization. The format of the numbers are usually consistent across language / protocol. The format of chars however, varies. For example, the [protobuf](https://protobuf.dev/programming-guides/encoding/#cheat-sheet) and [msgpack](https://github.com/msgpack/msgpack/blob/master/spec.md#type-system) defines string as UTF8 strings only. Other protocols rarely uses UTF32. The internal representation of MoonBit is also UTF16.

Also, it would be better that the character / string related functions being curated inside the encoding package, which is being developed at [x](https://github.com/moonbitlang/x/tree/main/encoding).

In conclusion, I think this API might be misleading, and we can alternatively add UTF32 encoding/decoding in package encoding.

PR #1510, opened 2025-01-20
to_string_by_radix method for standard integers

### LINKED ISSUE #1500: [Feature Request] `to_hex` Method for Standard Integer Types (Int, UInt, Int64, UInt64, etc.)
When dealing with large integers, such as those in IEEE754 double format, converting them to hexadecimal representation can be quite useful. Currently, using `BigInt::from_int64(...).to_hex()` is a bit cumbersome. It would be more convenient to have a `to_hex` method directly available for standard integer types like `Int, UInt, Int64, UInt64`, etc.

Example :

```moonbit
let ieee754_one = 1.0.reinterpret_as_int64().to_hex()
println("ieee754_one = 0x\{ieee754_one}") 
```
Expected Result:
```shell
ieee754_one = 0x3fff0000000000000
```

Current Workaround:
To achieve the same result, we currently need to use BigInt::from_int64(...).to_hex():

```moonbit
let ieee754_one = BigInt::from_int64(1.0.reinterpret_as_int64()).to_hex()
println("ieee754_one = 0x\{ieee754_one}") 
```

Adding a direct to_hex method for standard integer types would simplify this process and improve code readability.

Kaida-Amethyst:
This PR introduces the new feature described in #1500 .  

**Closes #1500 .**  

In JavaScript, the `toString` method allows converting numbers to strings. By default, `toString()` returns the decimal format, while `toString(16)` returns the hexadecimal format.  

However, directly modifying `to_string` in MoonBit can lead to issues. For example:  

```moonbit
let x = 1
println("x = \{x}")
// ^ This calls `x.to_string()`, which requires `to_string` keeping type `(Int) -> String`.
```

To address this, I've added a new function `to_string_by_radix`. If there's a better name for this function, I'm open to suggestions!

### DIFF
--- a/builtin/builtin.mbti
+++ b/builtin/builtin.mbti
@@ -454,7 +454,7 @@ impl Int {
   to_int16(Int) -> Int16
   to_int64(Int) -> Int64
   to_json(Int) -> Json
-  to_string(Int) -> String
+  to_string(Int, radix~ : Int = ..) -> String
   to_uint(Int) -> UInt //deprecated
   to_uint16(Int) -> UInt16
   to_uint64(Int) -> UInt64
@@ -466,7 +466,7 @@ impl Int16 {
   to_byte(Int16) -> Byte
   to_int(Int16) -> Int
   to_int64(Int16) -> Int64
-  to_string(Int16) -> String
+  to_string(Int16, radix~ : Int = ..) -> String
 }
 
 impl Int64 {
@@ -501,7 +501,7 @@ impl Int64 {
   to_int(Int64) -> Int
   to_int16(Int64) -> Int16
   to_json(Int64) -> Json
-  to_string(Int64) -> String
+  to_string(Int64, radix~ : Int = ..) -> String
   to_uint16(Int64) -> UInt16
   to_uint64(Int64) -> UInt64 //deprecated
   until(Int64, Int64, step~ : Int64 = .., inclusive~ : Bool = ..) -> Iter[Int64]
@@ -536,7 +536,7 @@ impl UInt {
   to_float(UInt) -> Float
   to_int(UInt) -> Int //deprecated
   to_json(UInt) -> Json
-  to_string(UInt) -> String
+  to_string(UInt, radix~ : Int = ..) -> String
   to_uint64(UInt) -> UInt64
   trunc_double(Double) -> UInt
   upto(UInt, UInt, inclusive~ : Bool = ..) -> Iter[UInt] //deprecated
@@ -546,7 +546,7 @@ impl UInt16 {
   to_byte(UInt16) -> Byte
   to_int(UInt16) -> Int
   to_int64(UInt16) -> Int64
-  to_string(UInt16) -> String
+  to_string(UInt16, radix~ : Int = ..) -> String
 }
 
 impl UInt64 {
@@ -580,7 +580,7 @@ impl UInt64 {
   to_int(UInt64) -> Int
   to_int64(UInt64) -> Int64 //deprecated
   to_json(UInt64) -> Json
-  to_string(UInt64) -> String
+  to_string(UInt64, radix~ : Int = ..) -> String
   to_uint(UInt64) -> UInt
   trunc_double(Double) -> UInt64
   upto(UInt64, UInt64, inclusive~ : Bool = ..) -> Iter[UInt64] //deprecated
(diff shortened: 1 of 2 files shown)


builtin/to_string.mbt
```diff
@@ -126,3 +126,185 @@ pub fn Int16::to_string(self : Int16) -> String {
 pub fn UInt16::to_string(self : UInt16) -> String {
   self.to_int().to_string()
 }
+
+///|
+pub fn Int64::to_string_by_radix(self : Int64, radix~ : Int64 = 10) -> String {
```
hackwaly, line 131: Why the type of radix use `Int64` not `Int`?

builtin/to_string.mbt
```diff
@@ -126,3 +126,185 @@ pub fn Int16::to_string(self : Int16) -> String {
...
+  guard radix <= 36 && radix >= 2 else {
+    abort("radix must be between 2 and 36")
+  }
+  let alphabet = "0123456789abcdefghijklmnopqrstuvwxyz"
```
hackwaly, line 135: Could lift as a `const alphabet = ...`?

builtin/to_string.mbt
Kaida-Amethyst: Yes, seems that setting `radix` to `Int` type for all interfaces is a better idea, I fixed this problem now. ^^

builtin/to_string.mbt
Kaida-Amethyst: done~ ^^

builtin/builtin.mbti
```diff
@@ -487,6 +487,7 @@ impl Int {
   to_int64(Int) -> Int64
   to_json(Int) -> Json
   to_string(Int) -> String
+  to_string_by_radix(Int, radix~ : Int = ..) -> String
   to_uint(Int) -> UInt //deprecated
```
bobzhang, line 491: how about 
```
fn Int::to_string(self : Int, radix : Int = 10) -> String { ...}
impl Show for Int with to_string(x) { x.to_string()}
```
cc @Guest0x0

Kaida-Amethyst: Thanks for the guidance! ^^  I tried implementing it this way, but there seems to be an issue with the `\{}` string formatting syntax.

Use `int64` as example:

```moonbit
pub fn Int64::to_string(self: Int64, radix~: Int = 10) -> String {
  // ...
}

pub impl Show for Int64 with to_string(self) {
  self.to_string()
}

test "Int64 to_string" {
  let x: Int64 = 16;
  assert_eq!(x.to_string(radix=16), "10");   // OK
  assert_eq!(x.to_string(), "16");  // OK
  // assert_eq!("\{x}", "16");
  //   ^ This will throw an error during compilation.
}
```

The problem arises when trying to use the `\{}` syntax for string formatting. While the `to_string` method works fine on its own, the `\{}` syntax fails to compile. I‘m afraid that it might require some adjustments in the compiler to properly support this use case. 🥲 

---------------

Full Implementation here:

```moonbit
pub fn Int64::to_string(self: Int64, radix~: Int=10) -> String {
  guard radix <= 36 && radix >= 2 else {
    abort("radix must be between 2 and 36")
  }
  fn abs(n : Int64) -> Int64 {
    if n < 0L {
      0L - n
    } else {
      n
    }
  }

  // max length should be over 64 + 1, 80 is enough
  let buf = StringBuilder::new(size_hint=80)
  if self < 0L {
    buf.write_char('-')
  }
  let radix : Int64 = radix.to_int64()
  fn write_digits(num : Int64) {
    let num2 = num / radix
    if num2 != 0L {
      write_digits(num2)
    }
    buf.write_char(ALPHABET[abs(num % radix).to_int()])
  }

  write_digits(abs(self))
  buf.to_string()
}

pub impl Show for Int64 with to_string(self) {
  self.to_string()
}

test "Int64 to_string" {
  let x : Int64 = 16
  assert_eq!(x.to_string(radix=16), "10");
  assert_eq!(x.to_string(), "16")
  // assert_eq!("\{x}", "16")
    // ^ this will throw error during compilation. 
}
```

bobzhang: would you try it again, it should be fixed now

Kaida-Amethyst: Yes! It has been fixed. 😄

PR #1517, opened 2025-01-21
make Failure pub(readonly)

tonyfettes:
This PR makes `Failure` has visibility `pub(readonly)`.


bobzhang:
what's the motivation here?

tonyfettes:
I'm not sure on this change as well. My motivations for this change is, providing a smart constructor to error type allow us to:

1. Change the inner implementation without breaking user code,
2. We can deprecate it.

However here are my cons for this change:

1. This will break user code anyway,
2. Poor performance (but it's less important since we're through error).

PR #1520, opened 2025-01-21
doc: error

tonyfettes:


tonyfettes:
Shall we document the implementation of a trait?

PR #1521, opened 2025-01-21
remove array capacity doc test

tonyfettes:
The doc test of `Array::capacity(...)` is not consistent across different backend.

cc @Guest0x0


bobzhang:
@tonyfettes shall we deprecate this function?

tonyfettes:
@bobzhang I think we should. However, `Array::capacity` is currently internally used by other array APIs, so IMO deprecating the function is not a trivial change that should go with this PR.

PR #1548, opened 2025-01-22
chore: ignore `@coverage.end` for coverage checks

rami3l:
As suggested by @lynzrand.


lynzrand (approved):
LGTM. It's even better since `@coverage.skip` might modify itself while outputting coverage data, so it's a (although relatively harmless) modify-when-reading, or race condition as some may classify as.

PR #1549, opened 2025-01-22
ci: use `ubuntu-latest` for coverage checks

rami3l:
This should help mitigate our CI bottleneck caused by the shortage of macOS runners.


lynzrand (approved):
\o/

PR #1557, opened 2025-01-23
ci: disable `dedup_wasm` checks on macOS runners

rami3l:
Similar to #1549, this should help mitigate our CI bottleneck caused by the shortage of macOS runners.


myfreess:
lgtm

Issue #1488, opened 2025-01-16
The implementation of trigonometric functions should be aligned with the JavaScript backend


Lampese:
I think the current trigonometric functions should be replaced with implementations consistent with V8, so that we can directly inline them into `Math` on the JavaScript backend.

Lampese:
cc @tonyfettes @Kaida-Amethyst Is this plan reasonable?

Kaida-Amethyst:
I think it would be beneficial. The current Moonbit implementation seems to have precision issues, as demonstrated by the following code and results --- the hex values of sin(π) and cos(π/2):

### Moonbit (`math` from `core`):
```moonbit
fn main {
  let x = 3.141592653589793  // π
  let sin_x = @math.sin(x)
  let isin_x = BigInt::from_int64(sin_x.reinterpret_as_int64()).to_hex(uppercase=false)
  println("sin(\{x}) = \{sin_x}(0x\{isin_x})")
  let x = 1.5707963267948966c   // π/2
  let cos_x = @math.cos(x)
  let icos_x = BigInt::from_int64(cos_x.reinterpret_as_int64()).to_hex(uppercase=false)
  println("cos(\{x}) = \{cos_x}(0x\{icos_x})")
}
```
**Output:**  
```
sin(3.141592653589793) = 1.2246703776017889e-16(0x3ca1a63c78000000)
cos(1.5707963267948966) = 6.123351888008944e-17(0x3c91a63c78000000)
```

### JavaScript ( run by node):
```javascript
function doubleToHex(double) {
    const buffer = new ArrayBuffer(8);
    const view = new DataView(buffer);
    view.setFloat64(0, double);

    let hex = '';
    for (let i = 0; i < 8; i++) {
        const byte = view.getUint8(i);
        hex += byte.toString(16).padStart(2, '0');
    }

    return hex;
}

let x1 = 3.141592653589793;   // π
let sin_x = Math.sin(x1);
let isin_x = doubleToHex(sin_x);
console.log('sin(' + x1 + ') = ' + sin_x + '(0x' + isin_x+')');

let x2 = 1.5707963267948966     // π/2
let cos_x = Math.cos(x2);
let icos_x = doubleToHex(cos_x);
console.log('cos(' + x2 + ') = ' + cos_x + '(0x' + icos_x+')');
```
**Output:**  
```
sin(3.141592653589793) = 1.2246467991473532e-16(0x3ca1a62633145c07)
[...]

PR #1559, opened 2025-01-23
better trigonometric functions' implementation

Kaida-Amethyst:
This PR aims to address the issue raised in #1488, where inconsistencies in trigonometric functions were identified. The current implementation shows deviations from JavaScript V8 and glibc, which could lead to precision-related issues in the future.

To resolve this, I have implemented custom versions of `sin`, `cos`, and `tan` based on the well-established [fdlibm](https://www.netlib.org/fdlibm/) library. These implementations have undergone extensive testing in the [Moonbit-Math](https://github.com/Kaida-Amethyst/Moonbit-Math) repository, demonstrating high consistency with JavaScript V8's results.

References:
- [fdlibm/sin](https://www.netlib.org/fdlibm/s_sin.c)
- [fdlibm/cos](https://www.netlib.org/fdlibm/s_cos.c)
- [fdlibm/tan](https://www.netlib.org/fdlibm/s_tan.c)

update in 01/31/25, add `asin`, `acos`, `atan`, `atan2`.
update in 02/01/25, add `sin`, `cos`, `tan` for floating-point


Lampese:
- [] sin
- [] cos
- [] tan
- [] atan
- [] asin
- [] acos
- [] atan2
- [] move algorithm implements to double/float
- [] add js api
- [] add test result to ensure consistency

bobzhang (approved):

double/trig_nonjs.mbt
```diff
@@ -0,0 +1,770 @@
+// Copyright 2025 International Digital Economy Academy
```
inkydragon, line 1: Note that the copyright for `fdlibm` explicitly requires that the copyright notice should be preserved.

```
/*
 * ====================================================
 * Copyright (C) 1993 by Sun Microsystems, Inc. All rights reserved.
 *
 * Developed at SunSoft, a Sun Microsystems, Inc. business.
 * Permission to use, copy, modify, and distribute this
 * software is freely granted, provided that this notice 
 * is preserved.
 * ====================================================
 */
```

It would be better that the copyright header be reserved for translated functions,
just like what rust do:
https://github.com/rust-lang/libm/blob/a89add35ebda33250cfeec0fdd942ea43a758120/src/math/acos.rs#L4-L9

Lampese: Thank you very much! This is a very, very helpful point. We'll go back to work when we get back from the Chinese New Year break.

Kaida-Amethyst: The copyright of fdlibm has been added ^^,

Lampese (approved):

PR #1566, opened 2025-01-24
test: improve coverage for `double/double.mbt`

rami3l:
**Disclaimer:** This PR was generated by an LLM agent as part of an experiment.

## Summary

```
coverage of `double/double.mbt`: 76.2% -> 95.2%
```


rami3l:
This seems like a `moon` bug instead of an actual failure in the added tests @Young-Flash

Young-Flash:
bleeding & nightly passed, stable would pass at next release

Issue #994, opened 2024-09-13
zip for arrays, iters, etc


NKID00:
Is seems zip is only implemented for `@immut/list.T` but not `Array` or `Iter`. There seems to be no way to iterate over multiple iterators, and zipping two iterators fits well.

I am willing to contribute this, please assign me if I may give it a try.

Guest0x0:
MoonBit's `Iter` and `Iter2` are *internal* iterators, where the user of iterators pass a callback to the provider of data. This is in contrast to *external* iterators, where the iterator produce one element every time the user requests, and save internal state about the current progress of iteration.

In internal iterators, the provider of data (e.g. `Array::iter`) owns the control flow, while for external iterators, the user of the iterator (e.g. `for .. in`) owns the control flow. This difference is subtle, because only the one who owns the control flow can stop the iterator in the middle. As a consequence, ***it is impossible to implement `zip` for internal iterators unless you have control effect constructs such as `call/cc`***. So sadly, the absence of `Iter::zip` is a theoretic limitation of the `Iter` type.

At this point, you may wonder why MoonBit chooses the less flexible internal style to implement iterators. This is because internal iterators are simpler to write and optimize, and is adequate for the most common scenario.

Of course there are many cases where it is handly or even necessary to zip two iterators, and a iterator type that support zipping (i.e. external iterator) would be a good addition to MoonBit's current `Iter` and `Iter2` types. If you are interested in this task, you may consider writing a package for external iterators on https://mooncakes.io , which is also a highly encouraged and appreciated way to contribute to the MoonBit community.

Guest0x0:
There is no problem with `Array::iter`` though, it would be great if you could implement it

hackwaly:
How about

```
fn Iter::zipWith[T1, T2](self: Iter[T1], seq: Unit -> T2) -> Iter[T1, T2]
```

PR #1575, opened 2025-01-26
add @array.zip and @array.zip_with

myfreess:


Lampese:
I have an API design concern, with Array zip_with, what about FixedArray? What about other linearly traversable structures? This leads to the discussion that Iter::zip, but our Iter can't do zip...

cc @bobzhang

FlammeShadow:
It seems `zip_with` can be easily composed with `zip` and `map`?

illusory0x0:
related issue.

- #994

Yoorkin:
There is a zip function in `immut/list`:

```
/// Zip two lists.
/// If the lists have different lengths, it will return None.
pub fn zip[A, B](self : T[A], other : T[B]) -> T[(A, B)]?
```

It's a method that returns an option. To make the API more consistent, consider changing the API like this:

```
// in immut/list
pub fn zip[A, B](self : T[A], other : T[B]) -> T[(A, B)]

// in array
pub fn zip[A, B](self : T[A], other : T[B]) -> T[(A, B)]
pub fn zip_with[A, B, C](self : T[A], other : T[B], merge~ : (A,B)->C) -> T[C]
```

Lampese:
> I have an API design concern, with Array zip, what about FixedArray? What about other linearly traversable structures? This leads to the discussion that Iter::zip, but our Iter can't do zip...
> 
> cc @bobzhang

So that the zip for FixedArray is a todo issue?

array/array.mbt
```diff
@@ -185,6 +185,73 @@ pub fn last[A](self : Array[A]) -> A? {
...
+///   let arr1 = [1, 2, 3]
+///   let arr2 = [4, 5, 6]
+///   let add = fn(a, b) { a + b }
+///   inspect!(zip_with(arr1, arr2, merge=add), content="[5, 7, 9]")
+/// }
+/// ```
+pub fn zip_with[A, B, C](
+  l : Array[A],
+  r : Array[B],
+  merge~ : (A, B) -> C
```
peter-jerry-ye, line 244: without labelled argument

peter-jerry-ye: Provide `zip_to_iter` for creating `Iter2`

peter-jerry-ye (commented):
drop the changes in `immut/list` as they are irrelevant and we are going to deprecate them.

PR #1576, opened 2025-01-26
change `{String|Bytes|Array}View` to `@{string|bytes|array}.View`

Yu-zh:
This PR:
1) deprecate `StringView`, `ArrayView` and `BytesView` and add `@string.View`, `@array.View`, `@bytes.View` as aliases for them.
2) Move all the definitions from `builtin/bytesview.mbt` to `bytes/view.mbt`. Move some of the definition from `builtin/arrayview.mbt` to `array/view.mbt`.


array/array.mbti
```diff
@@ -8,6 +9,7 @@ alias @moonbitlang/core/quickcheck as @quickcheck
 impl FixedArray {
   all[T](Self[T], (T) -> Bool) -> Bool
   any[T](Self[T], (T) -> Bool) -> Bool
+  blit_from_bytesview(Self[Byte], Int, @bytes.BytesView) -> Unit
```
Yu-zh, line 12: This method was in `builtin/bytesview.mbt`. It doesn't have to be in builtin so I moved it here when I deleted `builtin/bytesview.mbt`

array/moon.pkg.json
```diff
@@ -1,12 +1,12 @@
 {
   "import": [
     "moonbitlang/core/builtin",
-
+    "moonbitlang/core/bytes",
```
Yu-zh, line 4: This extra dep is used because `blit_from_bytesview` is moved from `@builtin` to `@array`.

PR #1577, opened 2025-01-26
add `ffi_util` package for FFI related util & raw function type

Guest0x0:
When binding FFI in wasm/native backend, it is often necessary to pass a callback function to FFI, and sometimes the FFI function must be capture free, for example Unix signal handler.

This PR adds a new package `@ffi_util` and a new type `@ffi_util.RawFunction[T]`. `@ffi_util.RawFunction[T]` represents capture-free toplevel function of type `T`. Note that the definition in core is merely a placeholder, the raw function type need special treatment from the compiler. So it is not usable until the compiler part is released.


bobzhang:
per offline discussion, let's make it builtin type

PR #1580, opened 2025-01-27
Revert "change `{String|Bytes|Array}View` to `@{string|bytes|array}.View`"

bobzhang:
Reverts moonbitlang/core#1576


bobzhang:
Note the compiler hardcoded BytesView in Basic_type_path, so we can not move BytesView from builtin to bytes without changes from upstream see #1581 
cc @Yu-zh

PR #1582, opened 2025-01-27
refactor json parser

Yu-zh:
The motivation for the refactoring:
1) avoid using `substring` and `write_substring` because the behavior of string APIs will be changed to unicode safe
2) use `StringView` to avoid manual utf16 surrogate pair handling
3) use peek and bump pattern to avoid backoff


string/view.mbt
```diff
@@ -447,3 +447,9 @@ pub fn StringView::iter(self : StringView) -> Iter[Char] {
     }
   })
 }
+
+///|
+/// Returns true if the view is empty.
+pub fn StringView::is_empty(self : StringView) -> Bool {
```
hackwaly, line 453: This seems unused in this PR? I think it's better to put it in separated PR.

Yu-zh: The usage is removed in this commit https://github.com/moonbitlang/core/pull/1582/commits/d099cae1f6fa136db058b433106934bc504fc7a3. I will clean it up.

json/lex_misc.mbt
```diff
@@ -13,37 +13,53 @@
...
+
+///|
+fn bump(self : ParseContext, builder? : Builder) -> Unit {
+  match self.view {
+    [] =>
+      // this should be a misuse of the peek & bump api
+      abort("empty view")
+    [c, .. rest] => {
+      if c == '\n' {
+        self.line += 1
```
hackwaly, line 31: The new version added overhead at every bump(s). Previous version is highly optimized and considered reasonable. In most cases, the inputs of `json.parse` are well-formed json. So we don't need report error with locations.  

Do we have some benchmark? I think the refactoring of one of motivations "avoid substring/write_substring" should not mixed with refactoring of location tracking. It's better to make these change into separated commits. So we can measure them separately.

hackwaly:
Do we have some benchmark for this refactoring? I'm guessing it will get lots of slowdown.

bobzhang:
> Do we have some benchmark for this refactoring? I'm guessing it will get lots of slowdown.

Do you have some more detailed input? Note we will optimize some common patterns

hackwaly:
> Do you have some more detailed input? 

No, I'm just guessing. The `peek` and `bump` is very very hot. The using of `StringView` first-classy added indirections and increases `copy` (see `ctx.view = rest`) cost unavoidable.

Yu-zh:
Initial benchmarking suggests that this PR approximately doubles the time required for JSON parsing on a 25Mb json file:
image

A profile looks like:
image

hackwaly:
lgtm for the code change.

bobzhang:
1. provide @string.RandomIndex and utils
2. provide StringView directly from RandomIndex
3. substr(s, RandomIndex, RandomIndex)

Yu-zh:
close as we are seeking a more efficient solution

PR #1584, opened 2025-01-27
fix wrong atan2 result on -0.0

codeworm96:
It is reported in https://taolun.moonbitlang.com/t/topic/1128, that the result of `atan2(-0, -1)` is wrong.

According to IEEE754, atan2(±0, x) is ±π for x < 0.

The root cause of the issue is that the old copysign just checks if y < 0 and does not distinguish ±0.

This PR fixes this by checking the sign bit instead.

### DIFF
--- a/math/trigonometric.mbt
+++ b/math/trigonometric.mbt
@@ -279,8 +279,12 @@ pub fn acos(x : Double) -> Double {
 }
 
 ///|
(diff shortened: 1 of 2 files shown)


Lampese (approved):

Lampese:
LGTM, and we are pushing new algorithm of those.

PR #1586, opened 2025-01-27
refactor parse_int64 and parse_uint64 in strconv

Yu-zh:
The main motivation is to avoid using `substring` which may unnecessarily create several copies.


strconv/uint_test.mbt
```diff
@@ -0,0 +1,92 @@
...
+  )
+
+  // Invalid hex digit (G is not a valid base-16 digit)
+  inspect!(@strconv.parse_uint64?("0xDEADBEG"), content="Err(invalid syntax)")
+}
+
+test "edge cases" {
+  // Invalid: Missing digits after hex prefix
+  // inspect!(@strconv.parse_uint64?("0x"), content="Err(invalid syntax)")
+
```
bobzhang, line 88: @Yu-zh should this error? currently it does not error

strconv/uint_test.mbt
```diff
@@ -0,0 +1,92 @@
...
+  // Invalid underscore placements in hexadecimal strings
+  // inspect!(@strconv.parse_uint64?("0x_DEADBEEF"), content="Err(invalid syntax)")
+
```
bobzhang, line 91: what's the desired behavior?

strconv/uint_test.mbt
Yu-zh: This is a bug. I think we could just follow the behavior of the strconv pkg in Go (see https://go.dev/play/p/jWnuL-QCwF-).

strconv/uint_test.mbt
Yu-zh: I'm not aware of a standard way for parsing the hex literals. So I think we could just follow other languages. This is valid in both Python and Go. See https://go.dev/play/p/jWnuL-QCwF- and 
image

I will add more documentation though.

strconv/uint_test.mbt
Yu-zh: This is invalid in both Python and Go. I will fix it.

strconv/uint_test.mbt
bobzhang: note in both python and go:
```
>>> int("0x_DAE_",0)
Traceback (most recent call last):
  File "<python-input-3>", line 1, in <module>
    int("0x_DAE_",0)
    ~~~^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 0: '0x_DAE_'
```

Yu-zh: Yes, trailing underscore is handled. See https://github.com/moonbitlang/core/pull/1586/files#diff-f33bf2098e6c91bdc93d15281b2f928dfe14c05b8b5158fc4caf1ab98404e909R249

PR #1600, opened 2025-02-07
docs: add docs for char to intrinsics

tonyfettes:
This PR add documentation to:

- `char.mbt`
- `console.mbt`
- `failure.mbt`
- `fixedarray.mbt`
- `fixedarray_block.mbt`
- `hasher.mbt`
- `int64_nonjs.mbt`
- `intrinsics.mbt`

Note: adding this PR cause context to exceed 128K when generating docs for the builtin package.

### DIFF
--- a/builtin/char.mbt
+++ b/builtin/char.mbt
@@ -13,6 +13,25 @@
 // limitations under the License.
 
 ///|
+/// Calculates the difference between the Unicode code points of two characters.
+///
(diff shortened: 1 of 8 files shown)


bobzhang:
> Note: adding this PR cause context to exceed 128K when generating docs for the builtin package.

We should be more creative in generating context

Issue #1599, opened 2025-02-06
Redundant loc in the assert_eq function


illusory0x0:
`assert_eq` call `fail` leads to redundant loc information.

```moonbit
pub fn fail[T](msg : String, loc~ : SourceLoc = _) -> T!Failure {
  raise Failure("FAILED: \{loc} \{msg}")
}
pub fn assert_eq[T : Eq + Show](a : T, b : T, loc~ : SourceLoc = _) -> Unit! {
  if a != b {
    fail!("FAILED: \{loc} `\{a} == \{b}`")
  }
}
pub fn my_assert_eq[T : Eq + Show](a : T, b : T, loc~ : SourceLoc = _) -> Unit! {
  if a != b {
    raise Failure("FAILED: \{loc} `\{a} == \{b}`")
  }
}
test "assert_eq" {
  assert_eq!(0,1)
}

test "my_assert_eq" {
  my_assert_eq!(0,1)
}
```

```
PS C:\illu\repo\dev-mbt\try> moon test
test username/hello/lib/max.mbt::assert_eq failed: FAILED: C:\Users\11378\.moon\lib\core\builtin\assert.mbt:51:5-51:43 FAILED: C:\illu\repo\dev-mbt\try\src\lib\max.mbt:9:3-9:18 `0 == 1`
test username/hello/lib/max.mbt::my_assert_eq failed: FAILED: C:\illu\repo\dev-mbt\try\src\lib\max.mbt:13:3-13:21 `0 == 1`
Total tests: 2, passed: 0, failed: 2.
```

peter-jerry-ye:
They should be:
```moonbit
pub fn assert_eq[T : Eq + Show](a : T, b : T, loc~ : SourceLoc = _) -> Unit! {
  if a != b {
    fail!("`\{a} == \{b}`", loc~)
  }
}
```

PR #1604, opened 2025-02-07
fix: remove redundant loc in assert_xx function

illusory0x0:
- #1599


peter-jerry-ye (approved):
I think you need to rerun `moon fmt`

PR #1609, opened 2025-02-07
add to_string method for StringView

Yu-zh:
This method will override the default trait implementation for `to_string` to avoid adding quotation marks to the result string.


bobzhang:
[386966d](https://github.com/moonbitlang/core/pull/1609/commits/386966d5746813bc7fd500cfb274b7bd396f8e1f)
cc @Yu-zh override Show::to_string instead

Issue #1608, opened 2025-02-07
undefined behavior: left shift of negative value in native backend


illusory0x0:
- #1607 

e.g.
```c
49964 int32_t $moonbitlang$core$builtin$rotl(int32_t x$1904, int32_t r$1905)       {
///============================================
49965   int32_t tmp$2578 = x$1904 << (r$1905 & 31);
///============================================
49966   uint32_t tmp$2581 = *(uint32_t*)&x$1904;
49967   int32_t tmp$2582 = 32 - r$1905;
49968   uint32_t tmp$2580 = tmp$2581 >> (tmp$2582 & 31);
49969   int32_t tmp$2579 = *(int32_t*)&tmp$2580;
49970   return tmp$2578 | tmp$2579;
49971 }
```

```
/home/w/dev-mbt/core/target/native/debug/test/builtin/builtin.whitebox_test.c:49978:11: runtime error: signed integer overflow: 97 * -1028477379 cannot be represented in type 'int'
/home/w/dev-mbt/core/target/native/debug/test/builtin/builtin.whitebox_test.c:49965:29: runtime error: left shift of negative value -603296558
/home/w/dev-mbt/core/target/native/debug/test/builtin/builtin.whitebox_test.c:49981:11: runtime error: signed integer overflow: -643450860 * 668265263 cannot be represented in type 'int'
/home/w/dev-mbt/core/target/native/debug/test/builtin/builtin.whitebox_test.c:49979:11: runtime error: signed integer overflow: -1846073424 + -1282304214 cannot be represented in type 'int'
skipped test block: uint_test.mbt: panic @strconv.parse_uint64/errors
skipped test block: int_test.mbt: panic @strconv.parse_int64/invalid_input
skipped test block: int_test.mbt: panic @moonbitlang/core/strconv.parse_int64/invalid_base
skipped test block: bytes_test.mbt: panic Fixed::blit_from_bytesview 1
skipped test block: bytes_test.mbt: panic Fixed::blit_from_bytesview 2
/home/w/dev-mbt/core/target/native/debug/test/builtin/builtin.internal_test.c:14145:11: runtime error: signed integer overflow: 105 * -1028477379 cannot be represented in type 'int'
/home/w/dev-mbt/core/target/native/debug/test/builtin/builtin.internal_test.c:14132:27: runtime error: left shift of negative value -241180998
skipped test block: deque_test.mbt: panic deque truncate
[...]

Issue #1605, opened 2025-02-07
chore: add sanitizer in native backend


illusory0x0:
- #1594

moon.pkg.json

```json
{
  "link": {
    "native": {
      "cc-flags": "-fsanitize=address -fsanitize=undefined"
    }
  }
}
```

illusory0x0:
Linking to libasan.so is very slow. link-core over 1min in builtin package in my machine.
Because builtin.blackbox_test.c is too large.

```
w@ew:~/dev-mbt/core/target/native/debug/test/builtin$ du -h builtin.blackbox_test.c
15M     builtin.blackbox_test.c
```

peter-jerry-ye:
Since core has little FFI (mainly in env), and adding sanitizer would slow down the CI, we are not adding the sanitizer here: it should be the compiler's responsiblity.

Issue #1594, opened 2025-02-06
signed integer overflow in native backend


illusory0x0:
Use -fsanitize=undefined in native backend, error detected

https://github.com/illusory0x0/moonbit-bug-report/tree/master/integer-overflow

```
w@ew:~/moonbit-bug-report/integer-overflow$ ./test.sh
/home/w/moonbit-bug-report/integer-overflow/target/native/debug/test/lib/lib.blackbox_test.c:4842:11: runtime error: signed integer overflow: 104 * -1028477379 cannot be represented in type 'int'
/home/w/moonbit-bug-report/integer-overflow/target/native/debug/test/lib/lib.blackbox_test.c:4829:27: runtime error: left shift of 787296381 by 17 places cannot be represented in type 'int'
/home/w/moonbit-bug-report/integer-overflow/target/native/debug/test/lib/lib.blackbox_test.c:4845:11: runtime error: signed integer overflow: 1627020762 * 668265263 cannot be represented in type 'int'
/home/w/moonbit-bug-report/integer-overflow/target/native/debug/test/lib/lib.blackbox_test.c:4843:11: runtime error: signed integer overflow: -1868146422 + -797000175 cannot be represented in type 'int'
Total tests: 1, passed: 1, failed: 0.
```

illusory0x0:
You're right.

https://en.cppreference.com/w/c/language/operator_arithmetic

> Unsigned integer arithmetic is always performed modulo 2n
 where n is the number of bits in that particular integer. E.g. for unsigned int, adding one to [UINT_MAX](https://en.cppreference.com/w/c/types/limits) gives ​0​, and subtracting one from ​0​ gives [UINT_MAX](https://en.cppreference.com/w/c/types/limits).

> When **signed integer** arithmetic operation overflows (the result does not fit in the result type), the behavior is **undefined**:

Guest0x0:
C generated by MoonBit should be compiled with `-fwrapv`, which will eliminate the undefined behavior here

PR #1610, opened 2025-02-07
Fix: signed integer overflow

illusory0x0:
The main difficulty in fixing this bug is locating it.
If we need to separate pull requests, I'll do that, but the process of fixing bugs is unlikely to be separated.

- #1605 
- #1594  
- #1608

### DIFF
--- a/builtin/byte.mbt
+++ b/builtin/byte.mbt
@@ -296,7 +296,7 @@ pub fn Byte::to_uint(self : Byte) -> UInt {
 ///
 /// Returns the resulting `Byte` value after the shift operation.
 pub fn Byte::op_shl(self : Byte, count : Int) -> Byte {
-  (self.to_int() << count).to_byte()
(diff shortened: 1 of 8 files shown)


illusory0x0:
this is another issus.

> D:\a\core\core\target\native\debug\test\builtin\builtin.blackbox_test.c(198789): error C2124: divide or mod by zero

Guest0x0:
the `-fwrapv` flag should be passed to C compiler when compiling C generated by MoonBit. So signed integer overflow is not UB in MoonBit. But the hasher should indeed use `UInt` instead of `Int`, so that part may still be an improvement (in terms of code readability)

illusory0x0:
FTR: [gcc option -fwrapv](https://gcc.gnu.org/onlinedocs/gcc/Code-Gen-Options.html#index-fwrapv)

PR #1612, opened 2025-02-08
remove unused package

bobzhang:
cc @Young-Flash any idea why this was not caught by the CI?


Yu-zh:
The package is used by the native backend but not other backends. `moon` should collect the warnings from `moonc` and when the unused package warning applies to all the backends, the warning should be reported to the user.

PR #1619, opened 2025-02-09
feat: Add join method for Array[String]

bobzhang:
cc @gmlewis , is this what you expected?


gmlewis (approved):
> cc @gmlewis , is this what you expected?

Yes, thank you, @bobzhang!
LGTM

PR #1627, opened 2025-02-10
fix intrinsic name

Yu-zh:
The automatically generated doc messed up the intrinsics. This PR fixed this problem.


hackwaly:
This hint us there's no tests to cover the intrinsics.

Yu-zh:
> This hints us there's no tests to cover the intrinsics.

These functions are primitive and they should not be annotated with intrinsics. I just realized this and removed the wrong intrinsics.

tonyfettes (approved):
LGTM

PR #1656, opened 2025-02-11
implement Eq and Compare for BytesView

Yu-zh:
The PR implements #1651


bobzhang:
@Yu-zh we are encouraging users to implement traits explicitly instead of using method implicit behaivor

PR #1665, opened 2025-02-12
use impl for Eq and Compare for Bytes

Yu-zh:
Following the suggested approach https://github.com/moonbitlang/core/pull/1656#issuecomment-2651059178.

There are still more methods in the builtin package and other packages that need to be changed.


Yu-zh:
The op_equal operator and the like are specialized in the compiler so this refactor cannot be done without changing the compiler simultaneously.

PR #1676, opened 2025-02-14
update Show::to_string  for char

bobzhang:
- **move to_string from method to trait Show for Char**
- **update test**
- **the last test inspect does not get right**

### DIFF
--- a/builtin/show_test.mbt
+++ b/builtin/show_test.mbt
@@ -197,38 +197,24 @@ test "Show for ArrayView" {
 
 ///|
 test "Show for Char" {
-  inspect!(Show::to_string('a'), content="'a'")
-  inspect!(Show::to_string('字'), content="'字'")
+  inspect!(Show::to_string('a'), content="a")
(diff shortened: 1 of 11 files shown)


builtin/show_test.mbt
```diff
@@ -237,8 +237,11 @@ test "Show for Char" {
 
 ///|
 test "char show hex digits" {
-  inspect!('\x01', content="'\\x01'")
-  inspect!('\x0b', content="'\\x0b'")
+  // inspect!('\x01', content="\x01")
+  // FIXME: inspect!('\x0b') will not work with update test
```
bobzhang, line 241: cc @lijunchen

lijunchen: two issues:
1. `"\x0b"` is not a valid JSON, causing `moon test -u` to fail parsing the inspect message.

2. `moon test—u` gets stuck in an unnecessary loop when it fails to parse the message.

PR #1679, opened 2025-02-17
Add string processing test cases with pattern matching

bobzhang:


string/string_test.mbt
```diff
@@ -579,3 +579,25 @@ test "length_eq" {
   assert_true!(str.length_eq(8))
   assert_false!(str.length_eq(9))
 }
+
+///|
+test {
```
hackwaly, line 584: add label `test "string pattern matching"`

PR #1713, opened 2025-02-26
refactor(string): update replace method to support count parameter

BigOrangeQWQ:


bobzhang:
do you have real world use case for such API? note we already had replace_all

BigOrangeQWQ:
In some string algorithm problems, I needed a replace method with a count parameter.
However, after searching through GitHub repositories for programming languages that implement a replace method with this parameter, I noticed it's rarely used. I apologize for any inconvenience caused and will close the PR.

PR #1714, opened 2025-02-27
refactor: Simplify FromJson implementations using guard syntax

bobzhang:


gmlewis:
Very nice examples for cleaning up my own `Json` handling... thank you, @bobzhang!

PR #1721, opened 2025-02-27
Refactor list implementation and mark init_ as deprecated

bobzhang:
- Move `init_` function to deprecated module
- Optimize some list functions by using `loop` instead of `match`
- Reimplement `repeat` function with a more imperative approach
- Add deprecation warning for `init_` function

### DIFF
--- a/immut/list/list.mbt
+++ b/immut/list/list.mbt
@@ -220,9 +220,9 @@ pub fn all[A](self : T[A], f : (A) -> Bool) -> Bool {
 ///|
 /// Test if any element of the list satisfies the predicate.
 pub fn any[A](self : T[A], f : (A) -> Bool) -> Bool {
-  match self {
+  loop self {
     Nil => false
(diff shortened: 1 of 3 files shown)


immut/list/list.mbt
```diff
@@ -594,11 +578,11 @@ pub fn nth[A](self : T[A], n : Int) -> A? {
 /// assert_eq!(@list.repeat(5, 1), @list.from_array([1, 1, 1, 1, 1]))
 /// ```
 pub fn repeat[A](n : Int, x : A) -> T[A] {
-  if n == 0 {
-    Nil
-  } else {
-    Cons(x, repeat(n - 1, x))
```
illusory0x0, line 600: old implementation is non tail recursion, when n is large wound encounter stackoverflow

bobzhang:
closed as immut/list is discouraged in favor of list

PR #1727, opened 2025-03-03
improve Array::flatten

illusory0x0:


builtin/array.mbt
```diff
@@ -1253,13 +1253,20 @@ pub fn Array::resize[T](self : Array[T], new_len : Int, f : T) -> Unit {
...
-    v.append(arr)
+  let mut len = 0
+  for x in self {
+    len += x.length()
   }
-  v
+  let res = Array::make_uninit(len)
+  let mut i = 0
+  for xs in self {
+    for x in xs {
```
Yu-zh, line 1264: I think there's a blit function you can use to avoid iterating through each element.

PR #1730, opened 2025-03-03
improve stringbuilder::output

illusory0x0:


Yu-zh:
The improvement does not look obvious to me. Can you explain it?

builtin/stringbuilder_buffer.mbt
```diff
@@ -87,11 +87,13 @@ pub fn StringBuilder::to_string(self : StringBuilder) -> String {
 }
 
 ///|
-/// TODO: improve perf
 pub impl Show for StringBuilder with output(self, logger) {
-  logger.write_string(
-    self.data.unsafe_to_bytes().to_unchecked_string(offset=0, length=self.len),
```
illusory0x0, line 93: `to_unchecked_string` just simply copy bytes into string

builtin/stringbuilder_buffer.mbt
```diff
@@ -87,11 +87,13 @@ pub fn StringBuilder::to_string(self : StringBuilder) -> String {
...
-  )
+  for i = 0; i < self.len; i = i + 2 {
+    let low = self.data[i].to_int()
+    let high = self.data[i + 1].to_int()
+    let code_unit = (high << 8) | low |> Char::from_int
```
illusory0x0, line 94: this `code_unit` is the same as `String::op_get` 's result

Yu-zh: I'm actually asking why this is more efficient than the original implementation?

builtin/stringbuilder_buffer.mbt
Yu-zh: So basically you want avoid the extra copy here. My suggestion is that we should change the semantics of `to_uncheck_string` since Bytes has become immutable. Then we can still enjoy the batch processing in `write_string`.

illusory0x0: When using `to_uncheck_string` with backends that require **extra copies**, such as the js backend and wasm-gc backend with the `js-string-builtin` option turned on, use this implementation

PR #1746, opened 2025-03-06
fix the order of elements in `from_iter`

Yu-zh:
The original implementation reverses the order. This PR fixes the order, and provides an efficient `from_iter_rev` for the use case where order is irrelevant.


bnse (approved):
r

PR #1748, opened 2025-03-07
add StringView::unsafe_make

hackwaly:
There missing a O(1) way to create `StringView` from `String` and UTF-16 positions. This PR added a `StringView::unsafe_make` for that.


peter-jerry-ye:
There exists `charcodes` now, so the corresponding one will be named as `unsafe_charcodes`.

Issue #1583, opened 2025-01-27
[Performance] @immut/array::from_iter


peter-jerry-ye:
The current implementation simply folds upon iter and uses `push`, which causes enumerous intermediate object allocations. This should be able to improved somehow by preallocating fixedarrays.

peter-jerry-ye:
If we change the `from_array` of `execute_array_test` to `from_iter`, and compare the time for `moon test -p immut/array` before and after #1752

Before: 2.25s
After: 0.92s

Kudos to @illusory0x0 !

PR #1752, opened 2025-03-09
improve @immut/array::from_iter

illusory0x0:
> That is exactly what Clojure’s persistent vector does, and it is done through balanced, ordered trees.

persistent vector is implemented using balanced tree and share Path during update/insert/remove, using Array as buffer avoids creating a lot of @immut/array[T].

- #1583

### DIFF
--- a/immut/array/array.mbt
+++ b/immut/array/array.mbt
@@ -126,46 +126,9 @@ pub fn from_iter[A](iter : Iter[A]) -> T[A] {
   if not(buf.is_empty()) {
     leaves.push(FixedArray::from_array(buf))
   }
-  fn dfs(leaves : ArrayView[FixedArray[A]], cap) -> Tree[A] {
-    if cap == branching_factor {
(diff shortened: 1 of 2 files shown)


peter-jerry-ye (commented):
This is not ideal when handling large Array as this would also introduce resizing.

It would be better to allocate an FixedArray of size `branching_factor`, and then insert it to the tree.

illusory0x0:
> This is not ideal when handling large Array as this would also introduce resizing.
> 
> It would be better to allocate an FixedArray of size `branching_factor`, and then insert it to the tree.

It need to improve `make` and `makei` in `@immut/array` package

peter-jerry-ye:
I updated this PR with:

- migrate `dfs` and other code snippet to the functions introduced in #1765
- rewrite using `FixedArray` instead of `Array`, so that there will be no copy for each leaf, and there will be no re-grow after the buffer gets cleared.

PR #1753, opened 2025-03-09
improve Array::makei

illusory0x0:
Uses the same implementation as Array::make, which is panic when len < 0.

### DIFF
--- a/array/array.mbt
+++ b/array/array.mbt
@@ -55,41 +55,6 @@ pub fn push_iter[T](self : Array[T], iter : Iter[T]) -> Unit {
 }
 
 ///|
-/// Creates a new array of the specified length, where each element is
-/// initialized using an index-based initialization function.
-///
-/// Parameters:
-///
-/// * `length` : The length of the new array. If `length` is less than or equal
-/// to 0, returns an empty array.
-/// * `initializer` : A function that takes an index (starting from 0) and
-/// returns a value of type `T`. This function is called for each index to
-/// initialize the corresponding element.
-///
-/// Returns a new array of type `Array[T]` with the specified length, where each
-/// element is initialized using the provided function.
-///
-/// Example:
-///
-/// ```moonbit
-/// test "Array::makei" {
-///   let arr = Array::makei(3, fn(i) { i * 2 })
-///   inspect!(arr, content="[0, 2, 4]")
-/// }
-/// ```
-pub fn Array::makei[T](length : Int, value : (Int) -> T) -> Array[T] {
-  if length <= 0 {
-    []
-  } else {
-    let array = Array::make(length, value(0))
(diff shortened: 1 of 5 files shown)


peter-jerry-ye (changes requested):
Please do not move the APIs.

If you want to use the intrinsics, just copy them to the `array` package.

Yu-zh:
Maybe you can just make `unsafe_set` public. Since the `unsafe_get` is already public, I don't see a reason why `unsafe_set` cannot be.

array/array.mbt
```diff
@@ -54,41 +54,6 @@ pub fn push_iter[T](self : Array[T], iter : Iter[T]) -> Unit {
...
-/// test "Array::makei" {
-///   let arr = Array::makei(3, fn(i) { i * 2 })
-///   inspect!(arr, content="[0, 2, 4]")
-/// }
-/// ```
-pub fn Array::makei[T](length : Int, value : (Int) -> T) -> Array[T] {
-  if length <= 0 {
-    []
-  } else {
-    let array = Array::make(length, value(0))
```
illusory0x0, line 84: I want to avoid redundent fill values

illusory0x0:
> Maybe you can just make `unsafe_set` public. Since the `unsafe_get` is already public, I don't see a reason why `unsafe_set` cannot be.

I also want to use `Array::make_uninit` to avoid redundant filling operations

illusory0x0:
> If you want to use the intrinsics, just copy them to the `array` package.

So how to copy them into the `array` package, If I want to use `Array::unsafe_set` which depend on **private** definitions.

peter-jerry-ye:
Copy them as well.

array/array.mbt
illusory0x0: `Array::make`  has two fill operations.

`Array::makei` call `Array::make` has three fill operations, 

```moonbit
pub fn Array::make[T](len : Int, elem : T) -> Array[T] {
  let arr = Array::make_uninit(len) // (1)
  for i in 0..<len {
    arr.unsafe_set(i, elem)  // (2)
  }
  arr
}

pub fn Array::makei[T](length : Int, value : (Int) -> T) -> Array[T] {
  if length <= 0 {
    []
  } else {
    let array = Array::make(length, value(0)) // (2)
    for i in 1..<length {
      array[i] = value(i) // (3)
    }
    array
  }
}
```

peter-jerry-ye: @Yu-zh Do we have intrinsics for these cases?

peter-jerry-ye:
Array related implementation will be moved to `@array`, which is WIP by @Guest0x0

bobzhang:
closed as the pr is outdated and too many conflicts

PR #1756, opened 2025-03-09
inconsistent hash result when derive Int

illusory0x0:
Int doesn't use default impl

### DIFF
--- a/builtin/hasher_test.mbt
+++ b/builtin/hasher_test.mbt
@@ -255,3 +255,18 @@ test "hash_combine for UInt" {
   Hash::hash_combine(value, hasher)
   inspect!(hasher.finalize(), content="1161967057")
 }
+
+///|
+priv type MyHashInt Int derive(Hash, Show)
+
+///|
+test "MyInt::hash" {
+  let my_int : MyHashInt = 42
+  let int : Int = 42
+  let uint : UInt = 42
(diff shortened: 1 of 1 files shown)


peter-jerry-ye:
How do you define "consistent"?

illusory0x0:
> How do you define "consistent"?

I think Int should also use the default implementation, but now Int uses a different set of implementations

```moonbit
pub impl Hash for Int with hash(self) {
  let self = self.reinterpret_as_uint()
  let mut x = self ^ (self >> 17)
  x = x * 0xed5ad4bb
  x = x ^ (x >> 11)
  x = x * 0xac4c1b51
  x = x ^ (x >> 15)
  x = x * 0x31848bab
  x = x ^ (x >> 14)
  x.reinterpret_as_int()
}
```


```moonbit
impl Hash with hash(self) {
  Hasher::new()..combine(self).finalize()
}
```

illusory0x0:
`derive(Hash)`  looks like it's using `hash_combine` to derive, and then using the default impl implementation, which should normally be the same as the hash of the reprentation

```moonbit
priv type MyHashInt Int derive(Hash, Show)
```

peter-jerry-ye:
What property are you expecting when you say "consistent" ? Like do you expect if `struct T {a : Int; b : Int}` 's hash has anything to do with the Int's hash?

PR #1769, opened 2025-03-11
fix(string): fix `rev` for non-ascii characters

Yu-zh:


hackwaly:
This may be not sufficient for unicode string.

Eg. Emoji "🏴‍☠️" is composed by two or more unicode characters. Simple reverse result is not as expected.

See https://github.com/marijnh/find-cluster-break

PR #1771, opened 2025-03-11
Breaking change: indexing string returns `Int` instead of `Char`

Yu-zh:
Breaking API change:
* `String::op_get` returns `Int` instead of `Char` to reflect that it reads the codeunit instead of the unicode codepoint

New API: 
* `String::unsafe_charcode_at` to avoid bound check

This PR also deprecates the following obsolete methods:
* `String::get()`
* `String::charcode_length()`

### DIFF
--- a/builtin/bigint_nonjs.mbt
+++ b/builtin/bigint_nonjs.mbt
@@ -1042,20 +1042,23 @@ pub fn BigInt::from_string(input : String) -> BigInt {
   if len == 0 {
     abort("empty string")
   }
-  let sign : Sign = if input[0] == '-' { Negative } else { Positive }
+  let sign : Sign = if Char::from_int(input.unsafe_charcode_at(0)) == '-' {
+    Negative
+  } else {
+    Positive
+  }
   let mut b_len = (
       (len.to_double() / decimal_ratio).to_unchecked_int() + 1 + radix_bit_len
     ) /
     radix_bit_len +
     1
   let b = FixedArray::make(b_len, 0U)
-  for i = (match sign {
-          Negative => 1
-          Positive => 0
-        })
-      i < input.length()
-      i = i + 1 {
-    let x = input[i].to_int() - 48 // ASCII value of '0'
+  for
+    i in (match sign {
+      Negative => 1
+      Positive => 0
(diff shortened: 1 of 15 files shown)


builtin/bigint_nonjs.mbt
```diff
@@ -1042,20 +1042,23 @@ pub fn BigInt::from_string(input : String) -> BigInt {
   if len == 0 {
     abort("empty string")
   }
-  let sign : Sign = if input[0] == '-' { Negative } else { Positive }
+  let sign : Sign = if Char::from_int(input.unsafe_charcode_at(0)) == '-' {
+    Negative
```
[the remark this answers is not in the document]
bobzhang: idea: it may make sense to allow ('x' : Int) works, since we already plan to make ('x' : Byte) work

builtin/intrinsics.mbt
```diff
@@ -1745,25 +1745,7 @@ pub fn FixedArray::make[T](len : Int, init : T) -> FixedArray[T] = "%fixedarray.
...
-/// Example:
-///
-/// ```moonbit
-/// test "String::charcode_length" {
-///   let s = "Hello🤣"
-///   inspect!(s.charcode_length(), content="7") // 5 ASCII chars + 2 surrogate pairs for emoji
-/// }
-/// ```
+#deprecated("use `length` instead")
 pub fn String::charcode_length(self : String) -> Int = "%string_length"
```
bobzhang, line 1749: mbti support is not there yet, you still need `/// @deprecated`

builtin/intrinsics.mbt
```diff
@@ -1796,6 +1778,7 @@ pub fn String::charcode_length(self : String) -> Int = "%string_length"
 /// ```
 ///
 /// @alert unsafe "Panic if index is out of bounds"
+#deprecated("use `charcode_at` instead")
 pub fn String::op_get(self : String, idx : Int) -> Char = "%string_get"
```
bobzhang, line 1782: same

bobzhang (commented):
I am thinking that we still support `s[i]` but its return type is `Int`, does it make sense?

PR #1774, opened 2025-03-11
feat: allow to get current working directory

ShellWen:
This PR adds a way to get the current working directory.

wasm and wasm-gc backend need moonrun changes in moonbitlang/moon#664
ref: moonbitlang/x#105


ShellWen:
After removing `moonbitlang/core/option` in `[moon.pkg.json](https://github.com/moonbitlang/core/blob/main/env/moon.pkg.json)` the test will fail. But it breaks `moon check`. I need help on it.

hackwaly:
This unused package warning might be false positive. Try using `warn-list` in moon.pkg.json to suppress the warning?

ShellWen:
P1

It seems like the `warn-list` has no effect.

hackwaly:
There's a subtle strange with `warn-list`. It needs `-29` instead of `-1029`. Shall we fix this? /cc @bobzhang

ShellWen:
moonbitlang/moon#664 has been merged

ShellWen:
It seems that we have to wait for the nightly build and can only continue tomorrow. 😭

ShellWen:
force push due to GPG

ShellWen:
ready for merge😉

bobzhang:
> There's a subtle strange with `warn-list`. It needs `-29` instead of `-1029`. Shall we fix this? /cc @bobzhang

I think so, @Yu-zh would you have a look?

Young-Flash (approved):
LGTM overall. we have a bleeding test on windows failed, wait for fixing

PR #1783, opened 2025-03-13
separate builtin package and prelude package

Guest0x0:
Currently, `@builtin` serves two purpose in core:

1. hold fundamental definitions, such as arithmetic operators on builtin types, to resolve cyclic definition between other core packages
2. serve as a prelude package. The content of `@builtin` can be used directly, unqualified

But these two purposes are contradictory: as a place to hold fundamental definitions, we want `@builtin` to be as small as possible. Meanwhile, due to `@builtin`'s role as a prelude package, we have put heavy things such as linked hash map into `@builtin`.

So, we intend to split `@builtin` into two packages:

- `@builtin` will hold fundamental definitions, and should be as small as possible
- a new package `@prelude` will be added, and its content can be used directly unqualified. `@prelude` should not contain any actual code, it should just hold alias to other core packages

This way, types like `Map`/`Set` can be placed in their own package, while users can still use them unqualified via aliases is `@prelude`. The dependency structure would be:

```
@prelude -> other core packages -> @builtin
```

This PR introduces the aforementioned `@prelude` package. However, this migration require change in the compiler as well. So currently, `@prelude` is just a useless new package, until a compiler with `@prelude` support is released.

### DIFF
--- a/prelude/moon.pkg.json
+++ b/prelude/moon.pkg.json
@@ -1,0 +1,5 @@
+{
+  "import": [
+    "moonbitlang/core/builtin"
+  ]
+}
--- a/prelude/prelude.mbt
+++ b/prelude/prelude.mbt
@@ -1,0 +1,79 @@
+// Copyright 2025 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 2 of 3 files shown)


Guest0x0:
This PR itself, without compiler support, is just a new package in core. So we can merge it first. cc @bobzhang

PR #1788, opened 2025-03-14
add stackoverflow test for @immut/list.T

illusory0x0:

PR #1789, opened 2025-03-14
refactor: use loop to implement rev_fold to avoid stack overflow

illusory0x0:
- #1788


Lampese (approved):

peter-jerry-ye:
Next time when creating a PR, you may merge the test and the fix in one PR, as long as the test comes first. For example:

1. add test to show bug
2. fix the test
3. update the snapshot (if using `inspect`) and cleanup, etc.

PR #1794, opened 2025-03-15
feat: add Compare::max and Compare::min

illusory0x0:


bobzhang:
@illusory0x0 we have a high bar when adding new traits or modify existing traits, can you elaborate why it is needed, does it justify?

illusory0x0:
> @illusory0x0 we have a high bar when adding new traits or modify existing traits, can you elaborate why it is needed, does it justify?

```moonbit

test {
  let x = 132
  let y = 2323
  let z = 2323 
  String::max(x, y) |> ignore // use monomorphic functions would help type infer
  (@math.maximum : (Int,Int) -> Int)(x,y) |> ignore

  x.max(y) |> ignore // infix syntax
  @math.maximum(x,y) |> ignore // prefix syntax

  x.max(y).max(x) |> ignore // easy to writing and avoid nesting
  @math.maximum(x,@math.maximum(y,z)) |> ignore
}
```

bobzhang:
the benefit is not clear, if you would like to chain, you can still implement it as a method, note `x.max(y)` is not that readable either

PR #1797, opened 2025-03-16
refactor: rewrite match with wildcard pattern to is pattern

hackwaly:
This PR rewrite match with wildcard pattern to is pattern, thus avoid fragile pattern matching warnings. And result's better readability.

### DIFF
--- a/immut/array/tree_utils.mbt
+++ b/immut/array/tree_utils.mbt
@@ -18,26 +18,17 @@
 ///|
 /// If the tree is a `Node`.
 fn Tree::is_node[A](self : Tree[A]) -> Bool {
-  match self {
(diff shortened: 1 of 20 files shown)


immut/array/tree_utils.mbt
```diff
@@ -18,26 +18,17 @@
 ///|
 /// If the tree is a `Node`.
 fn Tree::is_node[A](self : Tree[A]) -> Bool {
-  match self {
-    Node(_, _) => true
-    _ => false
-  }
+  self is Node(_, _)
 }
```
bobzhang, line 22: note these small utility functions may not be needed in the future

bobzhang (approved):

PR #1798, opened 2025-03-17
add trait definition for operators overloading

Guest0x0:
We plan to migrate the operator overloading mechanism from using methods to using traits. This would give:

- give better type inference for operators
- better documentation for operator overloading, the intent of overloading operators would become explicit
- allows user-defined traits to include builtin types' operators via super trait

This PR is a starter for this migration, it adds trait definitions for each overload-able MoonBit operator. Note that this PR alone has no effect yet, compiler changes are necessary too, which would hopefully be released next week.

After the compiler migrates operator overloading to traits, current user code that uses method for operator overloading will not receive hard break, because currently we still allow implementing traits with methods. But a warning will be emitted in this case, so there would be a graceful period of migration for free, until we completely remove the behavior of implementing traits with methods.

### DIFF
--- a/builtin/operators.mbt
+++ b/builtin/operators.mbt
@@ -14,66 +14,66 @@
 
 ///|
 /// types implementing this trait can use the `+` operator
-pub(open) trait OperatorAdd {
+pub(open) trait Add {
   op_add(Self, Self) -> Self
 }
 
 ///|
 /// types implementing this trait can use the `-` operator
-pub(open) trait OperatorSub {
+pub(open) trait Sub {
   op_sub(Self, Self) -> Self
 }
 
 ///|
 /// types implementing this trait can use the `*` operator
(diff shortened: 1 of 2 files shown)


bobzhang:
The original name is a bit too long, I removed the operator prefix to follow the Rust convention

builtin/operators.mbt
```diff
@@ -0,0 +1,79 @@
...
+
+///|
+/// types implementing this trait can use the `|` operator
+pub(open) trait OperatorBitOr {
+  lor(Self, Self) -> Self
+}
+
+///|
+/// types implementing this trait can use the `^` operator
+pub(open) trait OperatorBitXOr {
```
tonyfettes: @Guest0x0 Can we change this to `Xor`

PR #1810, opened 2025-03-19
migrate operator overloading to using traits

Guest0x0:
This PR migrate overloaded operators to use traits in `builtin/operators.mbt`. Since trait implementations can be called via dot syntax directly, this change is compatible with current version of MoonBit, and allows seamless migration to new operator overloading semantic based on traits.

The only exception that cannot be migrated is `Char::op_sub`, whose return type is `Int` rather than `Char`. This operator is already deprecated in #1806, and will be removed in the future.


bobzhang:
cc @gmlewis

PR #1822, opened 2025-03-21
feat: add @bytes.View::{bytes,start}

tonyfettes:
This PR adds the following two methods to `@bytes.View`:

1. `bytes`: Returns the underlying `Bytes`.
2. `start`: Returns the start index of this view.

These two functions can be helpful when passing MoonBit `@bytes.View` to the C side without creating a new `Bytes` object.

For example, the declaration of the `write` syscall on Unix is as follows:

```c
ssize_t write(int fd, const void *buf, size_t n_byte);
```

If we could retrieve the underlying bytes from `@bytes.View`, we can obtain a pointer to
the start of the `@bytes.View` on the C side:

```c
int moonbit_write_wrapper(int fd, const void *bytes, int start, int length) {
  const void *buf = bytes + start;
  return write(fd, buf, length);
}
```

And we can bind to such function on MoonBit side as:

```moonbit
extern "c" fn write_wrapper(fd : Int, bytes : Bytes, offset : Int, length : Int) -> Int = "moonbit_write_wrapper"

pub fn write(fd : Int, view : @bytes.View) -> Int {
  return write_wrapper(fd, view.bytes(), view.start())
}
```

And the user of the library could just use `@bytes.View` as the input, without additional cost of copying the bytes.

```moonbit
test {
  let bytes_view : @bytes.View = { ... }
  let fd : Int = { ... }
  write(fd, bytes_view)
}
```

### DIFF
--- a/buffer/buffer.mbt
+++ b/buffer/buffer.mbt
@@ -587,7 +587,7 @@ pub fn write_bytesview(self : T, value : @bytes.View) -> Unit {
   self.data.blit_from_bytes(
     self.len,
     value.data(),
-    value.start_of_data(),
+    value.start_offset(),
     value.length(),
   )
   self.len += val_len
--- a/buffer/buffer.mbt
+++ b/buffer/buffer.mbt
@@ -584,7 +584,12 @@ pub fn write_bytes(self : T, value : Bytes) -> Unit {
 pub fn write_bytesview(self : T, value : @bytes.View) -> Unit {
   let val_len = value.length()
   self.grow_if_necessary(self.len + val_len)
-  self.data.blit_from_bytesview(self.len, value)
+  self.data.blit_from_bytes(
+    self.len,
+    value.data(),
+    value.start_of_data(),
+    value.length(),
+  )
   self.len += val_len
 }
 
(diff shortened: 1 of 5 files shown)


bobzhang:
should this be exposed as C macros? we may change the bytes view layout in the future

tonyfettes:
Yes, adding C macros for these would also eliminate the copy and probably a better solution, if we are going to change the memory layout of the view.

bobzhang:
@tonyfettes I am convinced that API is generally useful in MoonBit itself, like what i did in @buffer.write_bytes_view, but tweaked the API name a little bit, let me know what you think

tonyfettes:
`start_of_data` sounds a little bit wired to me. If `data` is the underlying bytes, then `start_of_data` should be 0 by definition. Probably `start_offset`/`start_from_data` are better names for this?

bobzhang:
> `start_of_data` sounds a little bit wired to me. If `data` is the underlying bytes, then `start_of_data` should be 0 by definition? Probably `start_offset`/`start_from_data` are better names for this?

Modified to `start_offset` which is also consistent with @string.View

Issue #1830, opened 2025-03-25
Efficient set operations for `@immut/hash{map, set}`


waterlens:
There should be the following set operations on `@immut/hash{map, set}`:

```moonbit
// T is either T[K, V] or T[K]
fn intersection(T, T) -> T
fn union(T, T) -> T
fn difference(T, T) -> T
// only in hashmap for customed conflict resolution function
fn intersection_with(T, T, (V, V) -> V) -> T
fn union_with(T, T, (V, V) -> V) -> T
```

This clojure library might be a good reference:
https://github.com/cnuernber/ham-fisted

Lampese:
What about mutable hash{map,set}, mutable sorted{map,set}, @immut/sorted_{map, set}?

waterlens:
Leave space for other contributors :)

Lampese:
> Leave space for other contributors :)

If this API is necessary, we need to open another issue to mark it and avoid forgetting.

waterlens:
> > Leave space for other contributors :)
> 
> If this API is necessary, we need to open another issue to mark it and avoid forgetting.

Maybe we can leave this issue open?

PR #1835, opened 2025-03-27
Add @immut/hash{set,map}::union

waterlens:
A part of #1830.

### DIFF
--- a/immut/hashmap/HAMT.mbt
+++ b/immut/hashmap/HAMT.mbt
@@ -195,6 +195,26 @@ pub fn size[K, V](self : T[K, V]) -> Int {
 }
 
 ///|
+/// Union two hashmaps
+pub fn union[K : Eq + Hash, V](self : T[K, V], other : T[K, V]) -> T[K, V] {
+  match (self, other) {
+    (_, Empty) => self
+    (Empty, _) => other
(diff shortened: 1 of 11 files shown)


immut/hashmap/HAMT.mbt
```diff
@@ -194,6 +194,24 @@ pub fn size[K, V](self : T[K, V]) -> Int {
...
+///|
+/// Union two hashmaps
+pub fn union[K : Eq + Hash, V](self : T[K, V], other : T[K, V]) -> T[K, V] {
+  match (self, other) {
+    (_, Empty) => self
+    (Empty, _) => other
+    (_, Leaf(k, v)) => self.add(k, v)
+    (Leaf(k, v), _) =>
+      match other.find(k) {
+        Some(_) => other
```
Lampese, line 206: Why do the code behaviors of leaves differ between front and back?

Lampese (commented):
Maybe you need to refresh the mbti?

immut/hashmap/HAMT.mbt
waterlens: It's an asymmetric union where the value on the right-hand side is preferred.

Lampese: OK, I see.

Lampese (approved):

Issue #1626, opened 2025-02-10
Impl op_as_view for FixedArray


zxch3n:
It can reuse the `ArrayView` here, right?

Yu-zh:
There is a subtle issue with the wasm-gc backend. Because `FixedArray` cannot be casted to `UninitializedArray` so we need backend specific implementation for `ArrayView` instead of simply re-using the `ArrayView`. We will work on this.

PR #1840, opened 2025-03-28
feat(@array.View): Taking view from Fixedarray

Yu-zh:
Implementing #1626 

Still in draft because this PR needs update in moonc.

This PR separates the implementation of @array.View for wasm-gc backend and other backends because FixedArray and UninitializedArray are not convertible in wasm-gc backend. So the underlying data in wasm-gc backend could be either FixedArray or UninitializedArray.


peter-jerry-ye:
Will be replaced by a future PR.

PR #1852, opened 2025-03-30
refactor: reuse @math.maximum in @sorted_set

illusory0x0:


bobzhang:
@illusory0x0 it is intentional to avoid such small dependencies to keep packages independent when the effort is small

PR #1853, opened 2025-03-30
refactor: improve @string.View::output

illusory0x0:
feat: add default label parameter for String::output

### DIFF
--- a/builtin/builtin.mbti
+++ b/builtin/builtin.mbti
@@ -607,6 +607,7 @@ impl String {
   to_string(String) -> String
   unsafe_char_at(String, Int) -> Char
   unsafe_charcode_at(String, Int) -> Int
+  unsafe_output_substring(String, &Logger, start~ : Int = .., end~ : Int = ..) -> Unit
 }
 
 impl Option {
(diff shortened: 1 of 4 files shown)


builtin/builtin.mbti
```diff
@@ -609,6 +609,7 @@ impl String {
   to_string(String) -> String
   unsafe_char_at(String, Int) -> Char
   unsafe_charcode_at(String, Int) -> Int
+  unsafe_output_substring(String, &Logger, start~ : Int = .., end~ : Int = ..) -> Unit
```
Yu-zh, line 612: I don't think it is a good idea to make this a public API. We are trying to split some method implementations associated with a particular type from builtin to its own package. We will see if this method can be placed in the string package as a private method.

illusory0x0: > I don't think it is a good idea to make this a public API.

I think this API should be public like `Logger::write_substring`.

some string method actually need to be separate into `@string` package, `@builtin` package is too large.

Yu-zh: `Logger::write_substring` will be replaced by something like `write_view`. View is encouraged to use instead of  the UTF16 index

PR #1856, opened 2025-03-31
feat(env): make `now()` more precise on native targets

rami3l:
Upstreams https://github.com/moonbitlang/x/pull/122.

Manually tested on both Unix and Windows (in https://github.com/moonbitlang/core/actions/runs/14168177735/job/39685877322?pr=1856), both can provide the correct timestamp.

@Young-Flash suggests that this cannot be merged yet since `moon` will not recognize the native_stub.

### DIFF
--- a/env/env_native.mbt
+++ b/env/env_native.mbt
@@ -72,14 +72,8 @@ fn utf8_bytes_to_mbt_string(bytes : Bytes) -> String {
 }
 
 ///|
-extern "c" fn time(ptr : Int) -> UInt64 = "time"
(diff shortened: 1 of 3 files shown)


rami3l:
@Young-Flash suggests that this cannot be merged yet since `moon` will not recognize the native_stub.

peter-jerry-ye:
The original use case was for benchmark, but now benchmark has its own implementation in the moonbit c runtime. If there's any other usecases, we can add this to `x`.

PR #1892, opened 2025-04-08
migration: remove old [..] syntax

Guest0x0:
we are deprecating the old `..` and `.. as pat` syntax, in favor of `.. _` and `.. pat`. This PR migrates core.


bobzhang:
@Guest0x0 replace `_` with `_unused` should be good

bobzhang:
per discussion offline, we will support `[ .. x]`, `[ .. _]` and `[ .. _x]`, the `[.. _]` will be formatted into `[..]`, so this warning (on the bleeding) will be reverted. cc @Yu-zh

PR #1895, opened 2025-04-08
Benchmark support in core

waterlens:

PR #1898, opened 2025-04-09
fix: call obj.to_string() only once to avoid potential side effects

lijunchen:
Fix https://github.com/moonbitlang/moon/issues/731

tests will be added in `moon`


Young-Flash (approved):
wait https://github.com/moonbitlang/core/pull/1895 for merge to pass bleeding ci

PR #1901, opened 2025-04-09
test: add missing tests for benchmark

waterlens:


waterlens:
This should be merged only when we are going to release the new moon and new core to avoid possible breaks.

PR #1910, opened 2025-04-09
refactor: introduce Source trait for random generator

### LINKED ISSUE #1589: [Refactor API] Separate Rand and Source
Currently the `Rand` is bascially an implementation of ChaCha8, which is not ideal, as we want to provide multiple random number algorithms for different occasions. For example, we may want LCG as its fast and straightforward when we do not need cryptographic security.

A possible solution is to follow [go's path](https://pkg.go.dev/math/rand/v2#Source), defining a `Source` trait, and define `Rand` as a wrapper around it:

```
pub(open) trait Source {
  next() -> UInt64
}

type Rand &Source
```

The migration, however, will break a couple of APIs, notably the `new` function.

peter-jerry-ye:
Closes #1589

### DIFF
--- a/random/random.mbti
+++ b/random/random.mbti
@@ -7,6 +7,8 @@ import(
 // Values
 fn bigint(Rand, Int) -> @bigint.BigInt
 
+fn chacha8(seed~ : Bytes = ..) -> &Source
+
 fn double(Rand) -> Double
 
 fn float(Rand) -> Float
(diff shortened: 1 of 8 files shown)


random/random.mbti
```diff
@@ -24,7 +27,7 @@ impl Rand {
   float(Self) -> Float
   int(Self, limit~ : Int = ..) -> Int
   int64(Self, limit~ : Int64 = ..) -> Int64
-  new(seed~ : Bytes = ..) -> Self //deprecated
+  new(&Source) -> Self
```
peter-jerry-ye, line 30: add a default value

PR #1914, opened 2025-04-10
refactor(string): use View ubiquitously in public APIs

Yu-zh:
This PR significantly refactors the string package by using the View type whenever possible. The other packages should follow this as well. Currently there might performance regression caused by converting between String and View, but eventually we expect View to be used ubiquitously and String is only used for FFI boundaries. 

We will try to smooth the migration by providing implicit conversion from String to View when the expected type is View. However, method invocations cannot benefit from this conversion so all the methods will have two copies, one for String and the other for View.

There are four kinds of APIs changes:

* Both name and type are changed. Old API deprecated, new API added for both View and String. 
    * `s.ends_with(substr: String) -> Bool` -> `s.has_suffix(substr: View) -> Bool`
    * `s.starts_with(substr: String) -> Bool` -> `s.has_prefix(substr: View) -> Bool`
    * `s.index_of(substr: String, from~: Int) -> Int` -> `s.find(substr: View) -> Option[Int]`. We cannot use the word `index` here because we call it `offset` for utf16 index
    *  `s.last_index_of(substr: String, from~: Int) -> Int` -> `s.rfind(substr: View) -> Option[Int]`
    * `s.trim_end(charset: String) -> String` -> `s.trim_right(charset: View) -> View`
    * `s.trim_start(charset: String) -> String` -> `s.trim_left(charset: View) -> View`
 

* Only type is changed. I added the new APIs with correct type as methods to View. The original methods associated to String are deprecated, and once the deprecated methods are removed I will add back the methods to String.
    * `s.contains(substr: String) -> Bool` -> `s.contains(substr: View) -> Bool`
    * `s.pad_end(n: Int, c: Char) -> String` -> `s.pad_end(n: Int, c: Char) -> View`, and same for `pad_start`
    * `s.repeat(n: Int) -> String` -> `s.repeat(n: Int) -> View`
    * `s.rev() -> String` -> `s.rev() -> View`
    * `s.replace(old~: String, new~: String) -> String` -> `s.replace(old~: View, new~: View) -> View`
    * `s.replace_all(old~: String, new~: String) -> String` -> `s.replace_all(old~: View, new~: View) -> View`
    * `s.split(sep: String) -> Iter[String]` -> `s.split(sep: View) -> Iter[View]`
    * `s.to_lower() -> String` -> `s.to_lower() -> View`, and same for `to_upper`
    * `s.trim(charset: String) -> String` -> `s.trim(charset: View) -> View`
    * `s.trim_space() -> String` -> `s.trim_space() -> View`

* Both name and type are are kept. I added the API to View type, and deprecated the usage of methods as regular function calls.
    * `s.contains_char(c: Char) -> Bool`
    * `default() -> String/View`, `from_array(arr: Array[Char]) -> String/View`, `from_iter(iter: Iter[Char]) -> String/View`
    * `s.fold` and `s.rev_fold`
    * `s.is_empty()` and `s.is_blank()`
    * `s.iter()`, `s.iter2()`, `s.rev_iter()`

* Deprecated:
    * `s.to_bytes()`. I think it needs a better name like `to_underlying_bytes()`. Currently, user can use `Bytes::of_string(s)` instead.
    * `s.to_array()`. Should use `s.iter().collect()` instead.

### DIFF
--- a/json/json.mbt
+++ b/json/json.mbt
@@ -80,7 +80,7 @@ fn indent_str(level : Int, indent : Int) -> String {
   if indent == 0 {
     ""
   } else {
-    "\n" + " ".repeat(indent * level)
+    "\n" + String::make(indent * level, ' ')
   }
 }
 
--- a/string/methods.mbt
+++ b/string/methods.mbt
@@ -1,0 +1,821 @@
+// Copyright 2025 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+///|
+/// Returns the offset of the first occurrence of the given substring. If the
+/// substring is not found, it returns None.
+pub fn View::find(self : View, str : View) -> Int? {
+  let len = self.len()
+  let sub_len = str.len()
+  // Handle empty substring case
+  guard sub_len > 0 else { return Some(0) }
+  // If substring is longer than string, it can't be found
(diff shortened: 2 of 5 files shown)


bobzhang:
1. prefer `trim_end` to `trim_right`
2. str.view() -> str[:]
3. only change parameters from String to StringView

Issue #1887, opened 2025-04-06
Proposal to remove `next_positive_int` due to potential overflow and API improvements


FlyCloudC:
I propose deprecating the `next_positive_int` function in @quickcheck/splitmix

```moonbit
pub fn next_positive_int(self : RandomState) -> Int {
  let r = self.next_int()
  if r < 0 {
    -r
  } else {
    r + 1
  }
}
```

Reasons:

1. **Potential Overflow Vulnerability**  
   The current implementation contains `r + 1` which could overflow when `r` approaches `MAX` for its integer type. In a security-sensitive context, this could lead to severe vulnerabilities.

2. **API Misalignment**  
   All existing uses in the standard library follow the pattern `next_positive_int() % size`. This suggests the API should instead directly return values within `[0, size)` bounds.

![Image](https://github.com/user-attachments/assets/55da0377-44f7-42e0-b276-f37703de90f8)

**Proposed Solution**:
Introduce a replacement API:
```moonbit
fn next_int_mod(self : RandomState, p: Int) -> Int
```

CAIMEOX:
Thanks for the submission, but this API shouldn't be removed. There are probably other places to use besides the core. 
The overflow issue should be fixed though.

PR #1915, opened 2025-04-10
Fix next_positive_int() to properly get the next random number as a positive 32-bit signed integer

five-of-fuli:
Modified the function `next_positive_int()` function in `@quickcheck\splitmix\random.mbt`
Related #1887 
#### Original Code:
```
if r < 0 {
    -r
  } else {
    r + 1
  }
```
#### Modified Code:
```
match r {
    -2147483648 => 2147483647
    0 => 1
    r if r < 0 => -r
    r => r
  }
```

- Handled original code in edge cases `2147483647` causes overflow.
- Handled `0` by converting it to `1`.
- Converted negative values to their positive counterparts.
- Directly returned positive values.


CAIMEOX:
hello @five-of-fuli , please update the tests on the markdown file.

CAIMEOX (approved):

PR #1917, opened 2025-04-11
refactor: update visibility of Json enum to readonly for smart constructor usage

bobzhang:


builtin/json.mbt
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 ///|
+#visibility(change_to="readonly", "Use smart constructor to make json")
 pub(all) enum Json {
```
bobzhang, line 17: two issues I found:
1. #visibility(change_to="readonly") alone would trigger unused warning
2. the IDE does not show the warning

Yoorkin:
There are some warnings when running `moon check`:

```
moon check
[2000] Warning: Warning (Alert visibility): Use smart constructor to make json
[2000] Warning: Warning (Alert visibility): Use smart constructor to make json
[2000] Warning: Warning (Alert visibility): Use smart constructor to make json
```

These warnings don't have any location because they came from the code generated by `derive(ToJson,FromJson)`. 
We may need to fix the compiler first.

builtin/json.mbt
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 ///|
+#visibility(change_to="readonly", "Use smart constructor to make json")
```
gmlewis, line 16: I've seen the term "smart constructor" used in other places as well, but it is not clear to me what that means.
I also searched the official MoonBit docs for this term, but it doesn't come up with a definition.
Could you please define what this means somewhere?

smart-constructor-2025-04-17_08-02-17

Yoorkin: Smart constructor is a design pattern that creates the instances of enum types via helper functions... It's not documented because this concept comes from other programming languages, [possibly Haskell](https://wiki.haskell.org/index.php?title=Smart_constructors).

PR #1928, opened 2025-04-12
docs(SortedMap): add README

kesmeey:


kesmeey:
> Incorrect time complexity for range operation
> Missing example for corner case behavior

Thanks for the feedback! I've updated the docs to clarify range behavior, add performance notes, and include error-handling examples. Let me know if you'd like any adjustments.

PR #1933, opened 2025-04-13
refactor(iter): until, drop_while

FlyCloudC:
Improved readability, but need clarification on performance impact.

Is the current approach more/less efficient than the original? How does Moonbit handle for iteration internally?

### DIFF
--- a/builtin/iter.mbt
+++ b/builtin/iter.mbt
@@ -199,6 +199,72 @@ pub fn Iter::repeat[T](a : T) -> Iter[T] {
 }
 
 ///|
+/// require: step > 0
+fn iter_number_forward[A : Add + Compare](
+  begin : A,
+  end : A,
+  step~ : A,
+  inclusive~ : Bool
+) -> Iter[A] {
+  if inclusive {
+    fn(yield_) {
+      for i = begin; i <= end; {
+        guard yield_(i) is IterContinue else { break IterEnd }
+        let next = i + step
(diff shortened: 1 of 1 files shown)


bobzhang:
we prefer simple stuff, is there any tangible benefit after this commit from user point of view

FlyCloudC:
For `until`, only one check on `inclusive` and the sign of `step` needs to be performed, rather than multiple times.

For `drop_while`, `dropping = false` is executed once, rather than multiple times.  

For `take_while` and `map_while`, whether performance can be improved depends on how the compiler implements the `for` loop. I just tested it and found that the compiled JavaScript code actually became less efficient. The changes to these two functions should be canceled.

FlyCloudC:
By the way, is there a guide on how to write standard library code to leverage compiler optimizations, such as intrinsics? Or should this type of refactoring wait until compiler optimizations are more mature?

bobzhang:
@FlyCloudC The guideline is:
- Write simple code, only write generic code when it is really needed
    generic code may be as fast in runtime, but it could be slow in compile time.
- Performance optimization is okay but only when it is a clear win, in general,
   standard lib aims to provide reasonable performance, but it is a general lib so 
   it is hard to beat hand optimized lib for niche cases 
- Submit small patches so that it is easy to review

bobzhang:
I will close this now since it is not a clear win to me and we may provide a dedicated range package in the future

PR #1934, opened 2025-04-13
refactor(queue): optimize data structure and improve method efficiency

FlyCloudC:
1. Replaced `Cell[A]` with `Cons[A]?` to leverage compiler optimizations for nullable types.  
2. Refactored some methods to avoid redundant operations


peter-jerry-ye (approved):
LGTM

PR #1942, opened 2025-04-15
refactor(string): add `find`, `rev_find`, `has_suffix`, and `has_prefix`

Yu-zh:
This PR deprecates: `index_of`, `last_index_of`, `starts_with`, and `ends_with`. Replace them with `find`, `rev_find`, `has_prefix`, `has_suffix`.

Rationale:

* The word "index" is ambiguous since it could refer to either Utf16 index or char index, so it is inappropriate to use it in the API names. `find` and `rev_find` returns the Utf16 index.
* `s.starts_with("abc")` is unclear whether "abc" is used as a prefix substring or a char set. So it is renamed to `has_prefix` for clarity.

### DIFF
--- a/string/methods.mbt
+++ b/string/methods.mbt
@@ -1,0 +1,178 @@
+// Copyright 2025 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


bobzhang:
merged in #1943

PR #1946, opened 2025-04-16
fix(char): use Char::unsafe_from_int and deprecated Char::from_int

Yu-zh:
This PR deprecated `Char::from_int` and replaced it with `Char::unsafe_from_int`. Eventually we should change `Char::from_int` to `(Int) -> Option[Char]` after the grace period.


hackwaly:
How about `unsafe_from_codepoint`?

Yu-zh:
close due to #2007

PR #1947, opened 2025-04-16
refactor: separate `@bench` and `@test`

waterlens:
This is related to the improvement 2 in https://github.com/moonbitlang/moon/issues/749.
For benchmarking, we will use a separate `@bench.T` type. 

To maintain compatibility, the existing `@test.T::bench` method will remain until all changes to moon and core are stable. Its functionality is now delegated to `@bench.single_bench`~~, and it has been marked as deprecated~~.

The new `@bench.T::bench` method supports batch benchmarking. An example can be found in moonbitlang/moon#751.


waterlens:
The visibility of `Summary` was also changed as users may want to read its fields if they use `@bench.single_bench`

bench:
who wants to separate me from @test ?

bench/bench.mbti
```diff
@@ -0,0 +1,44 @@
...
+  median_abs_dev : Double
+  median_abs_dev_pct : Double
+  quartiles : (Double, Double, Double)
+  iqr : Double
+  batch_size : Int
+  runs : Int
+}
+impl ToJson for Summary
+
+pub(all) struct T {
```
bobzhang, line 31: do we need expose this to users?

waterlens: Definitely not what I want. Sry I missed this one.

bobzhang:
cc @Young-Flash

Issue #1941, opened 2025-04-15
[Refactor] migrate `immut/list` to destination passing style


peter-jerry-ye:
**Motivation**

In short: avoid stack overflow for operations such as `map`.

**Solution**

Make the structure mutable while not exposing the mutability to the outside world by using `pub` constructor (instead of `pub(all)`.

Ref: [tail modulo constructor](https://ocaml.org/manual/5.3/tail_mod_cons.html)
Impl: [Stack Safe List](https://github.com/Yu-zh/sslist)

**Migration**

- [ ] deprecate constructor, use smart constructor
- [ ] change visibility to `pub`, change internal representation, change all implementations that may stack overflow

hackwaly:
This will break a lot of code. Can we support `mut(priv)` so that mutability is only allowed in package level?

peter-jerry-ye:
What's the use case of `immut/list`, while there's `immut/array`, except for tutorial?

Yu-zh:
> What's the use case of `immut/list`, while there's `immut/array`, except for tutorial? It's less efficient and has fewer utilities, and suffers from the overflow.

`immut/list` has guaranteed O(1) complexity for cons and uncons instead of amortized, and it has pattern matching.

hackwaly:
I've tested list that perform nearly identical to array on `iter` and `cons` if it's elements constructed in a batch.

peter-jerry-ye:
So the current plan would be adding a new `list` package and deprecate the `immut/list` afterwards because:

1. It's impossible to migrate gradually since the new constructor would contain labelled argument and would break everything
2. We want to make it easier to access

hackwaly:
List constructing without cons operator is pailful. Will we consider add a cons operator for list like data structures?

illusory0x0:
> * [ ]  deprecate the `immut/list` package

I think it would be better to keep immut/list because immut/list is simpler to write like induction.

But we always recommend user use `@list` package for performance.

peter-jerry-ye:
We'll add deprecation on type (or methods)

PR #1950, opened 2025-04-17
refactor: add new stack safe list implementation

peter-jerry-ye:
Reference : #1941 

This PR rewrites the `immut/list` in a stack safe manner by using 'tail modulo constructor' pattern.

Notice that this introduces a few **breaking changes** compared to `immut/list`'s original implementation for API consistency with other packages:

- the argument ordering of the high order function parameter of `rev_foldi` `rev_fold` `scan_right` is changed so that the accumulator always come before the visited argument: `(i, accumulator, element) -> result` or `(accumulator, element) -> result`
- the semantic of the index of `rev_foldi` is changed: it used to be the index of the visited element in the list, and now it is the index of the visiting order.

### DIFF
--- a/list/list.mbt
+++ b/list/list.mbt
@@ -414,14 +414,11 @@ pub fn fold[A, B](self : T[A], init~ : B, f : (B, A) -> B) -> B {
 ///
 /// # Example
 /// ```
-/// let r = @list.of([1, 2, 3, 4, 5]).rev_fold(fn(x, acc) { x + acc }, init=0)
+/// let r = @list.of([1, 2, 3, 4, 5]).rev_fold(fn(acc, x) { x + acc }, init=0)
 /// assert_eq!(r, 15)
 /// ```
-pub fn rev_fold[A, B](self : T[A], init~ : B, f : (A, B) -> B) -> B {
-  match self {
-    Nil => init
-    Cons(head, tail~) => f(head, tail.rev_fold(f, init~))
-  }
+pub fn rev_fold[A, B](self : T[A], init~ : B, f : (B, A) -> B) -> B {
+  self.rev().fold(init~, fn(b, a) { f(b, a) })
 }
 
 ///|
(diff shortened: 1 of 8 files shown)


string/string_test.mbt
```diff
@@ -584,11 +584,11 @@ test "String::fold" {
   let s = "this is a long string"
   inspect!(s.fold(init=0, fn(acc, c) { acc + c.to_int() }), content="1980")
   inspect!(
-    s.fold(init=@list.Nil, @list.add) |> @list.rev(),
+    s.fold(init=@list.nil(), @list.add) |> @list.rev(),
```
hackwaly, line 587: Can we support `@list.nil` (not `@list.nil()`) as generic value? /cc @Yu-zh @bobzhang

Yu-zh: We need to check whether the generic value is deeply immutable to avoid the problem with value restriction. But I think we can only support simple cases like `nil` here.

list/list.mbt
```diff
@@ -414,14 +414,11 @@ pub fn fold[A, B](self : T[A], init~ : B, f : (B, A) -> B) -> B {
...
+/// let r = @list.of([1, 2, 3, 4, 5]).rev_fold(fn(acc, x) { x + acc }, init=0)
 /// assert_eq!(r, 15)
 /// ```
-pub fn rev_fold[A, B](self : T[A], init~ : B, f : (A, B) -> B) -> B {
-  match self {
-    Nil => init
-    Cons(head, tail~) => f(head, tail.rev_fold(f, init~))
-  }
+pub fn rev_fold[A, B](self : T[A], init~ : B, f : (B, A) -> B) -> B {
+  self.rev().fold(init~, fn(b, a) { f(b, a) })
```
FlyCloudC, line 421: Why not pass `f` directly instead of using `fn(b, a) { f(b, a) }`?

```moonbit
self.rev().fold(init~, f)
```

peter-jerry-ye: Because it wasn first written as such before I decided to break the API

list/list.mbti
```diff
@@ -0,0 +1,215 @@
...
+
+fn unzip[A, B](T[(A, B)]) -> (T[A], T[B])
+
+fn zip[A, B](T[A], T[B]) -> T[(A, B)]?
+
+// Types and methods
+pub enum T[A] {
+  Nil
+  Cons(A, mut tail~ : T[A])
+}
```
bobzhang, line 145: we should take this opportunity to make a more meaningful name, e.g, `Empty` `More`

Lampese: The key issue is that Nil and Cons are already very familiar names to everyone, and I don't think it's worth doing this for the sake of meaning. If that's the case, why not consider renaming Deque? The previous name was at least understandable to those who knew it before, but now it is incomprehensible to both those who know and those who don't.

peter-jerry-ye: Renamed to `Empty` and `Node`.

hackwaly: There's no `empty()` in core, we chose `new()` in favor of `empty()` which I don't like this for immutable collections.

Maybe we can provide both `nil()` and `empty()` for list, and mark `nil()` as deprecated in favor of `empty()`, and never remove `nil()` in the future.

peter-jerry-ye: added `new` as an alias for `empty`. And by the way, there are `empty` for `Iter` and `Option`

PR #1954, opened 2025-04-17
fix(bigint): correct to_int64

illusory0x0:


hackwaly:
The root cause is previous implementation of `BigInt::to_uint` is wrong. Under js target, UInt should be represented in -2147483648..=2147483647. But the previous implementation returns number in 0..=4294967295.

builtin/bigint_js.mbt
```diff
@@ -311,7 +311,7 @@ pub impl BitXOr for BigInt with lxor(self, other) {
 /// }
 /// ```
 pub extern "js" fn BigInt::to_uint(self : BigInt) -> UInt =
-  #|(x) => Number(BigInt.asUintN(32, x))
```
hackwaly, line 314: The root cause is here

hackwaly (approved):

illusory0x0:
> The root cause is previous implementation of `BigInt::to_uint` is wrong. Under js target, UInt should be represented in -2147483648..=2147483647. But the previous implementation returns number in 0..=4294967295.

We need add more test for `BigInt::to_uint`

hackwaly:
> > The root cause is previous implementation of `BigInt::to_uint` is wrong. Under js target, UInt should be represented in -2147483648..=2147483647. But the previous implementation returns number in 0..=4294967295.
> 
> We need add more test for `BigInt::to_uint`

I added one, this should cover the bug.
```
test "BigInt::to_uint" {
  assert_eq!(BigInt::from_int64(-1L).to_uint(), (-1).to_uint())
}
```

PR #1970, opened 2025-04-20
hongbo/modernize core

bobzhang:
- **refactor(strconv): simplify parsing functions by improving underscore handling and base determination logic**
- **refactor(string): replace `.view()` calls with slice syntax for improved performance and clarity in string methods**


bobzhang:
cc @Yu-zh

PR #1971, opened 2025-04-20
hongbo/view api design

bobzhang:
Note this PR is for discussion.
So here we consider @string.View short lived,

For parameters, always consider @string.View thanks to implicit type conversion
For return type,  the API have two choices:
1. return Self type
2. return String or View based on allocation

The choice one let you express more sharing, the choice two lacks such expressivity.


bobzhang:
cc @Yu-zh 
per discussion offline:

1. String -> String/View
2. View -> View
3. always take View as parameter

PR #1982, opened 2025-04-22
perf(string): optimize `View::find()`

rami3l:


rami3l:
Closed in favor of #1996.

PR #1990, opened 2025-04-23
refactor(immut/map): unify lookup to get

peter-jerry-ye:
We will unify the APIs to: 
- `get` : return `V?`
- `op_get` : return `V` directly

The `lookup` of `immut/hashmap` was just introduced so there's no need for deprecation phase.


hackwaly:
Does `@immut/list.T::lookup` need to rename?

peter-jerry-ye:
> Does `@immut/list.T::lookup` need to rename?

I think it's a difference concept for linear structures and maps.

peter-jerry-ye:
For the `lookup` in `list` where we use `list` as `assoclist`, we will handle that later. One possible solution is to introduce a newtype `type AssocList[K, V] List[(K, V)]`

PR #1993, opened 2025-04-23
Impl Compare for @immut/array, list and priority_queue

### LINKED ISSUE #1177: Missing common traits (Eq, Compare, Hash) impls for most @immut/xxx.T

juliano:
Fixes #1177


bobzhang:
cc @pd4d10 the implementation is correct but the performance could be improved, the priortiy queue comparison allocated a lot

Issue #1965, opened 2025-04-18
[Feature Request] add update for Map and hashmap


FlyCloudC:
```moonbit
pub fn update[K, V](self : T[K, V], f : (V?) -> V)
```

Like `get_or_init`, saves one hash computation.

PR #1998, opened 2025-04-24
implement Iter::group_by

### LINKED ISSUE #1201: provide Iter::group_by

juliano:
Fixes #1201

### DIFF
--- a/builtin/iter_test.mbt
+++ b/builtin/iter_test.mbt
@@ -811,14 +811,12 @@ test "group_by with strings" {
   let iter = ["apple", "avocado", "banana", "cherry", "blueberry"].iter()
   let grouped = iter.group_by(fn(s) { s[0] })
   let result = grouped.map(fn(g) { (g.0, g.1.collect()) }).collect()
-  let a_group = result.filter(fn(item) { item.0 == 'a' }).unsafe_get(0)
-  let b1_group = result.filter(fn(item) { item.0 == 'b' }).unsafe_get(0)
-  let c_group = result.filter(fn(item) { item.0 == 'c' }).unsafe_get(0)
-  let b2_group = result.filter(fn(item) { item.0 == 'b' }).unsafe_get(1)
-  assert_eq!(a_group.1, ["apple", "avocado"])
-  assert_eq!(b1_group.1, ["banana"])
-  assert_eq!(c_group.1, ["cherry"])
-  assert_eq!(b2_group.1, ["blueberry"])
+  assert_eq!(result, [
+    ('a', ["apple", "avocado"]),
+    ('b', ["banana"]),
+    ('c', ["cherry"]),
+    ('b', ["blueberry"]),
+  ])
 }
 
 ///|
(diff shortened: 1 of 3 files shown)


builtin/iter_test.mbt
```diff
@@ -748,3 +748,88 @@ test "Float::until negative step" {
...
+    Person::{ name: "Eve", age: 30 },
+  ].iter()
+  let grouped = people.group_by(fn(p) { p.age })
+  let groups = grouped
+    .map(fn(g) { (g.0, g.1.map(fn(p) { p.name }).collect()) })
+    .collect()
+  assert_eq!(groups, [
+    (25, ["Alice", "Bob"]),
+    (30, ["Charlie"]),
+    (35, ["Dave"]),
```
hackwaly, line 832: Why there's two `30`? It's not expected from users.

juliano: It's the same principle as the test above, it's generating a new group every time the value of the key function changes.

peter-jerry-ye (commented):
I have doubt about the type signature. I'm not quite sure whether we should have `Iter` inside `Iter`. It looks kind of weird to me. Maybe `Iter[(Key, Array[Value])]` is enough. I'm also not certain with the semantics. Are we expecting `group_by` to be an iterator that merges the adjacent items or are we expecting it to group everything visited, in which case we should have `Map[Key, Array[Value]]` as result.

hackwaly:
I think `Map[Key, Array[Value]]` make sense

juliano:
[This one was my reference](https://docs.python.org/3/library/itertools.html#itertools.groupby) for the implementation, but I probably should've questioned beforehand :)

peter-jerry-ye:
What I had in mind was
- this: https://learn.microsoft.com/en-us/dotnet/csharp/linq/standard-query-operators/grouping-data
- and this: https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/util/stream/Collectors.html#groupingBy(java.util.function.Function)

Seems we all have different understanding of what `group by` means

juliano:
It was actually my initial thought, but after reading a bit about iterators, that approach made sense.

No problem though, will come up with the new version soon

peter-jerry-ye:
FTR

The original implementation may be renamed with a more proper name such as:
- https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.chunk_by
- https://clojuredocs.org/clojure.core/partition-by
- https://ruby-doc.org/3.4.1/Enumerable.html#method-i-chunk

Though some uses `group_by`:
- https://hackage.haskell.org/package/base-4.21.0.0/docs/Data-List.html#v:groupBy
- https://docs.python.org/3/library/itertools.html#itertools.groupby

Also notice that
- `chunk` are usually considered as split into containers that has the fixed size
- `partition` are usually considered as splitting into 2

peter-jerry-ye (approved):
For better performance, we may need #1965

juliano:
> FTR
> 
> The original implementation may be renamed with a more proper name such as:
> 
> * https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.chunk_by
> * https://clojuredocs.org/clojure.core/partition-by
> * https://ruby-doc.org/3.4.1/Enumerable.html#method-i-chunk
> * https://hexdocs.pm/elixir/1.12/Enum.html#chunk_by/2
> 
> Though some uses `group_by`:
> 
> * https://hackage.haskell.org/package/base-4.21.0.0/docs/Data-List.html#v:groupBy
> * https://docs.python.org/3/library/itertools.html#itertools.groupby
> 
> Also notice that
> 
> * `chunk` are usually considered as split into containers that has the fixed size
> * `partition` are usually considered as splitting into 2

It seems to me that either `partition_by` and `chuck_by` are appropriate. Any preferences @peter-jerry-ye @hackwaly ?

hackwaly (approved):

hackwaly:
We approved on `group_by`. If this PR add `chunk_by`, it should not fix #1201

I suggest that make an separated PR for `chunk_by`.

peter-jerry-ye:
Can we keep the `group_by` for this PR and have `chunk_by` in another PR? I still have some doubt about the signature.

juliano:
> Can we keep the `group_by` for this PR and have `chunk_by` in another PR? I still have some doubt about the signature.

Done! https://github.com/moonbitlang/core/pull/2029

PR #2000, opened 2025-04-25
feat(stringview): add @buffer.write_stringview

tonyfettes:


buffer/buffer.mbt
```diff
@@ -673,6 +673,18 @@ pub impl Logger for T with write_substring(
...
+  let len = value.length()
+  self.grow_if_necessary(self.len + len * 2)
+  for i = 0, j = self.len; i < len; i = i + 1, j = j + 2 {
+    let c = value.unsafe_charcode_at(i).reinterpret_as_uint()
+    self.data[j] = (c & 0xff).to_byte()
+    self.data[j + 1] = (c >> 8).to_byte()
+  }
+  self.len += len * 2
+}
+
```
tonyfettes, line 687: I have to copy the code from FixedArray::blit_from_string here, as it is not possible to add a FixedArray::blit_from_stringview b/c of the cyclic depdency of `@builtin` and `@string`.

Issue #2002, opened 2025-04-25
@FlyCloudC can we use `prev = -1` for None to improve performance a little bit more


bobzhang:
@FlyCloudC can we use `prev = -1` for None to improve performance a little bit more

_Originally posted by @bobzhang in https://github.com/moonbitlang/core/issues/1984#issuecomment-2829571021_

Yu-zh:
close as completed in #2004

PR #2004, opened 2025-04-25
refactor(linked_hash_map): change type of prev and tail to Int

FlyCloudC:
Reference: #2002 

use `prev = -1` for None to improve performance a little bit more

### DIFF
--- a/builtin/linked_hash_map.mbt
+++ b/builtin/linked_hash_map.mbt
@@ -15,7 +15,7 @@
 ///|
 // Types
 priv struct Entry[K, V] {
-  mut prev : Int?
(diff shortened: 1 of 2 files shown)


hackwaly:
We already have this optimization at the compiler level.

hackwaly:
We have similar optimization at compiler level, but for Char, Byte etc. So it make sense.

FlyCloudC:
Do you mean that the compiler would optimize `Int32?` -> `Int64`? But what I actually want here is something like `Int31?` -> `Int32`, which seems impossible for the compiler to implement.

PR #2007, opened 2025-04-26
add Int::(unsafe)_to_char

bobzhang:
- **add Int::to_char safe API**
- **add Int::unsafe_to_char**


bobzhang:
cc @peter-jerry-ye we will use x.(unsafe)_to_char to replace `Char::from_int` since the former provide a better auto-completion experience

PR #2009, opened 2025-04-27
fix(strconv): allow "0b" to be parsed as digits when base is 16

### LINKED ISSUE #2006: `@strconv.parse_int!("0b", base=16)` fails to parse
Hi,

running

```moonbit
@strconv.parse_int!("0b", base=16)
```

with moonc v0.1.20250423+2af95d42c fails with `Err(invalid base)` even though the base is explicitely given so I would expect the base detection which is active when base=0 to not recognise "0b" as a base specifier and instead return 11 (hex B).

Yu-zh:
fix #2006

### DIFF
--- a/strconv/int.mbt
+++ b/strconv/int.mbt
@@ -41,10 +41,15 @@ fn check_and_consume_base(
       guard base is (0 | 8) else { base_err!() }
       (8, rest, true)
     }
-    ['0', 'b' | 'B', .. rest] => {
-      guard base is (0 | 2) else { base_err!() }
-      (2, rest, true)
-    }
+    ['0', 'b' | 'B', .. rest] =>
+      if base is (0 | 2) {
+        (2, rest, true)
+      } else if base is 16 {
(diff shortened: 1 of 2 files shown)


strconv/int.mbt
```diff
@@ -41,10 +41,15 @@ fn check_and_consume_base(
...
-    }
+    ['0', 'b' | 'B', .. rest] =>
+      if base is (0 | 2) {
+        (2, rest, true)
+      } else if base is 16 {
+        // 'b' is a legitimate hex digit
+        (16, view, false)
+      } else {
+        base_err!()
+      }
```
peter-jerry-ye, line 52: shouldn't it be allowed for base larger than 11?

peter-jerry-ye: `0 | 2 | 12 ..= 36` for `b`, `0 | 8 | 25..=36` for `o`, `0 | 16 | 34..=36` for `x`

peter-jerry-ye:
Replaced by #2015

Issue #1866, opened 2025-04-02
add `Iter::with_numbering(start~: Int)`


hackwaly:
Maybe `zip_with_numbering` better?

illusory0x0:
> Maybe `zip_with_numbering` better?

I think `with_index` would be better.

```moonbit
fn Iter::with_index[A](self : Iter[A]) -> Iter2[Int,A]
```

peter-jerry-ye:
What about `mapi` and `eachi`?

hackwaly:
`mapi` cannot fit my need.

`Iter:: zip_with_numbering(start~: Int = 0) -> Iter[(T, Int)]`

peter-jerry-ye:
Isn't that just `a.mapi(fn (index, value) { (value, index) } )`?

hackwaly:
Yes, you are right. But you can derive `mapi` like this `a. zip_with_numbering().map(fn { (v, i) => ... })`.

illusory0x0:
use Iter2 we can use `for syntax`to iterate index and element, and then use break statement, which eachi cann't do. 

use Iter2 can avoid destruct pair, use mapi has to construct pair and destruct  pair

peter-jerry-ye:
1. https://github.com/moonbitlang/core/issues/1319
2. the `construct pair and destruct pair` should be some optimization considered at compiler level

PR #2011, opened 2025-04-27
feat(iter): add mapi

peter-jerry-ye:
Related #1866

Note that we may want a labelled argument `offset` or something similar in the future to avoid computing the addition every time.

### DIFF
--- a/builtin/iter.mbt
+++ b/builtin/iter.mbt
@@ -416,6 +416,36 @@ pub fn Iter::map[T, R](self : Iter[T], f : (T) -> R) -> Iter[R] {
 }
 
 ///|
+/// Transforms the elements of the iterator using a mapping function.
+///
+/// # Type Parameters
+///
+/// - `T`: The type of the elements in the iterator.
+/// - `R`: The type of the transformed elements.
+///
+/// # Arguments
+///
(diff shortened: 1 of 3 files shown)


hackwaly:
There's both `map_with_index` and `mapi` in core. We should choose one to stick.

builtin/iter.mbt
```diff
@@ -415,6 +415,36 @@ pub fn Iter::map[T, R](self : Iter[T], f : (T) -> R) -> Iter[R] {
...
+/// * `self` - The input iterator.
+/// * `f` - The mapping function that transforms each element of the iterator with index.
+///
+/// # Returns
+///
+/// A new iterator that contains the transformed elements.
+pub fn Iter::mapi[T, R](self : Iter[T], f : (Int, T) -> R) -> Iter[R] {
+  fn {
+    yield_ => {
+      let mut i = -1
```
hackwaly, line 437: I prefer `i = 0`, but this doesn't matter.

hackwaly (approved):

builtin/iter.mbt
peter-jerry-ye: Just to simplify the writing. Though it may complicate the reasoning.

PR #2012, opened 2025-04-27
feat(iter): add iter2

### LINKED ISSUE #1319: [Feature Request] impl Iter2 for Iter[(A, B)]

peter-jerry-ye:
Closes #1319 

The naming is chosen as such because:

- we have `mapi` `eachi` where the `i` signifies the fact that an index is added as parameter
- we can't write `for i, v in [].iter()` anyway, so there's no ambiguity.

Update: we will use `iter2` to return `Iter2[Int, T]` to align with existing behaviors. As of `Iter[(K, V)]`, users can just use a pattern matching `let (k, v) = elem`.

### DIFF
--- a/builtin/builtin.mbti
+++ b/builtin/builtin.mbti
@@ -188,6 +188,7 @@ fn[T, K : Eq + Hash] Iter::group_by(Self[T], (T) -> K) -> Map[K, Array[T]]
 fn[A] Iter::head(Self[A]) -> A?
 fn[A] Iter::intersperse(Self[A], A) -> Self[A]
 fn[T] Iter::iter(Self[T]) -> Self[T]
+fn[T] Iter::iter2(Self[T]) -> Iter2[Int, T]
 fn Iter::join(Self[String], String) -> String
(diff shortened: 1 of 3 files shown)


builtin/builtin.mbti
```diff
@@ -236,6 +236,8 @@ impl Iter {
   head[A](Self[A]) -> A?
   intersperse[A](Self[A], A) -> Self[A]
   iter[T](Self[T]) -> Self[T]
+  iter2[K, V](Self[(K, V)]) -> Iter2[K, V]
+  iteri[T](Self[T], offset~ : Int = ..) -> Iter2[Int, T]
```
bobzhang, line 240: `iteri` looks reasonable. What's the benefit of having `iter2`?

peter-jerry-ye: We will use `iter2 -> Iter2[Int, T]`

peter-jerry-ye:
@Guest0x0 we need to let `for k, v in iter` find `iter2`

PR #2021, opened 2025-04-28
perf(@sorted_map.each) replace stack version with dfs version

illusory0x0:
sorted_map is BST, don't worry too deep cause stack overflow.

### DIFF
--- a/sorted_map/map.mbt
+++ b/sorted_map/map.mbt
@@ -124,39 +124,25 @@ pub fn clear[K, V](self : T[K, V]) -> Unit {
 ///|
 /// Iterates the map.
 pub fn each[K, V](self : T[K, V], f : (K, V) -> Unit) -> Unit {
-  let s = []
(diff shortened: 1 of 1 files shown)


sorted_map/map.mbt
```diff
@@ -123,39 +123,25 @@ pub fn clear[K, V](self : T[K, V]) -> Unit {
 ///|
 /// Iterates the map.
 pub fn each[K, V](self : T[K, V], f : (K, V) -> Unit) -> Unit {
```
hackwaly, line 125: This changed again and again in one of dfs and stack version. If there's no strong reason/support, I'd suggest that keep the old one.

illusory0x0: but this https://github.com/moonbitlang/core/pull/1456 is merged, I 'm confused about it...

hackwaly (approved):

PR #2029, opened 2025-04-28
implement Iter::chunk_by

juliano:
[As requested here](https://github.com/moonbitlang/core/pull/1998#issuecomment-2832915603)

### DIFF
--- a/builtin/builtin.mbti
+++ b/builtin/builtin.mbti
@@ -218,6 +218,7 @@ impl Iter {
   all[T](Self[T], (T) -> Bool) -> Bool
   any[T](Self[T], (T) -> Bool) -> Bool
   append[T](Self[T], T) -> Self[T]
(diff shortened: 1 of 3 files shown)


peter-jerry-ye:
I'm still not quite sure if the signature should be `-> Iter[(K, Iter[V])]` or `-> Iter[(K, Array[V])]`. Maybe we need some advices from the others here

bobzhang:
Iter was introduced as a data conversion layer between different collections. It was designed that compiler can optimize all the indirections so such conversion is almost free.

It is less powerful than other iterators (e.g, lazy lists) where meomizations and zip is easy to implement.
So if people want powerful iterators, I think the best practice is to roll a new one instead of build it on top of `Iter`, as its original scope is quite limited (no meomization intended).

I would be happy to hear your thoughts

bobzhang:
closed as suggested

Issue #2028, opened 2025-04-28
improve @list.rev_fold and @immut.list.rev_fold


illusory0x0:
The current implementation of fold_rev constructs a linked list that acts as a buffer, but both tracing GC and RC are slow to free memory, so it's better to use an array.

[benchmark](https://github.com/illusory0x0/bug_report.mbt/tree/master/bench)

```moonbit
@@ -2,6 +2,7 @@ fn main {
   let xs = Array::makei(2000,fn { x => x}) |> @immut/list.from_array
 
   for i in 0..<10000 {
-    xs.rev_fold(init=0,Add::op_add) |> ignore
+    let ys = xs.to_array()
+    ys.rev_fold(init=0,Add::op_add) |> ignore
   }
 }

```

```bash
bash bench.sh
```

```
native
rev_fold

real    0m0.282s
user    0m0.281s
sys     0m0.000s

real    0m0.268s
user    0m0.268s
sys     0m0.000s

real    0m0.267s
user    0m0.267s
sys     0m0.000s
rev_fold_array

real    0m0.180s
user    0m0.179s
sys     0m0.001s

real    0m0.178s
user    0m0.178s
sys     0m0.000s

real    0m0.179s
user    0m0.179s
sys     0m0.000s
wasm-gc
rev_fold

real    0m0.107s
user    0m0.097s
sys     0m0.012s

real    0m0.110s
user    0m0.100s
sys     0m0.011s

real    0m0.109s
user    0m0.100s
sys     0m0.010s
rev_fold_array

real    0m0.098s
user    0m0.089s
sys     0m0.012s

real    0m0.099s
user    0m0.097s
sys     0m0.005s

real    0m0.100s
user    0m0.097s
sys     0m0.005s
wasm
rev_fold

real    0m0.787s
user    0m0.786s
sys     0m0.005s

real    0m0.794s
user    0m0.794s
sys     0m0.004s

real    0m0.789s
user    0m0.786s
sys     0m0.007s
rev_fold_array

real    0m0.352s
user    0m0.350s
sys     0m0.009s

real    0m0.349s
user    0m0.348s
sys     0m0.008s

real    0m0.356s
user    0m0.357s
sys     0m0.005s
js
rev_fold

real    0m0.152s
user    0m0.150s
sys     0m0.011s

real    0m0.186s
user    0m0.186s
sys     0m0.009s

real    0m0.149s
user    0m0.141s
sys     0m0.018s
rev_fold_array

real    0m0.147s
[...]

PR #2030, opened 2025-04-28
perf: use array as buffer

illusory0x0:
- #2028


hackwaly:
Basically I think `rev_fold` should not be included in core. Any sane user should known list is not efficient to do this. And it just `list.rev().fold(...)` or `list.to_array().rev_fold(...)`

bobzhang:
note we will encourage people to use @list directly

PR #2032, opened 2025-04-28
Implement ViewArray::to_array

AlexandreRoba:
Simple implementation based on the filter method implementation

### DIFF
--- a/array/view.mbt
+++ b/array/view.mbt
@@ -334,7 +334,7 @@ pub fn View::mapi_inplace[T](self : View[T], f : (Int, T) -> T) -> Unit {
 /// assert_eq!(v, [4, 6])
 /// ```
 pub fn View::filter[T](self : View[T], f : (T) -> Bool) -> Array[T] {
-  let arr = []
+  let arr = Array::new(capacity=self.length())
   for v in self {
     if f(v) {
       arr.push(v)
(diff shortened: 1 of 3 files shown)


array/view.mbt
```diff
@@ -343,6 +343,23 @@ pub fn View::filter[T](self : View[T], f : (T) -> Bool) -> Array[T] {
...
+/// let view = [1, 2, 3, 4, 5, 6][2:4]
+/// let arr = view.to_array()
+/// assert_eq!(arr, [3, 4])
+/// ```
+pub fn View::to_array[T](self : View[T]) -> Array[T] {
+  let arr = []
+  for v in self {
+    arr.push(v)
+  }
+  arr
```
peter-jerry-ye, line 360: We may want to use `Array.unsafe_blit` for better performance.

@Yu-zh  Can we move the definition of `ArrayView` to `@array` since it's been deprecated for quite some time?

array/view.mbt
```diff
@@ -334,7 +334,7 @@ pub fn View::mapi_inplace[T](self : View[T], f : (Int, T) -> T) -> Unit {
 /// assert_eq!(v, [4, 6])
 /// ```
 pub fn View::filter[T](self : View[T], f : (T) -> Bool) -> Array[T] {
-  let arr = []
+  let arr = Array::new(capacity=self.length())
```
peter-jerry-ye, line 337: This may waste memory since filter are tends to take less space.

peter-jerry-ye (approved):
LGTM. Thank you for your contribution.

PR #2035, opened 2025-04-29
feat: @list.rev_unfold

illusory0x0:


FlyCloudC:
Is it better than the one below?

```moonbit
pub fn rev_unfold[A, S](f : (S) -> (A, S)?, init~ : S) -> T[A] {
  loop f(s), Empty {
    None, acc => acc
    Some((x, s)), acc => continue f(s), More(x, tail=acc)
  }
}
```

illusory0x0:
> Is it better than the one below?
> 
> ```moonbit
> pub fn rev_unfold[A, S](f : (S) -> (A, S)?, init~ : S) -> T[A] {
>   loop f(s), Empty {
>     None, acc => acc
>     Some((x, s)), acc => continue f(s), More(x, tail=acc)
>   }
> }
> ```

your impl

bobzhang:
unfold is quite common, rev_unfold is less common when we already had a stack safe version of unfold?

illusory0x0:
> unfold is quite common, rev_unfold is less common when we already had a stack safe version of unfold?

The reason for providing `rev_unfold` is the same reason we need `rev_fold`.

By providing constructors in two different directions, we can make it easier for programmers to switch, rather than constructing from left to right and then the reverse list.


If we directly use below  primitives to construct list, this is hard to switch construct direction. because the `@list` package only provides two construction primitives from **right to left**. but there isn't `snoc`

```moonbit

fn empty[A]() -> T[A]  // nil

fn construct[A](A, T[A]) -> T[A] // cons
```

PR #2036, opened 2025-04-29
perf: improve @list.from_iter use one pass

illusory0x0:


peter-jerry-ye:
How's the performance comparison?

illusory0x0:
> It looks drastically faster on wasm-gc. Seems to be a reasonable change.

benchmark result is unstable in js backend and wasm-gc backend due to GC.

https://github.com/moonbitlang/moon/issues/813

illusory0x0:
@peter-jerry-ye  Should we close this PR since this change is a weak performance boost?

peter-jerry-ye:
I'm fine with both implementation. If you are ready you can mark it as 'Ready for review'.

PR #2041, opened 2025-04-29
implement Map::update

### LINKED ISSUE #1965: [Feature Request] add update for Map and hashmap

juliano:
Fixes #1965 

And improves Iter::group_by

### DIFF
--- a/builtin/linked_hash_map.mbt
+++ b/builtin/linked_hash_map.mbt
@@ -263,6 +263,59 @@ pub fn Map::get_or_init[K : Hash + Eq, V](
 }
 
 ///|
+/// Updates the value associated with a key in the map using a function.
+///
+/// This method applies a function to the current value associated with the key
+/// (or `None` if the key does not exist) and sets the key to the result.
+///
+/// # Parameters
+///
+/// * `self` - The map to update.
+/// * `key` - The key whose value will be updated.
+/// * `f` - A function that takes the current value (or `None`) and returns the new value.
+///
+/// # Returns
+///
+/// The new value associated with the key after the update.
+///
+/// # Example
+///
+/// ```moonbit
+/// test "Map::update" {
+///   let map = { "a": 1, "b": 2 }
+///
+///   // Update an existing key
+///   let result1 = map.update("a", fn(v) {
+///     match v {
+///       Some(val) => val * 2
+///       None => 0
+///     }
+///   })
+///   assert_eq!(result1, 2)
+///   assert_eq!(map.get("a"), Some(2))
+///
+///   // Update a non-existing key
+///   let result2 = map.update("c", fn(v) {
+///     match v {
+///       Some(val) => val
+///       None => 42
+///     }
+///   })
+///   assert_eq!(result2, 42)
+///   assert_eq!(map.get("c"), Some(42))
+/// }
+/// ```
+pub fn Map::update[K : Hash + Eq, V](
+  self : Map[K, V],
+  key : K,
+  f : (V?) -> V
+) -> V {
+  let value = f(self.get(key))
+  self.set(key, value)
+  value
+}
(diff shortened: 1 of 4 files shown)


builtin/linked_hash_map.mbt
```diff
@@ -262,6 +262,59 @@ pub fn Map::get_or_init[K : Hash + Eq, V](
...
+pub fn Map::update[K : Hash + Eq, V](
+  self : Map[K, V],
+  key : K,
+  f : (V?) -> V
+) -> V {
+  let value = f(self.get(key))
+  self.set(key, value)
+  value
+}
+
```
peter-jerry-ye, line 317: The goal was to be efficient, which is not the case with this implementation.

I think you can copy the code of `set` and modify the line `curr_entry.value = value` to `curr_entry.value = f(Some(value))` and the `let entry = { .., value, ..}` to `let entry = { .., value : f(None), .. }`

peter-jerry-ye: Something like this:
```moonbit
  if self.size >= self.growAt {
    self.grow()
  }
  let hash = key.hash()
  let (idx, psl) = for psl = 0, idx = hash & self.capacity_mask {
    match self.entries[idx] {
      None => break (idx, psl)
      Some(curr_entry) => {
        if curr_entry.hash == hash && curr_entry.key == key {
          let new_value = f(Some(value))
          curr_entry.value = new_value
          return new_value
        }
        if psl > curr_entry.psl {
          self.push_away(idx, curr_entry)
          break (idx, psl)
        }
        continue psl + 1, (idx + 1) & self.capacity_mask
      }
    }
  }
  let new_value = f(None)
  let entry = { prev: self.tail, next: None, psl, key, value: new_value, hash }
  self.add_entry_to_tail(idx, entry)
  return new_value
```

peter-jerry-ye:
I also have some doubt about the function name and function signature. In Rust it might corresponds to `entry.and_modify.or_insert`. In Java the closest thing I find is `compute`.

However the return value of `compute` is also `V?`, where `None` means deleteting the value, which I think we can also have something similar. 

As of the name, I'm not quite sure whether it should be called `compute` or `update`. When updating a map, I tend to feel that we get the old value back, like `replace` in Java or `insert` in Rust; or self for chaining.

hackwaly:
> I also have some doubt about the function name and function signature
FYI: https://github.com/tc39/proposal-upsert

peter-jerry-ye:
> FYI: https://github.com/tc39/proposal-upsert

Thank you @hackwaly . But our operation is more complex than that, so maybe we need a name like `compute_then_insert_or_update_then_get`. Maybe it would make more sense if we simplify what this function does.

bobzhang:
FYI, the `update` in OCaml is more general:

```ocaml
  val update: key -> ('a option -> 'a option) -> 'a t -> 'a t
    (** [update key f m] returns a map containing the same bindings as
        [m], except for the binding of [key]. Depending on the value of
        [y] where [y] is [f (find_opt key m)], the binding of [key] is
        added, removed or updated. If [y] is [None], the binding is
        removed if it exists; otherwise, if [y] is [Some z] then [key]
        is associated to [z] in the resulting map.  If [key] was already
        bound in [m] to a value that is physically equal to [z], [m]
        is returned unchanged (the result of the function is then
        physically equal to [m]).
        @since 4.06.0
    *)

```

peter-jerry-ye:
We prefer the signature to be `update[K : Hash + Eq, V](Self[K, V], K, (V?) -> V?) -> Unit`

peter-jerry-ye:
Superseded by #2531

PR #2042, opened 2025-04-30
misc: reactivate disabled warning

waterlens:
Since `moon` upgraded, there are no false positive warnings for the extra `@bench` import. It's time to reactivate the warning now.


peter-jerry-ye:
CI is not passing. Do you know why? It seems that it didn't happen as you expected.

waterlens:
> CI is not passing. Do you know why? It seems that it didn't happen as you expected.

It's not expected because I wrongly assumed `moon check` would check the usefulness along with a generated test driver (where `@bench` is indeed used), but actually it doesn't work like that. The PR shows a false bleeding-CI-passing status when I opened this PR because bleeding-CI ignores the warning.

For now, I don't know what the correct approach to fix it. Some possible directions:
- Leave the disabled warning.
- Separate the bench driver and the test driver.
- Tell people to not use tests with parameters in core, and rewrite the only 1 test requiring it.

bobzhang:
superceded by #2507

PR #2046, opened 2025-04-30
refactor: update Show implementation for Char to use new syntax and improve output handling

bobzhang:
```
Error: Moonc.Basic_hash_string.Key_not_found("$@moonbitlang/core/builtin.Show::Char::to_string")
```


Yu-zh:
This intrinsic is only available for Js and wasm-gc backend.

PR #2068, opened 2025-05-07
Add `Array::windows`

MINGtoMING:
Add `Array::windows`.


FlyCloudC:
How about return an `Array[ArrayView[T]]`?

MINGtoMING:
I agree. `Array::windows`  is typically used in read-only scenarios as well. However, the func interface differs from existing `chunks`/`chunk_by` impl.

PR #2071, opened 2025-05-07
feat: add Byte::to_char method for byte to character conversion

bobzhang:
`moon test --enable-coverage` panic again

```
git/core$moon test --enable-coverage
thread 'main' panicked at /Users/runner/.cargo/git/checkouts/n2-e90a381ad9753ef2/3e74c79/src/work.rs:756:17:
BUG: no work to do and runner not running
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
cc @Young-Flash

### DIFF
--- a/builtin/builtin.mbti
+++ b/builtin/builtin.mbti
@@ -331,6 +331,7 @@ impl Byte {
   lsl(Byte, Int) -> Byte
   #deprecated
   lsr(Byte, Int) -> Byte
+  to_char(Byte) -> Char
   to_double(Byte) -> Double
   to_float(Byte) -> Float
   to_int(Byte) -> Int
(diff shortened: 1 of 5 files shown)


Young-Flash:
need pre-release moon and waiting https://github.com/moonbitlang/core/pull/2065 merged. cc @Guest0x0

peter-jerry-ye:
> We already have `Char::from_int`. If we get `Int::to_char` / `Byte::to_char`, I think it might make users get confused which method should be preferred.
> 
> I'd like to suggest that add implicit conversion for `Byte` to `Int`.
> 
> ```
> Char::from_codepoint(byte)
> ```
> 
> /cc @peter-jerry-ye

We've deprecated `Char::from_int`.

PR #2078, opened 2025-05-09
Conversion between byties and arrayview

ethereal-dream:


hackwaly:
There's `bytes[:]` to get `ByteView` for `Bytes`.

ethereal-dream:
need to implement conversion between ByteView and ArrayView?

bobzhang:
Bytes is immutable while Array is mutable so BytesView is preferred

PR #2084, opened 2025-05-12
feat(string): add find_by_charcode

Yu-zh:
Motivation: `find_by_charcode` will be more efficient than `find_by` because it skips the surrogate pair checks. This API is used for performance-critical applications.

Additionally, I'm not sure whether we want to expose an API that returns `Iter[Int]` or similar so that users care about performance can implement customized high performance string operations.


FlyCloudC:
How about naming it `find_charcode_by`?

bobzhang:
closing this since we already expose String as Immutable Array[Int] to users

PR #2087, opened 2025-05-13
feat: add BigInt::probably_prime for primality test

### LINKED ISSUE #2079: [Feature Request] BigInt::probably_prime
Large prime generation requires a primality test method.

refs: 
[https://en.wikipedia.org/wiki/Primality_test](https://en.wikipedia.org/wiki/Primality_test)
[Go's implementation](https://github.com/golang/go/blob/e282cbb11256db717b95f9d8cf8c050cd4c4f7c2/src/math/big/prime.go#L26)
[OpenSSL's implementation](https://github.com/openssl/openssl/blob/master/crypto/bn/bn_prime.c)

yj-qin:
Closes #2079 

This PR implements a primality check method for BigInt that uses trial division and the Miller-Rabin prime test algorithm.


hackwaly:
The modpow could be pub and split in another PR as someone request it.

yj-qin:
Should I put the change of `modpow_ffi` to another PR?

hackwaly:
> Should I put the change of `modpow_ffi` to another PR?

This should not block this PR. Can be done in later PR.

peter-jerry-ye:
Can we add only `ctz` `bit_length`? I don't think wee need `probably_prime`.

Random should probably be added in `random` package.

peter-jerry-ye:
The `probably_prime` should go to `math` package.

bigint/bigint_js.mbt
```diff
@@ -385,3 +397,108 @@ pub fn BigInt::to_int64(self : BigInt) -> Int64 {
...
+  #|(n) => {
+  #|  if (n === 0n) return 0;
+  #|  let absN = n < 0n ? -n : n;
+  #|  let count = 0;
+  #|  while ((absN & 1n) === 0n) {
+  #|    absN >>= 1n;
+  #|    count++;
+  #|  }
+  #|  return count;
+  #|}
```
peter-jerry-ye, line 456: @hackwaly would it be faster if use `BigInt.asUInt32` and then use ctz?

hackwaly: Maybe. But it's complex and error-prone. I think we can first land the correct version then consider optimization.

peter-jerry-ye (approved):

PR #2090, opened 2025-05-13
Add `data` and `start_offset` methods for `StringView`

MINGtoMING:
The `StringView` currently lacks methods to obtain raw data and starting offset, which hinders seamless switching between `StringView` and `String`.

Here is an example:

- Original:
```moonbit
///|
pub type Oid

///|
extern "C" fn oid_from_string_ffi(
  str : String,
  oid_buf : FixedArray[Oid?],
  err_buf : FixedArray[GitErrorMeta?]
) -> Unit = "oid_from_string_ffi"

///|
pub fn Oid::from_string(str : String) -> Oid!GitError {
  let oid_buf : FixedArray[Oid?] = [None]
  let err_buf : FixedArray[GitErrorMeta?] = [None]
  oid_from_string_ffi(str, oid_buf, err_buf)
  match (oid_buf, err_buf) {
    ([Some(oid)], [None]) => oid
    ([None], [Some(err)]) => raise GitError(err)
    _ => abort("unreachable")
  }
}
```
- After adding:
```moonbit
///|
pub type Oid

///|
extern "C" fn oid_from_string_ffi(
  str : String,
  start_offset : Int,
  length : Int,
  oid_buf : FixedArray[Oid?],
  err_buf : FixedArray[GitErrorMeta?]
) -> Unit = "oid_from_string_ffi"

///|
pub fn Oid::from_string(str : @string.StringView) -> Oid!GitError {
  let oid_buf : FixedArray[Oid?] = [None]
  let err_buf : FixedArray[GitErrorMeta?] = [None]
  oid_from_string_ffi(
    str.data(),
    str.start_offset(),
    str.length(),
    oid_buf,
    err_buf,
  )
  match (oid_buf, err_buf) {
    ([Some(oid)], [None]) => oid
    ([None], [Some(err)]) => raise GitError(err)
    _ => abort("unreachable")
  }
}
```

### DIFF
--- a/string/string.mbti
+++ b/string/string.mbti
@@ -7,6 +7,8 @@ fn contains(String, StringView) -> Bool
 
 fn contains_char(String, Char) -> Bool
 
+fn data(StringView) -> String
+
 fn default() -> String
 
 #deprecated
@@ -58,6 +60,8 @@ fn rev_iter(String) -> Iter[Char]
 
 fn split(String, StringView) -> Iter[StringView]
 
+fn start_offset(StringView) -> Int
+
 #deprecated
 fn starts_with(String, String) -> Bool
 
@@ -92,6 +96,7 @@ impl StringView {
   charcodes(Self, start~ : Int = .., end~ : Int = ..) -> Self
   contains(Self, Self) -> Bool
   contains_char(Self, Char) -> Bool
+  data(Self) -> String
   find(Self, Self) -> Int?
   find_by(Self, (Char) -> Bool) -> Int?
   fold[A](Self, init~ : A, (A, Char) -> A) -> A
@@ -119,6 +124,7 @@ impl StringView {
   rev_fold[A](Self, init~ : A, (A, Char) -> A) -> A
   rev_iter(Self) -> Iter[Char]
   split(Self, Self) -> Iter[Self]
+  start_offset(Self) -> Int
   to_lower(Self) -> Self
   to_upper(Self) -> Self
   trim(Self, Self) -> Self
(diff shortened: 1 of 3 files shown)


Yu-zh:
Is there a reason that you are using `String` and `@string.View` for C FFI instead of `Bytes` and `@bytes.View`? See https://github.com/moonbitlang/core/pull/1822

MINGtoMING:
The reason for using `@string.View` instead of `@bytes.View` here is as follows:  
1. Semantic consistency: The purpose of `Oid::from_string` is to create an object id from a hexadecimal git id string (or prefix slice). Using `@bytes.View` would essentially treat the parameter as a `CString`, which would require additional clarification to avoid ambiguity.  
2. Ergonomics: The caller can use it directly without worrying about whether the encoding of the passed `@bytes.View` is correct. For example: `let oid = Oid::from_string!("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")`

MINGtoMING:
Yes, the C prototype of `oid_from_string_ffi` accepts a hex string of length 40. Perhaps directly accepting a Bytes with `\x00` at the end would be more convenient, but on the Moonbit interface side, it feels somewhat inelegant and ambiguous. My previous solution was to place the encoding conversion in the C code to utilize the system-provided encoding conversion functions(iconv / win).

peter-jerry-ye:
Solely for your case, I think you don't even need the encoder or `iconv / win` since you should be accepting only asciis. A simple for loop should be enough for you.

MINGtoMING:
Right, in this func, it can be handled better as you suggested.

Yu-zh:
> I think it's fine to expose them. @Yu-zh

Then it looks good to me

PR #2092, opened 2025-05-14
feat: add size_hint for Iter::collect

illusory0x0:
The internal iterator doesn't know the number of elements, and it makes sense to add size_hint as a default parameter


peter-jerry-ye:
You can have:

```moonbit
let arr = Array::new(capacity=xxx)
iter.each(arr.push(_))
```

illusory0x0:
> You can have:
> 
> ```moonbit
> let arr = Array::new(capacity=xxx)
> iter.each(arr.push(_))
> ```

You're right, when using `collect` we should expect the content not to be particularly large.

PR #2102, opened 2025-05-15
feat(Byte): add count_ones method

BigOrangeQWQ:
Byte is widely used in files, protocols, and data structures. Counting the number of 1-bits in a byte is a common requirement in many scenarios where bitwise operations or bit statistics are needed, and similar functions are also available in the standard libraries of other programming languages.

[Rust count_ones](https://doc.rust-lang.org/std/primitive.u8.html#method.count_ones)
[C++ popcount](https://en.cppreference.com/w/cpp/numeric/popcount)

### DIFF
--- a/byte/byte.mbt
+++ b/byte/byte.mbt
@@ -51,12 +51,12 @@ pub fn to_uint64(self : Byte) -> UInt64 {
 /// Example:
 ///
 /// ```moonbit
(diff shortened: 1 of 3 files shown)


Yu-zh:
You can just call it `Bytes::popcnt` so that it's consistent with `Int::popcnt`.

PR #2104, opened 2025-05-16
refactor(bytesview): change bytesview back to concrete type

Yu-zh:
The compiler does not need `@bytes.View` to be abstract to perform optimizations. So it's changed back.


bobzhang:
is there any benefit changing it back? Note we still did some magic operations on View, so it may make sense to keep it abstract

Yu-zh:
> is there any benefit changing it back? Note we still did some magic operations on View, so it may make sense to keep it abstract

* To be consistent with `@string.View` and `@array.View` since there will be no benefit to change these two types to abstract type.
* Less hardcoded magic in the compiler, which makes it more maintainable. `@bytes.View` and other views are all treated as value types but they trigger different handling logic.

Guest0x0:
this can also help simplify debug info generation for bytes view in LLVM backend

PR #2105, opened 2025-05-16
ci: use prettier check fmt

illusory0x0:
`jq` only check the legal json, with prettier can be standardized formatting


bobzhang:
this would impose a burdern for new contributions who have to install the tools. The benefit is minor since our JSON is pretty small

PR #696, opened 2024-07-16
rename types to T in packages immut/x

Yoorkin:
#692

# changes

rename types in:

- immut/array
- immut/hashmap
- immut/hashset
- immut/list
- immut/priority_queue
- immut/sorted_map
- immut/sorted_set

PR #2109, opened 2025-05-17
refactor(list): replace `panic` with `abort`

illusory0x0:


peter-jerry-ye:
Replacing with `unreachable` is meaning less here.

illusory0x0:
> Replacing with `unreachable` is meaning less here.

### Meaning

Using abort internally improves code readability, is self-documenting, and can tell the user that some errors are internal to the library rather than the user violating the invariant.

### binary bloat (debug string)
The binary bloat caused by using the debug string can generally be optimized away by compilers in other languages.
But `moonc` is not optimized as such. e.g. [godbolt](https://godbolt.org/)

As you said in this https://github.com/moonbitlang/core/issues/1866#issuecomment-2800469792, These things should be done by improving the compiler and not overthinking it at coding time.



More often than not, resolving this [[moonc] generate duplicate code when type variable without constraint. #696](https://github.com/moonbitlang/moonbit-docs/issues/696)
 can drastically increase compilation speed and reduce binary size.

peter-jerry-ye:
> ### Meaning
> Using abort internally improves code readability, is self-documenting, and can tell the user that some errors are internal to the library rather than the user violating the invariant.

Yes, but `abort("unreachable")` is equivalent to `panic()`. It doesn't add any useful information.

> ### binary bloat (debug string)
> The binary bloat caused by using the debug string can generally be optimized away by compilers in other languages. But `moonc` is not optimized as such. e.g. [godbolt](https://godbolt.org/)
> 
> As you said in this [#1866 (comment)](https://github.com/moonbitlang/core/issues/1866#issuecomment-2800469792), These things should be done by improving the compiler and not overthinking it at coding time.

Yes, and I do believe that the compiler will be able to handle this. However, there's a subtle difference : the places where you are adding the abort are unreachable code. They are not some preconditions to be checked, such as #2110, but are some runtime invariant that will never be broken as long as they are properly implemented. Even if the compiler can do that, they shouldn't exist in the first place.

PR #2116, opened 2025-05-17
fix: issue #2114

HCHogan:


builtin/fixedarray_block.mbt
```diff
@@ -112,8 +112,11 @@ pub fn FixedArray::blit_to[A](
...
   dst_offset~ : Int = 0
 ) -> Unit {
-  if dst_offset + len > dst.length() || src_offset + len > self.length() {
-    panic()
+  guard len >= 0 else { abort("`len` is negative") }
+  guard dst_offset >= 0 else { abort("`dst_offset` is negative") }
+  guard src_offset >= 0 else { abort("`src_offset` is negative") }
+  guard dst_offset + len <= dst.length() && src_offset + len <= self.length() else {
+    abort("source or destination range exceeds array length")
   }
```
bobzhang, line 120: can we keep the condition concise and fail with  just single message, e.g
```
guard .... else {
  abort("bounds failed, dst: \{dst}, len: \{len} ,....")
}
```

bobzhang (changes requested):
see the comments

HCHogan:
I've refactored the guard checks and updated the error messages, let me know if there's anything else you'd like to tweak.

bobzhang (approved):

PR #2118, opened 2025-05-18
unify naming style

FlyCloudC:
1. rename `Just_One` to `JustOne`
2. rename `growAt` to `grow_at`


bobzhang:
@FlyCloudC would you do a rebase? the changes look good to me

PR #2122, opened 2025-05-18
ArrayView::join and FixedArray::join

ethereal-dream:
Adding ArrayView::join(366-381) and FixedArray::join(1177-1211) functions makes it convenient to concatenate strings within an array into a single complete string.

### DIFF
--- a/builtin/fixedarray_block.mbt
+++ b/builtin/fixedarray_block.mbt
@@ -112,11 +112,13 @@ pub fn FixedArray::blit_to[A](
   src_offset~ : Int = 0,
   dst_offset~ : Int = 0
 ) -> Unit {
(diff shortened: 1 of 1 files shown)


FlyCloudC:
**Problem Description**  
The files you modified were mistakenly placed in the root directory. This will cause Git to treat it as a new file, making your changes invisible in the history. Please place it in its original location.

**Suggested Solution**  
To resolve this cleanly, it is recommended to:  

1. Fork the repository again (if necessary) to start fresh.  
2. Make *in-place modifications* to the files you need to change without moving their locations.  
3. Submit a new Pull Request.  

This approach ensures that the repository structure remains correct and the contribution history stays organized.

Issue #2108, opened 2025-05-17
Inconsistent methods between immut/sorted_map and immut/hashmap


BigOrangeQWQ:
sorted_map methods should return an iter for better consistency and efficiency

[`sorted_map`](https://github.com/moonbitlang/core/blob/main/immut/sorted_map/sorted_map.mbti)
```moonbit
type T[K, V]
impl T {
  keys[K, V](Self[K, V]) -> Array[K]
  elems[K, V](Self[K, V]) -> Array[V]
}
```

[`hashmap`](https://github.com/moonbitlang/core/blob/main/immut/hashmap/hashmap.mbti)
```moonbit
type T[K, V]
impl T {
  elems[K, V](Self[K, V]) -> Iter[V]
  keys[K, V](Self[K, V]) -> Iter[K]
}
```

FlyCloudC:
Same applies to `sorted_map`

PR #2123, opened 2025-05-18
change return type of `@sorted_map.{keys, values}` to `Iter`

FlyCloudC:
see #2108

### DIFF
--- a/immut/sorted_map/README.mbt.md
+++ b/immut/sorted_map/README.mbt.md
@@ -24,7 +24,7 @@ Also, you can construct it from an array using `of()` or `from_array()`.
 test {
   let map = @sorted_map.of([("a", 1), ("b", 2), ("c", 3)])
   assert_eq(map.values().collect(), [1, 2, 3])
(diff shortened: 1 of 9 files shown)


peter-jerry-ye:
Same as #2124, this introduces breaking changes. It would be better to have some migration process for users.

FlyCloudC:
Thanks for the feedback. I’d like to ensure we handle the migration correctly. Since this is a return type change (not just a rename), could you clarify the preferred way to deprecate the old behavior while maintaining backward compatibility?

I noticed PR #1861 made a similar change directly—should we follow that precedent, or is there a different pattern you’d recommend for this case?

peter-jerry-ye:
Yeah, I admit that #1861 did it abruptly. If possible, I would like you to define a new method, probably something like `keys_as_iter`, and add deprecation to the `keys` and `values` to let people know about that this is actually happening.

FlyCloudC:
Waiting for #2124 to be merged first.

PR #2124, opened 2025-05-18
Rename `elems` to `values` in immut/sorted_map and hash_map

FlyCloudC:
+ Rename `elems()` to `values()` for consistency with `Map`.  
+ `(key, value)` pairs are called elements, but the second item is specifically called value
+ Other languages (e.g., Rust/JS/Python) also use `values()`

​​Note:​​
This PR is based on #2123 - If you want to merge this, please merge #2123 first

### DIFF
--- a/immut/hashmap/HAMT.mbt
+++ b/immut/hashmap/HAMT.mbt
@@ -348,8 +348,15 @@ pub fn[K, V] keys(self : T[K, V]) -> Iter[K] {
 
 ///|
 /// Returns all values of the map
-pub fn[K, V] elems(self : T[K, V]) -> Iter[V] {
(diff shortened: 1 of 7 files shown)


peter-jerry-ye (changes requested):
You are introducing a lot of breaking changes here.

Migrate them step by step: the old one should be kept and marked as deprecated.

peter-jerry-ye:
I think these two can be done in parallel if I understand correctly, or the other can depend on this PR since this one simply adds new functions (`values`). I think we can land this one first.

FlyCloudC:
> I think we can land this one first.

Done. It no longer depends on #2123.

PR #2125, opened 2025-05-18
ArrayView::join and FixedArray::join

ethereal-dream:
Adding ArrayView::join and FixedArray::join functions makes it convenient to concatenate strings within an array into a single complete string.

### DIFF
--- a/array/array.mbti
+++ b/array/array.mbti
@@ -64,6 +64,7 @@ impl FixedArray {
   stable_sort[T : Compare](Self[T]) -> Unit
   starts_with[T : Eq](Self[T], Self[T]) -> Bool
   swap[T](Self[T], Int, Int) -> Unit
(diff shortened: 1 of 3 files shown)


peter-jerry-ye (approved):

peter-jerry-ye:
@Yu-zh asked whether we should be using `View` instead of `String`.

I think maybe we can use `@string.View` for separator, but keep `T[String]` for this signature.

ethereal-dream:
When using @string.View for separator, what should I do with string.write_string(separator) since there doesn't seem to be a write_view function?

peter-jerry-ye:
I think you can simply call `to_string` on the view. The support for `view` on the string builder would come later as its currently blocked by how the core package is organize, which I assume would still need a few weeks.

PR #2136, opened 2025-05-19
perf(sorted_map/set): use tail call and reduce closure allocation in `iter`

FlyCloudC:
## bench

```moonbit
test (b : @bench.T) {
  let a : @sorted_set.T[Int] = @quickcheck.samples(1000).last().unwrap()
  fn old() {
    b.keep(a.iter().count())
  }

  fn new() {
    b.keep(a.iter_new().count())
  }

  b.bench(name="old", old)
  b.bench(name="new", new)
}
```

---

in `sorted_set`:

```
// bench moonbitlang/core/mytest/1.mbt::0
// name time (mean ± σ)         range (min … max)
// old   19.39 µs ±   1.62 µs    17.67 µs …  22.35 µs  in 10 ×   5262 runs
// new    9.70 µs ±   0.91 µs     8.48 µs …  11.02 µs  in 10 ×  10418 runs
// Total tests: 1, passed: 1, failed: 0. [wasm]
// bench moonbitlang/core/mytest/1.mbt::0
// name time (mean ± σ)         range (min … max)
// old   10.85 µs ±   0.61 µs     9.86 µs …  11.94 µs  in 10 ×  10485 runs
// new    5.26 µs ±   0.51 µs     4.32 µs …   6.02 µs  in 10 ×  21438 runs
// Total tests: 1, passed: 1, failed: 0. [js]
// bench moonbitlang/core/mytest/1.mbt::0
// name time (mean ± σ)         range (min … max)
// old    4.54 µs ±   0.21 µs     4.10 µs …   4.78 µs  in 10 ×  19873 runs
// new    2.01 µs ±   0.33 µs     1.59 µs …   2.54 µs  in 10 ×  64080 runs
// Total tests: 1, passed: 1, failed: 0. [wasm-gc]
```

---

in `immut/sorted_set`:

```
bench moonbitlang/core/mytest/1.mbt::0
name time (mean ± σ)         range (min … max)
old    8.75 µs ±   0.76 µs     7.61 µs …   9.88 µs  in 10 ×  12855 runs
new    2.62 µs ±   0.28 µs     2.31 µs …   3.09 µs  in 10 ×  32289 runs
Total tests: 1, passed: 1, failed: 0. [wasm-gc]
bench moonbitlang/core/mytest/1.mbt::0
name time (mean ± σ)         range (min … max)
old   17.24 µs ±   1.61 µs    14.46 µs …  18.94 µs  in 10 ×   7324 runs
new    5.70 µs ±   1.01 µs     4.49 µs …   7.48 µs  in 10 ×  18107 runs
Total tests: 1, passed: 1, failed: 0. [js]
bench moonbitlang/core/mytest/1.mbt::0
name time (mean ± σ)         range (min … max)
old   35.73 µs ±   4.30 µs    30.17 µs …  42.06 µs  in 10 ×   2590 runs
new   11.49 µs ±   0.76 µs    10.48 µs …  12.50 µs  in 10 ×   8856 runs
Total tests: 1, passed: 1, failed: 0. [wasm]
```

### DIFF
--- a/sorted_set/set.mbt
+++ b/sorted_set/set.mbt
@@ -433,22 +433,6 @@ pub fn iter[V](self : T[V]) -> Iter[V] {
 }
 
 ///|
-fn Node::iter[A](self : Node[A]) -> Iter[A] {
-  Iter::new(fn(yield_) {
-    if self.left is Some(l) {
-      guard l.iter().run(yield_) is IterContinue else { return IterEnd }
-
-    }
-    guard yield_(self.value) is IterContinue else { return IterEnd }
-    if self.right is Some(r) {
-      guard r.iter().run(yield_) is IterContinue else { return IterEnd }
-
-    }
-    IterContinue
-  })
-}
-
-///|
 pub fn from_iter[V : Compare](iter : Iter[V]) -> T[V] {
   let s = new()
   iter.each(fn(e) { s.add(e) })
@@ -458,10 +442,7 @@ pub fn from_iter[V : Compare](iter : Iter[V]) -> T[V] {
 ///|
 /// Converts the set to string.
 pub impl[V : Show] Show for T[V] with output(self, logger) {
-  match self.root {
-    None => logger.write_string("@sorted_set.of([])")
-    Some(node) => Show::output(node, logger)
-  }
+  logger.write_iter(self.iter(), prefix="@sorted_set.of([", suffix="])")
 }
 
 ///|
@@ -474,7 +455,8 @@ pub impl[X : @quickcheck.Arbitrary + Compare] @quickcheck.Arbitrary for T[X] wit
 
 ///|
 impl[T : Show] Show for Node[T] with output(self, logger) {
-  logger.write_iter(self.iter(), prefix="@sorted_set.of([", suffix="])")
+  let x = { root: Some(self), size: 0 } // Hack for test
+  logger.write_iter(x.iter())
 }
 
 ///|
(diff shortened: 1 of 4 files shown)


Yu-zh:
LGTM. Note that the local function `go` is not tail recursive though. This PR basically changes the iter process from recursively creating closures to creating a recursive closure.

sorted_set/set.mbt
```diff
@@ -797,29 +783,20 @@ test "join" {
...
-  inspect!(join(l.root, 6, r.root), content="@sorted_set.of([6])")
+  inspect!(join(l.root, 6, r.root), content="[6]")
   let l = from_array([])
   let r = from_array([7, 8, 9, 10, 11, 12])
-  inspect!(
-    join(l.root, 6, r.root),
-    content="@sorted_set.of([6, 7, 8, 9, 10, 11, 12])",
-  )
+  inspect!(join(l.root, 6, r.root), content="[6, 7, 8, 9, 10, 11, 12]")
 }
```
peter-jerry-ye, line 800: Why should the output gets changed?

Even though they are changed, I still think they shoulud be `@sorted_set.from_array([])` instead of just printing an array.

FlyCloudC: @sorted_set contains a `Node?` and a `size`, but `join` only returns a `Node`, so displaying it as `sorted_map([...])` here is inaccurate.

Maybe it should be changed to `Node([...])` or `Node::from_array([...])`?

peter-jerry-ye: The point I think is that it should represent how we actually write the MoonBit code. `Node([...])` is not MoonBit syntax, so it's not acceptable. `Node::from_array([...])` is fine but `Node` is not public.

FlyCloudC: But `Node` is not public, thus the `impl Show for Node` is not.

peter-jerry-ye: Oh, I see, it's the `Show` for `Node` that's changed. Sorry about that.

FlyCloudC:
It is indeed not tail-recursive. To be precise, the original second recursive call was moved to the tail position to eliminate the corresponding call stack frame. This is equivalent to code like:

```moonbit
    ...
    fn go(x : Node?) {
      loop x {
        None => IterContinue
        Some(n) =>
          if go(n.l) is IterEnd {
            IterEnd
          } else if yield_(n.x) is IterEnd {
            IterEnd
          } else {
            continue n.r
          }
      }
    }
    ...
    ```

**Comment by Yu-zh** [comment]
> To be precise, the original second recursive call was moved to the tail position to eliminate the corresponding call stack frame.

moonc itself does not apply such optimization. I guess LLVM backend will likely benefit from such change but that is backend specific.

**Comment by FlyCloudC** [comment]
After examining the generated code, I found that all four backends - wasm-gc, wasm, JS, and native - include this optimization in their output.

Test code:

```moonbit
struct Tree {
  x : Int
  l : Tree?
  r : Tree?
}

pub fn i1(x : Tree?) -> Iter[Int] {
  Iter::new(fn(yield_) {
    fn go(x : Tree?) {
      match x {
        None => IterContinue
        Some({ l, r, x }) =>
          if go(l) is IterEnd {
            IterEnd
          } else if yield_(x) is IterEnd {
            IterEnd
          } else {
            go(r)
          }
      }
    }

    go(x)
  })
}

fn main {
  ignore(i1(None))
}
```

FlyCloudC:
If this isn't standard behavior and requires us to convert it to a loop, then other changes would also be needed—such as the `dfs` in `each`.

Yu-zh:
> If this isn't standard behavior and requires us to convert it to a loop, then other changes would also be needed—such as the `dfs` in `each`.

My bad. It is indeed optimized.

PR #2139, opened 2025-05-20
refactor: char escaping

peter-jerry-ye:
This PR refactors the char escaping, by using `is_printable` to determine whether a char should be escaped or not.

Notice that the criteria of `is_printable` is different from language to language. Currently it is aligned to how Golang defines `is_printable`, as mentioned in the original doc comment.


peter-jerry-ye:
We would need another refactor in the future, by generating code based on the unicode database.

PR #2143, opened 2025-05-21
feat: add ArrayView::Iter2 and Bytes::Iter2

Erchiusx:
Part of the enhancement proposed in [this issue](https://github.com/moonbitlang/core/issues/1013)


peter-jerry-ye:
Based on the signature, I assume you are adding `iteri` instead of `iter2`.

peter-jerry-ye (approved):

PR #2146, opened 2025-05-22
fix FixedArray::get and corresponding doc test

myfreess:


peter-jerry-ye (changes requested):
Requires migration phase

peter-jerry-ye (approved):

PR #2145, opened 2025-05-21
Add set operations to `@immut/hash{map, set}` and `@internal/sparse_array` Summary

Asterless:
## Related Issue

Fixes [#1830: Efficient set operations for @immut/hash{map, set}](https://github.com/moonbitlang/core/issues/1830)

### `@immut/hashmap`
- **`union_with(f: (K, V, V) => V)`**  
  Merges two hashmaps, resolving key conflicts with a custom function `f`.
- **`intersection()`**  
  Returns a new hashmap containing keys present in both input maps, with values from the first map.
- **`intersection_with(f: (K, V, V) => V)`**  
  Computes intersection, resolving overlapping keys' values with function `f`.
- **`difference()`**  
  Returns entries present in the first map but not in the second.

### `@immut/hashset`
- **`intersection()`**  
  Returns a new set containing elements common to both input sets.
- **`difference()`**  
  Returns elements present in the first set but not in the second.

### `@internal/sparse_array`
- **`intersection()`**  
  Computes index-wise intersection of two sparse arrays.
- **`difference()`**  
  Computes index-wise difference between two sparse arrays.

## Motivation

These changes provide a more complete and consistent set of set operations for immutable collections, making it easier to perform common set algebra tasks and improving API parity across collection types.

## Tests
Added and updated unit tests for all new and modified methods to ensure correctness and expected behavior.

## Checklist
<input checked="" disabled="" type="checkbox"> All new and existing tests pass
<input checked="" disabled="" type="checkbox"> Code is formatted and documented where appropriate

PR #2147, opened 2025-05-22
Fix formatting of the code from [pull#2145](https://github.com/moonbitlang/core/pull/2145) 

Asterless:
Reformat the code from [pull#2145](https://github.com/moonbitlang/core/pull/2145)


peter-jerry-ye:
Please update directly in that branch otherwise that branch won't pass the CI tests.

PR #2151, opened 2025-05-22
remove usage of method as function

Guest0x0:
Previously, methods defined as `fn f(self : T, ..)` are considered **both** method and regular function. While this allows convenient ways of calling method via `@pkg.meth(..)`, it also complicates the semantic of MoonBit.

The latest release of MoonBit now supports partial dot application (`_.meth(..)`) and piping into dot application (`expr |> _.meth(..)`), which significantly reduce the need for calling methods as regular functions. So we decided to deprecate the behavior of calling methods as regular functions, to simplify the language.

This PR removes usage of calling method as regular function in core. In next week's release, we will start emitting warnings on such usage. We are also simplifying the syntax of MoonBit's method system, the simplified syntax will also be available next week.

For most cases, this PR just migrates regular function apply to dot apply. However, for math related methods (`Double::sin`, for example), people are more used to function call style `sin(..)` from math notation, compared of dot apply style `xx.sin()`. So an `fnalias` is provided for math related functions and several others, such as `Ref::swap`.

### DIFF
--- a/unit/README.mbt.md
+++ b/unit/README.mbt.md
@@ -23,7 +23,6 @@ Unit values can be converted to strings using either the standalone function or 
 test "unit string conversion" {
   let u = ()
   // Both ways produce the same result
-  inspect(@unit.to_string(u), content="()")
   inspect(u.to_string(), content="()")
 }
 ```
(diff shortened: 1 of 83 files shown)


unit/README.mbt.md
```diff
@@ -23,7 +23,7 @@ Unit values can be converted to strings using either the standalone function or
 test "unit string conversion" {
   let u = ()
   // Both ways produce the same result
-  inspect!(@unit.to_string(u), content="()")
+  inspect(u.to_string(), content="()")
```
FlyCloudC, line 26: missing `!`?

PR #2154, opened 2025-05-23
feat: introduce @list.sort_v2 to improve sort stack safety

Erchiusx:
Previous @list.sort would yield a stack overflow at a data scale of 100_00
introduced @list.sort_v2 which uses @array.FixedArray::sort to help sorting the list


peter-jerry-ye:
I think you can either:

- replace the existing implementation altogether
- try to find a boundary where the original is more efficient (which might not even exist) and keep both

Erchiusx:
> I think you can either:
> 
>     * replace the existing implementation altogether
> 
>     * try to find a boundary where the original is more efficient (which might not even exist) and keep both

The original implementation has a 3-constant factor, I think replacing it would count

Lampese (approved):
LGTM

bobzhang:
> The test seems a bit oversized, which may slow down the standard library's test suite. If it's for correctness, 100,000 might be overkill, and if it's only for speed measurements, maybe we can delete it after?

sorting 100_000 elements should be fine, we may add some attributes to allow test filtering mechanism in the future

Erchiusx:
100_000 is not a must, 100_00 is enough to show the difference

perhaps I should change it to 100_00?

Erchiusx:
The rebase is done, but seems still squashable

Lampese:
Later on, I will establish a core extra test with interns from @plct-lab to conduct some stress tests on the core. The testing in the core can be relatively reduced

PR #2159, opened 2025-05-23
fix: move type arguments after fn keyword in abort

tonyfettes:
/cc @Young-Flash `moon fmt` omitted the package.


Young-Flash:
fix by https://github.com/moonbitlang/moon/pull/827

PR #2183, opened 2025-05-29
refactor: update Array::each to return Unit?Error for improved error handling

bobzhang:


bobzhang:
cc @gmlewis now we have some support for error callbacks, feedback is welcome

gmlewis:
> cc @gmlewis now we have some support for error callbacks, feedback is welcome

Very nice, @bobzhang! Should the same feature be added to `eachi`?

PR #2184, opened 2025-05-30
impl ToJson for Error

Guest0x0:
the latest release of MoonBit introduces primitive for converting `Error` to json. This PR introduces this primitive to core


bobzhang:
note despite `impl ToJson for Error`, for suberror
 we can not use `@json.inspect` to inspect the error
```moonbit
 @json.inspect(ErrWithoutToJson(42)) // type error
```
This also applies to `impl Show for Error`, for suberror
 we can not use `inspect` to inspect the error
 we may auto-derive `ToJson` and `Show` for suberror automatically
 to enhance the user experience

PR #2191, opened 2025-05-31
Add truncate to array and join/drain/flatten to deque

ethereal-dream:


array/array.mbt
```diff
@@ -291,6 +291,43 @@ pub fn[A, B] zip_to_iter2(self : Array[A], other : Array[B]) -> Iter2[A, B] {
...
+///
+/// ```moonbit
+/// test "truncate" {
+///   let arr = [1, 2, 3, 4, 5]
+///   arr.truncate(3)
+///   inspect(arr, content="[1, 2, 3]")
+/// }
+/// ```
+pub fn[A] truncate(self : Array[A], len : Int) -> Unit {
+  guard len >= 0 && len < self.length() else { return }
```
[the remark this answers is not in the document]
ethereal-dream: I've updated the implementation to be a proper method on Array, using the signature `pub fn[A] Array::truncate`

array/array.mbt
```diff
@@ -291,6 +291,43 @@ pub fn[A, B] zip_to_iter2(self : Array[A], other : Array[B]) -> Iter2[A, B] {
...
+  if len == 0 {
+    self.clear()
+    return
+  }
+  let length = self.length() - len
+  for i = 0; i < length; i = i + 1 {
+    ignore(self.pop_exn())
+  }
```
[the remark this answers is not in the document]
ethereal-dream: how about creates an array view of the first len elements and then replaces the original array with the truncated version.

deque/deque.mbt
```diff
@@ -1243,6 +1243,141 @@ pub fn[A] to_array(self : T[A]) -> Array[A] {
...
+///   let deque = @deque.of(["a","b","c"])
+///   let s1 = deque.join("")
+///   inspect(s1, content="abc")
+/// 
+///   let s2 = deque.join(",")
+///   inspect(s2, content="a,b,c")
+/// }
+/// ```
+pub fn join(self : T[String], separator : @string.View) -> String {
+  let len = self.length()
```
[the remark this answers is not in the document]
ethereal-dream: I've updated the code to use deque.iter().join(",")

ethereal-dream:
@bobzhang,I just wanted to check if there are any unresolved issues with this PR.

array/array.mbt
```diff
@@ -328,6 +328,43 @@ pub fn[A, B] zip_to_iter2(self : Array[A], other : Array[B]) -> Iter2[A, B] {
...
+pub fn[A] Array::truncate(self : Array[A], len : Int) -> Unit {
+  guard len >= 0 && len < self.length() else { return }
+  if len == 0 {
+    self.clear()
+    return
+  }
+  let a = self[0:len]
+  let array = a.to_array()
+  self.clear()
+  self.append(array)
```
peter-jerry-ye: The implementation here looks weird to me. Shouldn't you use the `unsafe_truncate_to_length`?

deque/deque.mbt
```diff
@@ -1298,18 +1298,127 @@ pub fn[A] to_array(self : T[A]) -> Array[A] {
...
+/// test "join" {
+///   let deque = @deque.of(["a","b","c"])
+///   let s1 = deque.join("")
+///   inspect(s1, content="abc")
+/// 
+///   let s2 = deque.join(",")
+///   inspect(s2, content="a,b,c")
+/// }
+/// ```
+pub fn T::join(self : T[String], separator : String) -> String {
```
peter-jerry-ye: Should be accepting `@string.View` for the separator.

deque/deque.mbt
```diff
@@ -1298,18 +1298,127 @@ pub fn[A] to_array(self : T[A]) -> Array[A] {
...
+///   let deque = @deque.of([1,2,3,4,5,6,7,8,9])
+///   let deque_test = deque.drain(2, 6)
+///   inspect(deque_test, content="@deque.of([3, 4, 5, 6])")
+///   inspect(deque, content="@deque.of([1, 2, 7, 8, 9])")
+/// }
+/// ```
+pub fn[A] T::drain(self : T[A], begin : Int, end : Int) -> T[A] {
+  guard begin >= 0 && end <= self.length() && begin <= end
+  let deque = self.copy()
+  deque.clear()
```
peter-jerry-ye: Why not reserve capacity?

deque/deque.mbt
```diff
@@ -1298,18 +1298,127 @@ pub fn[A] to_array(self : T[A]) -> Array[A] {
...
+  let len = self.length()
+  let mut selflen = 0
+  for i = 1; i < len; i = i + 1 {
+    selflen = self[i].length()
+    for j = 0; j < selflen; j = j + 1 {
+      match self[i].pop_front() {
+        Some(pop) => deque.push_back(pop)
+        None => ()
+      }
+    }
```
peter-jerry-ye: The implementation is too weird for me. You should be creating an `UninstailizedArray` that is large enough for all of the deques (computing the sizes first), and then doing the blit since each deque can be divided into two parts where each part is ordered.

deque/deque.mbt
```diff
@@ -1298,18 +1298,127 @@ pub fn[A] to_array(self : T[A]) -> Array[A] {
...
+        if i > begin - 1 && i < end {
+          deque.push_back(pop)
+        } else {
+          self.push_back(pop)
+        }
+      None => ()
+    }
+  }
+  deque
+}
```
peter-jerry-ye: Should be implemented using blit.

peter-jerry-ye (changes requested):

deque/deque.mbt
ethereal-dream: The `join` method inside `self.iter().join(separator)` uses a `String` parameter. Is it necessary to use `@string.view` for the outer function parameter? Wouldn't that introduce an extra unnecessary conversion step?

deque/deque.mbt
ethereal-dream: I wasn't able to locate the `blit` you mentioned. Could you point me to the right place or provide more context? Array type does have a `blit` method — are you suggesting converting to an Array, using `blit`, and then converting back?

peter-jerry-ye: https://github.com/moonbitlang/core/blob/f0e81fc2543a1c77bef8d01f5f7c53955c9d8018/builtin/uninitialized_array.mbt#L101

deque/deque.mbt
peter-jerry-ye: Yes, but I think it's acceptable since you are only converting it once.

builtin/arraycore_nonjs.mbt
peter-jerry-ye, line 1: This should be put inside the `array.mbt` instead of `arraycore_nonjs.mbt`, because otherwise, it would not be compiled for the JS backend.

array/array.mbt
```diff
@@ -310,6 +310,7 @@ pub fn[A, B] zip_to_iter2(self : Array[A], other : Array[B]) -> Iter2[A, B] {
   })
 }
 
+
```
peter-jerry-ye: Maybe you would need to rerun the `moon fmt`.

deque/deque.mbt
```diff
@@ -1219,18 +1219,131 @@ pub fn[A] to_array(self : T[A]) -> Array[A] {
...
+/// Example:
+///
+/// ```moonbit
+/// test "drain" {
+///   let deque = @deque.of([1,2,3,4,5,6,7,8,9])
+///   let deque_test = deque.drain(2, 6)
+///   inspect(deque_test, content="@deque.of([3, 4, 5, 6])")
+///   inspect(deque, content="@deque.of([1, 2, 7, 8, 9])")
+/// }
+/// ```
```
peter-jerry-ye:
```suggestion
///   let deque = @deque.of([1,2,3,4,5,6,7,8,9])
///   let deque_test = deque.drain(2, 6)
///   inspect(deque_test, content="@deque.of([3, 4, 5, 6])")
///   inspect(deque, content="@deque.of([1, 2, 7, 8, 9])")
/// }
```

For the new convention

deque/deque.mbt
```diff
@@ -1219,18 +1219,131 @@ pub fn[A] to_array(self : T[A]) -> Array[A] {
...
 ///
 /// Example:
 ///
 /// ```moonbit
+/// test "flatten" {
+///   let deque = @deque.of([@deque.of([1,2,3]),@deque.of([4,5,6]),@deque.of([7,8])])
+///   let deque_test = deque.flatten()
+///   inspect(deque_test, content="@deque.of([1, 2, 3, 4, 5, 6, 7, 8])")
+/// }
+/// ```
```
peter-jerry-ye:
```suggestion
///   let deque = @deque.of([@deque.of([1,2,3]),@deque.of([4,5,6]),@deque.of([7,8])])
///   let deque_test = deque.flatten()
///   inspect(deque_test, content="@deque.of([1, 2, 3, 4, 5, 6, 7, 8])")
/// }
```

Same as above

peter-jerry-ye (approved):
LGTM. Thank you.

deque/deque.mbt
```diff
@@ -1236,18 +1236,122 @@ pub fn[A] to_array(self : T[A]) -> Array[A] {
...
+///
+/// Example:
+///
+/// ```moonbit
+///   let deque = @deque.of([1,2,3,4,5,6,7,8,9])
+///   let deque_test = deque.drain(2, 6)
+///   inspect(deque_test, content="@deque.of([3, 4, 5, 6])")
+///   inspect(deque, content="@deque.of([1, 2, 7, 8, 9])")
+/// ```
+pub fn[A] T::drain(self : T[A], begin : Int, end : Int) -> T[A] {
```
peter-jerry-ye:
```suggestion
pub fn[A] T::drain(self : T[A], start~ : Int, len~ : Int) -> T[A] {
```

peter-jerry-ye:
I just need more tests and verify the implementation of `blit`

ethereal-dream:
You suggested changing the parameter names from `begin`/`end` to `start`/`len`, so I updated the corresponding parameter names in the function.

peter-jerry-ye (approved):

peter-jerry-ye:
Thank you.

PR #2204, opened 2025-06-04
refactor: remove function not

Yoorkin:
We will introduce the `not expr` syntax to replace the `not()` function


Yoorkin:
Closed as Not Planned.

PR #2206, opened 2025-06-05
fix: disable bigint js backend

peter-jerry-ye:
Currently, `bigint` package has different implementation for JS backend, which uses `BigInt`, and other backends, which we implement on our own.

This difference has introduced some backend specific import, which caused 'unused import warning' when users are installing the toolchain, which is misleading and frustrating.

However, the moon build system currently doesn't support importing packages based on backend. Thus we disable this warning for now.


hackwaly:
This is not ideal: it results potential real unused package warnings be suppressed. 

I suggest that: when `--target all`, let the `moon` filter unused package warnings, and aggregate them to only report those packages which are marked unused under every targets.

PR #2208, opened 2025-06-05
remove use of old type alias syntax

Guest0x0:
We intend to deprecate the `typealias XX = YY` and `traitalias XX = YY` syntax, to make the language more uniform, and the various forms of alias (`typealias`/`traitalias`/`fnalias`) more consistent. This PR migrates core and remove usage of the old syntax.

Deprecating type alias is currently unsupported, however due to a compiler bug, the unused attribute warning is currently not reported for `typealias XX = YY`. But the warning will be correctly emitted for `typealias YY as XX`, so this PR need to temporarily comment out the deprecation attributes on type alias.

### DIFF
--- a/buffer/buffer.mbt
+++ b/buffer/buffer.mbt
@@ -13,8 +13,8 @@
 // limitations under the License.
 
 ///| Extensible buffer.
-#deprecated("Use type `T` instead")
-pub(all) typealias Buffer = T
+// #deprecated("Use type `T` instead")
+pub(all) typealias T as Buffer
 
 ///|
 /// Extensible buffer.
--- a/json/types.mbt
+++ b/json/types.mbt
@@ -13,8 +13,8 @@
 // limitations under the License.
 
 ///|
-#deprecated("Definition of json is moved to builtin package, use `Json` instead")
-pub(all) typealias JsonValue = Json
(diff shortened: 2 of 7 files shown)


buffer/buffer.mbt
```diff
@@ -13,8 +13,8 @@
 // limitations under the License.
 
 ///| Extensible buffer.
-#deprecated("Use type `T` instead")
-pub(all) typealias Buffer = T
+// #deprecated("Use type `T` instead")
+pub(all) typealias T as Buffer
```
bobzhang, line 18: this has been deprecated long time ago, maybe we can just remove it

json/types.mbt
```diff
@@ -13,8 +13,8 @@
 // limitations under the License.
 
 ///|
-#deprecated("Definition of json is moved to builtin package, use `Json` instead")
-pub(all) typealias JsonValue = Json
+// #deprecated("Definition of json is moved to builtin package, use `Json` instead")
+pub(all) typealias Json as JsonValue
```
bobzhang: same

bobzhang (approved):

buffer/buffer.mbt
Guest0x0: the problem is, the `deprecated` attribute is not taking effect. So users have never received the deprecation warning

PR #2209, opened 2025-06-05
Add tests for JSON parsing and nested JSON structures

bobzhang:


builtin/autoloc.mbt
```diff
@@ -36,7 +37,7 @@ pub(all) extern type SourceLoc
 /// Note: This function is primarily used internally by the compiler for error
 /// reporting and debugging purposes. Source locations are automatically created
 /// by the compiler when needed.
-pub fn SourceLoc::to_string(self : SourceLoc) -> String = "%loc_to_string"
+pub fn SourceLoc::to_string(self : Self) -> String = "%loc_to_string"
```
bobzhang, line 41: cc @Guest0x0 , we may need `SourceLoc::dummy` so that `fail(msg, loc=SourceLoc::dummy())` could work in some edge cases

Issue #1890, opened 2025-04-07
[Feature Request] add Map::map and Map::mapi


FlyCloudC:

peter-jerry-ye:
Looks like this can be simply replaced with:

```moonbit
a.iter().map(f) |> Map::from_iter()
a.iter().mapi(f) |> Map::from_iter()
```

FlyCloudC:
> Looks like this can be simply replaced with:
> 
> a.iter().map(f) |> Map::from_iter()
> a.iter().mapi(f) |> Map::from_iter()

The map operation does not alter the structure of the Map. Implementing it natively could allow mapping directly over the `entries: FixedArray[Entry[K, V]?]` field, avoiding redundant hash computations and collision resolution.

peter-jerry-ye:
Would you mind providing some usecases and your expected API?

FlyCloudC:
API:
```mbt
pub fn[K, V, U] Map::map(self : Map[K, V], f : (V) -> U) -> Map[K, U]
pub fn[K, V, U] Map::mapi(self : Map[K, V], f : (Int, V) -> U) -> Map[K, U]
```
usecase:
K = String (book name)
V = BookData (a struct)
U = AuthorData

I couldn't think of any real use cases for `mapi`—it was initially added just to maintain consistency with `Array`. Now I realize I was mistaken—the proper counterpart for `Map` should be `map_with_key`.
```mbt
pub fn[K, V, U] Map::map_with_key(self : Map[K, V], f : (K, V) -> U) -> Map[K, U]
```

peter-jerry-ye:
Any reason for having `map` and `map_with_key` as two separate APIs? I was thinking that 
- `map_with_key` covers the use case of `map` and they both can provide somewhat good performance (e.g. they have a 1:1 on data structures), and could be called as `map`.
- would anyone need a more general one where the map function would be `(K, V) -> (A, B)` (though it can be replaced as `map.iter().map() |> Map::from_iter`)

PR #2205, opened 2025-06-04
Refactor immut/{set, map, sparse_array}

FlyCloudC:
## Overview

I'm planning to refactor the immut/{set, map, sparse_array} modules. Due to the extensive changes, I've temporarily placed the new implementation in a separate repository: https://github.com/FlyCloudC/HAMT

## Bug Fixes
1. Singleton Leaf Issue​​
    + Fixed incorrect top-level Leaf creation in singleton, which previously forced subsequent add operations to degenerate into Collision nodes (causing O(n) performance).
2. Difference Operation​​
    + Resolved bugs in set difference (originally reported in #2145).

## Improvement
1. Path Compression​​
    + Introduced `Flat[K, Path]` to eliminate redundant single-leaf Branch nesting, reducing depth and improving lookup/insertion.
2. Bucket Consolidation​​
    + Replaced Bucket with `@list` to deduplicate method implementations.
    + Removed the previous optimization for single-element leaves (`Leaf[K]`). Now using `Leaf(K, List[K])` to represent non-empty lists. This adds an extra `@list.Empty` for each bottom-level node (depth=5), but since most leaves will be represented as `Flat` (without hash collisions), the overhead is acceptable
3. Internal `inplace`
    + Added internal `add_inplace` to reduce memory usage during `from_array` and `from_iter`

## Comapre
brfore:
```mbt
///|
/// A bucket is a non-empty linked list of key-value pair,
/// used to resolve hash collision in HAMT
priv enum Bucket[T] {
  JustOne(T) // must be non-empty
  More(T, Bucket[T])
}

///|
/// An immutable hash set data structure
enum T[A] {
  Empty
  Leaf(A) // optimize for the case of no collision
  Collision(Bucket[A]) // use a list of buckets to resolve collision
  Branch(@sparse_array.SparseArray[T[A]])
}
```

after:
```mbt
///|
/// Represents a non-empty set
/// 
/// Invariance:
/// - If it can be `Flat`, it must be `Flat`
/// - `Branch` nesting depth <= 5
priv enum Node[A] {
  Flat(A, Path)
  Leaf(A, @list.T[A])
  Branch(@sparse_array.T[Node[A]])
} derive(ToJson)

///|
type T[A] Node[A]? derive(Eq)
```

## Pending Work
+ Implementation of difference operations
+ immut/map module

## bench

```mbt
fn go(b : @bench.T, size : Int) -> Unit {
  let rs = @quickcheck/splitmix.new()
[...]

PR #2210, opened 2025-06-05
Add `map` and `map_with_key` Methods to `Map`

Asterless:
## Summary

Based on #1890
This PR adds two new methods to `Map` in `builtin/linked_hash_map.mbt`:

- **`map(f: (V) -> V2) -> Map[K, V2]`**  
  Applies a function to each value in the map and returns a new map with the same keys and transformed values.

- **`map_with_key(f: (K, V) -> V2) -> Map[K, V2]`**  
  Applies a function to each key-value pair and returns a new map with the same keys and transformed values.

## Implementation Details

- Both methods preserve the insertion order of keys.
- `map` is implemented by calling `map_with_key` and ignoring the key.
- If the map is empty, a new empty map is returned.
- The new map is pre-allocated with the same size as the original for efficiency.

## Example Usage

```moonbit
let m = {1: 10, 2: 20}
let m2 = m.map(fn(v) { v * 2 }) // {1: 20, 2: 40}
let m3 = m.map_with_key(fn(k, v) { k + v }) // {1: 11, 2: 22}
```

## Impact

- Backward compatible; does not affect existing APIs or behavior.
- Only adds new functionality to `Map`.


FlyCloudC:
There are many hash calculations that can be avoided here.

builtin/builtin.mbti
```diff
@@ -257,6 +257,8 @@ fn[K : Hash + Eq, V] Map::contains(Self[K, V], K) -> Bool
 fn[K : Hash + Eq, V : Eq] Map::contains_kv(Self[K, V], K, V) -> Bool
 fn[K, V] Map::each(Self[K, V], (K, V) -> Unit?Error) -> Unit?Error
 fn[K, V] Map::eachi(Self[K, V], (Int, K, V) -> Unit?Error) -> Unit?Error
+fn[K : Hash + Eq, V] Map::filter(Self[K, V], (K, V) -> Bool?Error) -> Self[K, V]?Error
+fn[K : Hash + Eq, V, A] Map::fold(Self[K, V], A, (A, K, V) -> A?Error) -> A?Error
```
peter-jerry-ye: Please focus on the functionalities that you described in the introduction; do not add too many stuff in one PR.

immut/hashmap/HAMT_test.mbt
```diff
@@ -673,6 +673,99 @@ test "HAMT::difference with branch" {
...
+test "HAMT::difference with branch" {
+  let m1 = @hashmap.of([(1, 1), (2, 2), (3, 3)])
+  let m2 = @hashmap.of([(2, 2), (3, 30), (4, 4)])
+  let diff = m1.difference(m2)
+  assert_eq!(diff.get(1), Some(1))
+  assert_eq!(diff.get(2), None)
+  assert_eq!(diff.get(3), None)
+  assert_eq!(diff.get(4), None)
+}
+
```
peter-jerry-ye, line 768: I don't understand the diff here.

builtin/builtin.mbti
Asterless: ok👍

immut/hashmap/HAMT_test.mbt
Asterless: The `diff` here refers to the result of the `difference` method between two hash maps, for example: `let diff = m1.difference(m2)`.

This produces a new map containing all key-value pairs from `m1` that do **not** have an identical key-value pair in `m2`.

FlyCloudC: 1. The tests for immut/map probably shouldn't be included in this PR about Map.  

2. The `difference` function in immut/map has a bug in cases where the tree is deep, but this test doesn't expose it. A week ago, I mentioned a failing test in a comment on PR #2145. I plan to fix it in PR #2205, which requires significant changes to `difference` to implement path compression.

Asterless: Oh no! It's my fault. I might have mistakenly established this branch. I will delete the test in this part.
Thank you for your hint🫡

peter-jerry-ye:
There are too many unrelated modifications. I'm closing this PR for the new one.

peter-jerry-ye:
Continue in #2228 #2229

PR #2219, opened 2025-06-06
feat(`@list.find_index`):  align `Array::find_index`

illusory0x0:


bobzhang:
cc @illusory0x0 added some tweaks for more efficiency. The other tweak is that we would like the method just to be used in one way: either function or method

PR #2224, opened 2025-06-07
refactor(`@list`): migrate to the new syntax, 

illusory0x0:
and add method alias to keep the behavior.

https://docs.moonbitlang.com/en/latest/language/methods.html#method-system

### DIFF
--- a/list/list.mbt
+++ b/list/list.mbt
@@ -14,29 +14,29 @@
 
 ///|
 /// Creates an empty list
-pub fn[A] new() -> T[A] {
+pub fn[A] T::new() -> T[A] {
   Empty
 }
 
 ///|
 /// Creates an empty list
-pub fn[A] empty() -> T[A] {
+pub fn[A] T::empty() -> T[A] {
   Empty
 }
 
 ///|
 /// Prepend an element to the list and create a new list.
-pub fn[A] construct(head : A, tail : T[A]) -> T[A] {
+pub fn[A] T::construct(head : A, tail : T[A]) -> T[A] {
   More(head, tail~)
 }
 
 ///|
-pub fn[A] prepend(self : T[A], head : A) -> T[A] {
+pub fn[A] T::prepend(self : T[A], head : A) -> T[A] {
   More(head, tail=self)
 }
 
 ///|
-pub fn[A] add(self : T[A], head : A) -> T[A] {
+pub fn[A] T::add(self : T[A], head : A) -> T[A] {
   More(head, tail=self)
 }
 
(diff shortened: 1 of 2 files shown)


illusory0x0:
cc @bobzhang Do we need to migrate the remaining functions?

FlyCloudC:
Could we use `f(...)` instead of `T::f(...)` for functions that don't take `self` as their first parameter? See #1472.

illusory0x0:
> Could we use `f(...)` instead of `T::f(...)` for functions that don't take `self` as their first parameter? See #1472.

This https://github.com/moonbitlang/core/pull/1472 proposal seems to have stale. We need to have new API design guidelines.

> Proposed new semantic
  methods can be declared with fn f(self : T, ..) or fn T::f(..)

illusory0x0:
@Guest0x0  @FlyCloudC 

> Could we use `f(...)` instead of `T::f(...)` for functions that don't take `self` as their first parameter? See #1472.

我认为总是定义方法会比较好，这是最新的API 设计指南， 但是并没有明确指出，当第一个参数不是 `self` 的时候，直接使用 `fn f(...)` 来定义函数.

> 4、“将方法当作普通函数使用” 的行为将被废弃

> 之前，使用 fn meth(self : T) 形式定义的方法既是方法也是函数，可以用 meth(..) 或 @pkg.meth(..) 的形式直接调用。这一 “将方法当作普通函数使用” 的行为将被废弃，目前编译器会对这一行为的使用提出警告。这一调整后，我们鼓励的 API 设计方式是：

> 永远使用 fn T::meth(..) 的形式定义方法，新代码中不再使用 fn meth(self : T)（在以后，fn meth(self : T) 语法本身也可能被废弃）

> 和某个类型绑定的 API 若无特殊理由，都鼓励设计成方法.

https://mp.weixin.qq.com/s/yUlMHxi4XmgffUZw9DT8dQ


即使第一个参数不是 `self`,  无法使用 `dot syntax`, 定义成方法仍然有好处， 用户不需要思考是使用 `@pkg.fun` 还是 `@pkg.T::meth`.

因为 `dot syntax` 是需要 `type directed resolution` , 当类型推导无法推导出结果的时候，是需要加类型注解，或者直接使用 `@pkg.fun` 或者 `@pkg.T::meth` 消除歧义.

FlyCloudC:
My question is, should these functions—like `new`, `from_iter`, `from_array`—be methods? I haven't seen these three usages in other packages of the core (except for `buildin`, which is to eliminate ambiguity). If so, then other packages would also need similar migration.

illusory0x0:
> My question is, should these functions—like `new`, `from_iter`, `from_array`—be methods? I haven't seen these three usages in other packages of the core (except for `buildin`, which is to eliminate ambiguity). If so, then other packages would also need similar migration.

I remove the method definition, e.g. `new`, `of`, `from_array`, look more details in  [list.mbti](https://github.com/moonbitlang/core/pull/2224/files#diff-12a42ef9c1970c5ec6ab01844c78c52ec3bbd0e1964f1c7c8371de1a30dc007eL38).


If we change the API design guild, then we can create a new PR to add new API usages.

PR #2226, opened 2025-06-08
feat(array): add unzip and tests  (closes #2187)

### LINKED ISSUE #2187: Add Array::unzip
There's Array::zip, but no Array::unzip. I think `map_unzip`, `map_unzip3`, ... might be useful than `unzip`.

YungjunCheng:
### Summary
Based on #2187  
#### This PR adds three new methods to Array in array.mbt:

- unzip: Splits an array of tuples into two arrays of the tuple’s components.
- map_unzip: Applies a function to each element and splits the result into two arrays.
- map_unzip3: Applies a function to each element and splits the result into three arrays.

#### Implementation Details

- All methods are generic and support tuple elements of any types.
- All new APIs are covered by comprehensive unit tests.
- No existing functionality is changed.

#### Example Usage

let arr = [(1, "a"), (2, "b")]
let (nums, strs) = arr.unzip()          // [1, 2], ["a", "b"]

let arr2 = [1, 2, 3]
let (x, y) = arr2.map_unzip(fn(v) { (v, v * 10) }) // [1,2,3], [10,20,30]

#### Impact

- Fully backward compatible.
- Only adds new API.
- All existing and new tests pass.

#### Closes #2187

### DIFF
--- a/array/array.mbt
+++ b/array/array.mbt
@@ -223,6 +223,83 @@ pub fn[A, B] zip(self : Array[A], other : Array[B]) -> Array[(A, B)] {
   }
 }
 
+///|
+/// Splits an array of pairs into two arrays, separating the first and second elements.
+///
+/// # Example
+/// ```moonbit
+/// test "Array::unzip" {
+///   let arr = [(1, "a"), (2, "b"), (3, "c")]
+///   let (nums, strs) = arr.unzip()
+///   inspect(nums, content="[1, 2, 3]")
+///   inspect(strs, content="[\"a\", \"b\", \"c\"]")
+/// }
+/// ```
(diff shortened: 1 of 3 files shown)


peter-jerry-ye:
I think having `unzip` might be fine, but `map_unzip` or `map_unzip3` are not proper for the core library. We don't want the core library to be filled with all kinds of utility functions.

YungjunCheng:
Makes sense, thanks for pointing it out. I added `map_unzip` and `map_unzip3` because #2187 mentioned them, but I understand your concern about core library size. Now I got it.I’ll take out `map_unzip` and `map_unzip3` and keep just `unzip`.

peter-jerry-ye (approved):
Seems the API is consistent with `@list.unzip`

PR #2228, opened 2025-06-09
feat(Map): add `map` and `copy` for `Map`

### LINKED ISSUE #1890: [Feature Request] add Map::map and Map::mapi

peter-jerry-ye:
Replaces #2210 
Closes #1890


peter-jerry-ye:
We will add `copy`: (Self, f?:(K, V) -> V). The benefit is that we can have a fast path when there's no mapping function, and it can perform all three functionalities.

We will merge `map_with_key` and `map` to have `(Self, f: (K, V) -> V2)` to simplify the APIs.

FlyCloudC:
Could we modify the trait bounds from `fn[K: Hash + Eq, V]` to `fn[K, V]`?

bobzhang:
> Could we modify the trait bounds from `fn[K: Hash + Eq, V]` to `fn[K, V]`?

done

PR #2229, opened 2025-06-09
feat(Map): add `filter` `fold` `partition`

peter-jerry-ye:
Replaces #2210


peter-jerry-ye:
We are closing this PR for now as this does not provide any benefit (efficiency) compared with using an external implementation. We have limited bandwidth so we are not adding this for now. We'll come back to it when we have more time.

PR #2231, opened 2025-06-09
refactor: use match instead of matrix functions

Yu-zh:


illusory0x0:
@Yu-zh  Why use `match` instead of `matrix functions`?

bobzhang:
@illusory0x0 we are planning to provide even shorter sugar for one liner, e.g, `a.map({x => x + 1})`, for large functions, we stick to verbose style since saving some chars won't help that much

PR #2241, opened 2025-06-10
clean @math

FlyCloudC:
1. Change the forwarding function to fnalias.  
    An added benefit: Some functions in `@math` now automatically include documentation (from `@double`).
2. Remove duplicate tests from `@double`.  
3. Since `@double` contains more precise trigonometric tests, remove the double-related trigonometric tests from `@math`.


peter-jerry-ye:
We will be migrating these complicated computations from `double` to `math`.

PR #2242, opened 2025-06-11
remove usage of complex default value

Guest0x0:
Currently, MoonBit allows default value of optional arguments to depend on previous parameters. However, this feature is in conflict with the virtual package feature, as we cannot tell the dependency of a default value from its `.mbti` signature. Meanwhile, the other kind of optional argument, `label? : T`, is expressive enough to cover all usage of complex default value (just pattern match `label : T?` in function body), and with MoonBit's unboxing optimization on option type, `label? : T` has little performance overhead. So we decided to deprecate complex default values that depends on previous parameters, this PR migrates core.

Note that simple cases where the default value does not depend on anything are still supported for convenience.


Guest0x0:
close via #2258

PR #2260, opened 2025-06-12
Add `blit_to()` to `@ArrayView`

Asterless:
### Summary
* Fixes [[Feature Request] add `ArrayView::bilt_to` #1958](https://github.com/moonbitlang/core/issues/1958)

### New Method Addition:
* [`array/view.mbt`](diffhunk://#diff-9788581ec781c0dbc9008639a013db6d990e29815f67170e75a42bd0b70e2fecR509-R522): Implemented the `blit_to` method with logic for iterating over the source `ArrayView` and copying its elements to the destination array.

### Testing Enhancements:
* [`array/view_test.mbt`](diffhunk://#diff-6fd45bc570c5817e4192acf323221b34be9d010946ba2526d690c1c8e18e730eR221-R260): Added multiple test cases for the `blit_to` method, covering scenarios such as basic usage, copying at the start or end of the destination array, handling empty source arrays, and full overwrites of the destination array.


array/view.mbt
```diff
@@ -494,3 +494,17 @@ pub impl[A : @quickcheck.Arbitrary] @quickcheck.Arbitrary for View[A] with arbit
 ) {
   Array::arbitrary(size, rs)[:]
 }
+
+///|
+/// Copies elements from this ArrayView to a destination array starting at a given index.
+pub fn[T] ArrayView::blit_to(
+  self : ArrayView[T],
+  dest : Array[T],
```
peter-jerry-ye: Any reason the dest is `Array`?

array/view.mbt
```diff
@@ -494,3 +494,17 @@ pub impl[A : @quickcheck.Arbitrary] @quickcheck.Arbitrary for View[A] with arbit
...
+  dest_index~ : Int = 0
```
peter-jerry-ye: `dest_index` is not the convention

array/view.mbt
```diff
@@ -494,3 +494,17 @@ pub impl[A : @quickcheck.Arbitrary] @quickcheck.Arbitrary for View[A] with arbit
...
+) -> Unit {
+  let mut i = 0
+  for v in self.iter() {
+    dest[dest_index + i] = v
+    i = i + 1
```
peter-jerry-ye: Not efficient at all. Use `Array::blit_to` at least.

peter-jerry-ye (dismissed):

array/view.mbt
Asterless: `Array` is mutable, so I used it. Hmm, If I don't use `Array`, what type should I use?

peter-jerry-ye: `ArrayView` is also mutable

Asterless: Ok. I have been using `ArrayView` now

array/view.mbt
```diff
@@ -494,3 +494,19 @@ pub impl[A : @quickcheck.Arbitrary] @quickcheck.Arbitrary for View[A] with arbit
...
+/// Copies elements from this ArrayView to a destination array starting at a given index.
+pub fn[T] ArrayView::blit_to(
+  self : ArrayView[T],
+  dst : ArrayView[T],
+  len~ : Int,
+  src_offset~ : Int = 0,
+  dst_offset~ : Int = 0
+) -> Unit {
+  let dst_arr = dst.to_array()
+  Array::blit_to(self.to_array(), dst_arr, len~, src_offset~, dst_offset~)
```
peter-jerry-ye: You should move it to `builtin/arrayview.mbt` so that you can get the underlying array without using `to_array()`. Also this would be a different semantic.

Asterless: I have moved this method to `bulitin/arrayview.mbt` and made corresponding changes👌

builtin/arrayview.mbt
```diff
@@ -260,3 +260,30 @@ pub fn[T] ArrayView::op_as_view(
...
+    } else {
+      len
+    }
+  } else if dst_len < len {
+    dst_len
+  } else {
+    len
+  }
+  for i in 0..<copy_len {
+    dst.op_set(dst_offset + i, self.op_get(src_offset + i))
```
peter-jerry-ye: Are you aware there's `Array::blit_to`?

builtin/arrayview.mbt
```diff
@@ -260,3 +260,30 @@ pub fn[T] ArrayView::op_as_view(
...
 }
+
+///|
+/// Copies elements from this ArrayView to a destination array starting at a given index.
+pub fn[T] ArrayView::blit_to(
+  self : ArrayView[T],
+  dst : ArrayView[T],
+  len~ : Int,
+  src_offset~ : Int = 0,
+  dst_offset~ : Int = 0
```
peter-jerry-ye: I'm not quite sure if the offset here is useful because one can always write `self[src_offset:].blit_to(dst[dst_offset:])`

builtin/arrayview_test.mbt
```diff
@@ -161,3 +161,43 @@ test "panic negative index8" {
...
+  assert_eq(dst, [7, 8, 9])
+}
+
+///|
+test "arrayview_blit_to partial overlapping" {
+  let arr = [1, 2, 3, 4, 5][:]
+  let dst = [9, 9, 9, 9, 9][:]
+  arr[1:4].blit_to(dst[2:], len=2)
+  assert_eq(dst, [9, 9, 2, 3, 9])
+}
```
peter-jerry-ye: As your example here showed, offset might not be necessary

builtin/arrayview_test.mbt
```diff
@@ -161,3 +161,43 @@ test "panic negative index8" {
...
+
+///|
+test "arrayview_blit_to zero length" {
+  let src : ArrayView[Int] = [1, 2, 3][:]
+  let dst : ArrayView[Int] = [4, 5, 6][:]
+  src[:].blit_to(dst[:], len=0)
+  assert_eq(dst, [4, 5, 6])
+}
```
peter-jerry-ye, line 203: You might want two other tests:

1. panic test to show what happens if the offset if out of bound or the length is too long
2. one test where the arrayview blit to itself, and maybe two areas overlap, e.g. `src[2:].blit_to(src[:], len=3`

builtin/arrayview.mbt
Asterless: Ok. Now I'll first copy the `ArrayView` to  `Array` and then use `Array::blit_to` for the underlying implementation. If there are any areas that can be optimized, please let me know👌

builtin/arrayview.mbt
Asterless: Now this method has been changed to `ArrayView::blit_to(Self[T], Self[T], Int) -> Unit`, so we can use it like `self[src_offset:].blit_to(dst[dst_offset:])`

builtin/arrayview_test.mbt
Asterless: The relevant test cases have been added. Please review them.

FlyCloudC:
The `bilt_to` function is expected to be more efficient than a `for` loop copy. However, the current implementation creates two new arrays and performs four copies, introducing additional space and time overhead. A more efficient approach might be to implement it as a direct `bilt_to` from `self.buf` to `dst.buf`.

Asterless:
> The `bilt_to` function is expected to be more efficient than a `for` loop copy. However, the current implementation creates two new arrays and performs four copies, introducing additional space and time overhead. A more efficient approach might be to implement it as a direct `bilt_to` from `self.buf` to `dst.buf`.

Ok, now I use `UninitializedArray::unsafe_blit` to achieve efficient `blit_to` method😊👍

builtin/arrayview.mbt
```diff
@@ -272,3 +272,57 @@ pub fn[T] ArrayView::op_as_view(
...
+/// ```moonbit
+///   let src = [1, 2, 3, 4, 5][:]
+///   let dst = [0, 0, 0, 0, 0][:]
+///   src[:].blit_to(dst[:], 3)
+///   // dst is now [1, 2, 3, 0, 0]
+/// ```
+pub fn[T] ArrayView::blit_to(
+  self : ArrayView[T],
+  dst : ArrayView[T],
+  len : Int,
```
peter-jerry-ye: We'll accept this PR. Before that, I'll modify it by removing this `len` parameter since it's contained in the `self`.

Issue #1893, opened 2025-04-08
Bytes Method


suiyunonghen:
Bytes no longer has an op_set interface. In addition, the performance of a series of initialization functions provided by Bytes is relatively low. I still hope that Bytes can provide a blit type method to quickly get the initialization content from a region (Bytes, FixedArray[Byte], Array[Byte], etc.). At the same time, I hope that the unsafe_to_bytes of FixedArray can be opened.

suiyunonghen:
Don't you think this function has poor performance? If we implement it this way, it doesn't make much sense. What we need is to know the Bytes capacity clearly and have a CopyMemory method to copy it directly, instead of transferring it twice in this mode.

PR #2261, opened 2025-06-12
Introduce `ByteSource` trait and generic `Bytes::blit_from` 

Asterless:
### Summary
Fixes [Bytes Method #1893](https://github.com/moonbitlang/core/issues/1893)

This pull request introduces the `ByteSource` trait and its implementations for different byte container types (`Bytes`, `FixedArray[Byte]`, and `Array[Byte]`). It also adds the `Bytes::blit_from` function for efficient byte manipulation and corresponding tests to validate functionality.

### New Trait and Functionality:

* **`ByteSource` trait**: Defines a common interface for retrieving a byte at a specific index from various byte container types (`Bytes`, `FixedArray[Byte]`, and `Array[Byte]`). [[1]](diffhunk://#diff-0794443e4fa71c411260d4ad0c9b16603fed3ee65f15323ad2f8da1663ea44c1R345-R392) [[2]](diffhunk://#diff-2f316f353185a37125bcb772553803bd37b395f256f55ba0705d5423f2592dbcR69-R74)
* **Implementations of `ByteSource`**: Added implementations of the `ByteSource` trait for `Bytes`, `FixedArray[Byte]`, and `Array[Byte]`. [[1]](diffhunk://#diff-0794443e4fa71c411260d4ad0c9b16603fed3ee65f15323ad2f8da1663ea44c1R345-R392) [[2]](diffhunk://#diff-2f316f353185a37125bcb772553803bd37b395f256f55ba0705d5423f2592dbcR69-R74)

### Byte Manipulation:

* **`Bytes::blit_from` function**: Allows copying a sequence of bytes from a `ByteSource` to a `Bytes` object, with configurable source and destination indices and length. [[1]](diffhunk://#diff-0794443e4fa71c411260d4ad0c9b16603fed3ee65f15323ad2f8da1663ea44c1R345-R392) [[2]](diffhunk://#diff-2f316f353185a37125bcb772553803bd37b395f256f55ba0705d5423f2592dbcR51)

### Testing:

* **Tests for `ByteSource` trait**: Added unit tests to verify the `get` method functionality for `Bytes`, `FixedArray[Byte]`, and `Array[Byte]`.
* **Tests for `Bytes::blit_from`**: Added unit tests to validate the behavior of `blit_from` with different `ByteSource` types (`Bytes`, `FixedArray[Byte]`, and `Array[Byte]`).

### DIFF
--- a/builtin/array.mbt
+++ b/builtin/array.mbt
@@ -1255,9 +1255,11 @@ pub fn[T] Array::swap(self : Array[T], i : Int, j : Int) -> Unit {
 /// TODO: perf could be improved
 #locals(f)
 pub fn[T] Array::retain(self : Array[T], f : (T) -> Bool raise?) -> Unit raise? {
-  for i = 0, j = 0; i < self.length(); {
-    if f(self.unsafe_get(i)) {
-      self.unsafe_set(j, self.unsafe_get(i))
+  let len = self.length()
+  for i = 0, j = 0; i < len; {
+    let item = self.unsafe_get(i)
+    if f(item) {
+      self.unsafe_set(j, item)
       continue i + 1, j + 1
     }
     continue i + 1, j
--- a/builtin/iter.mbt
+++ b/builtin/iter.mbt
@@ -739,7 +739,7 @@ pub fn[T] Iter::peek(self : Iter[T]) -> T? {
 /// # Returns
 ///
 /// Returns a new iterator with the element `a` prepended to the original iterator.
-#deprecated("Use `self + [a].iter()` or `self + Iter::singleton(a)` instead")
+#deprecated("Use `[a].iter() + self` or `Iter::singleton(a) + self` instead")
 pub fn[T] Iter::prepend(self : Iter[T], a : T) -> Iter[T] {
   fn(yield_) {
     if yield_(a) == IterContinue {
@@ -765,7 +765,7 @@ pub fn[T] Iter::prepend(self : Iter[T], a : T) -> Iter[T] {
 /// # Returns
 ///
 /// Returns a new iterator with the element `a` appended to the original iterator.
-#deprecated("Use `[a].iter() + self` or `Iter::singleton(a) + self` instead")
+#deprecated("Use `self + [a].iter()` or `self + Iter::singleton(a)` instead")
 pub fn[T] Iter::append(self : Iter[T], a : T) -> Iter[T] {
   fn(yield_) {
     if self.run(yield_) == IterContinue {


bytes/bytes.mbt
```diff
@@ -342,3 +342,51 @@ pub impl Add for Bytes with op_add(self : Bytes, other : Bytes) -> Bytes {
...
 }
+
+///|
+/// Retrieves a byte from the source at the specified index.
+///
+/// Parameters:
+///           
+/// * `self` : The byte source to retrieve the byte from.
+/// * `i` : The index of the byte to retrieve.
+pub trait ByteSource {
```
peter-jerry-ye: The others need not to know the content of the `ByteSource`, so it should be just `trait ByteSource`.

bytes/bytes.mbt
```diff
@@ -342,3 +342,51 @@ pub impl Add for Bytes with op_add(self : Bytes, other : Bytes) -> Bytes {
...
+  dest_index~ : Int = 0,
+  src_index~ : Int = 0,
+  len~ : Int
+) -> Bytes {
+  let arr = self.to_array()
+  for i in 0..<len {
+    arr[dest_index + i] = src.get(src_index + i)
+  }
+  Bytes::from_array(arr)
+}
```
peter-jerry-ye: Bytes are intended to be immutable. Maybe you should consider add this to `FixedArray[Byte]`.

bytes/bytes.mbt
Asterless: ok, I have removed the `pub` in [lastest commit](https://github.com/moonbitlang/core/pull/2261/commits/b0d936de38b55bf61850848398fab32936fb4339)😊

hackwaly:
This abstraction will damage the performance. As we do compile bytes blit to native memcpy intrinsic. We should investigate  another way to consolidate the api.

peter-jerry-ye:
> This abstraction will damage the performance. As we do compile bytes blit to native memcpy intrinsic. We should investigate another way to consolidate the api.

That's why I suggested to hide the actual interface of `ByteSource`. Right now it serves only as a marker. We should replace it with `blit_to_fixed_array` or something like that.

With that said, few bytes blit in the standard library uses native memcpy intrinsics right now. This is a TODO task and shouldn't block the API change.

peter-jerry-ye:
We will provide `unsafe_to_bytes` to solve the issue instead. We'll be closing this for now. You are welcome to provide some real world use cases to reason why this API is needed.

bytes/bytes.mbt
suiyunonghen: Don't you think this function has poor performance? If we implement it this way, it doesn't make much sense. What we need is to know the Bytes capacity clearly and have a CopyMemory method to copy it directly, instead of transferring it twice in this mode.

PR #2265, opened 2025-06-13
Improve list rev_fold performance

bobzhang:
## Summary
- avoid array allocation in list `rev_fold`
- implement direct recursion for `rev_foldi`
- run `moon fmt` and `moon info`

## Testing
- `moon fmt`
- `moon info`

------
https://chatgpt.com/codex/tasks/task_e_684b84f0d1148320bbfa1f922c8f1f42

### DIFF
--- a/list/list.mbt
+++ b/list/list.mbt
@@ -444,12 +444,14 @@ pub fn[A, B] T::fold(self : T[A], init~ : B, f : (B, A) -> B) -> B {
 /// }
 /// ```
 pub fn[A, B] T::rev_fold(self : T[A], init~ : B, f : (B, A) -> B) -> B {
-  let xs = self.to_array()
-  let mut acc = init
-  for x in xs.rev_iter() {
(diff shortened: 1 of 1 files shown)


list/list.mbt
```diff
@@ -444,12 +444,14 @@ pub fn[A, B] T::fold(self : T[A], init~ : B, f : (B, A) -> B) -> B {
...
 /// ```
 pub fn[A, B] T::rev_fold(self : T[A], init~ : B, f : (B, A) -> B) -> B {
-  let xs = self.to_array()
-  let mut acc = init
-  for x in xs.rev_iter() {
-    acc = f(acc, x)
+  fn go(xs : T[A], f : (B, A) -> B, acc : B) -> B {
+    match xs {
+      Empty => acc
+      More(x, tail=xs) => f(go(xs, f, acc), x)
```
hackwaly, line 450: This is not tail recursive. I don't think this change will improve performance.

PR #2269, opened 2025-06-13
internal: skip inconsistent test case

Young-Flash:
Skip some of the doc tests that behavior inconsistently across backends / OS, notably native backend on Windows.


peter-jerry-ye:
cc @Kaida-Amethyst need further investigation.

PR #2271, opened 2025-06-14
fix: update some deprecated usages

Topology2333:


builtin/intrinsics.mbt
```diff
@@ -1726,14 +1726,14 @@ pub fn String::op_get(self : String, idx : Int) -> Char = "%string_get"
 /// ```moonbit
 /// test "String::get" {
 ///   let s = "Hello, 世界"
-///   inspect(s.get(0), content="H")
-///   inspect(s.get(7), content="世")
+///   inspect(s.charcode_at(0), content="72")
+///   inspect(s.char_at(7), content="世")
```
Topology2333, line 1730: I'm not sure which method should be used for testing here. The original suggestion was to use `charcode_at()` instead of `get()`, but `charcode_at` returns an `Int`.

PR #2272, opened 2025-06-14
Perform immut/{set, sparse_array} 

FlyCloudC:
This PR is the first part of the split #2205, with clearer git history.

1. clean doc in sparse_array
2. using `Path` to avoid passing depth
3. replace `Collision(Bucket[A])` with `Leaf(A, @list.T[A])`
4. rename `T` to `Node`, remove `Node::empty`, define `T` as `Node?`
5. replace nested `Branch({ data:[Branch(...)], .. })` by `Flat(key, path)`

Parts from #2205 not included in this PR:
1. adopt CPS style in `add`/`remove` to reuse old nodes in special cases (maybe we don't need this)
2. optimize set operations: `union`, `intersection`, `difference`
3. Improve test coverage by using `MyInt`, which only use the lower 8 bits for hash
4. apply same improvements to immut/map

## Overview
```diff
-priv enum Bucket[T] {
-  JustOne(T) // must be non-empty
-  More(T, Bucket[T])
-}

-enum T[A] {
+priv enum Node[A] {
-  Empty

-  Leaf(A)
-  Collision(Bucket[A])
+  Leaf(A, @list.T[A])

+  Flat(A, Path)

-  Branch(@sparse_array.SparseArray[T[A]])
+  Branch(@sparse_array.SparseArray[Node[A]])
}

+ type T[A] Node[A]?
```


immut/internal/sparse_array/bitset.mbt
```diff
@@ -20,21 +20,22 @@ pub(all) type Bitset UInt derive(Eq)
...
 ///|
-/// `index_of(self: Bitset, idx: Int)`
-///
 /// Get the index of the bit in the bitset.
 pub fn Bitset::index_of(self : Bitset, idx : Int) -> Int {
   (self._ & ((1U << idx) - 1)).popcnt()
 }
 
+///|
+pub fn Bitset::fitst_idx(self : Bitset) -> Int {
```
peter-jerry-ye: Did you misspell here? Should it be `first_idx`?

immut/internal/sparse_array/sparse_array_test.mbt
```diff
@@ -12,27 +12,36 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+///|
+fn[A] SparseArray::get(self : SparseArray[A], idx : Int) -> A? {
```
peter-jerry-ye: I'm not quite sure why this is for test alone. Maybe you can also put it as a pub function to replace the old `op_get`?

peter-jerry-ye (commented):
I'll leave it to the experts for the implementation of the algorithm, but here's something I've observed.

immut/internal/sparse_array/bitset.mbt
FlyCloudC: yes

immut/internal/sparse_array/sparse_array_test.mbt
FlyCloudC: Originally, there was a `pub fn has(self, idx) -> Bool` and a `pub fn op_get(self, idx) -> A?`. These two functions had redundant functionality.

In my rewritten version, `op_get` returns an `A`, thus removing the redundancy.

At first, I simply ported my rewritten version, so I changed `op_get`'s return type to `A`. And I added `get`, which behaves the same as the original `op_get`, for test.

But you reminded me of a cleaner solution: leave `op_get` unchanged and simply remove `has`, since it was never used externally anyway.

immut/internal/path/path.mbt
```diff
@@ -0,0 +1,64 @@
...
+const SEGMENT_LENGTH : Int = 5
+
+///|
+const INDEX_MASK : UInt = (1 << SEGMENT_LENGTH) - 1
+
+///|
+const SEGMENT_NUM : Int = 32 / SEGMENT_LENGTH
+
+///|
+const HEAD_TAG : UInt = 0xffffffffU << (SEGMENT_LENGTH * SEGMENT_NUM)
```
Guest0x0, line 38: what's the purpose of this?

immut/hashset/HAMT.mbt
```diff
@@ -27,84 +27,74 @@
...
-        depth + segment_length,
-        key,
-        (hash.reinterpret_as_uint() >> segment_length).reinterpret_as_int(),
-      )
-      Branch(@sparse_array.singleton(idx, child))
+      (Flat(key1, path1.next()), Flat(key2, path2.next()))
     }
+    Branch(@sparse_array.doubleton(idx1, node1, idx2, node2))
   }
+}
```
Guest0x0, line 76: just a note: maybe it's possible to also do path compression for `Branch` (i.e. add an extra path segment (`Path` + length) parameter) too?

immut/internal/path/path.mbt
```diff
@@ -0,0 +1,64 @@
...
+
+///|
+/// If SEGMENT_LENGTH == 5, MAX_TAIL == 0b11_11111
+const MAX_TAIL : UInt = 0xffffffffU >> (SEGMENT_LENGTH * (SEGMENT_NUM - 1))
+
+///|
+pub fn Path::is_last(self : Path) -> Bool {
+  let Path(self) = self
+  self <= MAX_TAIL
+}
```
Guest0x0, line 56: what if the hash value is just zero?

immut/hashset/HAMT.mbt
```diff
@@ -310,12 +291,11 @@ pub impl[A : Hash] Hash for T[A] with hash_combine(self, hasher) {
...
 ///|
-pub impl[A : Eq] Eq for T[A] with op_equal(self, other) {
+impl[A : Eq] Eq for Node[A] with op_equal(self, other) {
   match (self, other) {
-    (Empty, Empty) => true
-    (Leaf(x), Leaf(y)) => x == y
-    (Collision(xs), Collision(ys)) =>
-      xs.size() == ys.size() && xs.iter().all(fn(x) { ys.find(x) })
+    (Leaf(x, xs), Leaf(y, ys)) =>
+      xs.length() == ys.length() && xs.add(x).iter().all(ys.add(y).contains(_))
```
Guest0x0: `ys.add(y)` will be run multiple times here

Guest0x0:
btw from #2205 the optimized set operations seem quite independent, I think we should include that part in this PR too, so that there would be no performance degeneration in those operations

immut/internal/path/path.mbt
FlyCloudC: `Path::of(x)` always has the form `0b11_xxxxx_xxxxx_xxxxx_xxxxx_xxxxx`.

- `Path::of(x).next().next().next()` is `0b11_xxxxx_xxxxx`, not satisfy `<= 0b11_11111`
- `Path::of(x).next().next().next().next()` is `0b11_xxxxx`, satisfy `<= 0b11_11111`

immut/internal/path/path.mbt
FlyCloudC: If `x.hash() == 0`, `Path::of(x)` is `0b11_00000_00000_00000_00000_00000`.

- `Path::of(x).next().next().next()` is `0b11_00000_00000`, not satisfy `<= 0b11_11111`
- `Path::of(x).next().next().next().next()` is `0b11_00000`, satisfy `<= 0b11_11111`

immut/hashset/HAMT.mbt
FlyCloudC: This is not the expected behavior for me. I thought `f(exp, _)` would desugar to

```mbt
{
  let v1 = exp
  hole => f(v1, hole)
}
```

I'll fix it. And I suggest documenting this behavior in the language's documentation.

immut/hashset/HAMT.mbt
FlyCloudC: > just a note: maybe it's possible to also do path compression for `Branch` (i.e. add an extra path segment (`Path` + length) parameter) too?

I think the benefit might be limited. With a good enough hash function, for k in 1~5, the chance of n different elements all have the same kth Path segment is about 1/(32^(n-1)) (assuming 5 bits per segment).

list/list.mbt
```diff
@@ -117,7 +117,7 @@ pub fn[A] length(self : T[A]) -> Int {
 ///   assert_eq(arr, [1, 2, 3, 4, 5])
 /// ```
 #locals(f)
-pub fn[A] each(self : T[A], f : (A) -> Unit) -> Unit {
+pub fn[A] each(self : T[A], f : (A) -> Unit raise?) -> Unit raise? {
```
Guest0x0: what's the purpose of this change? If this is necessary, let's have a dedicated PR for this

immut/internal/path/path.mbt
Guest0x0: So if I understand correctly, the new version only utilize the lower 30 bits of the hash value, and the highest 2 bits are used to identify end of hash value?

FlyCloudC: > So if I understand correctly, the new version only utilize the lower 30 bits of the hash value, and the highest 2 bits are used to identify end of hash value?

Yes

list/list.mbt
FlyCloudC: To support raising during the traversal of `bucket` in the `Leaf(k, bucket)` case within `immut/hashset.T::each`.

I'll remove it and create a PR later to add error polymorphism to `@list`. (#2353)

FlyCloudC:
(using arrow function)

immut/hashmap/HAMT.mbt
```diff
@@ -376,30 +376,13 @@ pub fn[K : Eq + Hash, V] T::intersection(
...
-        m
-      })
-  }
+  self
+  .iter()
+  .fold(init=Empty, (m, kv) => if other.get(kv.0) is Some(_) {
+    m.add(kv.0, kv.1)
+  } else {
+    m
+  })
```
Guest0x0, line 385: This PR should only affect `immut/hashset`, is this accidentally introduced?

Guest0x0 (commented):
This PR overall looks good to me, remaining TODO:

- accidental modification of `@immut/hashmap` as commented above
- add more comments to the `@path` package, explaining how the length of path is represented

immut/hashmap/HAMT.mbt
FlyCloudC: Not accidental. Since the types of `SparseArray::difference` and `SparseArray::intersection` were changed

```diff
-fn[X] SparseArray::difference(Self[X], Self[X]) -> Self[X]
+fn[X] SparseArray::difference(Self[X], Self[X], (X, X) -> X?) -> Self[X]?

-fn[X] SparseArray::intersection(Self[X], Self[X], (X, X) -> X raise?) -> Self[X] raise?
+fn[X] SparseArray::intersection(Self[X], Self[X], (X, X) -> X? raise?) -> Self[X]? raise?
```

the corresponding calls in immut/hashmap were also affected.

FlyCloudC:
> * add more comments to the `@path` package, explaining how the length of path is represented

Done.

Guest0x0:
cleanup the `@immut/hashmap` change and this PR should be good to go

FlyCloudC:
> cleanup the `@immut/hashmap` change and this PR should be good to go

How should I clean up? I replied above, this is not an accidental change.

Guest0x0:
> How should I clean up? I replied above, this is not an accidental change.

I cannot see your comment. If it is under my review comment, is it still pending and not submitted?

FlyCloudC:
> I cannot see your comment. If it is under my review comment, is it still pending and not submitted?

Copy here:

Not accidental. Since the types of `SparseArray::difference` and `SparseArray::intersection` were changed

```diff
-fn[X] SparseArray::difference(Self[X], Self[X]) -> Self[X]
+fn[X] SparseArray::difference(Self[X], Self[X], (X, X) -> X?) -> Self[X]?

-fn[X] SparseArray::intersection(Self[X], Self[X], (X, X) -> X raise?) -> Self[X] raise?
+fn[X] SparseArray::intersection(Self[X], Self[X], (X, X) -> X? raise?) -> Self[X]? raise?
```

the corresponding calls in immut/hashmap were also affected.

---

It seems I still need to learn how to use GitHub, XD

---

OK, now I know how to handle it. This "pending review" design is seriously user-unfriendly

PR #2274, opened 2025-06-15
opt: optimize union to avoid recalculating size in sorted_set

Topology2333:


Topology2333:
Just noticed there's a similar PR (#2267 ) addressing the same optimization.
Looking forward to feedback on which direction is preferred.

Issue #241, opened 2024-04-11
Checked arithmetic operations for `Int` and `Int64`


yj-qin:
We need some checked operations to prevent overflow when doing arithmetic operations, such as `checked_add` and `checked_sub`.

PR #2284, opened 2025-06-16
feat: add checked_add method with overflow detection

Topology2333:
### Summary

Add `checked_add` to `Rational` for safe addition with `Int64` overflow detection. Returns `None` if intermediate operations overflow, otherwise returns a reduced rational.

### Changes

* Add `T::checked_add(self, other) -> T?`
* Handle overflow in numerator and denominator computation
* Add unit tests for normal and overflow cases

### Motivation

`op_add` does not check for overflow. `checked_add` offers a safe alternative for critical use cases.

relates to #241

### DIFF
--- a/rational/rational.mbt
+++ b/rational/rational.mbt
@@ -103,12 +103,35 @@ fn new_unchecked(numerator : Int64, denominator : Int64) -> T {
 
 ///|
 /// NOTE: we don't check overflow here, to align with the `op_add` of `Int64`.
-/// TODO: add a `checked_add` method.
 pub impl Add for T with op_add(self : T, other : T) -> T {
   new_unchecked(
     self.numerator * other.denominator + other.numerator * self.denominator,
     self.denominator * other.denominator,
   )
+}
+
+/// Checked addition for rational numbers.
+/// Returns `None` if any intermediate calculation overflows,
+/// otherwise returns the correctly reduced rational.
+pub fn T::checked_add(self : T, other : T) -> T? {
+  let (a, b) = (self, other)
+  let lhs = a.numerator * b.denominator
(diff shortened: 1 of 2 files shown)


peter-jerry-ye:
Hi, thank you for your contribution. But here seems to be a conterexample

```moonbit
test {
  let a = new_unchecked(922337203685477587L, 4)
  let b= new_unchecked(922337203685477587, 6)
  let c = a.checked_add(b)
  inspect(c, content="None") // should be 4611686018427387935 / 12
}
```

I'm closing this PR for now because we are going to move this package to `x`. You are welcome to contribute there.

Topology2333:
> Hi, thank you for your contribution. But here seems to be a conterexample
> 
> ```moonbit
> test {
>   let a = new_unchecked(922337203685477587L, 4)
>   let b= new_unchecked(922337203685477587, 6)
>   let c = a.checked_add(b)
>   inspect(c, content="None") // should be 4611686018427387935 / 12
> }
> ```
> 
> I'm closing this PR for now because we are going to move this package to `x`. You are welcome to contribute there.

Thank you for your feedback!

You're absolutely right — the issue comes from the fact that I didn't reduce the operands before performing the intermediate multiplications. As a result, the unchecked multiplication overestimates the actual values, leading to a false overflow detection.

I'll work on improving the `checked_add` implementation by factoring out the GCD between the denominators before performing the multiplications, which should avoid the intermediate overflow in such cases.

Thanks again for pointing this out! Once the package is moved to `x`, I'll be happy to reopen a PR there with a corrected version.

peter-jerry-ye:
I have migrated it to x : https://github.com/moonbitlang/x/pull/161 and now it also supports int / int64 / bigint 

I think you can checkout https://github.com/rust-num/num-rational for the correct implementation.

PR #2301, opened 2025-06-20
feat(bytes): expose unsafe cast

### LINKED ISSUE #1893: Bytes Method

peter-jerry-ye:
Closes #1893


builtin/bytes.mbt
```diff
@@ -14,7 +14,12 @@
 
 ///|
 /// Reinterpret the byte sequence as Bytes.
-fn FixedArray::unsafe_to_bytes(self : FixedArray[Byte]) -> Bytes = "%identity"
+/// 
+/// Notice that this will make the `Bytes` object to be a view of the original
+/// byte sequence, so any modification to the original byte sequence will be
+/// reflected in the `Bytes` object.
+#internal(unsafe, "Creating mutable Bytes")
+pub fn FixedArray::reinterpret_as_bytes(self : FixedArray[Byte]) -> Bytes = "%identity"
```
peter-jerry-ye:
```suggestion
pub fn FixedArray::unsafe_reinterpret_as_bytes(self : FixedArray[Byte]) -> Bytes = "%identity"
```

PR #2305, opened 2025-06-20
refactor(map): align op_get with array

### LINKED ISSUE #1332: The `@hashmap.T[K, V]::op_get(K) ->Option[V]` method should not return a Option
The current `get` and `op_get` of `@hashmap.T[K, V]` type both return a `Option[V]`, but the methods of the same names of Array type return an `Option[T]` and a `T ` respectively. Their behavior should be aligned.

peter-jerry-ye:
Closes #1332


bobzhang:
I suggest we hold on this for 2 weeks to have AI updating most existing libs

PR #2325, opened 2025-06-24
separate 'moon test --target native --release' from other tests

myfreess:
Separate the most time-consuming test `moon test --target native --release` into a single job


peter-jerry-ye:
What's the motivation?

PR #2328, opened 2025-06-24
hongbo/more random tests

bobzhang:
- **tweak README**
- **tweak random interface**


peter-jerry-ye:
This is a breaking change and broke downstream users.

PR #2334, opened 2025-06-24
feat(json): expose decode_error for custom error handling

BigOrangeQWQ:
When users manually implement the FromJson trait, they may need to construct JsonDecodeError instances. Exposing the json library's decode_error function would help with this.

```moonbit
type NewType 
// impl @json.FromJson for NewType with from_json(json: Json, path: @json.JsonPath) -> NewType raise {
//    raise @json.JsonDecodeError((path, "JsonType::from_json: expected a string, got \{json}"))
//}

impl @json.FromJson for NewType with from_json(json: Json, path: @json.JsonPath) -> NewType raise {
    raise @json.decode_error(path, "JsonType::from_json: expected a string, got \{json}")
}

```

### DIFF
--- a/json/from_json.mbt
+++ b/json/from_json.mbt
@@ -30,7 +30,10 @@ pub fn[T : FromJson] from_json(
 }
 
 ///|
-fn[T] decode_error(path : JsonPath, msg : String) -> T raise JsonDecodeError {
+pub fn[T] decode_error(
+  path : JsonPath,
+  msg : String
+) -> T raise JsonDecodeError {
   raise JsonDecodeError((path, msg))
 }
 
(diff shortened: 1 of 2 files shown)


peter-jerry-ye:
Why would you need this when the `JsonDecodeError` itself is `pub(all)`?

BigOrangeQWQ:
> Why would you need this when the `JsonDecodeError` itself is `pub(all)`?

I think `decode_error` is clearly more user-friendly, when handling `FromJson` trait, people might write similar functions.

peter-jerry-ye:
We are controlling the scope of the core library and we're not adding utility functions.

PR #2337, opened 2025-06-24
docs: keep depreated message more simple,

illusory0x0:
`Iter::singleton(a) + self` is more clear than  `[a].iter() + self` and conforms to the conventions of other programming languages.


peter-jerry-ye:
There's extra costs.

PR #2339, opened 2025-06-24
docs(tuple): simplify deprecated message

illusory0x0:


FlyCloudC:
`tuple` may not be a variable, it could also be something like `f(x)`. This could lead to redundant computations.

illusory0x0:
> `tuple` may not be a variable, it could also be something like `f(x)`. This could lead to redundant computations.

You are right, but this is not automatical migration tools, and when we write too complicated expression, we will use `let x = expr` instead of writing a large expression.

peter-jerry-ye:
I would prefer rewriting using arrow function.

illusory0x0:
> `tuple` may not be a variable, it could also be something like `f(x)`. This could lead to redundant computations.

I rewrite via `block expression` without extra cost. It can help AI-agent generate correct PR QAQ.

PR #2333, opened 2025-06-24
refactor(random_source)

FlyCloudC:

PR #2344, opened 2025-06-25
chore(random_source)

FlyCloudC:
This PR is part of #2333

1. Flatten the structural. Remove `struct ChaCha8 { state : State }` and rename `State` to `ChaCha8`.
2. Improve api in ChaCha8::next. Return an `UInt64?` instead of `(UInt64, Bool)`.
3. Simplify.

### DIFF
--- a/random/internal/random_source/random_source_chacha.mbt
+++ b/random/internal/random_source/random_source_chacha.mbt
@@ -14,18 +14,6 @@
 
 ///|
 pub(all) struct ChaCha8 {
-  state : State
-}
-
-///|
-pub fn ChaCha8::new(seed : Bytes) -> ChaCha8 {
-  let c = { state: State::new() }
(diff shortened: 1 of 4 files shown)


random/internal/random_source/random_source_chacha.mbt
```diff
@@ -14,99 +14,59 @@
...
 ///|
-let ctrMax = 16U
+let ctr_max = 16U
 
 ///|
 let chunk = 32U
 
 ///|
-let chachaReseed = 4
+let chacha_reseed = 4
```
peter-jerry-ye, line 50: You can rewrite them as `const CTR_INC` if you'd like in another PR.

peter-jerry-ye (approved):
Thank you for your contribution. This PR is smaller and much more easier to verify the correctness.

PR #2349, opened 2025-06-25
perform(random_source): using FixedArray[UInt] for buffer and seed to avoid copy

FlyCloudC:
1. Using `FixedArray[UInt]` for `buffer` and `seed` to avoid copy.
2. Simplify.
3. replace magic numbers with constants.
    - In ChaCha8, a chunk = 64bit. buffer is 32 chunk, and seed is 4 chunk.

### DIFF
--- a/random/internal/random_source/random_source_chacha.mbt
+++ b/random/internal/random_source/random_source_chacha.mbt
@@ -14,10 +14,10 @@
 
 ///|
 struct ChaCha8 {
-  /// length = 32
-  buffer : FixedArray[UInt64]
-  /// length = 4
-  mut seed : FixedArray[UInt64]
+  /// length = 32 * 2
+  buffer : FixedArray[UInt]
+  /// length = 4 * 2
+  mut seed : FixedArray[UInt]
   mut i : UInt
   mut n : UInt
   mut c : UInt
@@ -26,8 +26,8 @@ struct ChaCha8 {
 ///|
 pub fn ChaCha8::new(seed : Bytes) -> ChaCha8 {
   let c = {
-    buffer: FixedArray::make(32, 0UL),
-    seed: FixedArray::make(4, 0UL),
+    buffer: FixedArray::make(32 * 2, 0U),
+    seed: FixedArray::make(4 * 2, 0U),
     i: 0U,
     n: 0U,
     c: 0U,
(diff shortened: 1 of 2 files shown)


random/internal/random_source/random_source_chacha.mbt
```diff
@@ -55,90 +55,60 @@ pub fn ChaCha8::next(self : ChaCha8) -> UInt64? {
...
-  chacha_block(self.seed, buffer, self.c)
-  // concat 2 u32 to u64 and store in self.buffer
-  for i in 0..<16 {
-    for j = 0; j < 4; j = j + 2 {
-      let lo = buffer[i][j]
-      let hi = buffer[i][j + 1]
-      let u64 : UInt64 = (hi.to_uint64() << 32) | lo.to_uint64()
-      self.buffer[i * 2 + j / 2] = u64
-    }
-  }
```
FlyCloudC: Note:
- in line 113: `lo` and `hi` are `UInt64`
- in line 124: `lo` and `hi` are `UInt32`, they form a UInt64 (stored in self.buffer[i * 2 + j / 2])

All operations follow little-endian.

random/internal/random_source/random_source_chacha.mbt
```diff
@@ -14,39 +14,36 @@
...
-let ctr_inc = 4U
+const COUNTER_INC = 4U
 
 ///|
-let ctr_max = 16U
+const COUNTER_MAX = 16U
 
 ///|
-let chunk = 32U
+/// Must be a power of 2
```
peter-jerry-ye, line 42: You can actually write a small test for this so that the invariant won't be broken by others, e.g. BUFFER_CHUNK_NUM.ctz() + BUFFER_CHUNK_NUM.clz() == 31

random/internal/random_source/random_source_chacha.mbt
```diff
@@ -14,39 +14,36 @@
 
 ///|
 pub(all) struct ChaCha8 {
```
peter-jerry-ye: I would suggest that you add another commit before everything: change the visibility to abstract (`struct ChaCha8`).

Then you would be warned whether the `mut` of `buffer` or `seed` is necessary, which can show your PR better.

random/internal/random_source/random_source_chacha.mbt
FlyCloudC: done

random/internal/random_source/random_source_chacha.mbt
FlyCloudC: done

peter-jerry-ye (approved):

peter-jerry-ye:
FTR: it's roughly 10 - 20% faster based on the backend.

PR #2360, opened 2025-06-26
chore(array): truncate

FlyCloudC:
The definition of `self.clear()` is just `self.unsafe_truncate_to_length(0)`.


peter-jerry-ye:
Thank you.

PR #2370, opened 2025-06-27
refactor: handle any json number

### LINKED ISSUE #1366: Extend JSON to include Int64, UInt64, Arbitrary(String)

peter-jerry-ye:
Closes #1366

This implementation will introduce new field in `Json::Number` to hold the representation of the Json number that can't be put into a `Double`.

The current implementation will decide during parsing, to use the `repr : String?` to hold the representations, if:
- it's an integer (not written with exponential or decimal point) that can't be safely represented by the `Double`
- it's a double that failed parsing

The corresponding value will be +/- inf, which aligns with the behavior of `Json.parse` in JS.

The implementation of `Eq` will compare only the number part and ignore the representation part.

### DIFF
--- a/builtin/json.mbt
+++ b/builtin/json.mbt
@@ -51,10 +51,14 @@ pub fn Json::null() -> Json {
 /// Example:
 ///
 /// ```moonbit
-///   inspect(Json::number(3.14), content="Number(3.14)")
+/// inspect(Json::number(3.14), content="Number(3.14)")
+/// inspect(
+///   Json::number(@double.infinity, repr="1e9999999999999999999999999999999").stringify(), 
+///   content="1e9999999999999999999999999999999"
+/// )
 /// ```
-pub fn Json::number(number : Double) -> Json {
-  return Number(number, repr=None)
+pub fn Json::number(number : Double, repr? : String) -> Json {
+  return Number(number, repr~)
 }
 
 ///|
--- a/builtin/json.mbt
+++ b/builtin/json.mbt
@@ -13,16 +13,29 @@
 // limitations under the License.
 
 ///|
-#visibility(change_to="readonly", "Use helper functions like `Json::object(...)` instead")
-pub(all) enum Json {
+pub enum Json {
   Null
   True
   False
-  Number(Double, repr~ : String?)
+  Number(Double, repr~ : String?) // 1.0000000000000000000e100 
   String(String)
   Array(Array[Json])
   Object(Map[String, Json])
-} derive(Eq)
+}
+
+///|
+pub impl Eq for Json with op_equal(a, b) {
+  match (a, b) {
+    (Null, Null) => true
+    (True, True) => true
+    (False, False) => true
+    (Number(a_num, ..), Number(b_num, ..)) => a_num == b_num
+    (String(a_str), String(b_str)) => a_str == b_str
+    (Array(a_arr), Array(b_arr)) => a_arr == b_arr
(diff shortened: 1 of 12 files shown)


json/lex_number_test.mbt
```diff
@@ -33,10 +33,7 @@ test "lex_zero invalid case" {
 
 ///|
 test "invalid number" {
-  inspect(
-    try? @json.parse("1e999999999"),
-    content="Err(Invalid number 1e999999999 at line 1, column 0)",
-  )
+  inspect(try? @json.parse("1e999999999"), content="Ok(Number(Infinity))")
```
peter-jerry-ye:
```suggestion
  inspect(try? @json.parse("1e999999999"), content="Ok(Number(Infinity, repr=Some("1e999999999"))")
```

FlyCloudC:
What should be the expected results of the following two expressions?

```mbt
@json.parse("1e2") == @json.parse("100")

@json.parse("1000000000000000000001") == 
@json.parse("1000000000000000000002")
```

peter-jerry-ye:
If we are lucky enough, the first should be equal, and the second would not be equal. Currently the parsing doesn't consider the `safe max integer`.

One thing we can do is to decide whether it is floating. If it's not, only convert the number that is safe.

But anyway, you are handling `Double`, so take caution.

bobzhang:
let's get this landed in the next release, we can make some time for the AI to upgrade most packages

builtin/json.mbt
```diff
@@ -13,16 +13,29 @@
...
+  match (a, b) {
+    (Null, Null) => true
+    (True, True) => true
+    (False, False) => true
+    (Number(a_num, ..), Number(b_num, ..)) => a_num == b_num
+    (String(a_str), String(b_str)) => a_str == b_str
+    (Array(a_arr), Array(b_arr)) => a_arr == b_arr
+    (Object(a_obj), Object(b_obj)) => a_obj == b_obj
+    _ => false
+  }
```
bobzhang, line 37: match a {
    Null => b is Null
    ...
}

PR #2379, opened 2025-06-29
rational: fix invariants, simplify implementation, clean up tests

FlyCloudC:
1. Fixed invariant violations in the `arbitrary` impl
2. Optimized implementation using invariants
3. Removed duplicate tests


peter-jerry-ye:
FYI: we are going to move this package out of core

PR #2383, opened 2025-06-30
feat(deque): add shuffle_in_place method

PingGuoMiaoMiao:
#### ​**​Summary​**
This pr adds the shuffle method, but does not write a test

### DIFF
--- a/bool/bool.mbt
+++ b/bool/bool.mbt
@@ -116,3 +116,53 @@ pub impl Hash for Bool with hash(self) {
 pub impl Hash for Bool with hash_combine(self, hasher) {
   hasher.combine_bool(self)
 }
+
+///|
+/// Converts a boolean value to an unsigned 16-bit integer.
+///
+/// Parameters:
+///
+/// * `self` : The boolean value to be converted.
+///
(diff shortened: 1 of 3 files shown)


PingGuoMiaoMiao:
I didn't expect how to write the test for this method.

deque/deque.mbt
```diff
@@ -1423,3 +1423,48 @@ pub fn[A] T::drain(self : T[A], start~ : Int, len? : Int) -> T[A] {
...
+/// and returns an integer.
+/// *rand n* is expected to return a uniformly distributed integer between 0 and n - 1
+/// 
+/// # Note
+/// This function handles the circular buffer nature of the deque internally.
+pub fn[A] T::shuffle_in_place(self : T[A], rand~ : (Int) -> Int) -> Unit {
+  let n = self.len
+  let buf_length = self.buf.length()
+  for i = n - 1; i > 0; i = i - 1 {
+    let j = rand(i + 1) % (i + 1)
```
FlyCloudC: As mentioned in the doc comment, `rand(i + 1)` already returns a number smaller than `i + 1`, so the `% (i + 1)` operation is unnecessary.

PR #2396, opened 2025-07-02
Perform immut/map

FlyCloudC:
This PR applies the same changes to immut/map as those made to immut/set in #2272.

1. using `Path` to avoid passing depth
2. replace `Collision(Bucket[K, V])` with `Leaf(K, V, @list.T[(K, V)])`
3. rename `T` to `Node`, remove `Node::empty`, define `T` as `Node?`
4. replace nested `Branch({ data:[Branch(...)], .. })` by `Flat(key, value, path)`


immut/hashmap/types.mbt
```diff
@@ -13,18 +13,16 @@
...
-/// An immutable hash-map data structure
-enum T[K, V] {
-  Empty
-  Leaf(K, V) // optimize for the case of no collision
-  Collision(Bucket[K, V]) // use a list of buckets to resolve collision
-  Branch(@sparse_array.SparseArray[T[K, V]])
+/// An non-empty immutable hash set data structure
+priv enum Node[K, V] {
+  Flat(K, V, Path)
+  Leaf(K, V, @list.T[(K, V)]) // use a list of buckets to resolve collision
```
Guest0x0, line 22: let's still keep the bucket abstraction, to avoid potential tuple boxing here

immut/internal/sparse_array/sparse_array.mbt
```diff
@@ -176,7 +176,6 @@ pub fn[X] SparseArray::difference(
   f : (X, X) -> X?
 ) -> SparseArray[X]? {
   let self_elem_info = self.elem_info
-  guard self_elem_info != empty_bitset else { return None }
```
Guest0x0, line 179: why this diff

immut/hashmap/HAMT.mbt
```diff
@@ -92,57 +97,67 @@ pub fn[K : Eq + Hash, V] op_get(self : T[K, V], key : K) -> V? {
...
       if key == key1 {
-        Leaf(key, value)
+        Leaf(key, value, bucket)
       } else {
-        Collision(More(key, value, JustOne(key1, value1)))
+        let new_bucket = match bucket.find_index(kv => kv.0 == key) {
+          None => bucket
+          Some(index) => bucket.remove_at(index)
+        }
+        Leaf(key, value, new_bucket.add((key1, value1)))
```
Guest0x0, line 144: avoid the unnecessary two traversal here

immut/hashmap/HAMT.mbt
```diff
@@ -189,20 +212,24 @@ pub fn[K, V, A] fold(
...
       match node {
-        Empty => continue (tail, acc)
-        Leaf(k, v) => continue (tail, f(acc, k, v))
-        Collision(bucket) => continue (tail, bucket.foldl_with_key(init=acc, f))
+        Leaf(k, v, bucket) =>
+          continue (
+              tail,
+              bucket.fold(init=f(acc, k, v), (acc, kv) => f(acc, kv.0, kv.1)),
+            )
+        Flat(k, v, _) => continue (tail, f(acc, k, v))
```
Guest0x0: TODO: rewrite this using recursive function

immut/hashmap/HAMT.mbt
```diff
@@ -216,81 +243,96 @@ pub fn[K, V, A] fold_with_key(
...
+      if key1 == key {
+        match bucket {
+          @list.Empty => None
+          More((key2, value2), tail~) => Some(Leaf(key2, value2, tail))
+        }
+      } else if bucket.find_index(kv => kv.0 == key) is Some(index) {
+        Some(Leaf(key1, value1, bucket.remove_at(index)))
+      } else {
+        Some(self)
+      }
```
Guest0x0, line 301: turn `Leaf` into `Flat` when the bucket contains only one element?

immut/internal/sparse_array/sparse_array.mbt
FlyCloudC: invariant: if `self` is an `SparseArray`, then `self.elem_info != empty_bitset`. So it should not check here

immut/hashmap/types.mbt
FlyCloudC: This boxing overhead only becomes significant (more than 24% elements in `Leaf`) when `self.size > 1000_0000`. This is a trade-off with maintainability. I prefer keeping the implementations of immut/hashmap and immut/hashset consistent. Is it worth supporting this scale in the standard library?

```mbt
fn in_leaf(self : T[Int, Unit], x : Int) -> Bool {
  loop (self.inner().unwrap(), @path.of(x)) {
    (Leaf(_), _) => true
    (Flat(_), _) => false
    (Branch(children), pathx) => {
      let child = children[pathx.idx()].unwrap()
      continue (child, pathx.next())
    }
  }
}

test {
  fn f(size) {
    let a : T[Int, Unit] = from_iter((0).until(size).map(x => (x, ())))
    let mut c = 0
    for x in 0..<size {
      if a.in_leaf(x) {
        c += 1
      }
    }
    (c.to_float() / size.to_float())
  }

  inspect(f(100), content="0")
  inspect(f(1000), content="0")
  inspect(f(1_0000), content="0")
  inspect(f(10_0000), content="0.003060000017285347")
  inspect(f(100_0000), content="0.02905300073325634")
  inspect(f(1000_0000), content="0.2445134073495865")
  inspect(f(1800_0000), content="0.3792286217212677")
}

```

FlyCloudC: Is using a value type possible here? (I'm not entirely sure)

```mbt
#valtype
priv struct Entry[K,V] {
  key : K
  value : V
}

priv enum Node[K, V] {
  Leaf(K, V, @list.T[Entry[K,V]])
  ...
}
```

immut/hashmap/HAMT.mbt
FlyCloudC: Maybe we can do this in a new PR. I'd prefer to keep hashmap and hashset consistent first, then modify them together

immut/hashmap/types.mbt
Guest0x0: `#valtype` is currently not applicable, this attribute can only be applied to types with only non-pointer field at the moment. But that would change in the future. Given that overhead of collision bucket is not significant, let's use `@list` now as you suggest, and apply value type optimization when applicable in the future

immut/hashmap/HAMT.mbt
Guest0x0: after second thought, this version has its own benefit, it can avoid allocation and copying if the element is not present (likely)

Guest0x0 (approved):

PR #2398, opened 2025-07-02
feat(deque): add shuffle_in_place and shuffle method

PingGuoMiaoMiao:
#### ​**​Summary​**

This pr adds the shuffle method, but does not write a test


PingGuoMiaoMiao:
Ask me about this PR and the above questions

peter-jerry-ye (changes requested):
Tests are missing.

PR #2402, opened 2025-07-03
fix(immut/hashmap): prioritize right values

### LINKED ISSUE #2395: [Consistency Review] Clarify the behavior of `union` and `intersection` for conflicting keys in `immut/hashmap`
What is the expected behavior of

```mbt
let m1 = singleton(0, "left")
let m2 = singleton(0, "right")
inspect(m1.union(m2).get(0))         // "left" or "right"?  
inspect(m1.intersection(m2).get(0))  // "left" or "right"?   
```

There’s a comment here stating that it should be right, but the line 337 has a bug where left is preserved in some cases.

https://github.com/moonbitlang/core/blob/046e877f6fec13aa8fd44249f76ba04cd4ccdafe/immut/hashmap/HAMT.mbt#L321-L339

And I've noticed the same issue with `intersection` as well.

So how should we design the API? Should we expose "right-hand side element is prioritized" to users?

FlyCloudC:
Close #2395

### DIFF
--- a/immut/hashmap/HAMT.mbt
+++ b/immut/hashmap/HAMT.mbt
@@ -319,23 +319,9 @@ pub fn[K, V] size(self : T[K, V]) -> Int {
 }
 
 ///|
-/// Union two hashmaps
+/// Union two hashmaps, right-hand side element is prioritized
 pub fn[K : Eq + Hash, V] T::union(self : T[K, V], other : T[K, V]) -> T[K, V] {
-  match (self, other) {
-    (_, Empty) => self
-    (Empty, _) => other
-    (_, Leaf(k, v)) => self.add(k, v)
-    (Leaf(k, v), _) =>
-      // right-hand side element is prioritized
-      // like Clojure's merge
-      match other.get(k) {
-        Some(_) => other
-        None => other.add(k, v)
-      }
-    (Branch(sa1), Branch(sa2)) =>
-      Branch(sa1.union(sa2, (m1, m2) => m1.union(m2)))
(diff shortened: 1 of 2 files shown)


immut/hashmap/HAMT.mbt
```diff
@@ -319,23 +319,9 @@ pub fn[K, V] size(self : T[K, V]) -> Int {
...
-      // right-hand side element is prioritized
-      // like Clojure's merge
-      match other.get(k) {
-        Some(_) => other
-        None => other.add(k, v)
-      }
-    (Branch(sa1), Branch(sa2)) =>
-      Branch(sa1.union(sa2, (m1, m2) => m1.union(m2)))
-    (_, _) => self.iter().fold(init=other, (m, kv) => m.add(kv.0, kv.1))
-  }
```
FlyCloudC, line 338: Let me explain why I removed the previous optimization for non-`(Branch(_), Branch(_))` cases.

The original implementation had a bug: `add` could only be called on top-level nodes, but line 336 performed recursive `union` on subtrees, and line 337 incorrectly called `add` on non-top-level node `other` during recursion.

Since these changes will eventually be superseded by #2396, I replace it with an inefficient but clearly correct implementation. It just a temporary solution.

peter-jerry-ye: Does this explain why the test on line 262 gave the correct behavior but the ones you modified didn't?

FlyCloudC: Yes, in that test, the recursive calls will eventually reach `Leaf(1, 1).union(Leaf(1, 10))`.

PR #2408, opened 2025-07-04
Bleeding: migrate multi-line string syntax

Yoorkin:
Migrate the multi-line string syntax. These patches should be shipped with the pre-release to avoid deprecation warnings when users compile the core.  

@rami3l  @Guest0x0


rami3l:
https://github.com/moonbitlang/core/tree/bleeding/fix-fmt should be used instead; closing.

PR #2413, opened 2025-07-05
feat(Array): add retain_map for in-place filter+map operations

PingGuoMiaoMiao:
### Changes Proposed

- Adds new `retain_map` method to `Array` type
- Combines filtering and mapping in a single in-place operation
- Matches behavior of similar functions in other languages (Rust's `retain_mut`, Swift's `removeAll(where:)` with mapping)

### DIFF
--- a/builtin/array.mbt
+++ b/builtin/array.mbt
@@ -1682,7 +1682,10 @@ pub fn[A] Array::truncate(self : Array[A], len : Int) -> Unit {
 /// })
 /// // arr is now [4, 8]
 /// ```
-pub fn[A] Array::retain_map(self : Array[A], f : (A) -> A?) -> Unit {
+pub fn[A] Array::retain_map(
+  self : Array[A],
+  f : (A) -> A? raise?
+) -> Unit raise? {
   if self.is_empty() {
     return
   }
@@ -1699,6 +1702,5 @@ pub fn[A] Array::retain_map(self : Array[A], f : (A) -> A?) -> Unit {
       None => ()
     }
   }
-  let _ = self.unsafe_truncate_to_length(write_idx)
-
+  self.unsafe_truncate_to_length(write_idx)
 }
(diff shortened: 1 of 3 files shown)


builtin/array.mbt
```diff
@@ -1678,3 +1678,39 @@ pub fn[A] Array::truncate(self : Array[A], len : Int) -> Unit {
...
+  }
+  let buf = self.buffer()
+  let len = self.length()
+  let mut write_idx = 0
+  for read_idx in 0..<len {
+    match f(buf[read_idx]) {
+      Some(new_val) => {
+        if write_idx != read_idx {
+          buf[write_idx] = new_val
+        }
```
FlyCloudC: Why not update `buf[write_idx]` when `write_idx == read_idx`?

Here is a test:

```mbt
test {
  let arr = [2, 3, 4]
  arr.retain_map(x => if x < 4 { Some(x + 1) } else { None })
  inspect(arr)
}

**Review by FlyCloudC — CHANGES_REQUESTED** [review]

**Review by FlyCloudC — APPROVED** [review]

File: builtin/array.mbt [hunk 2]
```diff
@@ -1678,3 +1678,41 @@ pub fn[A] Array::truncate(self : Array[A], len : Int) -> Unit {
...
+/// ```moonbit
+/// let arr = [1, 2, 3, 4, 5]
+/// arr.retain_map(fn(x) {
+///   if x % 2 == 0 {
+///     Some(x * 2)
+///   } else {
+///     None
+///   }
+/// })
+/// // arr is now [4, 8]
```
peter-jerry-ye: Please use `inspect` instead of writing comments

PingGuoMiaoMiao:
Ask me about this PR and the above questions

peter-jerry-ye (changes requested):
I would remove the error polymorphism for now as this would leave an array with broken state.

PR #2414, opened 2025-07-05
chore: replace deprecated @sorted_map.of with from_array

HCHogan:
since `@sorted_map.of` is deprecated, using it in `Show` will cause confusion


peter-jerry-ye:
We'll review the APIs in the next few months and see what we'll do with these output strings.

peter-jerry-ye:
After some discussion, we will not guarantee that the `Show` will always generate the code that executes.

peter-jerry-ye:
A counter example would be the derived `Show` implementation for user defined data structures, which might be readonly.

PR #2415, opened 2025-07-07
feat(cmp): Provide a recommended value for the comparison result

illusory0x0:


bobzhang:
The point of using Int as the returning value for comparsion is pragmatism, the benefit provided by this PR is not obvious.
Currently we are doing the clean up of core API, so we would prefer PR which is a clear win

PR #2429, opened 2025-07-11
feat(int16): add integer conversion

illusory0x0:
This behavior is consistent with Rust's `as`, which first expands to the corresponding bit and then reinterprets it with binary complement.

https://doc.rust-lang.org/reference/expressions/operator-expr.html#numeric-cast


bobzhang:
due to our limited bandwidth, we are only going to add new APIs which are a clear win, the `reinterpret_as_uint16` is uncontroversial, the other 2 apis like `to_unit` needs further discussions due to overflow(and there are easy work arounds)

PR #2430, opened 2025-07-11
feat(bigint): add from_int16 and from_uint16

illusory0x0:


bobzhang:
int16 is a fairly uncommon type, it was introduced mostly for completeness, and in Arrays, there is easy work around `BigInt::from_int((int16).to_int())`, so we are not going to add int16 support everywhere to cover very minimal use cases

Issue #1899, opened 2025-04-09
@prelude package doesn't export intrinsic types


illusory0x0:
In Haskell We can define the same name with intrinsic types, but MoonBit can't.

### Haskell 

```haskell
> newtype Int = Int(Prelude.Int) deriving(Show)
> x = Int(0)
> x
Int 0
```

### OCaml

```ocaml
utop # type int = Int of Int.t;;
type int = Int of int

utop # Int(0);;
- : int = Int 0
```

Guest0x0:
In MoonBit, `Int`/`Double` etc. are compiler builtin types, so there is no need to export them: they are universally available, unqualified.

Defining custom type with the same name as builtin types can cause a lot of confusion for code reader. So we may disable this in the future.

illusory0x0:
> In MoonBit, `Int`/`Double` etc. are compiler builtin types, so there is no need to export them: they are universally available, unqualified.
> 
> Defining custom type with the same name as builtin types can cause a lot of confusion for code reader. So we may disable this in the future.

I think it's still necessary to add these intrinsic types to prelude, because the templates used by moon now don't work with quanlified names, but they don't report compile errors now!

**related issues**

https://github.com/moonbitlang/moonbit-docs/issues/669

https://github.com/moonbitlang/moonbit-docs/issues/878

Issue #1167, opened 2024-10-25
add some explain of primitive&builtin API


Lampese:
Many people feedback that when they want to find some API of primitive types(like Byte), a part of API in the core/byte and other part of API in the core/builtin/core. If they don't know it before, they only can find one part. I think we should add some explain in each part to tell them "where is the another part".

peter-jerry-ye:
This should be fixed during gradual refactoring of organization, including the introduction of `@prelude` by @Guest0x0

PR #2432, opened 2025-07-12
feat(prelude): export intrinsic type name

illusory0x0:
In order for the template for the moon test deriver to use fully qualified names.

https://github.com/moonbitlang/moon/tree/main/crates/moonbuild/template/test_driver

related issues:

https://github.com/moonbitlang/core/issues/1167

https://github.com/moonbitlang/moonbit-docs/issues/878

https://github.com/moonbitlang/moonbit-docs/issues/669

https://github.com/moonbitlang/core/issues/1899

### DIFF
--- a/internal/intrinsic.mbt
+++ b/internal/intrinsic.mbt
@@ -1,0 +1,67 @@
+// Copyright 2025 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


bobzhang:
@illusory0x0 thanks for your contributions, but we have very limited bandwidth, we are currently foucsing on cleanup. So we tend to accept code which does not need lots of discussions, i.e obvious wins like bug fix and APIs which are conventional.

illusory0x0:
> @illusory0x0 thanks for your contributions, but we have very limited bandwidth, we are currently foucsing on cleanup. So we tend to accept code which does not need lots of discussions, i.e obvious wins like bug fix and APIs which are conventional.

This is indeed a **BugFix**, needing to refer to fully qualified names in the moon test deriver, respecting the current compiler implementation which allows the user to define types like `@lib.Char`, but now causes the moon test to fail to compile!

[bug example](https://github.com/moonbitlang/moonbit-docs/issues/878)

PR #2441, opened 2025-07-14
Deprecate immut list package

bobzhang:
## Summary
- mark `immut/list` type `T` and all functions as deprecated
- encourage using `list` instead
- update interface via `moon info`

## Testing
- `moon fmt`
- `moon info`
- `moon check`
- `moon test`

------
https://chatgpt.com/codex/tasks/task_e_68744707a8a08320af0964c7f38b09ab

### DIFF
--- a/immut/list/list.mbt
+++ b/immut/list/list.mbt
@@ -13,16 +13,19 @@
 // limitations under the License.
 
 ///|
+#deprecated("use `@list` instead")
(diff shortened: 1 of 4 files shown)


bobzhang:
note current there is not a easy way to deprecate `impl Trait for Type` cc @Yoorkin

PR #2442, opened 2025-07-14
hongbo/api cleanups

bobzhang:
- **tweak**
- **remove cascading deprecations**

### DIFF
--- a/immut/list/list.mbt
+++ b/immut/list/list.mbt
@@ -266,7 +266,6 @@ pub fn[A] unsafe_head(self : T[A]) -> A {
 }
 
 ///|
-#deprecated("Use `unsafe_head` instead")
 #coverage.skip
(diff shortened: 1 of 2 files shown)


immut/list/list.mbt
```diff
@@ -266,7 +266,6 @@ pub fn[A] unsafe_head(self : T[A]) -> A {
 }
 
 ///|
-#deprecated("Use `unsafe_head` instead")
 #coverage.skip
```
bobzhang, line 269: shall we detect duplicated deprecations? @Yoorkin

Yoorkin: I think yes. The overlapping warnings will be shadowed, meaning some deprecation messages will not be displayed. This behavior might confuse users.

PR #2455, opened 2025-07-15
feat(deque): implement extract_if method

PingGuoMiaoMiao:
## Feature Description
Implements a new `extract_if` method for deques that:
- Extracts elements matching a predicate from the original deque
- Modifies the original deque in-place
- Preserves original element order
- Handles circular buffer edge cases correctly

## Implementation
1. **Circular Buffer Handling**:

### DIFF
--- a/deque/deque.mbt
+++ b/deque/deque.mbt
@@ -1661,3 +1661,62 @@ pub fn[A] T::rev(self : T[A]) -> T[A] {
   // Create new deque with reversed elements
   T::{ buf: new_buf, len, head: 0, tail: if len == 0 { 0 } else { len - 1 } }
 }
+
+///|
+/// Extracts elements from a deque that satisfy a given predicate function. The
+/// extracted elements are removed from the original deque and returned as a new
+/// deque. The relative order of the extracted elements is preserved.
+///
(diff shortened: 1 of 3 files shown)


deque/deque.mbt
```diff
@@ -1661,3 +1661,62 @@ pub fn[A] T::rev(self : T[A]) -> T[A] {
...
+///
+/// Example:
+///
+/// ```moonbit
+///   let dq = @deque.of([1, 2, 3, 4, 5])
+///   let extracted = dq.extract_if((x) => { x % 2 == 0 })
+///   inspect(extracted, content="@deque.of([2, 4])")
+///   inspect(dq, content="@deque.of([1, 3, 5])")
+/// ```
+pub fn[A] extract_if(self : T[A], f : (A) -> Bool) -> T[A] {
```
Lampese: Can we try to solve this requirement without involving a large number of underlying operations as much as possible? Although there may be a small amount of performance waste.

The main issue is that Deque's non critical APIs are too prone to errors when operating at the bottom layer.

peter-jerry-ye: The current implementation is far too inefficient.

PR #2457, opened 2025-07-16
massive api refactoring using AI

bobzhang:
- **use length in favor of len**
- **rename length to len**


peter-jerry-ye:
Shall we make `len` return `Size` in the same way as Rust, so that we can support 64-bit systems in the future?

PR #2461, opened 2025-07-17
Add `strip_prefix` and `strip_suffix`

### LINKED ISSUE #2410: Add `remove_prefix` and `remove_suffix` methods for `String`
```moonbit
pub fn remove_prefix(self: String, prefix: String) -> String? {
  if self.has_prefix(prefix) {
    return Some(self.charcodes(start=prefix.length()).to_string())
  }
  return None
}

pub fn remove_suffix(self: String, suffix: String) -> String? {
  if self.has_suffix(suffix) {
    return Some(self.charcodes(end=self.length() - suffix.length()).to_string())
  }
  return None
}
```

Asterless:
### Summary

- Fixes #2410 

### New Methods:

- `strip_prefix`: Added a method to remove a specified prefix from a string if it exists. Returns `Some(suffix)` if the prefix is found and removed, or `None` otherwise.

-  `strip_suffix` :  Added a method to remove a specified suffix from a string if it exists. Returns `Some(prefix)` if the suffix is found and removed, or `None` otherwise.

### Test cases:

- Comprehensive test cases were added to ensure correct behavior for various edge cases, including empty strings and Unicode characters.

### DIFF
--- a/string/methods.mbt
+++ b/string/methods.mbt
@@ -230,26 +230,26 @@ pub fn strip_suffix(self : String, suffix : View) -> View? {
 
 ///|
 test "strip_prefix" {
-  inspect("hello world".strip_prefix("hello "), content="Some(world)")
+  inspect("hello world".strip_prefix("hello "), content="Some(\"world\")")
   inspect("hello world".strip_prefix("hi "), content="None")
-  inspect("hello".strip_prefix("hello"), content="Some()")
-  inspect("".strip_prefix(""), content="Some()")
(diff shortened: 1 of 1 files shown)


Asterless:
> Notice that the corresponding API in rust belongs to the `std::str` type, which is also called the `string slice`. So it should correpondingly return `@string.View`.

I'm very sorry I didn't notice this. I have already fixed this issue in the latest commit. Please review it.😊

PR #2464, opened 2025-07-18
feat: parse number for uint64/int64/bigint

peter-jerry-ye:
Now that we've changed the representation of `Number`, we can parse arbitrary length number. This PR relaxes the restriction upon the `from_int` implementation to accept numbers as well.

Notice though that the `BigInt::from_string` currently just panics if the string doesn't looks like a number. This should be adjusted in the future.


bobzhang:
> This PR relaxes the restriction upon the from_int implementation to accept numbers as well.

It is not cool to accept both, I think we need have a clear plan to pick one , either Number or String, and we need more docs about its precise semantics

PR #2466, opened 2025-07-18
refactor(random): use value type for performance

Yu-zh:
Before:
```
bench moonbitlang/core/random/random_test.mbt::bench random
time (mean ± σ)         range (min … max) 
 129.01 ms ± 362.74 µs   128.63 ms … 129.75 ms  in 10 ×      1 runs
```

After:
```
bench moonbitlang/core/random/random_test.mbt::bench random
time (mean ± σ)         range (min … max) 
  67.28 ms ± 381.11 µs    66.52 ms …  67.72 ms  in 10 ×      2 runs
```

### DIFF
--- a/random/moon.pkg.json
+++ b/random/moon.pkg.json
@@ -5,5 +5,8 @@
     "moonbitlang/core/random/internal/random_source",
     "moonbitlang/core/bigint"
   ],
-  "test-import": ["moonbitlang/core/float"]
+  "test-import": [
+    "moonbitlang/core/float",
+    "moonbitlang/core/bench"
+  ]
 }
--- a/random/random_test.mbt
+++ b/random/random_test.mbt
@@ -73,3 +73,12 @@ test "uint64 modulo bias handling with large limit" {
   let result = r.uint64(limit~)
   inspect(result < limit, content="true")
 }
+
+///|
+test "bench random" (b : @bench.T) {
+  let r = @random.Rand::new()
+  b.bench(() => for i in 0..<1000000 {
+    let _ = r.uint64(limit=10000000000000000000UL)
+
+  })
+}


random/moon.pkg.json
```diff
@@ -5,5 +5,9 @@
     "moonbitlang/core/random/internal/random_source",
     "moonbitlang/core/bigint"
   ],
-  "test-import": ["moonbitlang/core/float"]
+  "test-import": [
+    "moonbitlang/core/float",
+    "moonbitlang/core/bench",
+    "moonbitlang/core/test"
```
Yu-zh: It seems I have to add `moonbitlang/core/test` to the `test-import` so that `moon test` can run. But it's reported as an unused package.

bobzhang: cc @lynzrand it seems that the blackboxtest_driver used test:
```
let it = @moonbitlang/core/test.new(name)
```
do you have any idea to fix this?
This seems to be caused by using `@bench` in the test function

lynzrand: I can confirm that this was caused by the `with_args_driver_template`, which was used when the driver generator detected the presence of tests which has input parameters. I'm currently unsure why the input type of `@bench.T` didn't result in a type mismatch where the generated driver requires a `@test.T`.

As of the unused package issue, I suspect a quick fix would be adding the generated test driver (which actually uses `@test`) to the input file list of `moon check`.

The ultimate cause of this is probably the same thing that caused Go to use an import list at the top of each file -- **generated files that the user does not own** wants to import something in (and only in) them, but the user needs to add them manually into the import list (and this list of import might change from version to version). I currently do not have a solid idea on how this should be solved.

lynzrand: I suspect that this is a bug caused by https://github.com/moonbitlang/moon/blob/b9f040ed3aacccb81975151d670d6084ba3cbfb3/crates/moon/src/cli/generate_test_driver.rs#L211, which **searches a string** to determine whether only tests without args are used, while the string being searched is actually **a struct converted to string**. 

The code does not disambiguate between the presence **tests** with args and **benches** with args, while the later code only generates test **or** benches based on which kind of test the user wants to run. This caused any test or bench with arg to trigger the usage of `with_args_driver_template` even if the corresponding kind of test is not actually used.

I will try to fix it by refactoring the test driver generator to avoid such issues.

lynzrand: A fix can be found at https://github.com/moonbitlang/moon/pull/943

PR #2471, opened 2025-07-21
feat: Add `is_some` for option

HCHogan:


peter-jerry-ye:
We'll add this function but mark it as

```moonbit
#deprecate("use _ is Some(_) instead" )
```

illusory0x0:
> We'll add this function but mark it as
> 
> ```moonbit
> #deprecate("use _ is Some(_) instead" )
> ```

Using `_` as a variable name does not clearly express the intent, `x` is clearer.

`_` is wildcard which has special meaning in moonbit's pattern match.

```diff
- #deprecate("use _ is Some(_) instead" )
+ #deprecate("use `x is Some(_)` instead" )
```

PR #2474, opened 2025-07-22
refactor(strconv): accept string view for pub api

peter-jerry-ye:
and replace the string slice by string view.

TODO : Update the `FromStr` trait to accept the `@string.View` as well.


strconv/number.mbt
```diff
@@ -23,56 +23,46 @@ priv struct Number {
...
+      len += 1
+      x = x * 10UL +
+        UInt64::extend_uint((ch.to_int() - '0').reinterpret_as_uint()) // no overflows here
+      continue rest
     }
-    len += 1
-    x = x * 10UL + UInt64::extend_uint(to_digit(s[0]).reinterpret_as_uint()) // no overflows here
-    s = s.step_1_unchecked()
+    ['_', .. rest] => continue rest
+    s => return (s, x, len)
```
peter-jerry-ye, line 48: This is the only place where the new implementation differs from the original, as the original one looks like a bug to me where it didn't check again whether the first char is still a digit after having consumed a `-`. The test coverage suggests that such behavior was never tested.

PR #2488, opened 2025-07-25
feat: add min/max built-in function

quake:
This pull request introduces enhancements to the Compare trait by adding new methods for determining the maximum and minimum of two values.

Similar helper functions are commonly found in other modern programming languages, [rust](https://doc.rust-lang.org/std/cmp/trait.Ord.html#method.min) for example.

These methods serve as convenient and expressive shortcuts in everyday code, eliminating the need for verbose conditional logic like if-else or ternary expressions just to pick the smaller or larger of two values. They also improve readability and maintainability, especially when used in chains, comparisons, or when embedded in more complex expressions.

### DIFF
--- a/builtin/builtin.mbti
+++ b/builtin/builtin.mbti
@@ -594,6 +594,8 @@ impl BitXOr for UInt64
 
 pub(open) trait Compare : Eq {
   compare(Self, Self) -> Int
+  max(Self, Self) -> Self = _
+  min(Self, Self) -> Self = _
 }
 impl Compare for Bool
 impl Compare for Byte
--- a/builtin/traits.mbt
+++ b/builtin/traits.mbt
@@ -20,13 +20,34 @@ pub(open) trait Eq {
 
 ///|
 /// Trait for types whose elements are ordered
-///
-/// The return value of [compare] is:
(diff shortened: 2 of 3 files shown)


peter-jerry-ye:
Hi @quake,

Thank you for submitting this PR! The suggestion to add convenience methods to the Compare trait is valuable and would indeed improve code readability. However, after careful consideration, we've decided not to merge this PR at this time, based on the following considerations:

1. Existing functionality coverage. We already have @cmp.maximum and @cmp.minimum functions that serve this purpose well and adequately address the current use cases.

2. Trait stability concerns. As a foundational trait in our core library, the Compare trait needs to maintain high stability. Once trait interfaces change, that can't be undone easily (we are already struggling with the modifications around `Show`). We prefer to be conservative with trait design.

We greatly appreciate your contribution and welcome continued discussion on this topic. If there's strong community demand for this feature, we can reconsider in the future (e.g. around 1.0 release).

Thank you again for your time and effort in contributing to the project!

quake:
Thank you for the detailed explanation, I completely understand the need to maintain stability in core traits like `Compare`, will keep an eye on any future discussions around this topic

PR #2493, opened 2025-07-25
feat: add next_power_of_two fn

quake:
This PR adds a `next_power_of_two` fn to the `Int`, which is is particularly useful in systems programming and performance-sensitive applications. Two common use cases are:
1. Memory Allocation Alignment
2. HashMap Capacity Management

This PR also removed duplicate fn `power_2_above` in the codebase by utilizing this new fn

### DIFF
--- a/hashmap/hashmap.mbt
+++ b/hashmap/hashmap.mbt
@@ -13,31 +13,6 @@
 // limitations under the License.
 
 ///|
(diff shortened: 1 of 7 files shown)


bobzhang:
@quake thanks for your contribution, I tweaked the semantics of next_power_of_two per cursor's suggestions

PR #2497, opened 2025-07-26
[feature test] cfg test not working with test

bobzhang:


bobzhang:
this look harmless change

```
/Users/runner/work/core/core/target/native/debug/test/builtin/builtin.blackbox_test.c:573566: error: '_$$3c$String$2a$Int$2a$Bool$2a$Double$2a$Char$2a$Int$2a$String$2a$Bool$2a$Double$2a$Char$2a$UInt64$2a$UInt$2a$Int64$2a$Char$2a$Double$2a$$3c$String$2a$$moonbitlang$core$builtin$Array$3c$UInt64$3e$$3e$$3e$$as_$moonbitlang$core$builtin$ToJson$static_method_' defined twice
Failed to run the test: 
```

Looks like a bug already fixed, will retry later

PR #2513, opened 2025-07-30
fix: blit_from_string bound check bug

quake:
The boundary check in `FixedArray::blit_from_string` was incorrect, it allowed operations that would later cause array bounds violations. However, I haven't found how to create a unit test that can verify the guard condition: a guard statement without a return value will panic the whole program, but unit tests don't seem to have a way to assert what the panic is.


peter-jerry-ye:
You just need to define a test whose name has `panic` as prefix. E.g.

```
test "panic test" {}
```

PR #2516, opened 2025-07-31
tentative adding yaml support

bobzhang:


json/json.mbt
```diff
@@ -159,6 +159,145 @@ pub fn stringify(
...
+      return true
+    }
+
+    // Check if string looks like a number, boolean, or null
+    match s {
+      "true" | "false" | "null" | "~" => true
+      _ => {
+        // Check if it starts with special characters
+        let first_char = s[0]
+        if first_char == '-' ||
```
Lampese, line 177: May be this style is better?
```moonbit
if first_char
    is ('-'
    | ':'
    | '['
    | '{'
    | '!'
    | '&'
    | '*'
    | '|'
    | '>'
    | '\''
    | '"'
    | '#'
    | '@'
    | '`'){
    return true
}
```

PR #2538, opened 2025-08-05
fom_json flat trigger warnings

bobzhang:
```
Error: [0011] Error (warning): Partial match, some hints:
True
Null
False
error: failed when checking
```


lynzrand:
Bug. Will fix shortly.

PR #2544, opened 2025-08-07
add ToStringView and use it in Array::join

bobzhang:


peter-jerry-ye:
Shouldn't we just have a local sealed trait:

```moonbit
pub trait JoinStringLike {
  write_to(Self, StringBuilder)
  length(Self) -> Int
}

impl JoinStringLike for String with xxx
impl JoinStringView for @string.View with xxx

pub fn[T: JoinStringLike] join(self : Array[T], separator : @string.View) -> String {
  self[:].join(separator)
}
pub fn[T: JoinStringLike] View::join(self : ArrayView[T], separator : @string.View) -> String {
  ...
}
```

The current implementation have extrat cost for bound check during `write_substring`, and the current size_hint computation potentially introduces extra conversions

Yu-zh:
> The current implementation have extrat cost for bound check during `write_substring`, and the current size_hint computation potentially introduces extra conversions

I think the extra bound check cost is because we don't have `write_view` method.

string/string_like.mbt
```diff
@@ -0,0 +1,33 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+///|
+trait ToStringView {
+  to_string_view(Self) -> StringView
+}
+
+///|
+pub fn[T : ToStringView] to_stringview(x : T) -> StringView {
```
Yu-zh: I don't see the point of this wrapper. I think we can just use dot method like `s.to_string_view()`.

bobzhang:
> pub fn[T: JoinStringLike] join(self : Array[T], separator : @string.View) -> String {
  self[:].join(separator)
}
pub fn[T: JoinStringLike] View::join(self : ArrayView[T], separator : @string.View) -> String {
  ...

We did not go this route since there is a mental overhead interpreting `JoinStringLike`, the performance is marginal under this case. The current `join` is dead in-efficient, we will come with a better implementation later, you can also give it a try(using o3 model)

builtin/stringbuilder_buffer.mbt
```diff
@@ -129,3 +129,14 @@ pub impl Show for StringBuilder with output(self, logger) {
 pub fn StringBuilder::reset(self : StringBuilder) -> Unit {
   self.len = 0
 }
+
+// pub fn StringBuilder::write_stringview(
+//   self : StringBuilder,
```
bobzhang, line 134: note this is commented due to cycles

PR #2545, opened 2025-08-07
hongbo/flip order

bobzhang:
- **tweak the failure message so that it is easy to split using `FAILED:`**


peter-jerry-ye:
This is a breaking change and will block the `moon test --update` @lynzrand

PR #2547, opened 2025-08-07
feat: unify hash*

peter-jerry-ye:
Currently there are four hash related mutable data structures: [linked]hash(set/map). Each of them are similar to one another but different in some aspect.

This PR tries to merge shared codebase.

### DIFF
--- a/hashmap/hashmap.mbt
+++ b/hashmap/hashmap.mbt
@@ -39,6 +39,7 @@ pub fn[K, V] new(capacity~ : Int = 8) -> T[K, V] {
     capacity,
     entries: FixedArray::make(capacity, None),
     capacity_mask: capacity - 1,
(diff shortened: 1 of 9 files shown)


hashmap/hashmap.mbt
```diff
@@ -565,7 +493,9 @@ test "remove_unexist_key" {
 
 ///|
 test "clear" {
-  let m : T[MyString, Int] = of([("a", 1), ("b", 2), ("c", 3)])
+  let m : T[MyString, Int] = of(
+    ([("a", 1), ("b", 2), ("c", 3)] : FixedArray[(MyString, Int)]),
+  )
```
peter-jerry-ye, line 498: It failed to induce the type using tuple struct. Could be a bug @Yu-zh

PR #2552, opened 2025-08-10
clean up

bobzhang:


bytes/view.mbt
```diff
@@ -60,6 +57,9 @@ pub fn View::length(self : View) -> Int {
   self.len()
```
bobzhang, line 57:
```
pub fn View::length(self : View) -> Int = "%bytesview.len"
```
This will result in a crash @Guest0x0 for testing

PR #2554, opened 2025-08-11
feat: impl Hash for @bytes.View

tonyfettes:


Yu-zh (approved):

peter-jerry-ye:
Notice that this will make the `Bytes` and `@bytes.View` have different hash even if they share the same content:

```moonbit
test "impl Hash for View" {
  let b0 : Bytes = "12345678"
  inspect(b0.hash(), content="-1980736859")
  inspect(b0[:].hash(), content="-1397216998")
  inspect(b0[:] == b0, content="true")
}
```

bytes/view.mbt
```diff
@@ -633,3 +633,10 @@ pub fn View::to_bytes(self : View) -> Bytes {
   bytes.blit_from_bytes(0, self.bytes(), self.start_offset(), self.length())
   unsafe_to_bytes(bytes)
 }
+
+///|
+pub impl Hash for View with hash_combine(self : View, hasher : Hasher) {
+  for i in 0..<self.length() {
+    hasher.combine_byte(self.unsafe_get(i))
```
bobzhang, line 640:
```
let b : Bytes = ..
let b1 = b[:]
```
```
let s : String = ...
let s1 : StringView = ...
```
```
impl Hash for View with hash(self : View) {
   
}
```
cc @peter-jerry-ye

PR #2559, opened 2025-08-12
feat: align sets APIs

peter-jerry-ye:
This PR refactors the code base to reduce the maintenance cost by
1. Align the implementation between linked hash set and map
2. Align the implementation between linked hash set and hash set

This PR also adds the APIs for both linked hash set and hash set such that they share a similar interface.


peter-jerry-ye:
The `insert` remains deprecated because it might be difficult for people to tell the difference between `add` and `insert`.

PR #2561, opened 2025-08-12
refactor: remove direct hash impl

peter-jerry-ye:
Previously, we only had `hash` for `Hash` trait, until #713. The implementations made before them implemented through `hash` directly, which raised an inconsistency issue between the calculation through `hash_combine` and `hash`. This PR tries to show this issue by removing all the direct hash implementation.


peter-jerry-ye:
The current hehaviors are intended for performance. Closing this PR.

PR #2580, opened 2025-08-19
simplify(coverage)

FlyCloudC:
- simplifies the implementation
- add a test for empty case

### DIFF
--- a/coverage/coverage.mbt
+++ b/coverage/coverage.mbt
@@ -16,16 +16,14 @@
 /// The `CoverageCounter` structure is used for keeping track of the number of
 /// times each chunk of code is executed. It's not very useful outside of
 /// generated code.
-struct CoverageCounter {
-  counter : FixedArray[UInt]
-}
+struct CoverageCounter(FixedArray[UInt])
 
 ///|
 /// Create a new coverage counter with the given size.
 ///
 #coverage.skip
 pub fn CoverageCounter::new(size : Int) -> CoverageCounter {
-  { counter: FixedArray::make(size, 0) }
+  CoverageCounter(FixedArray::make(size, 0))
 }
 
 ///|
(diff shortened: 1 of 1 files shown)


coverage/coverage.mbt
```diff
@@ -16,73 +16,43 @@
...
-  logger.write_char('[')
-  for i in 0..<self.counter.length() {
-    if i != 0 {
-      logger.write_string(", ")
-    }
-    Show::output(self.counter[i], logger)
-  }
-  logger.write_char(']')
+  self.0.output(logger)
 }
```
lynzrand: This puts a dependency on the `Show` implementation of `(Array[Int] as Show)::output` being the same as JSON array's format. Although we are enforcing it via the tests below, shouldn't we at least add a note or something :)

lynzrand (approved):
LGTM as long as the output format doesn't change.

coverage/coverage.mbt
FlyCloudC: done

bobzhang:
this package was supposed to have minimal deps, let's not change it unless we have clear benefits

lynzrand:
> this package was supposed to have minimal deps, let's not change it unless we have clear benefits

AFAICT this does not change the dependency list of the package, but still makes it somewhat more readable?

---

Okay, maybe this comment is referring to the `Array::show` thing, which, in that case, I consider both sides reasonable -- rely less on "current" stdlib behavior that we don't have a guarantee on is definitely a good thing. The other parts are still reasonable changes, though.

peter-jerry-ye:
We will only accept the first commit ebdf4b506d508ad1e473b5774c925467754cbcce.

FlyCloudC:
> We will only accept the first commit ebdf4b506d508ad1e473b5774c925467754cbcce.

Should I submit a revert commit on this branch?

peter-jerry-ye:
> Should I submit a revert commit on this branch?

I think a simple `git rebase -i main` should do the trick to drop the other commits.

PR #2585, opened 2025-08-21
refactor: start api migrations

peter-jerry-ye:
The first commit add deprecation warnings for immutable maps. This start a migration phase and they would have the same APIs as the mutable ones after the migration.

The rest of the APIs adjust the constructor APIs, allowing the existence of `T::from_array`, which might be more intuitive to use when trying to define new instances.


peter-jerry-ye:
Wait for #as_free_fn @Guest0x0

PR #2586, opened 2025-08-21
Remove deprecated stringview

Yu-zh:


bigint/bigint_nonjs.mbt
```diff
@@ -373,8 +374,9 @@ fn BigInt::grade_school_mul(self : BigInt, other : BigInt) -> BigInt {
   { limbs, sign: Positive, len }
 }
 
-///|
 // Karatsuba multiplication
```
bobzhang, line 377: FTR: we should warn about the dangling comments(opt-in)?

PR #2590, opened 2025-08-24
hongbo/buffer

bobzhang:
- **adopt @buffer.Buffer**
- **Add ULEB128 encoding implementation for unsigned integers**
- **feat(buffer): implement ULEB128 encoding for unsigned integers**


buffer/sleb128.mbt
```diff
@@ -0,0 +1,77 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+///|
+trait Signed {
+  output_sleb128(Self, Buffer) -> Unit
+}
```
bobzhang, line 18: cc @peter-jerry-ye  we may only introduce one trait Leb128 and implement Int/UInt for it according to its signedness instead of two traits
```
trait Leb128 {
   output(Self, Buffer) -> Unit
}
```

peter-jerry-ye: Updated in #2592

PR #2593, opened 2025-08-25
feat(hashmap): add retain

peter-jerry-ye:


builtin/linked_hash_map.mbt
```diff
@@ -727,6 +735,51 @@ pub fn[K, V] Map::copy(self : Map[K, V]) -> Map[K, V] {
...
+///   inspect(map.size(), content="2")
+///   inspect(map.get("a"), content="None")
+///   inspect(map.get("b"), content="Some(2)")
+///   inspect(map.get("c"), content="None")
+///   inspect(map.get("d"), content="Some(4)")
+/// ```
+#locals(f)
+pub fn[K : Eq, V] Map::retain(
+  self : Map[K, V],
+  f : (K, V) -> Bool raise?,
```
bobzhang: We should be careful here, the current implementation does not look very efficient. 
Maybe we should not allow raise for in place operations

peter-jerry-ye: The implementations are updated with inplace modifications.

PR #2598, opened 2025-08-27
refactor: adjust fixed-array to bytes conversion

### LINKED ISSUE #2596: [performance] Bytes::from_fixedarray() should be a simple memcpy or no-op on native/llvm backend
For now the `Bytes::from_fixedarray` allocates closures because it uses `Bytes::makei`:

```moonbit
pub fn Bytes::from_fixedarray(arr : FixedArray[Byte], len? : Int) -> Bytes {
  let len = match len {
    None => arr.length()
    Some(x) => x
  }
  Bytes::makei(len, i => arr[i])
}
```

Using simple memcpy or, better, no-op will be much better as there will be no allocate when writing each individual element.

peter-jerry-ye:
Closes #2596 

This PR let the `FixedArray` to `Bytes` conversion use `FixedArray::blit` for better performance


tonyfettes (approved):

bytes/bytes.mbt
```diff
@@ -171,15 +173,17 @@ pub fn from_iter(iter : Iter[Byte]) -> Bytes {
 ///     ),
 ///   )
 /// ```
-/// TODO: marked as intrinsic, inline if it is constant
 pub fn Bytes::of(arr : FixedArray[Byte]) -> Bytes {
-  Bytes::makei(arr.length(), i => arr[i])
+  let len = arr.length()
+  let result = FixedArray::make(len, Byte::default())
```
peter-jerry-ye: Remove the init step

peter-jerry-ye:
@bobzhang Should we refactor the Buffer such that it uses `UninitializedArray` internally, similar to `Array`?

PR #2599, opened 2025-08-27
refactor(math): use libc trig functions

Yu-zh:


math/trig_double_nonjs.mbt
```diff
@@ -34,7 +34,11 @@
 /// inspect(@math.sin(@double.infinity), content="NaN")
 /// inspect(@math.sin(@double.neg_infinity), content="NaN")
 /// ```
-pub fn sin(x : Double) -> Double {
+pub fnalias sin_ as sin
```
peter-jerry-ye: Why do we need the extra `fnalias`?

Yu-zh: to avoid repeated docstring

PR #2603, opened 2025-08-27
test(buffer): remove duplicate tests

FlyCloudC:


buffer/buffer_test.mbt
```diff
@@ -62,98 +62,6 @@ test "expect method with matching content" {
...
-  inspect(buf, content="Test")
-}
-
-///|
-test "expect method with non-matching content" {
-  let buf = @buffer.new(size_hint=100)
-  buf.write_string("Test")
-  inspect(buf, content="Test")
-}
-
```
peter-jerry-ye, line 156: Did you accidentally removed this one?

FlyCloudC: No. It the same as test "expect method with matching content" at line 59

PR #2604, opened 2025-08-27
test(bool): remove duplicate tests

FlyCloudC:
The test "bool convert" duplicates doc comment.


peter-jerry-ye:
I personally think it's not **that** bad to repeat the doc test since they serve different purposes and are organized differently.

FlyCloudC:
> I personally think it's not **that** bad to repeat the doc test since they serve different purposes and are organized differently.

Okay. Thank you for your suggestion, I will close this PR.

Issue #2137, opened 2025-05-19
[Consistency Review] Aligning `Array` and `@deque.T` APIs


illusory0x0:
## **Current behaviors**

Most APIs are similar but some are inconsistent.

```moonbit
///|
test {
  let xs = [1, 2, 3, 4, 5]
  let ys = @deque.of([1, 2, 3, 4, 5])

  // push
  xs.push(0)
  ys.push_back(0)

  // truncate
  // Array missing `truncate` API
  ys.truncate(3)

  // get
  inspect!(xs[0], content="1")
  inspect!(ys[0], content="1")

  // set
  xs[0] = 0
  ys[0] = 0

  // join
  inspect!(xs.map(Int::to_string(_)).join(","), content="0,2,3,4,5,0")
  // @deque missing `join`

  // sort 
  xs.sort()
  // @deque missing `sort`
  // view 
  inspect!(xs[:], content="[0, 0, 2, 3, 4, 5]")
  inspect!(ys.as_views(), content="([0, 2, 3], [])")
  // https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.make_contiguous

  // append
  xs.append([1, 2, 3, 4])
  // @deque missing append

  // shrink_to_fit
  ys.shrink_to_fit()
  xs.shrink_to_fit()

  // drain
  inspect!(xs.drain(0, xs.length()), content="[0, 0, 2, 3, 4, 5, 1, 2, 3, 4]")
  // @deque missing drain

  // insert
  xs.insert(0, 99)
  // @deque missing insert

  // reserve_capacity
  xs.reserve_capacity(100)
  ys.reserve_capacity(100)

  // flatten
  {
    let xs = [[1, 2, 3], [4, 5, 6]]
    inspect!(xs.flatten(), content="[1, 2, 3, 4, 5, 6]")
    let ys = @deque.of([[@deque.of([1, 3, 4]), @deque.of([3, 4, 5])]])
    // @deque missing flatten

  }

  // blit_to
  xs.blit_to([], len=xs.length(), src_offset=0, dst_offset=0)
  // @deque missing blit_to

  // binary_search and binary_search_by
  xs.sort()
  xs.binary_search(5) |> ignore
  xs.binary_search_by(fn { x => 0 }) |> ignore
  // @deque missing binary_search

  // retain
  xs.retain(fn { x => x % 2 == 0 })
  ys.retain(fn { x => x % 2 == 0 })

  // retain_map
  // Array missing retain_map
[...]

PR #2597, opened 2025-08-26
Aligning array and deque api: Add deque.T::append

FrozenLemonTee:

PR #2608, opened 2025-08-28
Aligning array and deque api: Add deque.T::append

FrozenLemonTee:
Issue: #2137

### DIFF
--- a/deque/deque.mbt
+++ b/deque/deque.mbt
@@ -126,6 +126,52 @@ pub fn[A] copy(self : T[A]) -> T[A] {
 }
 
 ///|
+/// Appends all elements from one deque to the end of another deque. The elements
+/// are added in-place, modifying the original deque.
+///
+/// Parameters:
+///
+/// * `self` : The deque to append to.
+/// * `other` : The deque whose elements will be appended.
+///
+/// Example:
+///
+/// ```moonbit
+/// let v1 = @deque.of([1, 2, 3])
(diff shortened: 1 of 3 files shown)


FrozenLemonTee:
@peter-jerry-ye
Change to #2597:
1. Calculate the actual logical space when judging reallocation.
2. Clarify the intent of the test case and add as_views as the standard for test assertions.

deque/deque.mbt
```diff
@@ -125,6 +125,55 @@ pub fn[A] copy(self : T[A]) -> T[A] {
...
+/// let v2 = @deque.of([4, 5, 6])
+/// v1.append(v2)
+/// inspect(v1, content="@deque.of([1, 2, 3, 4, 5, 6])")
+/// let v1 = @deque.of([1, 2, 3])
+/// let v2 = @deque.of([])
+/// v1.append(v2)
+/// inspect(v1, content="@deque.of([1, 2, 3])")
+/// ```
+pub fn[A] append(self : T[A], other : T[A]) -> Unit {
+  guard other.len != 0 else {  }
```
bobzhang: FTR: this is correct, from the language design point of view, this is a code smell, we may prefer user to write

```
guard other.len != 0 else { return}
```

```
guard cond else { branch1}
... continuation
```
==>
```
if cond { ...contin }
else branch1
```
We may need lint such style

FrozenLemonTee: Thanks for the correction.I have fixed it.

PR #2610, opened 2025-08-28
feat: configurations for copilot

peter-jerry-ye:


bobzhang:
@peter-jerry-ye is this still relevant?

peter-jerry-ye:
> @peter-jerry-ye is this still relevant?

I think a clear Agents.md is better than using these small instructions in general, and the recent AIs do not seem to need these helpers.

PR #2623, opened 2025-08-29
hongbo/to json for bytes

bobzhang:
- **refactoring**
- **feat(json): add ToJson implementation for Bytes type**


peter-jerry-ye:
Do we need special treatment for alpha numerics within ascii?

peter-jerry-ye:
The from/to implementations are implemented

PR #2625, opened 2025-08-29
feat(buffer): add write char and string with encoding

peter-jerry-ye:


buffer/buffer.mbt
```diff
@@ -594,8 +594,149 @@ pub fn write_bytesview(self : Buffer, value : @bytes.View) -> Unit {
...
+
+///|
+pub fn Buffer::write_string_utf8(buf : Self, string : @string.View) -> Unit {
+  for ch in string {
+    buf.write_char_utf8(ch)
+  }
+}
+
+///|
+#alias(write_stringview, deprecated="use write_string_utf16 instead")
```
hackwaly: Where is `write_string_utf16`? Is it a typo?

hackwaly (approved):

buffer/pkg.generated.mbti
```diff
@@ -24,9 +24,15 @@ fn Buffer::is_empty(Self) -> Bool
 fn Buffer::length(Self) -> Int
 fn Buffer::reset(Self) -> Unit
 fn Buffer::to_bytes(Self) -> Bytes
+fn Buffer::write_bom_utf16be(Self) -> Unit
+fn Buffer::write_bom_utf16le(Self) -> Unit
+fn Buffer::write_bom_utf8(Self) -> Unit
```
bobzhang: let's not expose *bom* functions, this can be wrapped in a high level API, it is not 
performance sensitive either since they are called only once

peter-jerry-ye: Done.

PR #2627, opened 2025-08-29
implement the FJS string matching algorithm for Bytes & String in moonbitlang/core

0Ayachi0:
The figures below show the benchmark data of the old and new implementations, obtained by running the performance benchmark file "fjs_heavy_benchmark.mbt". "FJS_IMPLEMENTATION_SUMMARY.md" is the work summary.
4bb1c0bc-796a-4a1e-916c-73d88e774382


fjs_heavy_benchmark.mbt
```diff
@@ -0,0 +1,263 @@
+// Copyright 2025 International Digital Economy Academy
```
Lampese: What is the significance of Benchmark appearing in the standard library? You can mention the results of the benchmark in the "necessity" section of the PR, but there is no need to include the testing code as it has nothing to do with "correctness".

FJS_IMPLEMENTATION_SUMMARY.md
```diff
@@ -0,0 +1,172 @@
+# FJS 算法实现总结
```
Lampese: I am puzzled as to why this file appears here. Can you provide some explanation? If it is used as an explanation of the algorithm, does the information mentioned in the "概述" section contribute to what we are doing?

If you are using AI generated code, please verify at least whether it meets a rigorous project contribution standard.

Lampese (changes requested):
There is no problem with implementing the algorithm itself.

- Please revise some details, as I have already mentioned in the review comments.
- Please add more testing for Edge Cases to the new method to ensure foolproof performance.

bytes/moon.pkg.json
```diff
@@ -7,6 +7,8 @@
     "moonbitlang/core/uint64",
     "moonbitlang/core/test",
     "moonbitlang/core/quickcheck",
-    "moonbitlang/core/error"
+    "moonbitlang/core/error",
+    "moonbitlang/core/bench",
```
Lampese, line 11: Import `bench` and `string` for?

peter-jerry-ye:
The benchmark screenshot looks like it's vibed. We may need a proper moon bench result.

bytes/methods.mbt
```diff
@@ -13,21 +13,107 @@
...
+      k += 1
+    }
+    if j == pattern_len {
+      return Some(i) // Found match
+    }
+
+    // Mismatch occurred, use KMP shift or Boyer-Moore shift, whichever is larger
+    let kmp_shift = if j == 0 { 1 } else { j - border[j - 1] }
+    let bm_shift = 1 // Minimum shift for Boyer-Moore when last char matched
+    i += if kmp_shift > bm_shift { kmp_shift } else { bm_shift }
```
Guest0x0, line 114: I don't think this is correct. If `j > 0` after KMP match failure, we should retry KMP matching starting from `j <- border[j - 1]`. But the code here will repeat the comparison of `pattern[0:border[j-1]]`. This may lead to `O(n * m)` complexity in the worst case, for example when matching `a*ba` against `a*`.

If `j > 0`, we should just retry the KMP match with updated `j`, until `j == 0`.

bytes/methods.mbt
```diff
@@ -13,21 +13,107 @@
...
+    if target_byte != last_pattern_byte {
+      // Bad character skip
+      let skip = skip_table[target_byte.to_int()]
+      i += skip
+      continue
+    }
+
+    // Last character matches, now use KMP-style matching from left
+    let mut j = 0 // position in pattern
+    let mut k = i // current position in target
```
Guest0x0, line 100: There is no need for `k` here, we should just bump `i` directly. See the comments below at the end of loop

bobzhang:
For the core library, we need make a balance between efficiency and maintainability at this stage, @0Ayachi0 you are more than welcome to publish a separate package based on your findings, we may revisit this when MoonBit gets more stable in the future

PR #2628, opened 2025-08-30
hongbo/trait name migrations

bobzhang:
- **add duplicated names for migration**
- **fix all warnings**
- **update mi file**

```
Error: Moonc.Basic_hash_string.Key_not_found("$@moonbitlang/core/builtin.Add::Int16::op_add")
```
The compiler may hard coded Int16::op_add operator etc


Guest0x0:
Yes the ICE is due to the compiler hard-coding `op_add` for `Int16`. This and other similar issues can be circumvented by implementing *both* `op_add` and `add` for `Int16` etc. at first.

bobzhang:
remaining operators:

`op_notequal`, `op_le`

PR #2629, opened 2025-08-31
hongbo/immut map name

bobzhang:
- **tweak tojson for sorted_map**
- **change name to @sorted_map.Map**

Test failure due to drivers

### DIFF
--- a/immut/sorted_map/README.mbt.md
+++ b/immut/sorted_map/README.mbt.md
@@ -11,7 +11,7 @@ key-value pair using `singleton()`.
 
 ```moonbit
 test {
-  let map1 : @sorted_map.T[String, Int] = @sorted_map.new()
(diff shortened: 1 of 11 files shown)


bobzhang:
shadowing builtin names is a bad idea. for test driver we can be more conservative to use `@builtin.Map` cc @lynzrand

lynzrand:
> shadowing builtin names is a bad idea. for test driver we can be more conservative to use `@builtin.Map` cc @lynzrand

Could you please elaborate more? AFAICT all types in test drivers are fully qualified names like this:

```
typealias @moonbitlang/core/builtin.Map[
  String,
  @moonbitlang/core/builtin.Map[
    Int,
    (() -> Unit raise Error, @moonbitlang/core/builtin.Array[String]),
  ],
] as Moonbit_Test_Driver_Internal_No_Args_Map
```

PR #2636, opened 2025-08-31
hongbo/finish types renaming

bobzhang:
- **refactor(sorted_set): update deprecation message to suggest using `SortedSet` instead of `T`**
- **refactor(sorted_map): rename type T to SortedMap and update function signatures for consistency**


bobzhang:
cc @peter-jerry-ye

PR #2637, opened 2025-08-31
fix(immut/hashmap): hash_combine

### LINKED ISSUE #2606: [BUG] Hash for immut hashmap
In cases where hash collisions occur for elements, different insertion orders result in `immut/hashmap` instances having different hash values, even though they are `op_equal`. This violates the hash contract.

```moonbit
test "hash" {
  let s1 = MyString("1")
  let s2 = MyString("2")
  inspect(s1.hash() == s2.hash(), content="true") // collisions

  let mapa = new().add(s1, 1).add(s2, 2)
  let mapb = new().add(s2, 2).add(s1, 1)
  inspect(mapa == mapb, content="true")
  inspect(mapa.hash() == mapb.hash(), content="false")
}
```

Relevant code snippet

https://github.com/moonbitlang/core/blob/1d9d1720095e64c86fb5b7057954f11dd655c5d7/immut/hashmap/HAMT.mbt#L721-L723

We might need an order-independent `hash_combine`. Alternatively, similar to how `Hash` is implemented for immut hashset — though I'm not sure if this is a good idea.

https://github.com/moonbitlang/core/blob/1d9d1720095e64c86fb5b7057954f11dd655c5d7/immut/hashset/HAMT.mbt#L370-L372

FlyCloudC:
fix #2606

### DIFF
--- a/immut/hashmap/HAMT.mbt
+++ b/immut/hashmap/HAMT.mbt
@@ -754,5 +754,5 @@ pub impl[K : Hash, V : Hash] Hash for HashMap[K, V] with hash_combine(
   self,
   hasher,
 ) {
-  self.each((k, v) => hasher..combine(k)..combine(v))
+  hasher.combine(self.fold_with_key(init=0, (acc, k, v) => acc ^ (k, v).hash()))
 }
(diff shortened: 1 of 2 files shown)


peter-jerry-ye:
Could you please add a test to show the bug before the fix? It seems that the current tests didn't cover this case.

peter-jerry-ye:
Thanks

PR #2641, opened 2025-09-01
feat: utf8 encode / decode

peter-jerry-ye:
This PR adds simple encode/decode to/from utf8.

Based on #2625

### DIFF
--- a/encoding/utf8/decode.mbt
+++ b/encoding/utf8/decode.mbt
@@ -21,7 +21,15 @@ pub suberror Malformed @bytes.View
 const U_REP = '\u{FFFD}'
 
 ///|
-pub fn decode(bytes : @bytes.View) -> String raise Malformed {
+pub fn decode(
+  bytes : @bytes.View,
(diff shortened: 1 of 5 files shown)


encoding/utf8/decode.mbt
```diff
@@ -0,0 +1,179 @@
...
+
+///|
+#cfg(target="js")
+extern "js" fn decode_lossy_js(
+  bytes : Bytes,
+  start : Int,
+  length : Int,
+) -> String =
+  #|(bytes, start, length) => 
+  #|  new TextDecoder("utf-8", { fatal: false }).decode(new Uint8Array(bytes, start, length))
```
hackwaly: TextDecoder if not in Js standard. It's in web standard. Should provide fallback implementation.

peter-jerry-ye: Then, maybe just drop the js implementation at all?

hackwaly: It's ok. Optimization can be considered later.

PR #2646, opened 2025-09-02
refactor(buffer): remove initial_data from Buffer struct

illusory0x0:


buffer/buffer.mbt
```diff
@@ -735,7 +734,6 @@ pub fn write_iter(self : Buffer, iter : Iter[Byte]) -> Unit {
 ///   inspect(buf.is_empty(), content="true")
 /// ```
 pub fn reset(self : Buffer) -> Unit {
-  self.data = self.initial_data
   self.len = 0
 }
```
peter-jerry-ye, line 841: The size of the buffer is not shrinked and the data are not cleaned up in this case.

illusory0x0: `array.clear()` also doesn't shrink underlying memory.

if you want get a smaller buffer, I think create new buffer would be better.

peter-jerry-ye:
You need to update the doc comment as well

illusory0x0:
> You need to update the doc comment as well

Doc comment has been updated.

buffer/buffer.mbt
```diff
@@ -718,8 +717,7 @@ pub fn write_iter(self : Buffer, iter : Iter[Byte]) -> Unit {
...
 
 ///|
-/// Resets the buffer to its initial state by restoring its initial capacity and
-/// clearing its contents.
+/// Resets the buffer to an empty state by clearing all its contents.
```
peter-jerry-ye: I'm not quite sure if it is correct to say 'clearing all its contents' since the cleanup is not performed.

illusory0x0: Based on the test results from `inspect`, it does indeed clear the content.

peter-jerry-ye: The tests are blackbox tests. If you execute the white box tests, you'll see that the contents remained.

illusory0x0: I've updated the documentation, but I don't think exposing internal implementation details in the documentation is a good idea. Personally, I believe it's sufficient to describe the behavior of black-box testing within the public API.

PR #2663, opened 2025-09-03
hongbo/no fn alias

bobzhang:
- **refactor: update rounding functions to use #as_free_fn and clean up aliases**
- **refactor: replace fnalias with #as_free_fn for ceil function across multiple files**
- **refactor: update math functions to use #as_free_fn and clean up aliases**

### DIFF
--- a/math/pkg.generated.mbti
+++ b/math/pkg.generated.mbti
@@ -60,6 +60,7 @@ fn expm1(Double) -> Double
 
 fn expm1f(Float) -> Float
 
+#as_free_fn
 fn floor(Double) -> Double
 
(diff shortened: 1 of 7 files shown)


math/pkg.generated.mbti
```diff
@@ -41,6 +41,7 @@ fn cbrt(Double) -> Double
 
 fn cbrtf(Float) -> Float
 
+#as_free_fn
 fn ceil(Double) -> Double
```
bobzhang, line 45: cc @Guest0x0 `#as_free_fn` should not be inherited

PR #2667, opened 2025-09-04
refactor: deprecate mutable operations in ArrayView and update documentation

bobzhang:
cc @Yu-zh

### DIFF
--- a/array/view.mbt
+++ b/array/view.mbt
@@ -310,12 +310,7 @@ pub fn[T, U] View::map(self : View[T], f : (T) -> U raise?) -> Array[U] raise? {
 ///|
 /// Maps a function over the elements of the array view in place.
 ///
-/// # Example
-/// ```mbt
-///   let v = [3, 4, 5]
-///   v[1:].map_inplace((x) => {x + 1})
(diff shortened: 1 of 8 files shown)


array/view.mbt
```diff
@@ -310,12 +310,7 @@ pub fn[T, U] View::map(self : View[T], f : (T) -> U raise?) -> Array[U] raise? {
...
 ///
-/// # Example
-/// ```mbt
-///   let v = [3, 4, 5]
-///   v[1:].map_inplace((x) => {x + 1})
-///   assert_eq(v, [3, 5, 6])
-/// ```
+#deprecated("ArrayView will be immutable, use Array if you need mutation")
 pub fn[T] View::map_inplace(self : View[T], f : (T) -> T raise?) -> Unit raise? {
   for i, v in self {
```
bobzhang, line 315: cc @Guest0x0 this is another case where you want deprecation not working in the same package, we heave a function a deprecated, and a function b deprecated (b called a), but we can not simply remove `b` during migration period (similar to recursive ones). The workaround is the same, rename `a` to a private function, mark the public interface deprecated

PR #2674, opened 2025-09-04
change return type of string op_get to UInt16

Yu-zh:
This PR changes the type of `String::{op_get, unsafe_charcode_at, get}` and `StringView::{op_get, unsafe_charcode_at, get}` to UInt16. This PR also moved some `UInt16`'s methods to builtin package to resolve dependency issues.

### DIFF
--- a/bigint/bigint_nonjs.mbt
+++ b/bigint/bigint_nonjs.mbt
@@ -1002,7 +1002,7 @@ pub fn BigInt::from_string(input : String) -> BigInt {
     if x < 0 || x > 9 {
       abort("invalid character")
     }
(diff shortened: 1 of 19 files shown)


Yu-zh:
This PR requires update in moon.

bobzhang:
We need move forward, but this is a serious breaking change, we need provide a migration path:

1. provide an extra API:
String::code_unit_at(i : Int) -> UInt16
2. deprecated existing and encourage them to migrate to code_unit_at
#alias("_[_]")
#alias(charcode_at, deprecated)
fn String::at(String, Int) -> Int
3. #alias code_unit_at #"_[_]"

PR #2679, opened 2025-09-05
hongbo/at get at unsafe

bobzhang:
- **refactor: deprecate string functions and update aliases**
- **add `at`, `at_unsafe`, `get`**

### DIFF
--- a/immut/hashmap/HAMT.mbt
+++ b/immut/hashmap/HAMT.mbt
@@ -77,7 +77,7 @@ fn[K : Eq, V] Node::get_with_path(
       }
     (Branch(children), path) => {
       let idx = path.idx()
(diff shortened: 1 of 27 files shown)


immut/hashmap/HAMT.mbt
```diff
@@ -87,7 +87,8 @@ fn[K : Eq, V] Node::get_with_path(
 
 ///|
 #deprecated("Use `get` instead. `op_get` will return `V` instead of `Option[V]` in the future.")
-pub fn[K : Eq + Hash, V] op_get(self : HashMap[K, V], key : K) -> V? {
+#alias(op_get)
```
bobzhang: we should remove this operator first

PR #2680, opened 2025-09-05
deprecate: `ok`, `err`, `is_ok`, `is_err`

illusory0x0:
`T?` is similar to `Result[T,E]`, so I deprecate **tiny** function.

see also [Option deprecated functions](https://github.com/moonbitlang/core/blob/main/option/deprecated.mbt).


result/deprecated.mbt
```diff
@@ -0,0 +1,23 @@
+///|
+#deprecated("inline instead for such tiny function")
+pub fn[T, E] ok(value : T) -> Result[T, E] {
+  Ok(value)
+}
+
+///|
+#deprecated("inline instead for such tiny function")
```
peter-jerry-ye:
```suggestion
#deprecated("Use`Err(value)` instead")
```

illusory0x0: here https://github.com/moonbitlang/core/blob/main/option/deprecated.mbt#L15-L41 also need to be updated.

PR #2682, opened 2025-09-05
deprecate: `Result::fold`

illusory0x0:
`Result::fold` is inconsistent with other `fold` function such as `Array::fold` and `Iter::fold`.

```moonbit
pub fn[T, E, V] fold(self : Result[T, E], ok : (T) -> V, err : (E) -> V) -> V
```

here is corrected MoonBit version

```moonbit
fn [A,E,B] Result::fold_(self : Result[A,E], init~ : B, f : (B,A) -> B) -> B {
  match self {
    Ok(value) => f(init, value)
    Err(_) => init
  }
}
```

Haskell `Either` instance of `Foldable` is similar to this.

here is an example usage for `Either` foldl in Haskell

```haskell
foldl (\acc x -> acc + x) 10 (Right 5)
-- Result: 15
```

### DIFF
--- a/result/pkg.generated.mbti
+++ b/result/pkg.generated.mbti
@@ -17,6 +17,7 @@ fn[T, E] ok(T) -> Result[T, E]
 // Types and methods
 fn[T, E, U] Result::bind(Self[T, E], (T) -> Self[U, E]) -> Self[U, E]
 fn[T, E] Result::flatten(Self[Self[T, E], E]) -> Self[T, E]
+#deprecated
 fn[T, E, V] Result::fold(Self[T, E], (T) -> V, (E) -> V) -> V
 #deprecated
 fn[T, E] Result::is_err(Self[T, E]) -> Bool
--- a/result/result.mbt
+++ b/result/result.mbt
@@ -178,31 +178,12 @@ test "bind" {
 }
 
 ///|
-/// Folds a `Result` into a single value.
-///
-/// If the `Result` is an `Ok`, the `ok` function is applied to the value. If the `Result` is an `Err`, the `err` function is applied to the value.
-/// # Example
-///
-/// ```mbt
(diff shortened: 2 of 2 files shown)


peter-jerry-ye:
It was following the semantic of [scala](https://www.scala-lang.org/api/3.7.1/scala/util/Either.html#fold-7e4)

illusory0x0:
> It was following the semantic of [scala](https://www.scala-lang.org/api/3.7.1/scala/util/Either.html#fold-7e4)

However, Scala's [Array](https://www.scala-lang.org/api/3.7.1/scala/Array.html) does not have a `fold`, so there is no API consistency issue.

peter-jerry-ye:
> > It was following the semantic of [scala](https://www.scala-lang.org/api/3.7.1/scala/util/Either.html#fold-7e4)
> 
> However, Scala's [Array](https://www.scala-lang.org/api/3.7.1/scala/Array.html) does not have a `fold`, so there is no API consistency issue.

I think you are looking for [this](https://www.scala-lang.org/api/3.7.1/scala/collection/ArrayOps.html#fold-fffff9e1).

illusory0x0:
> > > It was following the semantic of [scala](https://www.scala-lang.org/api/3.7.1/scala/util/Either.html#fold-7e4)
> > 
> > 
> > However, Scala's [Array](https://www.scala-lang.org/api/3.7.1/scala/Array.html) does not have a `fold`, so there is no API consistency issue.
> 
> I think you are looking for [this](https://www.scala-lang.org/api/3.7.1/scala/collection/ArrayOps.html#fold-fffff9e1).

Scala has `foldLeft` and `foldRight`, so there's no API consistency issue. In MoonBit, `fold` generally refers to `fold left`, while `fold_rev` typically denotes `fold right` within the core library.

Issue #2681, opened 2025-09-05
[Consistency Review] `Option` and `Result` API


illusory0x0:
**Current behaviors**

MoonBit `Option` has the following three helper functions, while `Result` does not. These APIs need to be aligned.

```moonbit 
fn[T] Option::unwrap_or(T?, T) -> T
fn[T : Default] Option::unwrap_or_default(T?) -> T
fn[T] Option::unwrap_or_else(T?, () -> T raise?) -> T raise?
```

**Expected behaviors**

```moonbit
fn[T,E] Result::unwrap_or(Result[T,E], T) -> T
fn[T : Default,E] Result::unwrap_or_default(Result[T,E]) -> T
fn[T,E] Result::unwrap_or_else(Result[T,E], () -> T raise?) -> T raise?
```

**Justification**

`Option[T]` and `Result[T, Unit]` are equivalent.

PR #2688, opened 2025-09-06
Add missing unwrap methods to Result type for API alignment with Option

echozyr2001:
**Problem:** The `Result` type was missing three helper functions that exist in `Option`, creating an inconsistency in the API:

* unwrap_or
* unwrap_or_else
* unwrap_or_default

**Solution:** This PR adds the three missing methods to align the `Result` API with `Option`:

* fn[T,E] Result::unwrap_or(Result[T,E], T) -> T - Returns the contained Ok value or the provided default
* fn[T,E] Result::unwrap_or_else(Result[T,E], () -> T raise?) -> T raise? - Returns the contained Ok value or calls the provided closure (lazily evaluated)
* fn[T : Default,E] Result::unwrap_or_default(Result[T,E]) -> T - Returns the contained Ok value or T::default()

**Justification:** Since `Option[T]` and `Result[T, Unit]` are equivalent, having consistent APIs between these types improves developer experience and maintains API coherence.

**Testing:** All three methods include comprehensive unit tests covering both Ok and Err cases.

close: #2681

### DIFF
--- a/result/pkg.generated.mbti
+++ b/result/pkg.generated.mbti
@@ -30,6 +30,9 @@ fn[T, E] Result::or_else(Self[T, E], () -> T) -> T
 fn[T, E] Result::to_option(Self[T, E]) -> T?
 fn[T, E] Result::unwrap(Self[T, E]) -> T
 fn[T, E] Result::unwrap_err(Self[T, E]) -> E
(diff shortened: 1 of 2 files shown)


echozyr2001:
PTAL @illusory0x0

PR #2705, opened 2025-09-09
Taking view from Fixedarray

Yu-zh:
This PR implements taking view from FixedArray.

Solves #1626


peter-jerry-ye:
When will `ArrayView` become immutable?

Yu-zh:
> When will `ArrayView` become immutable?

The `op_set` method is already deprecated

Issue #2706, opened 2025-09-09
[Consistency Review]Aligning api for deque


FrozenLemonTee:
**Current behaviors**
Original issue: #2137

**Expected behaviors**
TODO list of missing functions/traits in deque:
- [ ] Add
- [ ] Hash
- [ ] blit_to
- [ ] insert
- [ ] remove
- [ ] chunks
- [ ] chunk_by
- [ ] filter
- [ ] extract_if
- [ ] sort

PR #2649, opened 2025-09-02
Aligning array and deque api: Add deque::blit_to

ethereal-dream:

PR #2709, opened 2025-09-09
Aligning deque api blit_to

FrozenLemonTee:
Issue: #2706
Add functions:
blit_to


Lampese:
cc @Guest0x0 
- #2649 
Which is better?

Guest0x0:
the implementation in #2649 seems more efficient because it utilizes batch array operations

PR #2721, opened 2025-09-11
feat: make copy faster

peter-jerry-ye:


Yu-zh:
This implementation is better than before. But there's still an unnecessary pass of initializing the array. I think it should be implemented as intrinsics from the moonc side.

So I'm not sure whether we should move forward with this PR.

PR #2722, opened 2025-09-11
fix: explicitly impl bitwise operators for `Int` and `UInt`

Lil-Ran:
Before patching, the last line below will fail to compile, with the error ``Type UInt does not implement trait BitXOr: no `impl` is defined (E4018)``

```moonbit
fn[T: BitXOr] foo(x : T, y : T) -> T {
  x ^ y
}

test {
  let x: UInt = 1
  let y: UInt = 2
  inspect(x ^ y, content="3")
  inspect(foo(x, y), content="3")
}
```

Given that `Int` and `UInt` already implement other arithmetic operators explicitly, e.g., `pub impl Add for Int with add(self, other) = "%i32_add"`, I propose also implementing the bitwise operators.


peter-jerry-ye:
Thanks

PR #2723, opened 2025-09-12
feat: enable abort for native backend

peter-jerry-ye:
Notice : this will cause the output be filled with the output from panic test. We may need to filter them @lynzrand @bobzhang

### DIFF
--- a/abort/abort.mbt
+++ b/abort/abort.mbt
@@ -23,8 +23,30 @@
 ///
 /// Returns a value of type `T`. However, this function never actually returns a
 /// value as it always causes a panic.
+#cfg(not(target="native"))
 pub fn[T] abort(msg : String) -> T {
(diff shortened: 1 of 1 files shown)


lynzrand:
I can mitigate this by capturing test outputs by default, only outputting when the relevant tests fail (we can't put tests on the panic message itself, unless we accept false negatives).

Changes needed:

- Capture stdout and associate each segment with each test result
- Hide the output by default
- Print the output on test failure, or if a flag like `--no-capture` is set.

I am currently unsure if this is the desired change.

lynzrand:
I am creating a new issue to track this change in `moon`.

peter-jerry-ye:
Since the capturing is still WIP, this PR will only enable the output of abort for native backend

PR #2735, opened 2025-09-16
hongbo/string view bytes view

bobzhang:
- **refactor: update function signatures to use StringView instead of @string.View for consistency**
- **refactor: replace View with StringView in method signatures for improved clarity and consistency**
- **refactor: replace @bytes.View with BytesView in function signatures for consistency and clarity**
- **refactor: update function signatures to use BytesView instead of View for consistency and clarity**
- **refactor: update function signatures to use StringView instead of @string.View for improved consistency and clarity**


bobzhang:
cc @Guest0x0 there seems to be a bug, that 
```
#deprecated("Use `BytesView` instead")
pub typealias BytesView as View
```
the deprecated attributes not revealed in mbti file

PR #2743, opened 2025-09-18
feat: encoding/utf16

peter-jerry-ye:
This PR adds UTF16 encoding/decoding, and deprecate any other `String` to `Bytes` conversions as they are misleading.

### DIFF
--- a/string/methods.mbt
+++ b/string/methods.mbt
@@ -463,7 +463,7 @@ pub fn StringView::to_array(self : StringView) -> Array[Char] {
 }
 
 ///|
-#deprecated("Use `@encoding/utf16le.encode` instead")
+#deprecated("Check `@encoding/utf8.encode`")
 #coverage.skip
 pub fn StringView::to_bytes(self : StringView) -> Bytes {
   let array = FixedArray::make(self.length() * 2, Byte::default())
--- a/builtin/deprecated.mbt
+++ b/builtin/deprecated.mbt
@@ -219,7 +219,7 @@ pub fn Double::upto(
 }
 
 ///|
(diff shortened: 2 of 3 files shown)


encoding/utf16le/pkg.generated.mbti
```diff
@@ -0,0 +1,19 @@
+// Generated using `moon info`, DON'T EDIT IT
+package "moonbitlang/core/encoding/utf16le"
+
+// Values
+fn decode(BytesView, ignore_bom? : Bool) -> String raise Malformed
+
+fn decode_lossy(BytesView, ignore_bom? : Bool) -> String
+
+fn encode(StringView, bom? : Bool) -> Bytes
+
```
bobzhang:
```
fn encode(StringView, bom? : Bool, endianness : Endian = Little) -> Bytes
```

peter-jerry-ye: This will likely double the code size, introduce unnecessary branching.

string/methods.mbt
```diff
@@ -463,9 +463,8 @@ pub fn StringView::to_array(self : StringView) -> Array[Char] {
 }
 
 ///|
-/// Converts the View into bytes using UTF-16 little endian format.
-/// 
-#deprecated("check @encoding/utf8.encode")
+#deprecated("Use `@encoding/utf16le.encode` instead")
+#coverage.skip
 pub fn StringView::to_bytes(self : StringView) -> Bytes {
   let array = FixedArray::make(self.length() * 2, Byte::default())
```
bobzhang, line 469: in most cases, this is not what we need, encode is mostly what we need

PR #2745, opened 2025-09-19
security(json): Add nesting depth limit to prevent stack overflow

quake:
This PR introduces a nesting depth limit to the JSON parser to fix a security vulnerability.

Without a depth limit, parsing a deeply nested JSON string can lead to a stack overflow, which crashes the program. This can be used as a DoS attack.

This is a common security measure found in many other languages and libraries (like Go, Rust, and Python) to ensure safe JSON parsing. This change makes moonbit json parser more robust and secure.

### DIFF
--- a/json/parse.mbt
+++ b/json/parse.mbt
@@ -23,6 +23,7 @@ pub fn valid(input : String) -> Bool {
 }
 
 ///|
+/// Parse a JSON input string into a Json value, with an optional maximum nesting depth (default is 1024)
 pub fn parse(
   input : String,
   max_nesting_depth? : Int = 1024,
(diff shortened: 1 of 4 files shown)


json/parse.mbt
```diff
@@ -59,15 +62,25 @@ fn ParseContext::parse_value2(
 
 ///|
 fn ParseContext::parse_object(ctx : ParseContext) -> Json raise ParseError {
+  if ctx.depth >= ctx.max_nesting_depth {
+    raise DepthLimitExceeded
+  }
+  ctx.depth += 1
```
FlyCloudC: How about replacing `max_nesting_depth` and `depth` with a single `mut remaining_available_depth : Int`, and using `remaining_available_depth >= 0` as the depth check?

quake: 👍 , let me fix it later

quake: resolved, please help to review again, thanks.

json/parse.mbt
```diff
@@ -23,8 +23,12 @@ pub fn valid(input : String) -> Bool {
 }
 
 ///|
-pub fn parse(input : String) -> Json raise ParseError {
-  let ctx = ParseContext::make(input)
+/// Parse a JSON input string into a Json value, with an optional maximum nesting depth (default is 1024)
+pub fn parse(
+  input : String,
+  max_nesting_depth? : Int = 1024,
```
peter-jerry-ye, line 29: Is there any reason for this number, i.e. what's the number picked by the others?

quake: follow python / java(Jackson), the default limit is around 1000.

PR #2747, opened 2025-09-19
hongbo/operator up

bobzhang:
- **op_get => "_[_]"**
- **we need change at => T::at**
- **more op_get update**
- **info**


bobzhang (commented):
note there is some weirdness that `_[_]` require `X::` syntax

PR #2753, opened 2025-09-22
Feat: Add `order_by` parameter to `@priority_queue`

### LINKED ISSUE #2613: Require newtype to specify ASC or DESC for priority_queue is inconvenient
**Is your feature request related to a problem? Please describe.**
Currently priority_queue is DESC by default, and no convenient way to change it to ASC. 

**Describe the solution you'd like**
Add explicit named parameter `order_by~: ASC | DESC` to `@priority.new`.

**Describe alternatives you've considered**
Or pass a custom `comparator` parameter to `@priority.new`.

**Additional context**

- [x] I do think that this should be part of the core library and should be delivered to all the MoonBit developers.

Asterless:
## Summary

Fixed #2613, this PR enhances the `@priority_queue` by adding an `order_by` parameter, enabling it to function as both a max-heap (DESC, the default behavior) and a min-heap(ASC).

## Key Changes:

- Added `order_by` Parameter: The `new()`, `from_array()`, and `of()` functions now accept an optional `order_by` parameter. This allows users to specify the ordering of elements.
    - `@priority_queue.desc` (default): The queue operates as a max-heap, with the largest element at the top.
    - `@priority_queue.asc`: The queue operates as a min-heap, with the smallest element at the top.
    
- Updated Internal Logic: The internal functions `meld()` and `merges()` have been modified to respect the `order_by` parameter, ensuring that elements are compared and ordered correctly.

- Exported `asc` and `desc`: To provide a clean public API, asc and desc values of type `OrderBy` are now exported from the `@priority_queue` package.

- Added Tests: New test cases have been added to `priority_queue_test.mbt` to verify the functionality of both ascending and descending order modes, ensuring that pop and peek return the correct elements in each case.

## Example
```moonbit
test "order by" {
  let pq = @priority_queue.new(order_by=@priority_queue.asc)
  pq.push(3)
  pq.push(1)
  pq.push(2)
  assert_eq(pq.pop(), Some(1))
  assert_eq(pq.pop(), Some(2))
  assert_eq(pq.pop(), Some(3))
  assert_eq(pq.pop(), None)
}
```


priority_queue/pkg.generated.mbti
```diff
@@ -6,24 +6,29 @@ import(
...
 
 // Values
+let asc : OrderBy
+
+let desc : OrderBy
 
 // Errors
 
 // Types and methods
+type OrderBy
```
hackwaly, line 16: Why make this type opaque/abstract?

quake:
In both Rust and Haskell, there is a common idiom called a “newtype” wrapper. The idea is to create a lightweight wrapper around an existing type, without adding runtime overhead, but allowing you to give it different traits/instances. 

- https://doc.rust-lang.org/std/cmp/struct.Reverse.html
- https://hackage.haskell.org/package/base-4.21.0.0/docs/Data-Ord.html#t:Down

They are zero-cost abstractions: at runtime, the wrapper disappears.

If moonbit's newtype is zero-cost,  I suggest adding a similar newtype in `core/cmp` instead of adding different sorting parameters to each container that supports sorting.

hackwaly:
Newtype is too tricky for newcomer. It should be considered as last resort.

Lampese:
It doesn't really matter to me whether or not there is an ORDER BY in actual user usage, but I am more concerned about the API consistency issues it brings.

The current export method is definitely not suitable. Even if this thing is added, it should be placed in a more generic place. I am actually thinking about whether it can be changed to a Boolean value?

The main issue is that our previous approach was similar to the implementation of overloading operators, which involved declaring a new type and overriding its Compare trait to reverse the order. I don't think this is intuitive either.

peter-jerry-ye:
I think the problem with newtype is the trade off between code size & performance @bobzhang 

The code size will be doubled due to monofy, but the performance would be greater.

peter-jerry-ye:
We would follow @quake 's suggestion and provide the `@cmp.Reverse`.

PR #2754, opened 2025-09-22
refactor: make arrayview immutable

peter-jerry-ye:


array/slice.mbt
```diff
@@ -67,3 +67,58 @@ fn[T] FixedArraySlice::slice(
 ) -> FixedArraySlice[T] {
   { array: self.array, start: self.start + start, end: self.start + end }
 }
+
+///|
+/// TODO: use specific type for non-js backend
+/// Notice: may need to move to builtin
+priv struct ArraySlice[T] {
+  array : Array[T]
```
bobzhang, line 75: cc @Yu-zh this is a valid use case where we would like it to be value type

PR #2759, opened 2025-09-23
remove unnecessary Compare constraints and correct typos

lue-bird:
Some functions like `@immut/sorted_set.SortedSet::fold` unnecessarily constrain type variables to be `Compare` even though the `compare` method is never called.

This PR removes such unnecessary constraints and also fixes a few typos.

Note that because some markdown code blocks had a typo in the language ("moobit") they weren't tested before and now are. One of them shows failing test for `tanhf(1)` (which likely doesn't signify an actual bug and just needs to be corrected).

Feel free to only take parts of this PR or disregard it entirely if for example the extra constraints were deliberately added.

Thanks for the already fairly polished language!


peter-jerry-ye:
Thank you for the PR. Looks like the unused trait bound check does not work due to recursion @Guest0x0

Issue #2758, opened 2025-09-23
[Consistency Review]JSON PATH


peter-jerry-ye:
**Current behaviors**

JSON Path is an abstract type used to represent the path where the `from json` error occured. It has a show type which is kind of randomly defined.

**Expected behaviors**

Migrate JSON Path to JSON pointer following RFC 6901

**Justification**

It will make the API more standarized

PR #2763, opened 2025-09-24
feat: add JsonPointer support and conversions from and to json

Erchiusx:
next move: add URI support and try to replace JsonPath with JsonPointer

trying to address issue https://github.com/moonbitlang/core/issues/2758


json/json_pointer.mbt
```diff
@@ -0,0 +1,277 @@
...
+      } else {
+        ObjectIndex(component)
+      }
+    _ => ObjectIndex(component)
+  }
+}
+
+///|
+pub fn JsonPointer::apply(self : JsonPointer, value : Json) -> Json? {
+  match (self, value) {
```
peter-jerry-ye: Shouldn't this be a loop at least?

peter-jerry-ye (commented):
The overall concept looks fine-ish, but the implementation doesn't look very idiomatic

json/json_pointer.mbt
Erchiusx: loop needs intermediate result collection, I think the cost is generally alike;
but it is true that implementations need improvement

I will use iterator to implement this apply, and I am considering to reverse the link direction as a refactor.

Erchiusx:
I see https://github.com/moonbitlang/core/pull/2764 implemented another JSON pointer feature
Seemingly more integrated, maybe this PR can be closed now.

peter-jerry-ye:
Yes, we will migrate the currently existing `JsonPath` to `JsonPointer`. Maybe we'll revisit this issue to see how to implement it closer to RFC. Thank you for your contribution.

PR #2767, opened 2025-09-24
feat：Add case-insensitive string comparison methods

xuxubo:


peter-jerry-ye:
I don't think we should allocate new strings during the comparison.

PR #2783, opened 2025-09-26
refactor: make hash random

peter-jerry-ye:
This PR makes the hash seed random on JavaScript backend, and cleanup the relevant tests.

This PR is broken into two steps:

1. Migrate `key.hash()` to `Hasher::new()..combine(key).finalize()`. This is to make sure that the hash values are computed based on the hash combine with a seed, which is fixed at this stage. The tests based on the specific hash values are cleaned up.
2. Make the `seed` in `Hasher::new()` random for JS backend. It will use `crypto.getRandomValue` if possible, and fallback to `Math.random`. The tests based on the fixed hash computations are cleaned up.

### DIFF
--- a/array/view_test.mbt
+++ b/array/view_test.mbt
@@ -195,8 +195,14 @@ test "arrayview_arbitrary" {
 ///|
 test "arrayview_hash" {
   let arr : Array[@array.View[Int]] = @quickcheck.samples(20)
-  inspect(arr[5:9].hash(), content="-966877954")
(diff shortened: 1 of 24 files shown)


peter-jerry-ye:
`Hasher::new()` -> key.hash

remove `impl hash`

peter-jerry-ye:
@quake Would you like to review this PR?

quake (approved):

quake:
> @quake Would you like to review this PR?

LGTM

Issue #2785, opened 2025-09-26
Proposal: Change Array::`chunks` and `chunk_by` to Return Views


quake:
The current implementations of `Array::chunks` and `Array::chunk_by` create a new `Array[T]` for each chunk. This has two main drawbacks:

1.  **Performance Overhead**: For each chunk, new memory must be allocated, and elements from the original array must be copied into the new array. This can be very inefficient, especially when dealing with large arrays or a large number of chunks.

2.  **Increased Memory Usage**: Creating copies of data for each chunk leads to a significantly higher memory footprint, as the data is duplicated in memory.

By returning an `Array[ArrayView[T]]`, we can adopt a **zero-copy** approach. `ArrayView` is a lightweight structure that simply points to a slice of the original array's data without taking ownership or creating a copy.

## Impact

This is a **breaking change**. Code that relies on the return type being `Array[Array[T]]` will no longer compile. For example, code that attempts to modify the returned chunks will fail, as `ArrayView` is immutable.

Migration is straightforward:
- If the user only needs to read from the chunks, no change is needed other than updating type annotations.
- If the user needs a mutable `Array[T]` for a chunk, they must now explicitly call `.to_array()` on the `ArrayView`.

quake:
btw, this proposal can be applied to other fn like `split` / `strip_prefix` and `strip_suffix` also.

FlyCloudC:
Returning an `Array` still requires allocating space roughly proportional to the array's length (especially in cases like `arr.chunks(2)`). Should we return `Iter[ArrayView[A]]` instead? If an `Array[ArrayView[A]]` is truly needed, `Iter::to_array` can be used.

quake:
Returning an `Iter` avoids the outer allocation but breaks indexed access. Returning Array<ArrayView[T]] gives us zero-copy element access and indexing, but still allocates an outer array of view-descriptors. Both sides of that debate are correct, I think the right choice depends on the access pattern, so the practical solution is adding a new zero-copy, indexable view API (and an iterator API for streaming).

PR #2786, opened 2025-09-26
refactor: array chunk should return view

quake:
resolve https://github.com/moonbitlang/core/issues/2785


bobzhang:
techinically this is a breaking change, but I think we should move forward with this

PR #2788, opened 2025-09-26
implement ImmutArray

Yu-zh:
This PR adds ImmutArray, which has the same representation as FixedArray but with methods like `set` removed.


hackwaly:
We could provide a shared `empty` `ImmutArray`. And the impl of `Default` could use the shared `empty`.

Yu-zh:
> We could provide a shared `empty` `ImmutArray`. And the impl of `Default` could use the shared `empty`.

The sharing has already been implemented in the compiler side

FlyCloudC:
Should we directly implement the trait using derive?

PR #2792, opened 2025-09-28
fix: resolve json stringify stack overflow issue

quake:
This pull request refactors the `Json.stringify` implementation to use an explicit stack instead of recursion, allowing safe stringification of deeply nested JSON structures without risking stack overflow. 

The unit test in first commit of this PR can be used to reproduced the stack overflow error.


json/json.mbt
```diff
@@ -85,77 +85,122 @@ fn indent_str(level : Int, indent : Int) -> String {
...
+          if members.is_empty() {
+            buf.write_string("{}")
+            continue (None, depth)
           } else {
-            buf.write_char(',')
-            buf.write_string(indent_str(level + 1, indent))
+            buf.write_char('{')
+            buf.write_string(indent_str(depth + 1, indent))
+            // After child value printed, we resume from this frame
+            stack.push(WriteFrame::Object(members.to_array(), 0, depth))
```
bobzhang: this seems to allocate a lot

quake: I think it can be resolved by the indexing json map operation, however, the built-in linked hashmap doesn't provide the indexing operation (map[index])

json/json_test.mbt
```diff
@@ -551,3 +551,29 @@ test "call stack overflow" {
   }
 
 }
+
+///|
+test "stringify deep nested array" {
+  // Build a deeply nested JSON array like [[[[[]]]]] with depth = 6000
+  let depth = 6000
```
peter-jerry-ye, line 558: For me the depth you used here to trigger the issue is far more than the bar you set previously (1024) for the `parse`. I think it might not be so realistic to have such a deeply nested structure and the user might need to be responsible for such structure which could only be constructed by themselves.

quake: Different OS have different default stack sizes, the maximum recursion depth before a stack overflow is not the same across systems, so I use a large depth (6000) to make sure it can be reproduced in most OS. On one of my machine, 1500 is enough to reproduce.

json/json.mbt
quake: btw, may I know why the `Iteartor` doesn't support `next` fn? The Iterator#next fn can be used to resolve issue also.

peter-jerry-ye: What we have right now is internal iterator, i.e. `Stream` if you have Java background. For external iterator #2800

quake: > What we have right now is internal iterator, i.e. `Stream` if you have Java background. For external iterator #2800

cool, I will rebase this PR after #2800 is merged

bobzhang: we exposed external iterator, would you give a try? let me know if you have any questions

quake: > we exposed external iterator, would you give a try? let me know if you have any questions

Updated the PR, please help to review again, thanks.

btw, why the `Iterator` related fn of `Array` are put in core/array instead of core/builtin/array.mbt since the `Iteartor` has been put in core/builtin/iterator.mbt already?

bobzhang: it is currently a bit messy for the core organization due to cycle dependencies. In general, we would like to keep the built in small

Issue #2796, opened 2025-09-28
[FR] `minmax` and `minmax_by_key` in `@cmp`


FlyCloudC:
**Is your feature request related to a problem? Please describe.**
Code like this will perform two comparisons:

```mbt
let x = @cmp.minimum(a, b)
let y = @cmp.maximum(a, b)
```

**Describe the solution you'd like**
I would like a new function, tentatively called `@cmp.minmax`, that takes two comparable values and returns a tuple containing both the minimum and the maximum value in a single comparison.

```mbt
let (x, y) = @cmp.minmax(a, b)
```

The same applies to `minmax_by_key`.

- [x] I do think that this should be part of the core library and should be delivered to all the MoonBit developers.

PR #2797, opened 2025-09-29
feat(cmp): min_max and min_max_by_key

FlyCloudC:
Associated FR: #2796


bobzhang:
https://doc.rust-lang.org/std/cmp/fn.minmax.html
can you follow the same naming convention

FlyCloudC:
> https://doc.rust-lang.org/std/cmp/fn.minmax.html can you follow the same naming convention

Ok, rename `min_max` to `minmax`, `min_max_by_key` to `minmax_by_key`.

PR #2798, opened 2025-09-29
fix(splitmix): remove derived Default for RandomState

FlyCloudC:
- It would generate zero `seed` and `gamma`, which affects randomness

### DIFF
--- a/quickcheck/README.mbt.md
+++ b/quickcheck/README.mbt.md
@@ -48,12 +48,11 @@ QuickCheck provides `Arbitrary` implementations for all basic MoonBit types:
 test "builtin types" {
   // Basic types
   let v : (Bool, Char, Byte) = @quickcheck.gen()
-  inspect(v, content="(true, '#', b'\\x12')")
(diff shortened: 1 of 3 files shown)


peter-jerry-ye:
What about an explicit implementation with a non-zero seed?

peter-jerry-ye:
And why did you say it would generate all zero seeds? It looks like it will have seed as zero at the beginning, but the next call with `next_uint64` it would be added with gamma so it would no longer be zero

FlyCloudC:
> And why did you say it would generate all zero seeds? It looks like it will have seed as zero at the beginning, but the next call with `next_uint64` it would be added with gamma so it would no longer be zero

You are right, my description was not accurate. What I meant to convey is that both its `seed` and `gamma` are 0.

peter-jerry-ye:
I see. Then maybe only the second commit is necessary?

FlyCloudC:
> I see. Then maybe only the second commit is necessary?

Ok

PR #2799, opened 2025-09-29
perf(priority_queue): better option handling

FlyCloudC:
Use option struct instead of enum to implement Node. Changed the type of meld from `(Node[A]?, Node[A]?) -> Node[A]?` to `(Node[A], Node[A]) -> Node[A]` to reduce branching.

### DIFF
--- a/priority_queue/priority_queue.mbt
+++ b/priority_queue/priority_queue.mbt
@@ -22,7 +22,7 @@
 /// ```
 #as_free_fn
 pub fn[A] T::new() -> T[A] {
-  { len: 0, top: Nil }
+  { len: 0, top: None }
 }
 
 ///|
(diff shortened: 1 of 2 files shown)


priority_queue/priority_queue.mbt
```diff
@@ -125,30 +125,34 @@ pub fn[K : Compare] T::from_iter(iter : Iter[K]) -> T[K] {
 
 ///|
 fn[A : Compare] meld(x : Node[A], y : Node[A]) -> Node[A] {
```
peter-jerry-ye, line 127: I didn't see the signature change described in the PR.

peter-jerry-ye:
It looks like the type is only redeclared in an equivalent manner. I don't really see why this is better.

priority_queue/priority_queue.mbt
FlyCloudC: The original `Node[A]` corresponds to the later `Node[A]?`, so the original `meld` is equivalent to having the signature `(Node[A]?, Node[A]?) -> Node[A]?` (i.e., the `meld_opt` in commit 898ff5c).

FlyCloudC:
> It looks like the type is only redeclared in an equivalent manner. I don't really see why this is better.

In some cases, it can be determined that the node is non-null, thus eliminating one match operation. For example, `from_array` originally had to check whether `acc` was null every time an element was added.

PR #2805, opened 2025-10-05
refactor(immut/sorted_set): Change Compare to compare by size first [BREAKING]

bobzhang:
## Summary

This PR refactors the `Compare` trait implementation for `SortedSet` to compare sets by **size first**, then by elements lexicographically if sizes are equal.

### Previous Behavior
- Compared sets element-by-element lexicographically
- Example: `{1,2}` vs `{3}` → compared as `[1,2]` vs `[3]` → result: `{1,2} < {3}` (because when first set has more elements, it was considered "greater")

### New Behavior
- Compares by **size first**, then by elements if sizes are equal
- Example: `{1,2}` vs `{3}` → size(2) > size(1) → result: `{1,2} > {3}`
- Only compares elements lexicographically when sizes are equal

## Rationale

1. **More intuitive**: Larger sets are naturally "greater than" smaller sets
2. **Performance**: Size comparison is O(1), avoids unnecessary element iteration for different-sized sets
3. **Consistency**: Matches common set comparison semantics in other languages/libraries

## Changes

- ✨ Refactored `Eq` and `Compare` implementations to `generic.mbt`
- ✨ Extracted `InorderIterator` to `external_iterator.mbt` for better code organization
- ✅ Added comprehensive test coverage for:
  - Equal sets
  - Same-size sets with different elements
  - Different-size sets (new behavior)
  - Empty sets
  - Edge cases

## Impact

This affects anyone using:
- `SortedSet::compare()`
- Ordering operations (`<`, `>`, `<=`, `>=`) on `SortedSet`
- Sorting collections of `SortedSet`

## Testing

All 5634 tests pass. New tests specifically cover the size-first comparison behavior.

## Request for Review

Please review whether this semantic change is acceptable for the standard library. While it's a breaking change, it provides more intuitive behavior and better performance.

### DIFF
--- a/immut/sorted_set/external_iterator.mbt
+++ b/immut/sorted_set/external_iterator.mbt
@@ -1,0 +1,64 @@
+// Copyright 2025 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+///|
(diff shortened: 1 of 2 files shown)


bobzhang:
cc @hackwaly this will make the comparison faster but semantics will change a little bit

PR #2824, opened 2025-10-14
fix(linked_hash_map/set): from_array, of

FlyCloudC:
Fix bug in #2728.

Here, two bugs canceled each other out, which is why they weren't detected in the last test.

One bug is that the `capacity` was calculated but never actually used.  

The other bug is that the condition for `capacity *= 2` was incorrect. It should follow the precondition of `add`: `!(self.size > self.grow_at)`, not the condition checked in the `if` statement at the beginning of `add`: `self.size >= self.grow_at`.


peter-jerry-ye:
Do you have a test case to demonstrate the bug that you described?

FlyCloudC:
> Do you have a test case to demonstrate the bug that you described?

This test is the edge case where `self.size == self.grow_at`

https://github.com/moonbitlang/core/blob/main/builtin%2FLinkedHashMap.mbt.md#L75-L82

PR #2831, opened 2025-10-16
fix: access crypto with globalThis

peter-jerry-ye:
```
qjs > globalThis.crypto?.getRandomValues
undefined
qjs > crypto?.getRandomValues
ReferenceError: 'crypto' is not defined
    at <eval> (<evalScript>:1:1)
```

### DIFF
--- a/builtin/hasher.mbt
+++ b/builtin/hasher.mbt
@@ -84,9 +84,9 @@ let seed : Int = random_seed()
 #cfg(target="js")
 extern "js" fn random_seed() -> Int =
   #|() => {
-  #|  if (crypto?.getRandomValues) {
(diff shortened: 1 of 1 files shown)


builtin/hasher.mbt
```diff
@@ -84,7 +84,7 @@ let seed : Int = random_seed()
 #cfg(target="js")
 extern "js" fn random_seed() -> Int =
   #|() => {
-  #|  if (crypto?.getRandomValues) {
+  #|  if (globalThis.crypto?.getRandomValues) {
```
hackwaly, line 87:
```
typeof crypto == "object" && typeof crypto.getRandomValues === "function"
```

globalThis is a recent feature maybe unavailable in some js engine.

PR #2841, opened 2025-10-18
feat: add a method to return a random boolean value

xuxubo:
- Implements a new method to generate a random boolean (true or false).
- The value is determined by generating a random integer in the range [0, 2), and returning true if the result is 0, and false otherwise.


FlyCloudC:
I suggest directly using `(self.next() & 1) == 1`, which has fewer function calls.

Additionally, you need to run `moon info` before pushing to avoid CI failures.

PR #2858, opened 2025-10-27
Optimize Json stringify frame reuse

bobzhang:
## Summary
- reuse the active WriteFrame when resuming Json::stringify iteration
- avoid extra pop/push churn while preserving existing formatting behavior

## Testing
- moon test


bobzhang:
cc @quake I did some changes to avoid some allocations

Issue #2850, opened 2025-10-21
[Consistency Review] Deque::rev_iter2


peter-jerry-ye:
**Current behaviors**

The `rev_iter2`'s first element, index, is counted from zero

**Expected behaviors**

It should correspond to the actual index in the data structure

**Justification**

Currently, all the other apis, such as `rev_eachi` or `rev_iteri` has the index corresponding to the actual index.

bobzhang:
> Currently, all the other apis, such as rev_eachi or rev_iteri has the index corresponding to the actual index.

`Array::rev_eachi` does not follow this?

PR #2862, opened 2025-10-27
fix: update `rev_iter2` to yield indices in descending order

Asterless:
Based on #2850. This pull request updates the `rev_iter2`  for the `Deque` data structure to ensure that the index values correctly reflect their position when iterating in reverse order. The index now starts from `length - 1` and decrements with each step, rather than starting from 0 and incrementing. This change makes the reverse iterator's behavior more intuitive and consistent with typical expectations.


peter-jerry-ye:
We need to discuss this furthur before making changes.

PR #2868, opened 2025-10-28
feat: optimize unsafe_substring for full string return

bobzhang:
@Yu-zh can you check if intrinsic also optimize such case


Yu-zh:
@hackwaly This intrinsic is only used by js backend

PR #2870, opened 2025-10-28
feat: replacer for Json::stringify

peter-jerry-ye:
This PR adds the replacer for Json::stringify.

Notice that compared with JS's version, there's no initial first call with `"", object` in the current implementation
Also notice that it is possible to have stack overflow with the replacer : just creates new object for any visit

### DIFF
--- a/json/pkg.generated.mbti
+++ b/json/pkg.generated.mbti
@@ -82,6 +82,7 @@ fn Json::output(Self, &Logger) -> Unit // from trait `Show`
 fn Json::stringify(Self, escape_slash? : Bool, indent? : Int, replacer? : Replacer) -> String
 fn Json::to_json(Self) -> Self // from trait `ToJson`
 fn Json::to_string(Self) -> String // from trait `Show`
+fn Json::transform(Self, Replacer) -> Self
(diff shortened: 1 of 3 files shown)


json/pkg.generated.mbti
```diff
@@ -74,7 +88,7 @@ fn Json::as_string(Self) -> String?
 fn Json::from_json(Self, JsonPath) -> Self raise JsonDecodeError // from trait `FromJson`
 fn Json::item(Self, Int) -> Self?
 fn Json::output(Self, &Logger) -> Unit // from trait `Show`
-fn Json::stringify(Self, escape_slash? : Bool, indent? : Int) -> String
+fn Json::stringify(Self, escape_slash? : Bool, indent? : Int, replacer? : Replacer) -> String
 fn Json::to_json(Self) -> Self // from trait `ToJson`
 fn Json::to_string(Self) -> String // from trait `Show`
```
bobzhang, line 84: 1. Json::transform(Self, replacer) -> Json
2. may be should not transform replaced result
3. Idx -> String
4. ArrayView[String]

PR #2872, opened 2025-10-28
Improve performance of `Int::to_string`

### LINKED ISSUE #2833: [performance] Improve performance of `Int::to_string`
**Background** `Int::to_string` is a commonly used API and it's implemented by calling `Int::output`

**But** `Int::output` is not implemented very efficiently for two reasons: 1. the inner function `write_digits` is not a tail-recursive function; 2. it calls `logger.write_char`, which is a dynamic object method call, repeatedly.

**Therefore** I would like it to be improved in several ways: 1. using a tail-recursive helper function or a loop to write each digit; 2. write to a buffer of type `FixedArray[Byte]` or `StringBuilder` first and then dump it to the logger. This should improve the performance.

Yu-zh:
close #2833

This PR implements various optimizations for `Int::to_string`:
* Avoid extra allocation when converting from `FixedArray[Byte]` to `String`. Instead, I use a buffer of type `FixedArray[UInt16]` and cast it to `String`.
* Specialize the common case where radix=10 and radix=16. For radix=10, we can calculate each digit by simple arith without having to lookup the alphabetic table.
* Loop unrolling techniques.
* Use loop instead of the original non-tail-recursive recursive function
* Avoid using trait object calls.
* Use JS native `int.to_string()` for js backend

---
Benchmark:
```
test "bench int_to_string" (b : @bench.T) {
  let i1 = @int.max_value
  let i2 = @int.min_value
  let i3 = 123456
  let i4 = -123456
  b.bench(() => b.keep(for i in 0..=1000 { ignore(i1.to_string())},))
  b.bench(() => b.keep(for i in 0..=1000 { ignore(i2.to_string())},))
  b.bench(() => b.keep(for i in 0..=1000 { ignore(i3.to_string())},))
  b.bench(() => b.keep(for i in 0..=1000 { ignore(i4.to_string())},))
}
```
Result:
| backend | input | before (μs) | after (μs) | speedup | perf improvement |
|------|------|------------|------------|--------|----------|
| **WASM** | max_value | 345.45 ± 5.70 | 38.61 ± 1.15 | 8.9x | 88.8% |
| **WASM** | min_value | 365.72 ± 1.57 | 39.17 ± 0.98 | 9.3x | 89.3% |
| **WASM** | 123456 | 209.91 ± 0.62 | 34.59 ± 0.25 | 6.1x | 83.5% |
| **WASM** | -123456 | 223.02 ± 0.46 | 35.03 ± 0.09 | 6.4x | 84.3% |
| **WASM-GC** | max_value | 106.91 ± 0.20 | 15.92 ± 0.03 | 6.7x | 85.1% |
| **WASM-GC** | min_value | 111.17 ± 0.30 | 16.28 ± 0.11 | 6.8x | 85.4% |
| **WASM-GC** | 123456 | 60.31 ± 0.14 | 12.08 ± 0.03 | 5.0x | 80.0% |
| **WASM-GC** | -123456 | 63.41 ± 0.18 | 12.08 ± 0.04 | 5.2x | 80.9% |
| **JS** | max_value | 163.21 ± 0.38 | 4.41 ± 0.02 | 37.0x | 97.3% |
| **JS** | min_value | 153.56 ± 0.35 | 4.40 ± 0.01 | 34.9x | 97.1% |
| **JS** | 123456 | 84.73 ± 0.27 | 4.39 ± 0.01 | 19.3x | 94.8% |
| **JS** | -123456 | 86.22 ± 0.35 | 4.41 ± 0.06 | 19.5x | 94.9% |
| **Native** | max_value | 116.87 ± 0.70 | 18.22 ± 0.14 | 6.4x | 84.4% |
| **Native** | min_value | 119.81 ± 0.66 | 15.92 ± 0.03 | 7.5x | 86.7% |
| **Native** | 123456 | 64.73 ± 0.27 | 15.68 ± 0.07 | 4.1x | 75.8% |
| **Native** | -123456 | 69.56 ± 0.27 | 16.52 ± 0.19 | 4.2x | 76.2% |

---
This PR only handles `Int::to_string` case. I will handle `Int64`, `UInt`, `UInt64` similarly in a separate PR.

### DIFF
--- a/builtin/to_string.mbt
+++ b/builtin/to_string.mbt
@@ -29,6 +29,7 @@ pub impl Show for Int64 with to_string(self) {
 
 ///|
 /// Converts an unsigned 32-bit integer to hexadecimal
+#cfg(not(target="js"))
 fn int_to_string_hex(
   buffer : FixedArray[UInt16],
   num : UInt,
@@ -67,6 +68,7 @@ fn int_to_string_hex(
 
 ///|
 /// Generic radix conversion for any base 2-36
+#cfg(not(target="js"))
 fn int_to_string_generic(
   buffer : FixedArray[UInt16],
   num : UInt,
@@ -107,6 +109,7 @@ fn int_to_string_generic(
 
 ///|
 /// Converts an unsigned 32-bit integer to decimal string
+#cfg(not(target="js"))
 fn int_to_string_dec(
   buffer : FixedArray[UInt16],
   num : UInt,
@@ -164,6 +167,7 @@ fn int_to_string_dec(
 
 ///|
 /// Calculates the number of decimal digits in a u32 value
+#cfg(not(target="js"))
 fn dec_count32(value : UInt) -> Int {
   if value < 100000U {
     if value < 100U {
@@ -184,6 +188,7 @@ fn dec_count32(value : UInt) -> Int {
 
 ///|
 /// Calculates the number of hex digits needed for a u32 value
+#cfg(not(target="js"))
 fn hex_count32(value : UInt) -> Int {
   if value == 0U {
     1
@@ -195,6 +200,7 @@ fn hex_count32(value : UInt) -> Int {
 
 ///|
 /// Calculates the number of digits needed for a u32 value in any radix
+#cfg(not(target="js"))
 fn radix_count32(value : UInt, radix : Int) -> Int {
   if value == 0U {
     return 1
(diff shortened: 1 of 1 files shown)


builtin/to_string.mbt
```diff
@@ -35,11 +217,65 @@ pub impl Show for Int64 with to_string(self) {
...
+
+  // Write minus sign if negative
+  if is_negative {
+    buffer.unsafe_set(0, 0x002D)
+  }
+  unsafe_fixedarray_to_string(buffer)
 }
 
+///|
+fn unsafe_fixedarray_to_string(buffer : FixedArray[UInt16]) -> String = "%identity"
```
peter-jerry-ye, line 285: Doesn't looks good for JS backend

PR #2880, opened 2025-10-30
feat: implement mutarrayview

Yu-zh:
This PR Implements `MutArrayView[T]`, a zero-copy mutable view type that provides read-write access to array slices without copying data.

Major Changes:

- New type: MutArrayView[T] - a builtin valtype representing a mutable window into arrays
- Core operations:
  - Element access: `at()/[]` and `set()/[]=` with bounds checking
  - Unsafe operations: `unsafe_get()` and `unsafe_set()` for performance-critical code
  - View creation: `mut_sub()` method for Arrays, FixedArrays, and MutArrayViews
  - Read-only conversion: `sub()/[:]` to convert to immutable ArrayView
- Iterator support: Added `iterator()`, `rev_iterator()`, and `iterator2()` implementations
- Trait implementations: `Show`, `Eq`, `Compare`, and `Hash`
- Negative index support: Enables Python-style negative indexing (e.g., `arr.mut_sub(start=-3, end=-1)`)
- For the rest of APIs that `ArrayView[T]` has but not implemented for `MutArrayView[T]`, user can first take view from `MutArrayView[T]` and then use `ArrayView[T]`'s API.

### DIFF
--- a/builtin/pkg.generated.mbti
+++ b/builtin/pkg.generated.mbti
@@ -117,7 +117,7 @@ fn[T] Array::map_in_place(Self[T], (T) -> T raise?) -> Unit raise?
 fn[T, U] Array::mapi(Self[T], (Int, T) -> U raise?) -> Self[U] raise?
 #alias(mapi_inplace, deprecated)
 fn[T] Array::mapi_in_place(Self[T], (Int, T) -> T raise?) -> Unit raise?
-fn[T] Array::mut_sub(Self[T], start? : Int, end? : Int) -> MutArrayView[T]
+fn[T] Array::mut_view(Self[T], start? : Int, end? : Int) -> MutArrayView[T]
 fn[T] Array::new(capacity? : Int) -> Self[T]
 fn[T] Array::pop(Self[T]) -> T?
 fn[T] Array::push(Self[T], T) -> Unit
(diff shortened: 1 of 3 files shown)


Yu-zh:
Update: 
* Added a commit that moves some relevant definition of ArrayView to `builtin`
* Moves all the MutArrayView definition to `builtin` as well

builtin/pkg.generated.mbti
```diff
@@ -117,6 +117,7 @@ fn[T] Array::map_in_place(Self[T], (T) -> T raise?) -> Unit raise?
 fn[T, U] Array::mapi(Self[T], (Int, T) -> U raise?) -> Self[U] raise?
 #alias(mapi_inplace, deprecated)
 fn[T] Array::mapi_in_place(Self[T], (Int, T) -> T raise?) -> Unit raise?
+fn[T] Array::mut_sub(Self[T], start? : Int, end? : Int) -> MutArrayView[T]
```
peter-jerry-ye: `mut_sub` sounds wierd to me. How about `slice` or something like that?

bobzhang: let's conslidate to `view` and  `mut_view`

Issue #2874, opened 2025-10-29
Provide `Map[String]::get_from_string(Self, StringView)` and Bytes


bobzhang:
```moonbit
/// The `view` function and the `Hash` implementation of `View` should satisfy:
/// ```
/// view(key).hash() == key.hash() 
/// ```
fn get_by_view[K, View : Hash + Eq, Value](
  self : Self[K, Value],
  key : View,
  view~ : (K) -> View
) -> Value?
```
It is hard to reason about if the variant is met or not. for String, we added an optimization that `"abc"[:].to_string()` does not allocate, maybe we can add a special function:

```
Map[String]::get_from_string(self : Self, StringView)
Map[Bytes]::get_from_bytes(self : Self, BytesView)
```

_Originally posted by @bobzhang in https://github.com/moonbitlang/core/issues/2082#issuecomment-3459488020_

bobzhang:
note it is not straightforward  due to cylic dependencies we may need move some stringview funcs into builtin first

PR #2883, opened 2025-10-30
Feat: Add `Map::get_from_string` and `Map::get_from_bytes`

Asterless:
## Summary
Based on #2874. This PR introduces `Map::get_from_string` and `Map::get_from_bytes` functions to allow direct lookup of values in maps using `StringView` and `BytesView` keys, and move some `StringView` and `BytesView` funcs into `builtin`

### New Functions
* Introduced two special functions:
```moonbit
pub fn[V] Map::get_from_string(self : Map[String, V], key : StringView) -> V?
pub fn[V] Map::get_from_bytes(self : Map[Bytes, V], key : BytesView) -> V?
```

### Cleanup and Refactoring

* Moved some `StringView` and `bytesView` functions into `builtin/stringview.mbt` and `builtin/bytesview.mbt` from `string/view.mbt` and `Bytes/view.mbt`
* Removed redundant definitions of `StringView` accessors (`str`, `start`, `end`) from `builtin/stringbuilder.mbt`, as these are now handled centrally.

### DIFF
--- a/string/pkg.generated.mbti
+++ b/string/pkg.generated.mbti
@@ -1,6 +1,10 @@
 // Generated using `moon info`, DON'T EDIT IT
 package "moonbitlang/core/string"
 
+import(
+  "moonbitlang/core/builtin"
+)
+
 // Values
 fn default() -> String
 
@@ -138,6 +142,7 @@ impl Show for StringView
 impl ToJson for StringView
 
 // Type aliases
+pub using @builtin {type StringView as View}
 
 // Traits
 pub trait ToStringView {


peter-jerry-ye (changes requested):
You are moving **everything** to builtin, which is not ideal. Only the implementation of `Hash` and `Eq` is necessary to be moved.

Asterless:
> You are moving **everything** to builtin, which is not ideal. Only the implementation of `Hash` and `Eq` is necessary to be moved.

Received. Now I have moved most of the methods back to the original `view.mbt`, retaining only impl `Hash` and `Eq` as well as the underlying methods necessary to implement them.

peter-jerry-ye:
See #2886

Asterless:
> See #2886

i see, closing.👌

PR #2886, opened 2025-10-31
feat: add `map::get_from_string` and `map::get_from_bytes`

### LINKED ISSUE #2874: Provide `Map[String]::get_from_string(Self, StringView)` and Bytes

peter-jerry-ye:
Closes #2874

### DIFF
--- a/builtin/pkg.generated.mbti
+++ b/builtin/pkg.generated.mbti
@@ -664,6 +664,8 @@ fn Bytes::unsafe_get(Bytes, Int) -> Byte
 
 
 
+
+
 // Type aliases
 
 // Traits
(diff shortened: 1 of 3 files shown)


peter-jerry-ye:
@copilot Add tests, docs, and format & update moon info

PR #2898, opened 2025-10-31
refactor: use ArrayView in Map::from_array

Yu-zh:
Use `ArrayView` as the parameter for `Map::from_array` for better optimization when the argument is an array literal that can be statically initialized.


bobzhang:
@copilot run `moon info` to update

PR #2913, opened 2025-11-05
fix: replace deprecated hashmap methods with their new names

bobzhang:
- Replace filter_with_key with filter (17 occurrences)
- Replace fold_with_key with fold (4 occurrences)
- Replace map_with_key with map (6 occurrences)

All deprecated method calls in HAMT.mbt and HAMT_test.mbt have been updated to use the new non-deprecated method names.


bobzhang:
cc @peter-jerry-ye

PR #2915, opened 2025-11-05
[DONT MERGE]chore: improve coverage from 89.93% to 93.37%

Lampese:


math/trig_test.mbt
```diff
@@ -227,3 +227,294 @@ test "trig reduce methods simple" {
...
+///|
+test "asin double - large values coverage" {
+  // Test values close to 1 with ix >= 0x3FEF3333
+  // This triggers line 116-119
+  inspect(@math.asin(0.9999), content="1.5566540733173846")
+  inspect(@math.asin(-0.9999), content="-1.5566540733173846")
+
+  // Test values in range 0.5 <= absx < ~0.9375 
+  // This triggers the else branch (line 121-128)
+  inspect(@math.asin(0.6), content="0.6435011087932844")
```
Lampese, line 367: The CI machine on Github says that the result calculated here should be 0.6435011087932843, but I obtained the result using the latest version of MoonBit (moonc v0.6.30+07d9d2445 (2025-10-30)) and `moon test -u`. As far as I know, this should not be machine related?

This problem can be stably reproduced.

Lampese: I have checked and it is a JS issue not a machine issue. The calculated results have very, very small errors. My Node version is `v24.10.0`. Given that our current numerical algorithms are actively maintaining consistency in backend results, I believe this issue should be discussed and resolved with @Kaida-Amethyst.

bobzhang:
@Lampese can you split the non ci failure changes in separate PR (to avoid conflicting PRs in the future)

Lampese:
> @Lampese can you split the non ci failure changes in separate PR (to avoid conflicting PRs in the future)

OK, I will do that soon.

hashmap/hashmap_test.mbt
```diff
@@ -432,3 +432,12 @@ fn[K : Hash + Eq + Show, V : Eq + Show] verify_content(
   }
   assert_eq(map.length(), expected.length())
 }
+
+///|
+test "hashmap.to_json" {
```
Lampese, line 437: For this test:
```moonbit
test "hashmap.to_json" {
  let map = @hashmap.from_array([(1, "one"), (2, "two"), (3, "three")])
  inspect(
    map.to_json().stringify(),
    content="{\"1\":\"one\",\"2\":\"two\",\"3\":\"three\"}",
  )
}
```

WASM backend will output `{"1":"one","2":"two","3":"three"}`, but in JavaScript backend, it will output `"{"1":"one","3":"three","2":"two"}"`.

I think this is understandable. The order of JSON serialization does not affect the final result, but I will change this test to a key value comparison test, so I will make a note here.

Lampese:
- #2920

math/trig_test.mbt
Lampese: After discussion, these errors can be ignored, and I will increase the error range to 1e-15 and initiate a new PR.

PR #2918, opened 2025-11-06
fix: mark deprecated functions and update usage in tests

bobzhang:


bobzhang:
cc @peter-jerry-ye we should mark `#as_free_fn` as deprecated, adn move them into builtin package later

PR #2941, opened 2025-11-12
hongbo/ice

bobzhang:
- **move fixedarray**
- **move readonly array**
- **info**


bobzhang:
9d45890d95942b610601826ae01d6221798b6e43 triggered an ICE
cc @Yu-zh

Yu-zh:
I just pushed a fix. The ICE is because `char_length_eq` is not in the builtin package but one of the FixedArray APIs uses it implicitly by pattern matching on StringView.

Yu-zh (approved):

PR #2943, opened 2025-11-12
ci: cleanup test script

Yu-zh:
Two changes:
1. `moon test --target all` includes native backends now, so there's no need to run native backend separately
2. add CI test with `use_js_builtin_string` on
3. https://github.com/moonbitlang/core/pull/2906 introduced redundant jobs, and this PR removed them.

And a refactor:
factored out some common testing script to `actions/test/action.yml` and `actions/test-native-release/action.yml` and share them across stable/pre-release/nightly CI workflows.

### DIFF
--- a/.github/actions/test-native-release/action.yml
+++ b/.github/actions/test-native-release/action.yml
@@ -1,0 +1,21 @@
+name: "Run MoonBit Native Release Tests"
+description: "Run MoonBit tests with --target native --release"
+runs:
+  using: "composite"
+  steps:
+    - name: Set ulimit and run moon test (--release + --target native)
+      if: ${{ runner.os != 'Windows' }}
+      shell: bash
+      run: |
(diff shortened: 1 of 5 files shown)


bobzhang:
hold on a bit, note we need a bit more discussions about this

Yu-zh:
@bobzhang This should be good now. All the testing jobs should take ~2mins to run.

bobzhang:
```
 385 │         attr.length() >= 5 && attr[0] == 'p' && attr[1] == 'a' && attr[2] == 'n' && attr[3] == 'i' && attr[4] == 'c'
     │                                   ─┬─  
     │                                    ╰─── Warning (Alert deprecated): Use `code_unit_at` instead which returns UInt16
─────╯
```
cc @Yu-zh  we need fix the driver like `if attr is [.."panic",..]`?

PR #2964, opened 2025-11-17
fix: Ensure List::zip is side-effect free by using immutable rev()

xuxubo:


peter-jerry-ye:
Unnecessary because the zipped list is freshly created.

PR #2965, opened 2025-11-17
adapt latest compiler release

Guest0x0:
This PR migrates core to the latest compiler release (v0.6.32):

- the default semantic of `#deprecated` has been migrated to `skip_current_package=false`, which means local usage of deprecated item will be reported by default
- `moon info` now displays content of `pub using` properly

### DIFF
--- a/math/pkg.generated.mbti
+++ b/math/pkg.generated.mbti
@@ -41,6 +41,9 @@ fn cbrt(Double) -> Double
 
 fn cbrtf(Float) -> Float
 
+#as_free_fn
+fn ceil(Double) -> Double
+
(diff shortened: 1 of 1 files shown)


math/pkg.generated.mbti
```diff
@@ -41,6 +41,9 @@ fn cbrt(Double) -> Double
 
 fn cbrtf(Float) -> Float
 
+#as_free_fn
```
Guest0x0, line 44: It seems that the `#as_free_fn` attribute from the target of `using` is not properly cleaned. Since this is not critical, let's fix this in the next release. cc @bobzhang

PR #2966, opened 2025-11-17
fix: Fix memory safety bug in Deque::flatten

xuxubo:
When the elements of an inner deque are not wrapped around its buffer, this head_len will be greater than the actual number of valid elements.

### DIFF
--- a/bigint/deprecated.mbt
+++ b/bigint/deprecated.mbt
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 ///|
-#deprecated("Use infix bitwise operator `>>` instead", skip_current_package=false)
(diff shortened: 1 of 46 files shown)


deque/deque.mbt
```diff
@@ -1941,25 +1941,39 @@ pub fn[A] Deque::chunk_by(
...
-  let mut len = 0
+  let mut total_len = 0
   for deque in self {
-    len += deque.length()
+    total_len += deque.length()
   }
+  if total_len == 0 {
+    return Deque::new()
+  }
+  let tail = if total_len == 0 { 0 } else { total_len - 1 }
```
peter-jerry-ye, line 1951: The second if condition is useless

PR #2990, opened 2025-11-25
Switch bit extraction to shift/mask and cover signed edges

bobzhang:
The branchless path lifts the byte to UInt/Int, shifts to the requested bit and masks with 1, which preserves the previous semantics without building a temporary mask byte. Byte::to_uint and Byte::to_int keep the 0-255 value so the shift works for both unsigned and signed projections, and the signed helper still maps set bits to -1 via the multiply/OR trick.\n\nNew tests exercise ArrayView and BytesView bit-sign extraction on 0x80/0xFF boundaries to prove the new logic keeps MSB handling intact.


bobzhang:
@copilot can you fix the CI failure by running `moon fmt`

PR #2995, opened 2025-11-26
hongbo/string get alias

bobzhang:
- **move UInt16 from its package to builtin**
- **migrate some tests**


bobzhang:
cc @peter-jerry-ye

PR #3008, opened 2025-12-01
promote pre-release

illusory0x0:
[DON'T MERGE] sed: /```moonbit/```mbt check/

### DIFF
--- a/array/pkg.generated.mbti
+++ b/array/pkg.generated.mbti
@@ -7,62 +7,63 @@ import(
 )
 
 // Values
-fn[A, B, C] zip_with(Array[A], Array[B], (A, B) -> C raise?) -> Array[C] raise?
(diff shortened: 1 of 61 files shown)


array/blit_js.mbt
```diff
@@ -25,7 +25,7 @@
 ///
 /// Example:
 ///
-/// ```moonbit
+/// ```mbt check
 ///   let original = [1, 2, 3]
```
bobzhang, line 29: this will be skipped by the toolchain -> `mbt test`
cc @bzy-debug we should emit a warning for this unexpected missing

array/README.mbt.md
```diff
@@ -6,7 +6,7 @@ This package provides array manipulation utilities for MoonBit, including fixed-
 
 There are several ways to create arrays in MoonBit:
 
-```moonbit
+```mbt check
 ///|
```
bobzhang, line 10: this is ok
`mbt check` and `mbt test` both work

PR #3015, opened 2025-12-03
refactor: move array to builtin

peter-jerry-ye:


bobzhang:
[trace.json](https://github.com/user-attachments/files/24035401/trace.json)
I attached a trace file and found something interesting:

1. one redundant command https://github.com/moonbitlang/moon/issues/1295

2. the overall 200ms looks acceptable to me

note this could be optimized further: -no-mi is not applied to whitebox/blackbox

```
git/core$ls -al /Users/hongbozhang/git/core/target/wasm-gc/release/check/builtin/builtin.whitebox_test.mi
-rw-r--r--@ 1 hongbozhang  staff  518629 Dec  8 22:58 /Users/hongbozhang/git/core/target/wasm-gc/release/check/builtin/builtin.whitebox_test.mi
```
we could generate an empty file here instead(not generating it could bring some trouble in build graph though)

PR #3027, opened 2025-12-09
fix(@json.parse): remove max_nesting_depth default parameter

illusory0x0:
`max_nesting_depth` is OS system setting, only should be set by `moonbitlang/core`, the user should not worry about it.


illusory0x0:
continue in https://github.com/moonbitlang/core/pull/3028/files

PR #3035, opened 2025-12-12
improve `Show` impl for `SourceLoc`

Guest0x0:
Previously, the `Show` implementation of `SourceLoc` displays something like:
```
@monbitlang/core/builtin:autoloc.mbt:1:1-2:2
```
One problem with this format is: tools like VSCode will recognize the whole `@package_path:file.mbt` as a file name, so users cannot jump to the corresponding location by clicking on test diff.

This PR tweaks the format to:
```
builtin/autoloc.mbt:1:1-2:2@moonbitlang/core
```
The package full path is split into the module part (i.e. content before the second slash) and package part (everything else), the package path is combined with the file name, and the module part is moved to the end. This way, the file name part (`builtin/autoloc.mbt`) is a valid file path relative to module root directory in a typical MoonBit project, and users can jump to the location by clicking.

Note that the change in this PR only affects the `Show` implementation of `SourceLoc`. Interaction with `moon` (via `inspect`, for example) is not affected.

### DIFF
--- a/builtin/autoloc.mbt
+++ b/builtin/autoloc.mbt
@@ -32,16 +32,16 @@ pub(all) type SourceLoc
 /// file path, line number, and column number.
 ///
 /// Returns a string representation of the source location, typically in the
-/// format "file:line:column".
+/// format "@package:file:start_line:start_column-end_line:end_column".
 ///
 /// Note: This function is primarily used internally by the compiler for error
 /// reporting and debugging purposes. Source locations are automatically created
 /// by the compiler when needed.
-pub fn SourceLoc::to_string(self : Self) -> String = "%loc_to_string"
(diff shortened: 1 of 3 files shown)


illusory0x0:
FTR: `inspect`, `@json.inspect`, `@test.Test::snapshot` doesn't directly call `SourceLoc::to_string` so this changes won't influence the `moon test --update`.

bobzhang:
@Guest0x0 As discussed offline, we need the relative location to project_root_path be reliable(regardless of whether using `src` field or not), since both moon and moonc konw such info, this is doable, having reliable error location is helpful for AI Agent to quick locate the failure place.

we are going to merge this, it does not do any harm, but we need follow up work to clean this up

PR #3040, opened 2025-12-14
BREAKING CHANGE: String::at and StringView::at now return UInt16

bobzhang:
## Breaking Changes

This PR changes the return type of `String::at` and `StringView::at` from `Int` to `UInt16`.

### Changes:
- `String::at` and `StringView::at` now return `UInt16` instead of `Int`
- These methods are now the primary interface for UTF-16 code unit access
- `code_unit_at` is now an alias for `at` (previously it was the other way around)
- The old Int-returning behavior is preserved in `String::charcode_at` (deprecated)

### Rationale:
This change aligns the API with the actual UTF-16 representation of strings and makes the type system more accurate.

### Migration Guide:
If you were using `String::at` or `StringView::at` and expecting an `Int`:
- Use `String::charcode_at` (deprecated) for temporary compatibility
- Or convert to `UInt16`: `str.at(i).to_int()`

### Updated:
- CHANGELOG.md with breaking change documentation

### DIFF
--- a/builtin/intrinsics.mbt
+++ b/builtin/intrinsics.mbt
@@ -1558,7 +1558,7 @@ pub fn String::length(self : String) -> Int = "%string_length"
 /// * `index` : The position in the string from which to retrieve the code unit.
 ///
 /// This method has O(1) complexity.
-#deprecated("Use `code_unit_at` instead which returns UInt16")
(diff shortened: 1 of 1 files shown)


bobzhang:
cc @peter-jerry-ye

PR #3048, opened 2025-12-16
feat: iterate SortedMap and SortedSet in reverse order

mndrix:
These methods follow the pattern used by the other rev_* methods in the standard library.


peter-jerry-ye:
I would like to postpone this until #3050 is settled

mndrix:
I've updated this pull request to use external iterators, according to #3050, and rebased it on the main branch.

mndrix:
I've rebased this pull request against `main`. I'll plan to keep doing this every couple weeks so this pull request stays valid until https://github.com/moonbitlang/core/pull/3099 is finished

PR #3050, opened 2025-12-18
deprecate `Iter::new`, migrate its usage in `moonbitlang/core`

Guest0x0:
This PR deprecates `Iter::new` & `Iter2::new`, and migrates their usage in `moonbitlang/core`. This is the next step in the process of migrating from internal iterator to external iterator, starting from #2851.

## Background
We have decided to migrate from internal iterator to external iterator for various reasons:

- `for .. in` body can perform `async` operations
- richer API such as `zip` and `equal`

One challenge here is incremental migration: we don't want to break user code hard during the transition from internal iterator to external iterator.

## The migration plan

Our migration plan from internal iterator to external iterator is as follows:

- [X] introduce external iterator type `Iterator` and `Iterator2` (#2851, #2859)
- [X] make `for .. in` loop use external iterator by default, fallback to internal iterator only when no `.iterator()` method is found
- [x] (this PR) deprecate `Iter::new` and `Iter2::new`. This is the only internal iterator API that external iterator cannot implement. Fortunately, end users do not invoke `Iter::new` directly in general. People who do uses `Iter::new`, such as author of a data structure package, should create an external iterator `Iterator[_]` first, and convert the external iterator to internal iterator via `Iterator::iter`.
- [ ] once  `Iter::new` is completely removed, `Iter` will become an alias of `Iterator`, and `Iterator::iter` will become the identity function. This change should be non-breaking, as all methods on `Iter` are supported by `Iterator`. During this stage, `Iterator::iter` may get deprecated to cleanup redundant APIs

## What users should do to adapt this PR

- for end users who do not invoke `Iter::new` directly, this PR should have no effect
- users who do invoke `Iter::new`, such as authors of data structure packages, should replace `Iter::new` with `Iterator::iter(xxx)`, where `xxx` is an external iterator with the same semantic as the previous internal one created via `Iter::new`
    - for `.iter()` methods used by `for .. in` loops, a corresponding external iterator version called `.iterator()` should be added

Creating external iterators is harder than creating internal iterators, especially when the data structure is complex. If you are not sure how to create an external iterator for your own data structure, changes in this PR is a good reference:

- for array-like structure, see `ArrayView::iterator` in `builtin/arrayview.mbt`, `Map::iterator` in `builtin/linked_hash_map.mbt`, and `StringView::iterator` in `builtin/stringview.mbt`
- for binary tree shaped structures, see `@sorted_map.SortedMap::iterator` in `sorted_map/map.mbt`. You can also find a more complex example in `@sorted_map.SortedMap::range`, which has an addition range limitation on iterated elements
- for rose tree like structure, see `@immut/array.Tree::iterator` in `immut/array/tree.mbt`

## One breaking behavior change
Although this PR should be non-breaking in terms of type signature, it introduces a breaking behavior change that is hard to avoid. Previously, the `Iter[_]` type, and various `.iter()` methods that create internal iterators, are *immutable*. The iterator can be iterated from the beginning multiple times. Iterating a iterator multiple times will result in duplicated computation, and side effects in the iterator will get triggered multiple times. So although possible, the pattern of traversing an iterator multiple times is discouraged for both performance and robustness reasons.

After this PR, `.iter()` methods in `moonbitlang/core` are implemented via the external iterator type `Iterator[_]`, which is mutable and can be traversed only once. Subsequent traversal will continue from where the last traversal ends, typically the end of the whole iterator. So the behavior of code that traverses a iterator multiple times will change.

### DIFF
--- a/builtin/iter.mbt
+++ b/builtin/iter.mbt
@@ -145,7 +145,7 @@ pub fn[T] Iter::count(self : Iter[T]) -> Int {
 
 ///|
 /// Do not use this method, it is for internal use only.
-#deprecated("`Iter::new`, and the internal iterator type `Iter`, is deprecated. Create an external iterator `Iterator`, and use `Iterator::iter` to migrate usage of `Iter::new`. `Iter` will become an alias of `Iterator` in the future.")
+#deprecated("`Iter::new`, and the internal iterator type `Iter`, is deprecated. Create an external iterator `Iterator`, and use `Iterator::iter` to migrate usage of `Iter::new`. `Iter` will become an alias of `Iterator` in the future. See https://github.com/moonbitlang/core/pull/3050 for more details.")
 pub fn[T] Iter::new(f : ((T) -> IterResult) -> IterResult) -> Iter[T] {
   Iter(f)
 }
--- a/builtin/iter2.mbt
+++ b/builtin/iter2.mbt
@@ -55,7 +55,7 @@ pub impl[A : Show, B : Show] Show for Iter2[A, B] with output(self, logger) {
 }
 
 ///|
-#deprecated("`Iter2::new`, and the internal iterator type `Iter2`, is deprecated. Create an external iterator `Iterator2`, and use `Iterator2::iter2` to migrate usage of `Iter2::new`. `Iter2` will become an alias of `Iterator2` in the future.")
+#deprecated("`Iter2::new`, and the internal iterator type `Iter2`, is deprecated. Create an external iterator `Iterator2`, and use `Iterator2::iter2` to migrate usage of `Iter2::new`. `Iter2` will become an alias of `Iterator2` in the future. See https://github.com/moonbitlang/core/pull/3050 for more details.")
 pub fn[A, B] Iter2::new(
   f : ((A, B) -> IterResult) -> IterResult,
 ) -> Iter2[A, B] {


builtin/array_test.mbt
```diff
@@ -1156,19 +1156,6 @@ test "zip_to_iter2 early termination" {
...
-  let arr = [1, 2, 3]
-  let arr2 = ['a', 'b', 'c']
-  let iter = arr.zip_to_iter2(arr2)
-  arr.push(4)
-  arr2.push('d')
-  inspect(iter.to_array(), content="[(1, 'a'), (2, 'b'), (3, 'c'), (4, 'd')]")
-  arr.clear()
-  inspect(iter.to_array(), content="[]")
-}
-
```
Guest0x0, line 1171: this test pushes elements into the array during iteration, this is undefined behavior, and the actual behavior changes in the new implementation. In the worst case modifying data structure while iterating can crash the program, so we can provide no guarantee is this case

immut/array/tree.mbt
```diff
@@ -304,6 +290,7 @@ fn[A] Tree::iterator(self : Tree[A]) -> Iterator[A] {
       Node(children, _) as t if curr_index < children.length() => {
         let child = children.unsafe_get(curr_index)
         parents.push((t, curr_index + 1))
+        curr_tree = child
```
Guest0x0, line 293: this is a bug in the previous implementation. After this PR, tests that previously use internal iterator now use external iterator under the hood, triggering the bug

gmlewis:
@peter-jerry-ye - are you planning on updating your utf8 streaming iterator here: https://github.com/peter-jerry-ye/jstream ?

peter-jerry-ye:
> @peter-jerry-ye - are you planning on updating your utf8 streaming iterator here: https://github.com/peter-jerry-ye/jstream ?

No, I don't think so. I think the streaming encoder in x is good enough.

PR #3059, opened 2025-12-21
run `moon fmt`

bobzhang:


sorted_map/map.mbt
```diff
@@ -47,14 +47,16 @@ pub fn[K : Compare, V] SortedMap::from_array(
...
 ///|
-/// Inserts a key-value pair.
-/// ```mbt test
-/// let map = @sorted_map.new()
-/// map.set(1, "a")
-/// map.set(2, "b")
-/// map.set(2, "c") // updates value for key 2
-/// inspect(map.get(1), content="Some(\"a\")")
-/// inspect(map.get(2), content="Some(\"c\")")
+/// FIXME: (remove this line will break formatter)
```
bobzhang, line 50: cc @Yoorkin this seems to be a bug in moonfmt

builtin/iter.mbt
```diff
@@ -435,14 +435,16 @@ pub fn[T, R] Iter::flat_map(self : Iter[T], f : (T) -> Iter[R]) -> Iter[R] {
...
-/// iter.map(f).flatten() == iter..flat_map(f)
-/// ```mbt test 
-/// // ignore this test case for now since it will cause ice in release build
-/// // test {
-/// //   fn f(n : Int) { Int::until(0,n) }
-/// //   let xs = f(10)
-/// //   assert_eq(xs.map(f).flatten().to_array(),xs.flat_map(f).to_array())
-/// // }
+/// FIXME:(remove this line will break formatter)
+/// ```mbt check
```
bobzhang, line 439: same thing @Yoorkin

PR #3074, opened 2025-12-29
deque: align head/tail for empty constructors

bobzhang:
Single-commit fix for review.

Commit: 242550ba


bobzhang:
fixed in another incoming PR

PR #3105, opened 2026-01-04
refactor: use js stdlib for double2str

peter-jerry-ye:


mndrix:
Thank you for this change. I've been doing the same thing in my code to reduce .js size.

bobzhang:
FTR: we are going to align with JS double2str in the future when any inconsistency found

PR #3115, opened 2026-01-06
Use lexmatch for json, autoloc, and strconv parsing

bobzhang:


builtin/autoloc.mbt
```diff
@@ -87,34 +81,23 @@ impl Show for SourceLocRepr with output(self, logger) {
...
-  fn parse_loc(view : StringView) -> (StringView, StringView)? {
-    guard view.find(":") is Some(i) else { None }
-    guard i > 0 && i + 1 < view.length() else { None }
-    Some((view.view(end_offset=i), view.view(start_offset=i + 1)))
+  lexmatch repr[:] {
+    (
+      "@"
+      ("[^:]*" as pkg)
+      ":"
+      (".+" as filename)
```
hackwaly: This will eat all chars. Should be ".*?".

builtin/autoloc.mbt
```diff
@@ -87,34 +81,23 @@ impl Show for SourceLocRepr with output(self, logger) {
...
 ///|
 fn SourceLocRepr::parse(repr : String) -> SourceLocRepr {
-  fn parse_loc(view : StringView) -> (StringView, StringView)? {
-    guard view.find(":") is Some(i) else { None }
-    guard i > 0 && i + 1 < view.length() else { None }
-    Some((view.view(end_offset=i), view.view(start_offset=i + 1)))
+  lexmatch repr[:] {
+    (
+      "@"
+      ("[^:]*" as pkg)
```
hackwaly, line 87: Could use ".*?".

builtin/autoloc.mbt
```diff
@@ -87,34 +81,23 @@ impl Show for SourceLocRepr with output(self, logger) {
...
+      ":"
+      (".+" as filename)
+      ":"
+      ("[^:]+" as start_line)
+      ":"
+      (".+" as start_column)
```
hackwaly: ".+" will eat '-'. Should use ".*?".

builtin/autoloc.mbt
```diff
@@ -87,34 +81,23 @@ impl Show for SourceLocRepr with output(self, logger) {
 
 ///|
 fn SourceLocRepr::parse(repr : String) -> SourceLocRepr {
```
hackwaly, line 83: This function may have no test coverage?

builtin/autoloc.mbt
bobzhang: Updated to non-greedy .+? for filename and digit-only line/column so separators do not get consumed; added regression test with colons in filename. Commit 740be3b3.

builtin/autoloc.mbt
bobzhang: There is coverage in builtin/source_loc_wbtest.mbt; I also added a new case with colons in the filename to exercise this path. Commit 740be3b3.

bobzhang:
landed separately

PR #3125, opened 2026-01-07
add `@debug.inspect`

Yoorkin:
- add `@debug.inspect`
- move tests to `debug_test.mbt`


hackwaly:
Will we have @debug.assert_eq?

bobzhang:
> Will we have @debug.assert_eq?

Yes

PR #3129, opened 2026-01-07
refactor(debug): removing unnecessary conversions

FlyCloudC:
This PR refactors some details in the debug pkg.

1. Beta reduction
2. Since `ArrayView::map` returns an `Array`, `view.to_array().map(f)` can be replaced with `view.map(f)`.

### DIFF
--- a/debug/debug.mbt
+++ b/debug/debug.mbt
@@ -106,24 +106,21 @@ pub impl[T : Debug] Debug for Array[T] with to_repr(self) {
 
 ///|
 pub impl[T : Debug] Debug for ArrayView[T] with to_repr(self) {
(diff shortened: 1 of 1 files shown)


FlyCloudC:
By the way, could we mention the equivalence between `view.to_array().map(f)` and `view.map(f)` in the documentation for `ArrayView::to_array`, or rename `ArrayView::map` to a clearer name, so that AI can generate more efficient code?

bobzhang:
> By the way, could we mention the equivalence between view.to_array().map(f) and view.map(f) in the documentation for ArrayView::to_array, or rename ArrayView::map to a clearer name, so that AI can generate more efficient code?

lgtm

PR #3137, opened 2026-01-09
fix: add missing return for timsort shortcut

Yu-zh:
Found this when I was playing with AI agent on the loop invariants and reasoning


bobzhang:
let me try to ask AI to create a test revealing the bug and fix it later

PR #3145, opened 2026-01-12
remove unused string/bytes in deps

hackwaly:
This PR removes unused `moonbitlang/core/string` and `moonbitlang/core/bytes` packages from deps. This is a major blocker for #3141 , which expect to put regex surface api directly inside `string` and `bytes` packages.

Other packages for builtin types in deps should cleanup this way as well, but there's no convenient way for now, as the unused package warning not working. /cc @peter-jerry-ye @Yu-zh 

I didn't investigate why unused package warning not working. This should raise another issue.

### DIFF
--- a/bigint/moon.pkg.json
+++ b/bigint/moon.pkg.json
@@ -6,8 +6,6 @@
     "moonbitlang/core/json",
     "moonbitlang/core/quickcheck",
     "moonbitlang/core/quickcheck/splitmix",
-    "moonbitlang/core/string",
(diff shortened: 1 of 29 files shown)


hackwaly:
@copilot create issue for "unused package warning not working for packages in moonbitlang/core".

Yu-zh:
Looks like they are indirect dependencies, and we currently don't report warnings on imports that could be omitted because they are indirect dependency.

PR #3157, opened 2026-01-16
Fix/regex engine warnings

bobzhang:


bobzhang:
Note we should ask AI first qualify these binders and remove `using` after

PR #3174, opened 2026-01-21
fix alias of deprecated stuff in `@prelude`

Guest0x0:
Currently, the behavior of `using` on deprecated stuff is:

- the `#deprecated` attribute is inherited by the new alias
- `using` itself never report a deprecation warning

but this behavior is not desirable. It is not intuitive, and may lead to warning that cannot be fixed for downstream of `pub using`. Current behavior also forbids deprecating the alias generated by `using` only, which is useful for `pub using`. So we plan to change the behavior of `using` on deprecated stuff to:

- the `#deprecated` attribute is not inherited by the new alias in `using`
- `using` itself will show a deprecation warning when the alias target is deprecated
- `using` itself can be annotated with `#deprecated`, which will deprecate the alias and only the alias

This PR prepares `moonbitlang/core` for this semantic change by temporarily avoid `using` for deprecated stuff in `@prelude`, so that we won't run into invalid state when the MoonBit compiler is performing the migration.

---

<a href="https://app.devin.ai/review/moonbitlang/core/pull/3174">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/prelude/pkg.generated.mbti
+++ b/prelude/pkg.generated.mbti
@@ -87,6 +87,12 @@ pub using @builtin {type Iter2}
 
 pub using @builtin {type IterResult}
 
+#deprecated
+pub using @builtin {type Iter as Iterator}
+
+#deprecated
+pub using @builtin {type Iter2 as Iterator2}
+
 pub using @builtin {type Json}
 
 pub using @builtin {type Map}
--- a/prelude/prelude.mbt
+++ b/prelude/prelude.mbt
@@ -33,7 +33,6 @@ pub using @builtin {
   println,
   not,
   repr,
-  dump,
   trait Eq,
   trait Compare,
   trait Hash,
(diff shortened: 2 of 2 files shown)


prelude/pkg.generated.mbti
```diff
@@ -87,6 +87,12 @@ pub using @builtin {type Iter2}
 
 pub using @builtin {type IterResult}
 
+#deprecated
+pub using @builtin {type Iter as Iterator}
+
+#deprecated
+pub using @builtin {type Iter2 as Iterator2}
+
```
bobzhang, line 95: curious why this has a diff? @Guest0x0

Guest0x0: these two alias were previously missing from `.mbti`, could be a bug in `moon info`, will investigate later

PR #3213, opened 2026-02-10
feat: argparse

peter-jerry-ye:
This PR adds an arg parser that is inspired by clap but with reduced API because we want to focus on the clarity of the behavior. A lot of extra checks were imposed.

The API of this PR is roughly determined, but the internal implementation can still be improved drastically.

@FlyCloudC Do you think you would like to give it a review in terms of the API? (I know the internal implementation is a bit messy).

---

<a href="https://app.devin.ai/review/moonbitlang/core/pull/3213" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/argparse/README.mbt.md
+++ b/argparse/README.mbt.md
@@ -6,6 +6,8 @@ This package is inspired by [`clap`](https://github.com/clap-rs/clap) and keeps 
 small, predictable feature set.
 
 `long` defaults to the argument name. Pass `long=""` to disable long alias.
+Argument constructors are named `FlagArg` / `OptionArg` / `PositionArg` to
+avoid shadowing built-in types like `Option`.
 
 
 ## Quick Start
@@ -17,8 +19,8 @@ both the parse error and the full contextual help text.
 ///|
 test "basic option + positional success snapshot" {
   let matches = @argparse.parse(
-    Command("demo", options=[Optional("name")], positionals=[
-      Positional("target"),
+    Command("demo", options=[OptionArg("name")], positionals=[
+      PositionArg("target"),
     ]),
     argv=["--name", "alice", "file.txt"],
     env={},
@@ -33,8 +35,8 @@ test "basic option + positional success snapshot" {
 
 ///|
 test "basic option + positional failure snapshot" {
-  let cmd = @argparse.Command("demo", options=[Optional("name")], positionals=[
-    Positional("target"),
+  let cmd = @argparse.Command("demo", options=[OptionArg("name")], positionals=[
+    PositionArg("target"),
   ])
   try cmd.parse(argv=["--bad"], env={}) catch {
     err =>
@@ -69,8 +71,8 @@ contextual help text.
 ```mbt nocheck
 ///|
 async fn main {
-  let cmd = @argparse.Command("demo", options=[@argparse.Optional("name")], positionals=[
-    @argparse.Positional("target"),
+  let cmd = @argparse.Command("demo", options=[@argparse.OptionArg("name")], positionals=[
+    @argparse.PositionArg("target"),
   ])
   let _ = cmd.parse()
 }
@@ -85,7 +87,7 @@ states.
 ```mbt check
 ///|
 test "negatable flag success snapshot" {
-  let cmd = @argparse.Command("demo", flags=[Flag("cache", negatable=true)])
+  let cmd = @argparse.Command("demo", flags=[FlagArg("cache", negatable=true)])
   inspect(
     cmd.render_help(),
     content=(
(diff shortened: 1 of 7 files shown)


FlyCloudC:
I have the following suggestions:

The parameter `(args: Array[&ToArg])` in `Command::new` could be replaced with `(flags: Array[FlagArg], options: Array[OptionArg], positionals: Array[PositionalArg])`(or `flags` and `options` could be merged into `nameds : Array[&ToArg]`). This avoids mixing three different types of parameters together. After all, these three are not grouped together in the help display either. Additionally, this approach eliminates the need to write package names since the types are already determined.

Another optional suggestion is to consider merging the three parameters (`required`, `env`, and `default_values`) using sum type modeling (though the current approach seems acceptable as well)

FlyCloudC:
A question: Why are errors generated when building a Command deferred until argument parsing time to be raised?

It would be better if they could be raised immediately. I mean: changing the signature of `Command::new` to raise `ArgBuildError`.

FlyCloudC:
`Positional::new` already accepted `num_args`, so there is no need to accept `required`

peter-jerry-ye:
> I have the following suggestions:
> 
> The parameter `(args: Array[&ToArg])` in `Command::new` could be replaced with `(flags: Array[FlagArg], options: Array[OptionArg], positionals: Array[PositionalArg])`(or `flags` and `options` could be merged into `nameds : Array[&ToArg]`). This avoids mixing three different types of parameters together. After all, these three are not grouped together in the help display either. Additionally, this approach eliminates the need to write package names since the types are already determined.
> 
> Another optional suggestion is to consider merging the three parameters (`required`, `env`, and `default_values`) using sum type modeling (though the current approach seems acceptable as well)

The rational was to avoid making them to verbose, because we would have

```moonbit
Command(flags=[FlagArg(xxx)], options=[OptionArg(xxx)])
```

instead of having just

```moonbit
Command(args=[FlagArg(xxx), OptionArg(xxx)])
```

peter-jerry-ye:
> A question: Why are errors generated when building a Command deferred until argument parsing time to be raised?
> 
> It would be better if they could be raised immediately. I mean: changing the signature of `Command::new` to raise `ArgBuildError`.

The rational is that the final check involves checking the configuration from the parent + the subcommand. And it is not expected to be raised too often because when we finish defining the arg parser, it is likely to be stable. I think for clap in Rust it's actually checked only under debug mode.

FlyCloudC:
> > I have the following suggestions:
> > The parameter `(args: Array[&ToArg])` in `Command::new` could be replaced with `(flags: Array[FlagArg], options: Array[OptionArg], positionals: Array[PositionalArg])`
> 
> The rational was to avoid making them to verbose, because we would have
> 
> ```moonbit
> Command(flags=[FlagArg(xxx)], options=[OptionArg(xxx)])
> ```
> 
> instead of having just
> 
> ```moonbit
> Command(args=[FlagArg(xxx), OptionArg(xxx)])
> ```

Actually, here's how it looks when users use it:

```moonbit
@argparse.Command(args=[@argparse.FlagArg(xxx), @argparse.OptionArg(xxx)])
```

instead of

```moonbit
@argparse.Command(flags=[FlagArg(xxx)], options=[OptionArg(xxx)])
```

Unless the user imports it with `using`

FlyCloudC:
> > A question: Why are errors generated when building a Command deferred until argument parsing time to be raised?
> > It would be better if they could be raised immediately. I mean: changing the signature of `Command::new` to raise `ArgBuildError`.
> 
> The rational is that the final check involves checking the configuration from the parent + the subcommand. ...

I'm not entirely sure what you mean. Do you mean that this situation might exist?
```
let sub1 = Command(...) // invalid
let sub2 = Command(...) // invalid
let parent = Command(..., subcommands=[sub1, sub2]) // valid
```
So the error cannot be raised in the first two lines

FlyCloudC:
> > A question: Why are errors generated when building a Command deferred until argument parsing time to be raised?
> > It would be better if they could be raised immediately. I mean: changing the signature of `Command::new` to raise `ArgBuildError`.
> 
> The rational is that the final check involves checking the configuration from the parent + the subcommand. And it is not expected to be raised too often because when we finish defining the arg parser, it is likely to be stable. I think for clap in Rust it's actually checked only under debug mode.

I just thought about it again, and I believe `ArgBuildError` is not an Error. If it occurs, it indicates a bug in the program that defines the Command (which shouldn't exist in a released CLI tool) and is unrecoverable, so it should just abort instead.

peter-jerry-ye:
@FlyCloudC I agree. Having `&ArgLike` makes it more verbose. I also agree that it should be a panic, but I would like to leave it for the user to handle.

PR #3223, opened 2026-02-12
chore(debug): use gumtree diff in @debug.assert_eq

myfreess:


bobzhang:
diff is in the core now, user may adapt it if they need tell the difference

PR #3230, opened 2026-02-25
refactor(argparse): use sum type in Arg

FlyCloudC:
---

<a href="https://app.devin.ai/review/moonbitlang/core/pull/3230" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/argparse/arg_spec.mbt
+++ b/argparse/arg_spec.mbt
@@ -53,15 +53,14 @@ priv enum ArgInfo {
     long~ : String?,
     action~ : OptionAction,
     default_values~ : Array[String]?,
-    allow_hyphen_values~ : Bool,
-    last~ : Bool
+    allow_hyphen_values~ : Bool
   )
   Positional(
     index~ : Int?,
     num_args~ : ValueRange?,
+    last~ : Bool,
     default_values~ : Array[String]?,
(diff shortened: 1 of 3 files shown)


argparse/parser_lookup.mbt
peter-jerry-ye, line 1: I did consider using sum type in `Arg`, but the result is that we'll have a lot of things as

```moonbit
    if arg.info is (Flag(long~, ..) | Option(long~, ..)) && long is Some(name) {
```

instead of

```moonbit
    if arg.long is Some(name)
```

so I hesitated

FlyCloudC: Yes, I also find this strange. Maybe I should add a `long() -> String?` method to `Arg`, but that doesn't seem much different from before. I hope the domain model should be clearer, for example, instead of `Command` storing an `Array[Arg]`, it could store multiple arrays like `Array[FlagArg]`, `Array[OptionArg]`, etc. But I'm not sure how to model it.

peter-jerry-ye: I think the problem is that sometimes we need to distinguish between them, while sometimes we don't really care what they are and we only want a uniformed handling. Anyway I'm merging this and we'll focus on the API first.

PR #3244, opened 2026-03-04
refactor: use `ref::Ref` instead of compiler defined Ref

peter-jerry-ye:
We will make the `Ref` in the core point to a type defined here instead of embedded in the compiler.

This will be a breaking change for people using plain `{ val : T }` without any type annotation, but this has been warned for quite some time.

---

<a href="https://app.devin.ai/review/moonbitlang/core/pull/3244" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


Guest0x0:
Currently struct constructor does not play well with `pub using` re-export. We should fix that later in the compiler.

PR #3247, opened 2026-03-05
feat(diff): add Myers/patience diff with hunk grouping

myfreess:
This PR introduces a Myers-based diff package to provide efficient sequence diffing, structured edit scripts, and hunk-level grouping for downstream diff-style output.

---

<a href="https://app.devin.ai/review/moonbitlang/core/pull/3247" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/diff/pkg.generated.mbti
+++ b/diff/pkg.generated.mbti
@@ -1,0 +1,42 @@
+// Generated using `moon info`, DON'T EDIT IT
+package "moonbitlang/core/diff"
+
+import {
+  "moonbitlang/core/debug",
+}
+
+// Values
+
+// Errors
+
+// Types and methods
+pub struct Diff[T] {
+  old : ArrayView[T]
+  new : ArrayView[T]
+  edits : Array[Edit]
+}
+pub fn[T : Hash + Eq] Diff::Diff(old~ : ArrayView[T], new~ : ArrayView[T], cutoff? : Int, algorithm? : DiffAlgorithm) -> Self[T]
+pub fn[T] Diff::group(Self[T], radius? : Int) -> Array[Hunk[T]]
+
+pub(all) enum DiffAlgorithm {
+  Myers
+  Patience
+}
+
+pub enum Edit {
+  Delete(old_index~ : Int, new_index~ : Int, len~ : Int)
+  Insert(old_index~ : Int, new_index~ : Int, len~ : Int)
+  Equal(old_index~ : Int, new_index~ : Int, len~ : Int)
+} derive(@debug.Debug)
+
+pub struct Hunk[T] {
+  edits : ArrayView[Edit]
+  old : ArrayView[T]
+  new : ArrayView[T]
+}
+pub impl[T : Show] Show for Hunk[T]
+
+// Type aliases
+
+// Traits
+
--- a/diff/pkg.generated.mbti
+++ b/diff/pkg.generated.mbti
@@ -34,6 +34,7 @@ pub struct Hunk[T] {
   old : ArrayView[T]
   new : ArrayView[T]
 }
+pub fn[T : Show] Hunk::to_string(Self[T]) -> String
 pub impl[T : Show] Show for Hunk[T]
 
 // Type aliases
--- a/diff/pkg.generated.mbti
+++ b/diff/pkg.generated.mbti
@@ -10,32 +10,28 @@ import {
 // Errors
 
 // Types and methods
-pub struct Diff[T] {
-  old : ArrayView[T]
-  new : ArrayView[T]
-  edits : Array[Edit]
-}
+type Diff[T]
 pub fn[T : Hash + Eq] Diff::Diff(old~ : ArrayView[T], new~ : ArrayView[T], cutoff? : Int, algorithm? : DiffAlgorithm) -> Self[T]
-pub fn[T] Diff::group(Self[T], radius? : Int) -> Array[Hunk[T]]
+pub fn[T] Diff::edits(Self[T]) -> ArrayView[Edit]
+pub fn[T] Diff::group(Self[T], context? : Int) -> Array[Hunk[T]]
+pub fn[T] Diff::new_view(Self[T]) -> ArrayView[T]
+pub fn[T] Diff::old_view(Self[T]) -> ArrayView[T]
 
 pub(all) enum DiffAlgorithm {
   Myers
   Patience
-}
+} derive(Eq, @debug.Debug)
+pub fn DiffAlgorithm::equal(Self, Self) -> Bool
 
 pub enum Edit {
-  Delete(old_index~ : Int, new_index~ : Int, len~ : Int)
-  Insert(old_index~ : Int, new_index~ : Int, len~ : Int)
+  Delete(old_index~ : Int, old_len~ : Int, new_index~ : Int)
+  Insert(old_index~ : Int, new_index~ : Int, new_len~ : Int)
   Equal(old_index~ : Int, new_index~ : Int, len~ : Int)
-} derive(@debug.Debug)
+} derive(Eq, @debug.Debug)
+pub fn Edit::equal(Self, Self) -> Bool
 
-pub struct Hunk[T] {
-  edits : ArrayView[Edit]
-  old : ArrayView[T]
-  new : ArrayView[T]
-}
-pub fn[T : Show] Hunk::to_string(Self[T]) -> String
-pub impl[T : Show] Show for Hunk[T]
+type Hunk[T]
+pub fn[T] Hunk::render(Self[T], show~ : (T) -> String) -> String
 
 // Type aliases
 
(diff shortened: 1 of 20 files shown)


diff/pkg.generated.mbti
```diff
@@ -0,0 +1,40 @@
...
+}
+pub fn[T] Edit::take_tagged_view_from(Self, old~ : MutArrayView[T], new~ : MutArrayView[T]) -> (EditTag, ArrayView[T])
+pub impl @debug.Debug for Edit
+
+pub enum EditTag {
+  Delete
+  Insert
+  Equal
+}
+
```
bobzhang, line 19:
```
struct Edits (ArrayView[Edit]))

fn Edits::to_hunk_string(..)
fn Edits::to_string(style=Diff| Patch| Unified_diff)
```

myfreess:
@codex review

myfreess:
@codex review

bobzhang (commented):
Review focused on `diff/pkg.generated.mbti` — the shape is broadly right (one constructor, one grouping method, three types), and `Diff::Diff(old~, new~, ...)` correctly follows the `Type::Type` convention. Comments below are mostly about the public surface.

## 1. `Hunk.old` / `Hunk.new` are the whole sequences, not the hunk's slices

`edit.mbt:119` constructs `{ edits: pending, old, new }` where `old`/`new` are `Diff::group`'s full views, so `hunk.old` is the entire old sequence. But the mbti reads as "this hunk's old lines", and the doc comment says "with its edit script and the matching source ranges" (`hunk.mbt:74`), which reinforces the wrong reading. Someone writing `for line in hunk.old` gets the whole input back.

Making `Hunk` an abstract type resolves this cleanly: expose `Hunk::old_view()` / `new_view()` returning the hunk's actual slice and keep the full-source views private.

## 2. Please don't `impl Show for Hunk[T]`

`Show` here is a *renderer*: it emits unified-diff text with a trailing newline. Core is actively moving in the opposite direction — on `main`, every collection-shaped `Show` impl is now deprecated: 39 of them across `builtin` (`Array`, `ArrayView`, `FixedArray`, `Map`, `Iter`, …), all 16 container packages, and all tuple arities, plus the matching `extend` entries. All carry the message:

> Use `@debug.Debug` instead of `Show` for collections

This is the Rust `Display`/`Debug` split: `Debug` is implemented for all public types and shows internal state; `Display` is *"not expected that all types implement"* and is reserved for values with one obvious faithful string form. `Vec<T>` implements `Debug` and deliberately not `Display`, because there's no single correct separator/bracket choice. Rust's idiom for Display-ish output of a structured value is an explicit adapter — `path.display()`, `slice.join(sep)` — which is what `@debug.Repr(x)` is in core.

Concretely, that suggests:

- `derive(@debug.Debug)` on `Hunk` and `Diff` (as `Edit` already does) for the repr,
- an explicit renderer instead of `Show`: `Hunk::to_string(style? : Style, render? : (T) -> String) -> String`.

Two things fall out of dropping `Show`. It frees the `to_string` name for the parameterized version — you can't have both a promoted `Show::to_string` and a `to_string(style~)`, they collide — which is what makes @bobzhang's suggested `to_string(style=Diff|Patch|Unified_diff)` signature possible. And the `T : Show` bound stops smuggling in an assumption: unified-diff output is line-oriented, so `Show for Hunk[T]` looks generic while quietly requiring `T` to render as a line. An explicit `render? : (T) -> String` makes that honest.

Rendering probably also belongs at the `Diff` level rather than `Hunk` — a unified diff is a sequence of hunks plus optional headers, and today printing a whole diff means `group()` plus a manual join.

## 3. Nothing public supports rendering in any other format

`Edit::old_range`, `Edit::new_range` and `Edit::view_from` are all private, so `Show` is the only supported output. `diff_test.mbt:18-27` hand-rolls its own renderer by pattern-matching and calling `.view(start=…, end=…)` — that's a sign the surface is a function short. Suggest making `Edit::old_range` / `new_range` public.

## 4. `Edit` derives

Currently `derive(@debug.Debug)` only. Adding `Eq` would let consumers (and tests) compare edit scripts directly — the README currently has to assert with `d.edits[:] is [Equal(old_index=0, …), …]` because `assert_eq` isn't available. `@debug.Debug` is the right repr trait here, so no need for `Show`.

## 5. `Edit` is `pub`, so it can be matched but not constructed

`DiffAlgorithm` is `pub(all)`; `Edit` isn't. For a plain-data edit script, `pub(all)` is what lets downstream code build and transform scripts (applying a patch, test fixtures, round-tripping). If read-only is deliberate, worth saying why in the doc comment.

## 6. `Diff.edits : Array[Edit]` exposes mutable internal state

`pub struct` blocks construction and field assignment but not mutation of the array itself, so a caller can `d.edits.clear()` or push an inconsistent edit and `group()` will consume it. It's also inconsistent with `Hunk.edits : ArrayView[Edit]`. Suggest `ArrayView[Edit]` in both.

More broadly, consider making `Diff` abstract too. The representation isn't settled — @bobzhang's `Edits` wrapper proposal would change what `Diff` holds — and publishing fields is a one-way door: you can always expose them later, you can't retract them. Note `new` doesn't survive as a method name (it collides with the conventional zero-arg constructor), so accessors would need names like `old_seq()` / `new_seq()`.

## 7. Missing `extends.mbt`

Every other package now declares trait-method promotions explicitly. `moon check -p diff` currently reports:

```
diff/edit.mbt:40  implicit_impl_as_method: to_repr from `impl @debug.Debug for Edit` implicitly promoted
diff/hunk.mbt:82  implicit_impl_as_method: to_string, output from `impl Show for Hunk` implicitly promoted
```

Per AGENTS.md and the `strconv/extends.mbt` precedent, `@debug.Debug::{to_repr}` wants the `#deprecated(…, skip_current_package=true)` + `#doc(hidden)` form. There are also implicit-promotion and `prefer_readonly_array` warnings in the test files.

## Smaller notes

- `DiffAlgorithm` stutters as `@diff.DiffAlgorithm`; `@diff.Algorithm` reads better. (`@diff.Diff` is fine — that's the `@bigint.BigInt` pattern.)
- `radius` is called *context* by git and friends, and `radius < 0` aborts (`edit.mbt:93`) but only the README says so — `Diff::group`'s doc comment should carry the panic note, or clamp.
- `cutoff? : Int` is opaque from the mbti alone; the constructor doc explains it, but something like `max_cost` would self-document.
- `Diff` retains `ArrayView`s into caller-owned arrays — if the caller mutates them afterwards the stored edits go stale and rendering can index out of bounds. Worth a line in the doc.
- `Hash + Eq` is required even for `Myers` because `make_indexer` (`diff.mbt:432`) hashes for the presence filter. Defensible, just noting it excludes `Eq`-only element types.
- README: the `[:]` in `["apple", …][:]` isn't needed — `Array[T]` coerces to `ArrayView[T]` at the call site (verified against this package). Dropping it makes the examples match the other packages' READMEs.

bobzhang:
API redesign pushed (`fe5e4b35`), following maintainer review + two rounds of codex design review. Final surface:

```
type Diff[T]
pub fn[T : Hash + Eq] Diff::Diff(old~ : ArrayView[T], new~ : ArrayView[T], cutoff? : Int, algorithm? : DiffAlgorithm) -> Self[T]
pub fn[T] Diff::edits(Self[T]) -> ArrayView[Edit]
pub fn[T] Diff::group(Self[T], context? : Int) -> Array[Hunk[T]]
pub fn[T] Diff::new_view(Self[T]) -> ArrayView[T]
pub fn[T] Diff::old_view(Self[T]) -> ArrayView[T]

pub(all) enum DiffAlgorithm { Myers; Patience } derive(Eq, @debug.Debug)

pub enum Edit {
  Delete(old_index~ : Int, old_len~ : Int, new_index~ : Int)
  Insert(old_index~ : Int, new_index~ : Int, new_len~ : Int)
  Equal(old_index~ : Int, new_index~ : Int, len~ : Int)
} derive(Eq, @debug.Debug)

type Hunk[T]
pub fn[T] Hunk::render(Self[T], show~ : (T) -> String) -> String
```

Key decisions (details in the commit message): `Diff`/`Hunk` abstract (closes the mutable-`edits` invariant hole), rendering is hunk-only via a `show~` callback instead of a `Show` bound, per-variant edit lengths (`old_len`/`new_len`) with a documented cursor contract, `group(context?)` per unified-diff terminology, `cutoff` kept with precise edit-distance semantics, and a consciously recorded **no-`Replace`-variant** decision (pub enums are matched exhaustively, so adding it later would be breaking).

Also added an exhaustive property test: replaying the edit script over `old` reconstructs `new` (monotone cursors, no zero-length edits) across all {0,1}-array pairs up to length 4 × both algorithms × default and tiny `cutoff` — verifying validity independently of which minimal script the deterministic tie-breaking picks, including on the heuristic path.

Verified: `moon check --deny-warn --target all` clean; diff tests 84/84 on wasm/wasm-gc/js/native; full suite green.

PR #3317, opened 2026-03-19
fix(buffer): prevent integer overflow in grow_if_necessary

bobzhang:
## Summary

The `Buffer::grow_if_necessary` doubles the capacity in a loop until it meets the required size. When the current capacity exceeds `Int.max_value / 2` (~1 billion), `space * 2` overflows to a negative value, causing either an infinite loop or a panic when allocating with a negative size.

### Fix
Add a guard: when `space > 1073741823` (Int max / 2), fall back to allocating exactly the required size instead of doubling.

## Test plan
- [x] All 96 buffer tests pass

---

<a href="https://app.devin.ai/review/moonbitlang/core/pull/3317" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/buffer/buffer.mbt
+++ b/buffer/buffer.mbt
@@ -49,7 +49,9 @@ fn Buffer::grow_if_necessary(self : Buffer, required : Int) -> Unit {
     if space >= required {
       break space
     }
-    if space > 1073741823 { // Int max / 2, prevent overflow
+    // Guard against integer overflow: when space > Int max / 2 (1073741823),
+    // doubling would exceed 32-bit Int range. Fall back to exact allocation.
+    if space > 1073741823 {
       break required
     }
     continue space * 2
--- a/buffer/buffer.mbt
+++ b/buffer/buffer.mbt
@@ -49,6 +49,9 @@ fn Buffer::grow_if_necessary(self : Buffer, required : Int) -> Unit {
     if space >= required {
       break space
     }
+    if space > 1073741823 { // Int max / 2, prevent overflow
+      break required
+    }
     continue space * 2
   }
   if enough_space != self.data.length() {


buffer/buffer.mbt
```diff
@@ -49,6 +49,11 @@ fn Buffer::grow_if_necessary(self : Buffer, required : Int) -> Unit {
     if space >= required {
       break space
     }
+    // Guard against integer overflow: when space > Int max / 2 (1073741823),
+    // doubling would exceed 32-bit Int range. Fall back to exact allocation.
+    if space > 1073741823 {
```
peter-jerry-ye, line 54: We might use a constant instead of a random number. Also the current runtime should be put into account -> https://github.com/moonbitlang/moonbit-docs/issues/1155

bobzhang:
Superseded — recommend closing.

`b903d495` "fix: make buffer growth overflow-safe" (zihang, 2026-07-28) restructured this code into `buffer_growth_capacity` (`buffer/buffer.mbt:46`) and fixed the same bug:

```moonbit
let next = space * 2
if next <= space { break required }
```

Rebasing this PR onto current `main` makes **both commits empty** — the conflict is head-on (main's overflow test and this PR's `space > 1073741823` occupy the same lines for the same purpose), and taking main's side leaves `git diff origin/main` empty.

To be precise about the difference, since I got this wrong at first: the two are **behaviorally identical**. `space > 1073741823` is false at `space == 1073741823`, so this PR also permits the last valid doubling to `2147483646`; both break exactly when `space >= 2^30`. Main is better only in that it has no magic constant — which is the substance of @peter-jerry-ye's first comment.

Main also fixes a **second** overflow this PR never touched: `required < len` aborts when a caller's `self.len + n` wraps, which the old code would have treated as already-satisfied and left the buffer undersized before the write.

Verified on current main by probing `buffer_growth_capacity` directly:

| current | required | result |
| ---: | ---: | ---: |
| 2^30 | 1073741828 | 1073741828 |
| 1073741823 | 1073741824 | 2147483646 |
| 1 | 2147483647 | 2147483647 |
| 8 | 9 | 16 |

No infinite loop, no negative capacity, ordinary doubling preserved.

## Review comments

**@peter-jerry-ye** — both halves resolved:

- *"a constant instead of a random number"*: main needs no constant at all.
- *"the current runtime should be put into account"* → moonbit-docs#1155 (native packs array length into 28 bits, so ≥ 2^28 truncates and SIGABRTs). **That issue is stale for the current toolchain.** `moonbit.h:151` now reads `Moonbit_array_length(obj) ((int32_t)(header->meta))` — the full 32-bit field — and `moonbit_malloc_array` (`runtime.c:487`) stores the complete length via `Moonbit_set_meta(obj, (uint32_t)len)`. `String` and `Bytes` share that allocation path. Probed on native: 2^28, 2^28+12345 and 2^29 all report correct lengths with readable last elements. moonbit-docs#1155 is still marked OPEN and should probably be updated or closed.

**coveralls** — moot, and my earlier claim that main lacks a regression test here was wrong: `b903d495` added `buffer/buffer_growth_wbtest.mbt`, which covers the overflowing double (`:22`), the exact-requirement fallback at 2^30 (`:26`), and a panic test for a wrapped `required` (`:34`).

## One thing that is still live, separately

`moonbit_malloc_array` computes `((len + padding) << elem_size_shift)` with `int` arithmetic (`runtime.c:491`). For a byte array `padding` is 1, so `len == Int::max_value` overflows that addition before the `malloc`. So `buffer_growth_capacity` returning `Int::max_value` is arithmetically correct but not actually allocatable. That is a runtime endpoint bug, unrelated to this PR and not fixed by it — worth a separate report rather than a cap in `buffer`.

Reviewed by Codex CLI at `xhigh`, which confirmed the obsolescence, the correctness of the overflow test for every `space >= 1`, and caught both of the errors I corrected above.

bobzhang:
Closing — superseded by `b903d495`, as detailed above. Rebasing onto current `main` empties both commits.

The one live thread that came out of this triage is filed separately on the compiler tracker (`compiler/ideas#2583`): `moonbit_malloc_array` computes `((len + padding) << elem_size_shift)` in `int`, so the size computation overflows and `malloc` returns `NULL` with no check before the header write. That segfaults rather than panicking, at thresholds well below `Int::max_value` and depending on element width — verified on native, `FixedArray[Int64]` of `2^28 - 1` allocates fine while `2^28` crashes, and `FixedArray[Int]` crashes at `2^29`.

So `buffer_growth_capacity` can legitimately return a capacity the runtime cannot allocate. That is a runtime bug rather than something `buffer` should cap around, which is why it is filed there rather than fixed here.

PR #3381, opened 2026-03-30
refactor: avoid string interpolation

peter-jerry-ye:
For the `abort` message, we avoid using string interpolation, as they will introduce the `StringBuilder` APIs, and the `Show` APIs. We also use `SourceLoc::repr` directly instead of using `SourceLoc::to_string`. Overall this should reduce the size around `abort`.

This should fix the #3325 better with the following statistics:

```
=== Raw JS output ===
  abort: 811 bytes (4 functions)
  panic: 550 bytes (2 functions)

=== Minified (esbuild --bundle --minify) ===
  abort: 299 bytes
  panic: 136 bytes
  diff:  +163 bytes (119x overhead)

=== Gzipped ===
  abort: 241 bytes
  panic: 145 bytes
  diff:  +96 bytes
```

---

<a href="https://app.devin.ai/review/moonbitlang/core/pull/3381" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/builtin/array.mbt
+++ b/builtin/array.mbt
@@ -798,9 +798,7 @@ pub fn[T] Array::rev(self : Array[T]) -> Array[T] {
 pub fn[T] Array::split_at(self : Array[T], index : Int) -> (Array[T], Array[T]) {
   if index < 0 || index > self.length() {
     let len = self.length()
-    abort(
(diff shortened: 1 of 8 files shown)


hackwaly:
String concat is O(n^2) compared to StringBuilder's O(n).

PR #3402, opened 2026-03-31
refactor: make show implement one of the methods

peter-jerry-ye:
We are going to change the behavior of `Show` so that it is purely for displaying. So there will not be special handling for `String` or `Char` any more. As a result, the `output` and `to_string` would behave the same. Thus people only need to implement one of the methods.

This PR also migrates some of the builtin types' implementations to `to_string`, because use `output` would forcefully upcast `StringBuider` to `&Logger`, which will increase code in some cases.

### DIFF
--- a/builtin/traits.mbt
+++ b/builtin/traits.mbt
@@ -121,18 +121,25 @@ impl Logger with write_char(self, value) {
 
 ///|
 /// Trait for types that can be converted to `String`
+#must_implement_one(output, to_string)
 pub(open) trait Show {
   // `output` is used for composition of aggregate structure.
   // `output` writes a string representation of `self` to a logger.
   // `output` should produce a valid MoonBit-syntax representation if possible.
   // For example, `Show::output` for `String` should be quoted
-  output(Self, &Logger) -> Unit
+  output(Self, &Logger) -> Unit = _
   // `to_string` should be used by end users of `Show`,
   // for printing, interpolation, etc. only, and should not be used for composition.
   // By default `to_string` is implemented using `output` and a buffer,
   // but some types, such as `String`, may override `to_string`,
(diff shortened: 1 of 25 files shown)


peter-jerry-ye:
Note : it seems that this requires compiler fix for the display of errors

builtin/traits.mbt
```diff
@@ -123,22 +123,32 @@ impl Logger with write_char(self, value) {
...
+#must_implement_one(output, to_string)
 pub(open) trait Show {
   // `output` is used for composition of aggregate structure.
   // `output` writes a string representation of `self` to a logger.
   // `output` should produce a valid MoonBit-syntax representation if possible.
   // For example, `Show::output` for `String` should be quoted
-  output(Self, &Logger) -> Unit
+  // By default `output` is implemented using `to_string`,
+  // but some types, such as `String` and `Char`, may override `output`,
+  // for different (escaped) behavior when composed into larger structures.
```
bobzhang: this is no longer the case? we have two purely for efficiency?

peter-jerry-ye: It depends on how we are going to phase out the transition.

error/error.mbt
```diff
@@ -13,7 +13,9 @@
 // limitations under the License.
 
 ///|
-fn Error::to_string(self : Error) -> String = "%error.to_string"
+fn Error::to_string(self : Error) -> String {
+  @debug.to_string(self)
+}
```
peter-jerry-ye: We are using `debug.to_string` to display. But this can cause major breakage for errors that have display, such as the errors in the argparse. Users might find that the program no longer prints the errors well. Though no one should depend on the error message's content.

PR #3465, opened 2026-04-24
[codex] Fix dense symbol map offsets

bobzhang:
## Summary

Fix dense symbol-map finalization for non-zero lower bounds in `bytes/internal/regex_engine/symbol_map`.

Dense tables are allocated as `ub - lb + 1`, but the old implementation indexed them with absolute `Rechar` values. Calling `finalize(10, 12)` could therefore index position `10` in a length-3 table and trap. The dense table now stores `lb`, writes entries at `c - lb`, and maps with the same offset.

## Validation

- `moon fmt bytes/internal/regex_engine/symbol_map`
- `moon info bytes/internal/regex_engine/symbol_map`
- `moon test bytes/internal/regex_engine/symbol_map`
- `moon test bytes/internal/regex_engine`
- `moon check`
- `moon test`

### DIFF
--- a/bytes/internal/regex_engine/compile.mbt
+++ b/bytes/internal/regex_engine/compile.mbt
@@ -27,7 +27,7 @@ pub fn compile(profile~ : Profile, ast : Pattern) -> Regex {
   }
   let symbol_map = @symbol_map.new()
   symbolize(profile~, symbol_map~, ast)
-  let (symbol_table, symbol_repr) = symbol_map.finalize(profile.lb, profile.ub)
+  let (symbol_table, symbol_repr) = symbol_map.finalize(profile.ub)
   let ctx = @automata.Context::new()
   let tc = TranslateContext::new(ctx, symbol_table)
   let (expr, pref) = translate(tc, ast)
--- a/bytes/internal/regex_engine/symbol_map/dense_table.mbt
+++ b/bytes/internal/regex_engine/symbol_map/dense_table.mbt
@@ -13,9 +13,9 @@
 // limitations under the License.
 
 ///|
-priv struct DenseTable(@shared_types.Rechar, FixedArray[@shared_types.Rechar])
+priv struct DenseTable(FixedArray[@shared_types.Rechar])
 
 ///|
 impl Table for DenseTable with map(self, c) {
-  self.1.unsafe_get(c - self.0)
+  self.0.unsafe_get(c)
 }
--- a/bytes/internal/regex_engine/symbol_map/pkg.generated.mbti
+++ b/bytes/internal/regex_engine/symbol_map/pkg.generated.mbti
@@ -12,7 +12,7 @@ pub fn new() -> SymbolMap
 
 // Types and methods
 type SymbolMap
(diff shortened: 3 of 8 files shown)


hackwaly:
The `lb` is redundant indeed. There are only two known profiles, so assume lb is 0 is safe. If store `lb` in dense table, it will hurt performance.

bobzhang:
@hackwaly You're right that `lb` is empirically always 0 — I checked all `Profile(...)` constructions across the repo:

- `bytes/regex_profile.mbt:18` — `valid = char_range(0, 255)`
- `bytes/internal/regex_engine/regex_engine_test.mbt:16,32` — `0..0x10FFFF`, `0..0x7F`
- `string/regex_profile.mbt:22,36` — same shape
- `string/internal/regex_engine/regex_engine_test.mbt:16` and `string/internal/regex_parser/parser_test.mbt:17,36`

Eight profiles, all with `lb = 0`. And `Profile::Profile` is only re-exported through `bytes/internal/regex_engine` and `string/internal/regex_engine` — `bytes/pkg.generated.mbti` and `string/pkg.generated.mbti` don't expose it, so external users of `moonbitlang/core` can't construct a profile with non-zero `lb` either. So the invariant holds from inside the crate.

On perf: the lookup at `dense_table.mbt:19` runs once per input byte through `execute.mbt`. This PR adds one subtraction plus a field load. Real, but I'd want a benchmark before treating it as decisive — the dominant cost in regex execution is the state-machine work and memory traffic, not arithmetic on a single index.

Where I'd push back: the configuration that produced this bug is exactly *"`finalize(lb, ub)` accepts `lb` as a parameter, but ignores it and requires it to be 0."* Reviewer attention is the only thing keeping `lb=0` true; nothing in the type system enforces it. If we accept this PR's approach, we pay the per-byte subtraction. If we reject it on perf grounds, we should at minimum **remove `lb` from `finalize`'s signature** so the assumption is enforced by the API rather than by convention. Keeping a parameter you intend to ignore is the trap.

Three options I see, in increasing order of structural safety:

1. **Status quo + comment** — fastest, but the next contributor adding a profile with `lb≠0` (UTF-16 surrogate range, non-ASCII script) will reintroduce this bug silently. The compiler won't help them.
2. **This PR** — correct under any `lb`, costs one sub per lookup. Defensive but pays for a generality that, per your point, no caller actually uses.
3. **Drop `lb` from `finalize` and `DenseTable`** — fastest *and* makes the invariant unbypassable. `finalize(ub : Rechar)` and `DenseTable(FixedArray[Rechar])`, indexed absolutely. Anyone who tries to construct a profile starting above 0 has to update this API, which is the right place for that conversation to happen.

I lean toward (3): it gets your perf back and removes the footgun. If you'd prefer (1) for now I can document the invariant explicitly with an assert in `finalize` (debug-only or always-on), but I'd rather not leave a parametric API that quietly requires a specific argument value.

Happy to rework either way — what's your call?

hackwaly:
I agree 3) is best.

hackwaly:
I'll file a new PR for 3).

PR #3467, opened 2026-04-24
Fix BigInt JSON parse errors

bobzhang:
## Summary
- Validate `BigInt` JSON decimal strings before calling the aborting parser
- Return `JsonDecodeError` for malformed string payloads instead of aborting
- Add bigint package tests for empty, sign-only, and non-digit JSON strings

## Tests
- `moon fmt bigint`
- `moon info bigint`
- `moon test bigint`
- `moon test json`
- `moon check`
- `moon test`

### DIFF
--- a/bigint/bigint.mbt
+++ b/bigint/bigint.mbt
@@ -52,8 +52,36 @@ pub impl @json.FromJson for BigInt with from_json(json, path) {
       (path, "BigInt::from_json: expected number in string representation"),
     )
   }
-  // TODO: Capture parse failure
+  BigInt::validate_json_decimal(s, path)
   BigInt::from_string(s)
+}
+
+///|
+fn BigInt::validate_json_decimal(
+  s : String,
+  path : @json.JsonPath,
+) -> Unit raise @json.JsonDecodeError {
+  let len = s.length()
+  let parse_error = @json.JsonDecodeError(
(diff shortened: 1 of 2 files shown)


bobzhang:
Superseded by #3679 (`9d0169eb`, merged 2026-07-17), which removed the
`// TODO: Capture parse failure` this PR targeted and replaced it with:

```moonbit
parse_bigint(s.view()) catch {
  _ => raise JsonDecodeError(
    (path, "BigInt::from_json: invalid number in string representation"))
}
```

That handles every malformed payload in a single pass, rather than pre-validating
with a separate digit scan and then parsing again.

Verified on current `main` that the reported failures no longer reproduce — on
native, js and wasm-gc, all of `""`, `"-"`, `"12a"`, `"+5"`, `" 1"`, `"1 "`,
`"--1"`, `"０"`, `"0x10"` return `JsonDecodeError`, and `"-0"` returns `Ok(0)`.

The regression tests are already covered too: `bigint/parse_bigint_test.mbt`
pins `""` / `"-"` / `"12a"` individually, `bigint/quickcheck_deep_test.mbt` has a
cross-target rejection table containing them, and `json/from_json_test.mbt`
covers the adapter turning a parse failure into `JsonDecodeError`.

Closing — nothing left to land.

PR #3553, opened 2026-05-07
feat(lazy_list): add re-traversable lazy list package

bobzhang:
## Summary

`Iter[A]` is pull-based and single-use, so consumers cannot re-traverse it. This adds `@lazy_list.LazyList[A]`: a persistent, head-strict / tail-lazy linked list whose tails are memoized thunks. Once a cell is forced its result is cached, so multiple traversals share evaluation work.

- Type: `enum LazyList[A] { Empty | Cons(A, tail~ : Lazy[LazyList[A]]) }` (odd-style)
- Memoization is in-place mutation of the `Lazy` cell on first force; not thread-safe (consistent with the rest of core).

### API surface

- **Constructors:** `empty`, `cons`, `lazy_cons`, `singleton`, `repeat`, `iterate`, `unfold`, `from_array`, `from_list`, `from_iter`
- **Queries:** `head`, `tail`, `is_empty`, `length`, `nth` (negative index returns `None` without forcing)
- **Lazy transformations:** `map`, `filter`, `take`, `drop`, `take_while`, `drop_while`, `concat`, `flat_map`, `zip` (zip short-circuits on the empty side without forcing the other)
- **Strict consumers:** `each`, `fold`, `to_array`, `to_list`, `iter` (the `iter` bridge forces at most one cell per `next()` call)
- **Trait impls:** `Eq`, `Show`, `ToJson`, `@debug.Debug`

### Notes / pre-review

- Pre-reviewed with `codex review`; three P2 findings addressed before pushing:
  1. `iter()` no longer eagerly forces the tail when returning the head.
  2. `zip` no longer forces the longer side after the shorter side ends.
  3. `nth(-1)` returns `None` without forcing any tail.
- `Show` / `ToJson` / `length` / `to_array` / `to_list` force the entire spine — documented as a hazard for infinite streams.
- No `Hash`, no `Compare`, no `quickcheck.Arbitrary` in this first cut; easy to add later.

## Test plan

- [x] `moon check lazy_list` — clean (no warnings/errors in the new package)
- [x] `moon test lazy_list --target all` — 26 tests pass on wasm, wasm-gc, js, and native
- [x] `moon fmt`, `moon info` (mbti regenerated and committed)
- [x] Memoization regression test (counts pulls across a re-traversed `from_iter`)
- [x] Regression tests for the three Codex fixes

### DIFF
--- a/lazy_list/lazy_list.mbt
+++ b/lazy_list/lazy_list.mbt
@@ -204,11 +204,18 @@ pub fn[A] LazyList::take_while(
 }
 
 ///|
-/// Skips elements while `pred` holds, returning the rest.
+/// Skips elements while `pred` holds, returning the rest. `pred` is
+/// called eagerly during the skip; a raising `pred` propagates here.
+///
+/// (Contrast with `map` / `filter` / `take_while` / `flat_map`, whose
+/// callbacks also run inside lazy thunks. `@lazy.Lazy` thunks are
+/// non-raising by design — see `@lazy`'s rationale — so those
+/// combinators cannot accept a raising callback. `drop_while` only runs
+/// `pred` during the strict skip phase, so it can.)
 pub fn[A] LazyList::drop_while(
   self : LazyList[A],
-  pred : (A) -> Bool,
-) -> LazyList[A] {
+  pred : (A) -> Bool raise?,
+) -> LazyList[A] raise? {
   let mut cur = self
   while cur.cell is Cons(x, tail~) && pred(x) {
     cur = tail.force()
--- a/lazy_list/lazy_list.mbt
+++ b/lazy_list/lazy_list.mbt
@@ -1,0 +1,535 @@
+// Copyright 2026 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// =====================================================================
+// Constructors.
+// =====================================================================
+
+///|
+/// Creates an empty lazy list.
+///
+/// # Example
+///
+/// ```mbt check
+/// test {
+///   let xs : @lazy_list.LazyList[Int] = @lazy_list.empty()
+///   inspect(xs.is_empty(), content="true")
+/// }
+/// ```
+#as_free_fn
+pub fn[A] LazyList::empty() -> LazyList[A] {
+  Empty
+}
+
+///|
+/// Prepends an element with a strict tail. The tail is treated as already
+/// forced — useful when you have an existing `LazyList` value to put behind
+/// the head.
+///
+/// # Example
+///
+/// ```mbt check
+/// test {
+///   let xs = @lazy_list.cons(1, @lazy_list.cons(2, @lazy_list.empty()))
+///   inspect(xs, content="@lazy_list.from_array([1, 2])")
+/// }
+/// ```
+#as_free_fn
+pub fn[A] LazyList::cons(head : A, tail : LazyList[A]) -> LazyList[A] {
+  Cons(head, tail=@lazy.Lazy::ready(tail))
+}
+
+///|
+/// Prepends an element with a lazy, memoized tail. This is the primary
+/// builder for streams whose continuation should not be evaluated until
+/// it is observed. The tail thunk is invoked at most once.
+///
+/// # Example
+///
+/// ```mbt check
+/// test {
+///   fn nat_from(n : Int) -> @lazy_list.LazyList[Int] {
+///     @lazy_list.lazy_cons(n, () => nat_from(n + 1))
+///   }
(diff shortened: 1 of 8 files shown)


lazy_list/types.mbt
```diff
@@ -0,0 +1,45 @@
...
+///|
+/// A persistent linked list with a memoized lazy tail. The head is always
+/// strict; the tail is computed on first access and cached, so repeated
+/// traversals share evaluation work.
+///
+/// Unlike `Iter[A]`, which is a single-use pull-based iterator, a
+/// `LazyList[A]` can be re-traversed any number of times.
+pub enum LazyList[A] {
+  Empty
+  Cons(A, tail~ : Lazy[LazyList[A]])
```
hackwaly: `mut tail~`? This can avoid an indirection.

lazy_list/lazy_list.mbt
```diff
@@ -0,0 +1,558 @@
...
+  match self {
+    Empty => Empty
+    Cons(x, tail~) =>
+      if pred(x) {
+        LazyList::lazy_cons(x, () => tail.force().take_while(pred))
+      } else {
+        Empty
+      }
+  }
+}
```
CAIMEOX: same as `take`: no tail form

lazy_list/lazy_list.mbt
```diff
@@ -0,0 +1,558 @@
...
+
+// =====================================================================
+// Strict consumers.
+// =====================================================================
+
+///|
+/// Applies `f` to every element in order. Forces the entire spine.
+pub fn[A] LazyList::each(self : LazyList[A], f : (A) -> Unit) -> Unit {
```
CAIMEOX: To align with other API, we can use `f : (A) -> Unit raise?` here, same in the `fold` function

lazy_list/types.mbt
```diff
@@ -0,0 +1,45 @@
...
+///|
+/// A memoized thunk. The first call to `force` runs the thunk and replaces
+/// the internal state with its result; subsequent calls return the cached
+/// value. Treated as opaque outside this package — construct via the
+/// `lazy_cons` / `cons` smart constructors.
+///
+/// Not thread-safe.
+struct Lazy[A] {
+  mut state : LazyState[A]
+}
```
CAIMEOX: Lazy should be opaque?

lazy_list/lazy_list.mbt
```diff
@@ -0,0 +1,558 @@
...
+// =====================================================================
+
+///|
+/// Lazily applies `f` to every element. `f` runs the first time a given
+/// cell of the result is forced, and the mapped value is then memoized.
+pub fn[A, B] LazyList::map(self : LazyList[A], f : (A) -> B) -> LazyList[B] {
+  match self {
+    Empty => Empty
+    Cons(x, tail~) => LazyList::lazy_cons(f(x), () => tail.force().map(f))
+  }
```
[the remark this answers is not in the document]
CAIMEOX: As the copilot pointed out, there is a discrepancy here with the documentation, but since we are considering "head strict, tail lazy," this behavior is correct, and the documentation should be updated.

bobzhang:
## Public API shrunk; review comments addressed

Pushed `525282dd`. Net change to `pkg.generated.mbti`: down from ~30 public items to 20 + opaque type + `Debug`.

**Review comments addressed:**

- @CAIMEOX ("Lazy should be opaque?"): `LazyList` is now an opaque struct wrapping a private `LazyListCell` enum. `@lazy.Lazy` no longer leaks through the public type; users go through `head` / `tail` / `iter` / etc.
- @CAIMEOX / @Copilot (`map` docstring): now correctly describes the head-strict / tail-lazy semantics.
- @CAIMEOX (`each` / `fold` `raise?`): already addressed during a previous rebase.
- @Copilot (`to_list` via `to_array` is wasteful): rewritten to walk-and-prepend then `rev` — single pass, no intermediate array.
- @Copilot (`ToJson` allocates twice): removed entirely (see below).
- @Copilot (test comment "length+1 pulls"): comment corrected — the counter only increments on `Some(x)`, so it equals `length`.
- @hackwaly (`mut tail~` to avoid indirection): N/A since memoization moved into `@lazy.Lazy` (its own package) — the indirection lives there.

**Cuts** (with rationale; all reversible if users ask):

- **`Eq`, `Show`, `ToJson`, `to_json`** — each silently forces the entire spine. `Show` and `ToJson` hang on infinite lists; `Eq` can diverge. `Debug` (which uses `@lazy.peek` to walk only the *forced* prefix) is the safe replacement and stays.
- **`singleton`, `repeat`, `iterate`, `unfold`** — all expressible in a few lines with `lazy_cons`. `Iter` also has the relevant combinators.
- **`from_array`, `from_list`** — use `arr.iter() \|> from_iter` or `list.iter() \|> from_iter`. `from_iter` is the unique value-add (memoizes the source pulls so re-traversal is free).
- **`length`, `nth`** — use `xs.iter().count()` / `xs.iter().nth(n)`.
- **`cons`** demoted to a private helper.

**Kept:**

- Constructors: `empty`, `lazy_cons`, `from_iter`
- Queries: `is_empty`, `head`, `tail`
- Transforms: `map`, `filter`, `take`, `take_while`, `drop`, `drop_while`, `concat`, `flat_map`, `zip`
- Strict consumers: `each`, `fold`, `to_array`, `to_list`
- Bridge: `iter`
- Trait: `Debug`

Transforms are kept (rather than always going through `iter() + from_iter`) because keeping them on `LazyList` directly preserves the re-traversable memoized cell structure with one less layer of indirection. Same big-O, fewer allocations.

All 16 tests pass on wasm / wasm-gc / js / native locally.

PR #3556, opened 2026-05-08
Refactor array traversals to use for-in loops

Yu-zh:
## Summary
- replace straightforward indexed traversals with for-in loops
- keep behavior and public API unchanged

## Validation
- moon check --target all --target-dir /tmp/core-forin-post-rebase-check
- moon test --target all --target-dir /tmp/core-forin-review-test
- moon fmt
- moon info


builtin/array_sort_impl.mbt
bobzhang, line 1: blit

bobzhang (approved):

PR #3575, opened 2026-05-12
refactor: make String and Char Show output raw

Yoorkin:
This PR will change the behaviors for `impl Show for String` and `impl Show for Char`.

## Summary

- Remove the specialized `Show::output` implementations for `String` and `Char`, so they use the default raw `to_string()` output.
- Update affected `Show`/`to_string` tests and snapshots to match the new raw string/char output semantics.
- Keep deprecated composite `Show` coverage where the tests intentionally verify `to_string()` behavior.

## Tests

- `moon test`

### DIFF
--- a/builtin/array_test.mbt
+++ b/builtin/array_test.mbt
@@ -254,9 +254,9 @@ test "array_rev_eachi" {
   inspect(
     buf,
     content=(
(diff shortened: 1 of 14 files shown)


Yoorkin:
@codex review

PR #3588, opened 2026-05-14
refactor(builtin): share BytesView/StringView cross-type equality helpers

bobzhang:
## Summary

Lift the cross-type view-vs-owned equality helpers out of `Map`'s internals and into `builtin` so the same code can serve every keyed container.

- **New primitive** `Bytes::unsafe_range_equal(self, self_off, other, other_off, len) -> Bool` in `bytes_unsafe.mbt` — package-private, `#borrow(self, other)`, documented as a future `memcmp`-style intrinsic candidate.
- **`Eq for BytesView`** now delegates to the primitive (length check + one call).
- **New** `pub BytesView::equal_to_bytes(self, other : Bytes) -> Bool` — built on the primitive; the cross-type counterpart of `==`.
- **New** `pub StringView::equal_to_string(self, other : String) -> Bool` — preserves the `physical_equal` fast path for views that span their entire backing string.
- `linked_hash_map.mbt` drops its private duplicates of these helpers.

Direct unit tests in `bytesview_test.mbt` and `stringview_test.mbt` exercise length-mismatch, content-mismatch, empty, full-view, and (for strings) multi-byte / physical-equal paths.

## Motivation

`Map` already has these helpers (private, baked into `linked_hash_map.mbt`). #3586 is about to add the same helpers for `HashMap` — pulling them out into `builtin` first removes the duplication and gives downstream callers (e.g. user-defined containers, or `view == owned` comparisons in user code) a zero-alloc equality primitive without forcing them to wrap one side in a fresh view.

## Test plan
- [x] `moon check --target all` clean
- [x] `moon test --target all` green: 6480/6480 wasm, 6480/6480 wasm-gc, 6439/6439 js, 6391/6391 native
- [x] New direct tests for `BytesView::equal_to_bytes` and `StringView::equal_to_string`

### DIFF
--- a/hashmap/hashmap.mbt
+++ b/hashmap/hashmap.mbt
@@ -218,6 +218,78 @@ pub fn[K : Hash + Eq, V] HashMap::get(self : HashMap[K, V], key : K) -> V? {
 }
 
 ///|
+/// Retrieves the value associated with a `BytesView` key in a hash map with
+/// `Bytes` keys.
+///
+/// This allows efficient lookups using a `BytesView` (e.g. a sub-slice of a
+/// larger byte buffer) without allocating a fresh `Bytes`.
+///
+/// Returns `Some(value)` if a matching key exists in the map, `None` otherwise.
+///
+/// Example:
+///
+/// ```mbt check
+/// test {
+///   let map = @hashmap.from_array([(b"hello", 1), (b"world", 2)])
+///   let bytes = b"prefix_hello_suffix"
+///   let view = bytes[7:12] // view of "hello"
+///   @debug.debug_inspect(map.get_from_bytes(view), content="Some(1)")
+/// }
+/// ```
+pub fn[V] HashMap::get_from_bytes(
+  self : HashMap[Bytes, V],
+  key : BytesView,
+) -> V? {
+  let hash = key.hash()
+  for i = 0, idx = hash & self.capacity_mask {
+    guard self.entries[idx] is Some(entry) else { break None }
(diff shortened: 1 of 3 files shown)


builtin/bytes_unsafe.mbt
```diff
@@ -543,6 +543,40 @@ pub fn BytesView::unsafe_read_uint16_be(
...
+/// - `len`: number of bytes to compare
+///
+/// This is the shared primitive backing view/owned equality comparisons
+/// (e.g. `BytesView == BytesView`, `BytesView::equal_to_bytes`). It is a
+/// candidate for an `memcmp`-style intrinsic in the future.
+#borrow(self, other)
+fn Bytes::unsafe_range_equal(
+  self : Bytes,
+  self_off : Int,
+  other : Bytes,
```
bobzhang, line 568: to be improved later cc @Yu-zh

builtin/stringview.mbt
```diff
@@ -291,6 +291,45 @@ pub impl Eq for StringView with equal(self, other) {
...
+///     content="true",
+///   )
+///   inspect(
+///     s.view(start_offset=4, end_offset=9).equal_to_string("world"),
+///     content="false",
+///   )
+/// }
+/// ```
+pub fn StringView::equal_to_string(self : StringView, other : String) -> Bool {
+  let len = self.length()
```
bobzhang, line 320: optimize string equality as an intrinsic cc @Yu-zh

PR #3619, opened 2026-05-22
Optimize ChaCha8 random source next

Yu-zh:
## Summary
- add a non-optional ChaCha8::next_uint64 path that refills internally
- use it for the ChaCha8 Source implementation to avoid per-value Option matching in Rand hot paths
- update generated interfaces for the touched random packages

## Performance
Before this change, repeated local runs of the existing benchmark were around 38.3-38.5 ms:

```
moon bench -p moonbitlang/core/random -f random_test.mbt
bench random: ~38.3-38.5 ms
```

With this change, repeated runs are around 31.1-31.4 ms:

```
moon bench -p moonbitlang/core/random -f random_test.mbt
bench random: 31.16 ms ± 578.23 us
```

## Validation
- `moon fmt --check random/random.mbt random/internal/random_source/random_source_chacha.mbt`
- `moon check random random/internal/random_source`
- `moon test random random/internal/random_source`
- `moon check`
- `moon bench -p moonbitlang/core/random -f random_test.mbt`

### DIFF
--- a/random/internal/random_source/random_source_chacha.mbt
+++ b/random/internal/random_source/random_source_chacha.mbt
@@ -53,11 +53,11 @@ const SEED_CHUNK_NUM = 4
 ///|
 /// Return the next random `UInt64`, refilling the internal buffer when needed.
 pub fn ChaCha8::next_uint64(self : ChaCha8) -> UInt64 {
-  let i = self.i
+  let mut i = self.i
   if i >= self.n {
     self.refill()
+    i = 0
   }
-  let i = self.i
   self.i = i + 1
   let index = i.reinterpret_as_int() & (BUFFER_CHUNK_NUM - 1)
   let lo = self.buffer[index * 2].to_uint64()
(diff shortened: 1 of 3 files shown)


bobzhang:
Refactoring note for head `100bb6c0`: the old refill loop has been folded into `ChaCha8::next_uint64()`, and `ChaCha8::next() -> UInt64?` was removed from the internal `random/internal/random_source` API.

This is stream-preserving: when the buffer has data, both paths read the same buffered `i` entry, increment `i`, and combine the same two `UInt` words into a `UInt64`; when `i >= n`, the old loop called `refill()` and retried, while `next_uint64()` calls `refill()` before reading. The output test now drives `next_uint64()` directly against the existing 372-value expected sequence.

I also checked references after the change: there are no remaining `ChaCha8::next` call sites, and `next_uint64()` is used by the output test plus the `Rand` `Source` implementation. Local validation passed: `moon info && moon fmt`, `moon check`, and `moon test` (`6500` tests passed).

Caveat: this intentionally changes the internal package interface from `ChaCha8::next(Self) -> UInt64?` to `ChaCha8::next_uint64(Self) -> UInt64`; since this package is internal and all in-tree call sites are migrated, it should not affect public `@random` behavior.

random/internal/random_source/random_source_chacha.mbt
```diff
@@ -51,17 +51,18 @@ const BUFFER_CHUNK_NUM = 32
...
-/// Function `next`.
-pub fn ChaCha8::next(self : ChaCha8) -> UInt64? {
+/// Return the next random `UInt64`, refilling the internal buffer when needed.
+pub fn ChaCha8::next_uint64(self : ChaCha8) -> UInt64 {
   let i = self.i
   if i >= self.n {
-    return None
+    self.refill()
   }
+  let i = self.i
```
bobzhang: This second `self.i` read is correct because `refill()` mutates `self.i`, but the shadowing makes it easy to read as accidental duplication. Consider making that explicit with a mutable local, e.g. `let mut i = self.i; if i >= self.n { self.refill(); i = self.i }`, then use `i` below.

bobzhang (commented):
Left one non-blocking readability suggestion.

PR #3634, opened 2026-05-26
perf(builtin): override Hash::hash for primitive types to skip Hasher alloc

mizchi:
## Summary

The default `Hash::hash` builds a fresh `Hasher` ({ mut acc : UInt }),
combines, finalizes — **one heap allocation per call**. For primitive
types the math is small enough to inline, so the allocation is pure
overhead.

This PR overrides `hash` for **eight primitive types** that already
have `hash_combine` impls:

| type | strategy |
|---|---|
| `Int`, `UInt` | inline single-word combine + avalanche |
| `Int64`, `UInt64` | inline two-half combine + avalanche |
| `String`, `StringView` | inline per-codepoint combine loop + avalanche |
| `Char` | forward to `UInt.hash` via `to_uint()` |
| `Bool` | forward to `UInt.hash` via 0 / 1 |

The default `hash_combine` impls stay unchanged (they're still needed
for composite hashing where `hash` of a sub-value can't be used
directly).

## Numbers

Measured on a native-target alloc profiler over mizchi/pprof-mbt's
bench suite (`--sample-rate 100`):

| bench | before | after | Δ |
|---|---|---|---|
| `hashset_ops` (Int keys, 10 k + 20 k probes × 50) | 2 665 300 allocs | 1 165 300 | **−56 %** |
| └ `contains` site | 1 M / 3.81 MB | gone from top sites | — |
| `hashmap_update` (Int → Int) | 620 100 allocs | 10 100 | **−98 %** |
| `hashmap_ops` (Int keys) | `Hash::hash` 100 % of probes | gone from top | — |
| `hashmap_string` (String keys, 10 k × 30) | 845 700 allocs | 395 700 | **−53 %** |
| └ `Hash::hash` site | 450 k / 1.72 MB | gone from top sites | — |

## Why this is safe

The override produces the same hash value as the default path on every
backend. The default is:

```moonbit
impl Hash with fn hash(self) {
  let h = Hasher::new()
  h.combine(self)
  h.finalize()
}
```

For `Int`, expanding `Hasher::new` → `Hasher.combine_int` → `finalize`
on `Hasher { mut acc : UInt }` gives:

```text
acc  = seed + GPRIME5            // Hasher::new()
acc += 4                         // combine_uint length
acc  = rotl(acc + v * GPRIME3, 17) * GPRIME4
acc ^= acc >> 15; acc *= GPRIMES2;
acc ^= acc >> 13; acc *= GPRIME3;
acc ^= acc >> 16
return acc.reinterpret_as_int()
```

— which is exactly what the inline override computes. `UInt` skips the
`reinterpret_as_uint` hop. `UInt64` combines low / high 32-bit halves
the same way `Hasher::combine_int64` does. `Int64` forwards to
`UInt64`. `String`/`StringView` keep the same per-codepoint loop body
as `Hasher::combine_string`, just with the accumulator in a local
`UInt`. `Char` and `Bool` forward to `UInt.hash` via their existing
conversion (matching `combine_char` / `combine_bool`).

The global `seed` (0 on native / wasm / wasm-gc, JS-randomized on the
js target) is shared with `Hasher::new()` so the output matches per-
backend.

## Test plan

- [x] `moon test --target native -p builtin -p hashset -p hashmap` — 3175 / 3175 pass
- [x] `moon test --target wasm-gc -p builtin -p hashset -p hashmap` — 3217 / 3217 pass
- [x] `moon fmt --check` clean

Independent of #3632 (json lex_skip_whitespace) and #3633 (json
lex_number tuple). Same workflow (native-target alloc profile of a
real bench, then attack the top site), different file.

### DIFF
--- a/builtin/bool.mbt
+++ b/builtin/bool.mbt
@@ -121,6 +121,14 @@ pub impl Hash for Bool with fn hash_combine(self, hasher) {
 }
 
 ///|
+// Forward to the inline `UInt` hash to skip the per-call
+// `Hasher::new()` allocation. See `Hash for Int with fn hash` in
+// `hasher.mbt` for the rationale.
+pub impl Hash for Bool with fn hash(self : Bool) -> Int {
+  Hash::hash(if self { 1U } else { 0U })
+}
+
+///|
 /// Converts a boolean value to an unsigned 16-bit integer.
 ///
 /// Parameters:
--- a/builtin/char.mbt
+++ b/builtin/char.mbt
@@ -45,6 +45,14 @@ pub impl Hash for Char with fn hash_combine(self, hasher) -> Unit {
 }
 
 ///|
+// Forward to the inline `UInt` hash to skip the per-call
+// `Hasher::new()` allocation. See `Hash for Int with fn hash` in
+// `hasher.mbt` for the rationale.
+pub impl Hash for Char with fn hash(self : Char) -> Int {
+  Hash::hash(self.to_uint())
+}
+
+///|
 /// Checks if the value is within the ASCII range.
 pub fn Char::is_ascii(self : Self) -> Bool {
   self is ('\u{00}'..='\u{7F}')
--- a/builtin/hasher.mbt
+++ b/builtin/hasher.mbt
@@ -498,6 +498,29 @@ pub impl Hash for String with fn hash_combine(self, hasher) {
 }
 
 ///|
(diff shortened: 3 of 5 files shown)


bobzhang:
lgtm, can you add some tests to verify the hash value is the same using hash_combine vs using the hash

PR #3648, opened 2026-06-01
perf(list): add #inline to List::new

bobzhang:
Adds an `#inline` attribute to `List::new`, a trivial constructor that just returns `Empty`. Inlining lets the compiler elide the call at use sites.


bobzhang:
cc @Yu-zh

PR #3679, opened 2026-06-10
feat(string): add parse_bigint (Fixes #3566)

### LINKED ISSUE #3566: [Consistency Review] BigInt::from_string should raise or add a @string.parse_bigint
**Current behaviors**

There's only BigInt::from_string, it will panic if input is malformed. It's impossible to catch if input is malformed.

**Expected behaviors**

We should have a @string.parse_bigint or make BigInt::from_string raises.

**Justification**

rainhuang0220:
## Summary
- Add `@string.parse_bigint` with raising error handling
- Keep `BigInt::from_string` panic behavior for compatibility
- Update `BigInt::from_json` to use `parse_bigint`
Fixes #3566
## Test plan
- [x] `moon test -p bigint -p string`
- [x] `moon check bigint string --deny-warn`


rainhuang0220:
All checks passed, please review and merge this PR when you have time, thanks. @bobzhang

bigint/bigint_nonjs.mbt
```diff
@@ -1240,54 +1240,49 @@ fn BigInt::from_string_radix_pow2(
...
-pub fn BigInt::from_string(input : String, radix? : Int = 10) -> BigInt {
-  if radix < 2 || radix > 36 {
-    abort("radix must be between 2 and 36")
+pub fn parse_bigint(str : StringView, base? : Int = 10) -> BigInt raise {
+  if base < 2 || base > 36 {
+    base_err()
   }
-  if radix != 10 {
-    return BigInt::from_string_radix(input, radix)
+  let input = str.to_owned()
```
Yu-zh: This creates a copy of the string, which seems unnecessary here.

bigint/bigint_js.mbt
```diff
@@ -19,54 +19,30 @@ type BigInt
...
-  if radix != 10 {
-    return BigInt::from_string_radix(str, radix)
+/// Returns an error if the input is malformed or the base is invalid.
+pub fn parse_bigint(str : StringView, base? : Int = 10) -> BigInt raise {
+  if base < 2 || base > 36 {
+    base_err()
   }
-  if str == "-" {
-    abort("invalid character")
+  let input = str.to_owned()
```
Yu-zh: same here

bigint/pkg.generated.mbti
```diff
@@ -9,6 +9,7 @@ import {
 }
 
 // Values
+pub fn parse_bigint(StringView, base? : Int) -> BigInt raise
```
Yu-zh: I don't think we should add a new API. Instead, we could deprecate `BigInt::from_string` and encourage users to use `@strconv.parse_bigint` instead.

Yu-zh (changes requested):
Thanks for your contribution and I apologize for a late response.

Please also take a look at the review comments from copilot and resolve them if possible.

rainhuang0220:
Thanks for the review @Yu-zh 

I've addressed your comments:

Removed unnecessary `str.to_owned()` — `parse_bigint` and the internal parsers (`from_string_dec`, `from_string_radix`, etc.) now work directly on `StringView`.

Moved the user-facing API out of `@bigint` — `@bigint.parse_bigint` is now `#internal` and no longer appears in `bigint/pkg.generated.mbti`. The public entry points are `@string.parse_bigint` and a deprecated `@strconv.parse_bigint` re-export (following the same pattern as `parse_int`).

`BigInt::from_string` deprecation — I initially deprecated it as suggested, but `moon check --deny-warn` failed with ~270 warnings across the repo (tests, docs, examples). I've kept `from_string` non-deprecated for now so CI passes; happy to do the deprecation in a follow-up PR that migrates call sites.

Also merged latest `main` to resolve the conflict in `bigint/bigint_js.mbt`.
Deprecating `from_string` in this PR alone triggers deprecation warnings in `bigint_test.mbt`, `panic_test.mbt`, `json/`, `math/`, etc. under `--deny-warn`. I'd prefer to land the `parse_bigint` API first and handle deprecation + migration separately

Could you take another look when you have time? Thanks!

bigint/bigint_js.mbt
```diff
@@ -19,54 +19,31 @@ type BigInt
...
-  if radix != 10 {
-    return BigInt::from_string_radix(str, radix)
-  }
-  if str == "-" {
-    abort("invalid character")
+  if str.length() == 0 {
+    syntax_err()
   }
-  BigInt::js_from_string(str)
+  BigInt::from_string_radix(str, base)
```
hackwaly: The new implementation has performance regression. The general path is 138x slower than native BigInt(...) path.

We should use native path where possible.

bigint/parse_bigint_test.mbt
```diff
@@ -0,0 +1,69 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+///|
+test "parse_bigint valid" {
+  inspect(@bigint.parse_bigint("12345"), content="12345")
+  inspect(@bigint.parse_bigint("-12345"), content="-12345")
+  inspect(@bigint.parse_bigint("ff", base=16), content="255")
+  inspect(@bigint.parse_bigint("-0"), content="0")
+}
```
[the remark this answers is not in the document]
hackwaly: This should be considered seriously.

Yu-zh (approved):

Yu-zh:
@rainhuang0220 Sorry again for delayed response. We have reverted the unnecessary changes to the strconv package (my bad for suggesting it) and fixed a regression problem. Now this PR is good to merge.

PR #3689, opened 2026-06-16
Fix rand seed

cybershang:
# Fix `Rand::new()` hard-coded default seed

`Rand::new()` currently uses a hard-coded ChaCha8 seed, producing the same sequence everywhere. This PR seeds it from OS entropy, while keeping `Rand::chacha8(seed~)` for deterministic use.

### `env/` — cross-platform entropy API `@env.rand(n)`

| File | Backend |
|---|---|
| `env/env.mbt` | Public API: `pub fn rand(Int) -> Bytes?` |
| `env/env_js.mbt` | JS: `crypto.getRandomValues` |
| `env/env_wasm.mbt` | WASM/WASM-GC: returns `None` (no WASI raw pointer FFI yet) |
| `env/env_native.mbt` | **See Windows note below** |

### `random/` — use `@env.rand` in `Rand::new()`

- Prefer OS entropy via `@env.rand(32)`, fall back to `@env.now()` + counter.
- Tests updated to use explicit seeds via `Rand::chacha8(seed=fixed_test_seed)`.
- Added regression test: two fresh `Rand::new()` instances differ.

## Windows CI note

The initial native backend used MoonBit's `#cfg(target_os="windows")` to select between `rand_s` and `getentropy`. **MoonBit's `#cfg` doesn't support `target_os`** — only `target` (`native`, `js`, ...). So the POSIX path was compiled everywhere, and MSVC can't find `getentropy`.

Fix: add a single FFI function `moonbit_rt_get_random` to the MoonBit C runtime (`runtime.c`) that uses C-level `#ifdef _WIN32` to pick the right API:

```c
#ifdef _WIN32
  rand_s(...)
#else
  getentropy(...)
#endif
```

The patch (`patches/add-moonbit_rt_get_random-to-runtime.patch`) needs to be applied by the MoonBit team and shipped in a toolchain release. Until then, Windows falls back to the time+counter seed (still non-deterministic).

## Testing

```
native:  6460 passed
js:      6505 passed
wasm:    6549 passed
wasm-gc: 6549 passed
```


cybershang:
@bobzhang  Hi, could you please review this PR when you have a moment? Thanks!

env/env_js.mbt
```diff
@@ -119,3 +119,22 @@ extern "js" fn unset_env_var_ffi(key : String) -> Unit =
...
+    return Some(b"")
+  }
+  get_random_values_js(n)
+}
+
+///|
+extern "js" fn get_random_values_js(n : Int) -> Bytes? =
+  #| function(n) {
+  #|   if (typeof crypto === "undefined" || typeof crypto.getRandomValues !== "function") {
+  #|     return null;
```
peter-jerry-ye: @hackwaly Is the ABI stable here?

patches/add-moonbit_rt_get_random-to-runtime.patch
peter-jerry-ye, line 1: We need to modify this upstream @Guest0x0

env/env_js.mbt
hackwaly: Yes, but there's user reports that API may be defined but not usable on cloudflare worker at some situations.

patches/add-moonbit_rt_get_random-to-runtime.patch
peter-jerry-ye: The patch is applied upstream with similar concept but a rewrite. We'll need to wait for next release.

peter-jerry-ye:
I think we'll change the FFI shape into `fn rand(to_be_filled : FixedArray[Byte]) -> Bool`.

env/env_js.mbt
peter-jerry-ye: Adjusted the shape. The current code has proper try catch

peter-jerry-ye:
Follow-up summary for the pushed update:

- Kept the public API unchanged; the internal `rand_internal` path now fills a caller-provided `FixedArray[Byte]` and returns `Bool` for success.
- Native uses the current runtime ABI, `moonbit_rt_get_random(buf) == 0`.
- JS fills the caller buffer with `globalThis.crypto.getRandomValues`, chunked in 64 KiB slices, and returns `false` when crypto is missing or throws.
- wasm and wasm-gc always return `false`, so `@env.rand` reports `None` there. `Rand::new()` now falls back to the fixed ChaCha seed instead of using a timer fallback (primarily because this would pull in timer FFI dependency that might be unexpected).
- Removed the runtime patch file and updated `random/README.mbt.md` to document the unsupported-target fallback.

Wrangler check:

- A Worker probe imported the compiled MoonBit JS at module initialization and called `@env.rand(0)` plus `Rand::new()`.
- Under `wrangler dev`, init-time randomness reported unavailable (`randAvailable: false`), matching Cloudflare behavior where `crypto.getRandomValues` is not usable during initialization.
- `Rand::new()` still completed through the deterministic fixed-seed fallback; request-time `crypto.getRandomValues` was available.

This validates that the JS FFI catches the Cloudflare init-time failure and exposes it as `false` instead of crashing module initialization.

PR #3691, opened 2026-06-17
fix(math): use unsigned compare for cosh near-overflow boundary

bobzhang:
## Summary

Follow-up to #3690. While reviewing that fix, an independent review (codex CLI) and I found a second, pre-existing bug in the same `cosh(x : Double)` function (non-JS backend), in the near-overflow boundary branch.

The branch that handles `|x|` between `log(maxdouble)` and the overflow threshold compared the **low word** of `|x|` as a **signed** `Int`:

```moonbit
let lx = get_low_word(x).reinterpret_as_int()
if ix < 0x408633ce || (ix == 0x408633ce && lx <= 0x8fb9f87d) { ... }
```

The threshold low word `0x8fb9f87d` is **negative** when interpreted as a signed `Int` (`-1883637635`). So for any `x` whose high word is exactly `0x408633ce` and whose low word is in `[0x00000000, 0x7fffffff]` (positive when signed), the test `lx <= 0x8fb9f87d` is false and the code falls through to `Infinity` — even though those values are below the true overflow threshold and have a **finite** `cosh`.

For example, `cosh(710.4757080078125)` (bit pattern `0x408633ce40000000`) returned `Infinity` instead of the correct `~1.7974e308`.

## Fix

Replace the split high/low-word comparison with a single **unsigned 64-bit** comparison of `|x|` against the threshold, mirroring the analogous `sinh` branch in the same file:

```moonbit
if x.abs().reinterpret_as_uint64() <= 0x408633ce8fb9f87d { ... }
```

`ix` already has its sign bit cleared, so `x.abs().reinterpret_as_uint64()` reproduces the intended `(ix << 32) | low_word` magnitude pattern; the inclusive `<=` matches the fdlibm/FreeBSD `e_cosh.c` reference. A regression test exercising the affected high-word bucket (for both signs) is added.

The JS backend (`Math.cosh` delegation) and the `float` variant are unaffected.

## Validation

- [x] `moon fmt`
- [x] `moon check math --deny-warn`
- [x] `moon test math --target all` (139/139 passed on wasm, wasm-gc, js, native)
- [x] `moon info` (no `.mbti` signature change)
- [x] Independently reviewed with codex CLI (no functional issues; one comment-wording nit addressed)

### DIFF
--- a/math/hyperbolic_double_nonjs.mbt
+++ b/math/hyperbolic_double_nonjs.mbt
@@ -141,8 +141,13 @@ pub fn cosh(x : Double) -> Double {
   if ix < 0x40862E42 {
     return 0.5 * exp(x.abs())
   }
-  let lx = get_low_word(x).reinterpret_as_int()
-  if ix < 0x408633ce || (ix == 0x408633ce && lx <= 0x8fb9f87d) {
+  // |x| in [log(maxdouble), overflow threshold]: compute via the scaled form
+  // 0.5 * exp(0.5*|x|) * exp(0.5*|x|) to avoid overflow in the intermediate exp.
+  // The threshold bit pattern 0x408633ce8fb9f87d must be compared as an unsigned
+  // 64-bit integer (mirroring `sinh`); a split high/low-word comparison with a
+  // signed low word misclassifies low words in [0, 0x7fffffff] and wrongly
+  // overflows to infinity for finite values just below the threshold.
+  if x.abs().reinterpret_as_uint64() <= 0x408633ce8fb9f87d {
     let w = exp(0.5 * x.abs())
     let t = 0.5 * w
     return t * w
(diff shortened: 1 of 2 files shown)


bobzhang:
cc @connect0459

connect0459:
@bobzhang 

Thanks for the cc! The fix looks good to me. The Copilot comments are worth addressing, but I'll leave that to your discretion.

bobzhang:
Thanks for the review! Addressed both Copilot comments in 73dfcfb3:

- **cosh recomputed `x.abs()`**: now computed once into a local `abs_x` at the top of `cosh` and reused in every branch, matching `sinh`.
- **Test calling `cosh` repeatedly / no NaN check on `-x`**: results are now stored and both `!is_inf()` and `!is_nan()` are asserted for each sign. The boundary value is also built from its exact bit pattern `0x408633ce40000000` (with a round-trip `assert_eq`) so the regression can't be silently weakened by decimal-literal rounding.

`moon fmt` / `moon check --deny-warn` / `moon test math --target all` (139/139 on wasm, wasm-gc, js, native) / `moon info` (no `.mbti` change) all pass.

PR #3692, opened 2026-06-17
fix(math): make sinh overflow-threshold boundary inclusive

bobzhang:
## Summary

Follow-up to #3690 / #3691. While fixing the `cosh` near-overflow boundary, codex CLI and I noticed the analogous `sinh` branch in the same file (`math/hyperbolic_double_nonjs.mbt`, non-JS backend) had a related off-by-one boundary bug.

The `sinh` near-overflow branch used a **strict** `<` against the overflow threshold `0x408633ce8fb9f87d` (`|x| ≈ 710.4758600739439):

```moonbit
if abs_x.reinterpret_as_uint64() < 0x408633ce8fb9f87d { ... scaled exp ... }
x * shuge   // overflow
```

The fdlibm `e_sinh.c` reference treats this boundary as **inclusive** (`lx <= 0x8fb9f87d`). At exactly the threshold value, `sinh` is still finite (`~1.7977e308`, just below `DBL_MAX`), so it must take the scaled-exp branch rather than fall through to `x * shuge` and overflow.

As a result, `sinh(710.4758600739439)` returned `±Infinity`, even though `cosh` at the same point correctly returns the finite `1.7976931348621744e+308`.

## Fix

Change `<` to `<=`, matching the reference and the analogous (now-consistent) `cosh` branch:

```moonbit
if abs_x.reinterpret_as_uint64() <= 0x408633ce8fb9f87d { ... }
```

A regression test exercises the exact threshold (both signs) plus a larger value that must still overflow. The scaled computation has finite intermediates at the threshold (`w ≈ 1.896e154`), and the next representable value above the threshold (`0x408633ce8fb9f87e`) correctly overflows. The JS backend (`Math.sinh` delegation) and the `float` variant are unaffected.

## Validation

- [x] `moon fmt`
- [x] `moon check math --deny-warn`
- [x] `moon test math --target all` (139/139 passed on wasm, wasm-gc, js, native)
- [x] `moon info` (no `.mbti` signature change)
- [x] Independently reviewed with codex CLI: *"None. The `<=` change is correct."* — verified the threshold bit pattern, finite value, reference, and cross-target behavior.

### DIFF
--- a/math/hyperbolic_double_nonjs.mbt
+++ b/math/hyperbolic_double_nonjs.mbt
@@ -81,7 +81,11 @@ pub fn sinh(x : Double) -> Double {
   if ix < 0x40862E42 {
     return h * exp(abs_x)
   }
-  if abs_x.reinterpret_as_uint64() < 0x408633ce8fb9f87d {
+  // |x| in [log(maxdouble), overflow threshold]: the boundary is inclusive, so
+  // the exact threshold value 0x408633ce8fb9f87d (~710.4758600739439) still has a
+  // finite sinh and must take the scaled branch. Matches the fdlibm `e_sinh.c`
+  // reference and the analogous `cosh` branch.
+  if abs_x.reinterpret_as_uint64() <= 0x408633ce8fb9f87d {
     let w = exp(0.5 * abs_x)
     let t = h * w
     return t * w
--- a/math/hyperbolic_test.mbt
+++ b/math/hyperbolic_test.mbt
@@ -196,3 +196,20 @@ test "cosh near-overflow low-word boundary" {
   assert_true(!pos.is_inf() && !pos.is_nan() && pos > 1.79e308)
   assert_true(!neg.is_inf() && !neg.is_nan() && neg > 1.79e308)
 }
+
+///|
+test "sinh overflow-threshold boundary" {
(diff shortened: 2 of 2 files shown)


bobzhang:
Addressed the Copilot review comment in 6b73ef6e: the regression test no longer depends on decimal-literal rounding — the threshold is now built from its exact bit pattern `0x408633ce8fb9f87d.reinterpret_as_double()` with a round-trip `assert_eq`, and the `sinh` results are stored with `!is_inf()` + `!is_nan()` checked for each sign.

`moon fmt` / `moon check --deny-warn` / `moon test math --target all` (139/139 on all targets) pass.

bobzhang:
Rebased onto latest `main` (the sibling cosh fix #3691 has since merged) — now conflict-free and a single clean commit. Also switched the regression test to use MoonBit's **Double hex-float literal** for the threshold, which is exact and reads better than reinterpreting an integer:

```moonbit
let threshold : Double = 0x1.633ce8fb9f87dp9  // bit pattern 0x408633ce8fb9f87d
assert_eq(threshold.reinterpret_as_uint64(), 0x408633ce8fb9f87d)
```

`moon fmt` / `moon check --deny-warn` / `moon test math --target all` (140/140 on wasm, wasm-gc, js, native) / `moon info` (no `.mbti` change) all pass.

PR #3702, opened 2026-06-25
Optimize BytesView::find on Wasm targets

mizchi:
## Motivation

This PR is intended as a concrete example of what the core library can do if the `V128` type is available to core implementations. It uses `V128` to express the wasm SIMD byte-search fast path in MoonBit code instead of inline wasm snippets, while keeping the public `BytesView::find` API unchanged.

It also shows one remaining abstraction gap: the wasm-gc path currently needs to copy from `Bytes` to `FixedArray[Byte]` before loading vector lanes. A zero-copy way to load `V128` from `Bytes` / `BytesView` would make this style of implementation more generally useful.

## Summary
- split `BytesView::find` into scalar, wasm, and wasm-gc target-specific implementations
- use SIMD byte scanning for wasm and wasm-gc, with scalar verification for substring matches
- share small private `V128` helpers in `builtin` instead of inline wasm snippets
- add regression tests and benchmarks for single-byte, substring, dense-miss, offset-view, and small-input cases

## Benchmarks
Measured with `moon bench --release -p builtin -f bytes_find_bench_test.mbt --no-parallelize`. The baseline is the scalar implementation before this change.

| Case | wasm baseline | wasm PR | wasm-gc baseline | wasm-gc PR |
| single byte hit end, n=4096 | 2.61 µs | 171.98 ns | 2.93 µs | 1.79 µs |
| single byte miss, n=4096 | 2.76 µs | 176.66 ns | 3.12 µs | 1.86 µs |
| substring rare hit end, n=4096 | 2.72 µs | 227.98 ns | 2.89 µs | 1.77 µs |
| substring dense miss, n=4096 | 9.88 µs | 230.14 ns | 11.43 µs | 1.71 µs |
| offset view rare hit end, n=4096 | 2.61 µs | 244.07 ns | 3.12 µs | 1.71 µs |

Small-input checks on the PR branch:

| Case | wasm PR | wasm-gc PR |
| --- | ---: | ---: |
| single byte miss, n=32 | 15.44 ns | 35.11 ns |
| substring miss, n=32 | 26.79 ns | 26.91 ns |

## Notes
The wasm implementation is now written through private `V128` helper methods rather than inline wasm snippets. In local measurements, the inline-wasm version was slightly faster on wasm, but this version remains much faster than the scalar baseline and is easier to maintain.

The small private `V128` wrappers use direct intrinsics for the target-only operations. This avoids coverage instrumentation hitting fallback helper bodies that are replaced by target intrinsics.

The wasm-gc implementation currently converts `Bytes` to `FixedArray[Byte]` before loading V128 lanes. That keeps the implementation correct and still improves the measured cases, but a zero-copy abstraction for viewing `Bytes` as `FixedArray[Byte]` on wasm-gc would be preferable. It would avoid the copy and make reusable SIMD helpers easier to write.

I also tried using the same first+last-byte SIMD prefilter on wasm-gc, but it regressed substring cases to around 4.4-4.5 µs, so the wasm-gc path keeps the cheaper last-byte anchor scan.

## Validation
- `moon fmt`
- `moon info`
- `moon check --target wasm`
- `moon check --target wasm-gc`
- `moon check --target native` (existing missing_doc warnings only)
- `moon check --target js` (existing missing_doc warnings only)
- `moon test --enable-coverage`
- `moon test --target wasm builtin/bytes_test.mbt`
- `moon test --target wasm-gc builtin/bytes_test.mbt`
- `moon bench --target wasm --release -p builtin -f bytes_find_bench_test.mbt --no-parallelize`
- `moon bench --target wasm-gc --release -p builtin -f bytes_find_bench_test.mbt --no-parallelize`


mizchi:
### Coverage ICE repro

I hit a reproducible compiler ICE while this PR had `#intrinsic` wrappers with MoonBit fallback bodies in `builtin/bytes_find_v128.mbt`.

The triggering shape was:

```moonbit
fn bytes_find_v128_lo(value : V128) -> UInt64 = "%v128.lo"
fn bytes_find_v128_hi(value : V128) -> UInt64 = "%v128.hi"

fn bytes_find_u8_lane(value : V128, lane : Int) -> Byte {
  let word = if lane < 8 {
    bytes_find_v128_lo(value)
  } else {
    bytes_find_v128_hi(value)
  }
  ((word >> (lane % 8 * 8)) & 0xffUL).to_byte()
}

#intrinsic("%v128.i8x16_bitmask")
fn V128::i8x16_bitmask(self : V128) -> Int {
  let mut result = 0
  for i in 0..<16 {
    result = result | (((bytes_find_u8_lane(self, i).to_int() >> 7) & 1) << i)
  }
  result
}
```

Similar fallback bodies for `i8x16_splat` and `i8x16_eq` also reproduced it.

Repro commands:

```bash
moon test --target wasm --enable-coverage -p v128
moon test --target wasm-gc --enable-coverage -p v128
```

Observed errors:

```text
Moonc.Basic_hash_string.Key_not_found("$_M0FPB21bytes__find__u8__lane")
Moonc.Basic_hash_string.Key_not_found("$_M0FPB31bytes__find__bool__to__u8__mask")
```

`moon test --target wasm-gc --enable-coverage -p builtin` passed, but linking the `v128` tests with coverage enabled failed. Adding `#coverage.skip` to these helpers did not avoid the ICE.

The current patch avoids this by using direct intrinsic definitions for these target-only wrappers, so `moon test --enable-coverage` now passes locally.

Yu-zh:
I have cleaned up the implementation here and opened PR #3710 . I'd appreciate if you take some time to review.

Several changes that I want to bring up:
1. The v128 package apply to native backends as well so the optimization for wasm applies to native backend as well.
2. I added Rabin-Karp approach as fallback when the first-byte-match candidates failed to match many times. This follows golang's implementation

mizchi:
Thanks for taking this over and cleaning it up, @Yu-zh — much appreciated. This PR turned into a bit of a mess from all the trial-and-error on my side (sorry about that), so I'm happy to defer to #3710. Reviewing it now from the perf side.

mizchi:
Superseded by #3710 (cleaner, adds native v128 + a Rabin-Karp worst-case fallback). Thanks @Yu-zh — closing this in favor of that. Reviewed #3710 and it looks good to me.

PR #3704, opened 2026-06-26
fix(json): reject backtick as hex digit in \u escapes

bobzhang:
## Summary

`@json.parse` accepts backtick as hex digit 9 in `\uXXXX` escapes: the invalid JSON `"\u`bcd"` parses successfully to `"\u{9BCD}"` (`"鯍"`) instead of raising `ParseError`.

The hex letter branch of `ParseContext::lex_hex_digits` case-folds with `c & ~32` and only checks the upper bound:

```moonbit
let d = (c.to_int() & (32).lnot()) - 'A'.to_int() + 10
if d > 15 { ctx.invalid_char(shift=-1) }
```

Backtick (0x60) folds to `'@'` (0x40), yielding `d == 9`, which passes `d > 15`. It is the only character in the codespace with this property (0x41–0x5F fold to themselves; of 0x60–0x7F only 0x60 lands below `'A'` after folding). Fix: also reject `d < 10`, since every valid letter digit folds into 10–15.

## Relationship to #3678

This is the same fix as #3678, restructured into a clean **3-commit, test-driven** history:

1. **`test`** — add the regression test asserting `InvalidChar` for `` \u`bcd ``. It is **red** against current `main` (the invalid input parses), pinning down the bug.
2. **`fix`** — add the `d < 10` lower bound; turns the failing test green.
3. **`test`** — broaden coverage: `\u@bcd`, `\u1G23`, `\u1g23` rejections plus positive boundary tests that all valid hex ranges (`0-9`, `a-f`, `A-F`) still parse.

No public API change (`pkg.generated.mbti` untouched).

## Testing

- `moon test json` — 174/174 passing
- `moon fmt` — clean

### DIFF
--- a/json/lex_string.mbt
+++ b/json/lex_string.mbt
@@ -90,8 +90,10 @@ fn ParseContext::lex_hex_digits(
     match ctx.read_char() {
       Some(c) =>
         if c >= 'A' {
+          // Case-folding maps '`' (0x60) to '@' (0x40), i.e. d == 9, so
+          // reject anything outside the letter range 'A'..'F' (10..15).
           let d = (c.to_int() & (32).lnot()) - 'A'.to_int() + 10
-          if d > 15 {
+          if d < 10 || d > 15 {
             ctx.invalid_char(shift=-1)
           }
           continue (r << 4) | d
--- a/json/lex_string_test.mbt
+++ b/json/lex_string_test.mbt
@@ -31,3 +31,17 @@ test "lex_hex_digits incomplete hex digits" {
     ),
   )
 }
+
+///|
+test "lex_hex_digits rejects backtick as hex digit" {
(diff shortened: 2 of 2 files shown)


bobzhang:
**Fresh `codex` CLI review** (`codex exec review --base main`, model `gpt-5.5`, ran the json tests itself — 174/174):

> The change correctly tightens hex digit validation and the added tests cover the previously accepted invalid characters as well as valid mixed-case escapes. Targeted json tests pass.

---

_(Edited: this PR has since **merged**.)_ Follow-up: #3705 now sits on top of this fix and **rewrites** `lex_hex_digits` with char-range patterns, removing the `c & ~32` case-folding that the `d < 10` guard here was protecting — so the bug becomes structurally impossible. Pure refactor, no behavior change.

PR #3706, opened 2026-06-26
refactor(json): use char-range/equality patterns in number + string lexers

bobzhang:
## Summary

Follow-up cleanup to #3705, applying the same "let the pattern be the check" idea across the rest of the JSON lexer. Two behavior-preserving refactors:

### 1. Match digit ranges directly in the number lexer
The number lexer repeatedly did `Some(c) => if c is ('0'..='9') { ... }`. Fold the range into the match pattern (`Some('0'..='9') => ...`) across `lex_decimal_integer`, `lex_decimal_point`, `lex_decimal_fraction`, `lex_decimal_exponent`, `lex_decimal_exponent_sign`, `lex_decimal_exponent_integer`, `lex_zero`, and the `'1'..='9'` arm in `lex_value`. These arms don't use the digit's value (accumulation happens in `scan_json_number`), so no binding is needed. Also merges `Some('+') | Some('-')` → `Some('+' | '-')`.

### 2. Drop redundant control-char checks in `lex_string`
Two spots named `'\n'`/`'\r'` explicitly even though the general control-char guard already covers them:
- Fast path: `c == '\n' || c == '\r' || c < ' '` is just `c < ' '` (both 0x0A and 0x0D are `< ' '` = 0x20).
- Slow path: the `Some('\n' | '\r') => invalid_char(shift=-1)` arm is subsumed by the catch-all `Some(ch) => if ch.to_int() < 32 { ... }`.

Added a slow-path regression test (raw newline after a valid `\t` escape), since that branch was previously only reached via the now-removed explicit arm.

## Scope
- Independent of #3705 (different functions) — they touch `lex_string.mbt` in non-overlapping places.

## Testing
- `moon test json` — 175/175 passing (added 1 regression test)
- `moon check json --target all` — 0 errors / `moon fmt` clean

### DIFF
--- a/json/lex_main.mbt
+++ b/json/lex_main.mbt
@@ -52,13 +52,11 @@ fn ParseContext::lex_value(
           let (n, repr) = ctx.lex_zero(start=ctx.offset - 2)
           return Number(n, repr.map(repr => repr.to_owned()))
         }
-        Some(c2) => {
-          if c2 is ('1'..='9') {
-            let (n, repr) = ctx.lex_decimal_integer(start=ctx.offset - 2)
-            return Number(n, repr.map(repr => repr.to_owned()))
-          }
-          ctx.invalid_char(shift=-1)
+        Some('1'..='9') => {
+          let (n, repr) = ctx.lex_decimal_integer(start=ctx.offset - 2)
+          return Number(n, repr.map(repr => repr.to_owned()))
         }
+        Some(_) => ctx.invalid_char(shift=-1)
         None => raise InvalidEof
       }
     Some('0') => {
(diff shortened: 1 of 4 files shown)


bobzhang:
**Fresh `codex` CLI review** (`codex exec review --base main`, model `gpt-5.5`, ran `moon test`/`moon check` itself — 175/175):

> No actionable correctness issues were found in the diff. The changes are behavior-preserving refactors plus an added regression test, and the json test/check targets pass.

Follow-up to #3705; independent of it (non-overlapping functions). Group C (the hot `scan_json_number` / `lex_skip_whitespace` loops) was intentionally left out pending a benchmark run.

PR #3707, opened 2026-06-26
refactor(json): use match in the hot number-scan and whitespace loops

bobzhang:
## Summary

Group C of the JSON-lexer pattern cleanups (follow-up to #3705 / #3706): convert the two remaining char `if`/`else if` cascades — both on the parse hot path — to `match`.

- **`scan_json_number`**: `if c is ('0'..='9') { … } else if c == '.' { … } else if c == 'e' || c == 'E' { … }` → `match` with `'0'..='9' as c`, `'.'`, and `'e' | 'E'` arms (digit case still checked first).
- **`lex_skip_whitespace`**: `c == ' ' || c == '\t' || c == '\r' || c == '\n'` → a `' ' | '\t' | '\r' | '\n'` match arm.

## Benchmark (this is why Group C was held back)

These sit on the hot path, so I measured before/after with `moon bench -p json --target native --release` (n=1000 mixed object/array), including an **interleaved A/B** to control for thermal drift:

| | mean |
|---|---|
| baseline (main) | ~0.98–1.02 ms |
| after (this PR) | ~0.95–0.96 ms |

The `match` form is **neutral-to-slightly-faster, never slower** — one interleaved block showed ~5% (1.00 ms → 0.95 ms), another showed parity. So no regression; readability is the real win.

## Scope
- `moon test json` — 175/175 passing; `moon check --target all` clean; `moon fmt` clean.

### DIFF
--- a/json/lex_misc.mbt
+++ b/json/lex_misc.mbt
@@ -138,11 +138,9 @@ fn ParseContext::lex_skip_whitespace(ctx : ParseContext) -> Unit {
     if offset >= end {
       break offset
     }
-    let c = ctx.input.unsafe_get(offset)
-    if c == ' ' || c == '\t' || c == '\r' || c == '\n' {
-      continue offset + 1
-    } else {
-      break offset
+    match ctx.input.unsafe_get(offset) {
+      ' ' | '\t' | '\r' | '\n' => continue offset + 1
+      _ => break offset
     }
   }
 }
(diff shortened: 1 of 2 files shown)


bobzhang:
**Fresh `codex` CLI review** (`codex exec review --base main`, model `gpt-5.5`, ran `moon test`/`moon check` across targets itself — 175/175):

> No correctness issues were found in the diff. The changes are equivalent conditional-to-pattern-match rewrites, and the json package checks/tests pass across targets.

Final piece of the lexer pattern-cleanup series (#3705 → #3706 → this). Benchmark-checked since both loops are hot; see the PR description for the interleaved A/B numbers (neutral-to-slightly-faster, never slower).

PR #3710, opened 2026-06-26
Optimize bytes find

Yu-zh:
## Summary

- Move `Bytes::find` and `BytesView::find` from `builtin` into the `bytes` package.
- Add a bytes search implementation with trivial-case dispatch, SIMD byte search on linear-memory backends, short-pattern candidate filtering, and a Rabin-Karp fallback for long dense-candidate cases.
- Reuse `Bytes::unsafe_range_equal` from the bytes package through a hidden public API after view-relative bounds are validated by the search logic.
- Add edge-case tests and an all-backend benchmark file.

## Validation

- `git diff --check`
- `moon fmt`
- `env MOONC_RC_CONVENTION=borrow moon info --target wasm,wasm-gc,js,native`
- `moon check --target all bytes --no-render`
- `moon test --target all bytes/find_test.mbt --no-render`
- `moon bench --target all --release -p moonbitlang/core/bytes -f find_bench_test.mbt --no-render --warn-list=-74-20 --no-parallelize`

## Performance

Baseline is current `origin/main` with the same `bytes/find_bench_test.mbt` harness copied into a temporary worktree. Benchmarks were run with `--release`; speedup is `main mean / PR mean`.

| backend | workload | main | PR | speedup |
| --- | --- | ---: | ---: | ---: |
| wasm | byte miss n=4096 | 4.05 us | 274.65 ns | 14.7x |
| wasm | substring dense miss n=4096 | 14.29 us | 378.09 ns | 37.8x |
| wasm | long dense miss n=4096 | 102.18 us | 4.78 us | 21.4x |
| wasm | substring miss n=32 | 119.24 ns | 35.02 ns | 3.4x |
| wasm-gc | byte miss n=4096 | 4.04 us | 2.04 us | 2.0x |
| wasm-gc | substring dense miss n=4096 | 14.21 us | 2.01 us | 7.1x |
| wasm-gc | long dense miss n=4096 | 98.98 us | 4.74 us | 20.9x |
| wasm-gc | substring miss n=32 | 104.09 ns | 24.58 ns | 4.2x |
| js | byte miss n=4096 | 3.45 us | 2.27 us | 1.5x |
| js | substring dense miss n=4096 | 12.98 us | 2.27 us | 5.7x |
| js | long dense miss n=4096 | 96.24 us | 4.74 us | 20.3x |
| js | substring miss n=32 | 96.82 ns | 25.48 ns | 3.8x |
| native | byte miss n=4096 | 1.78 us | 136.15 ns | 13.1x |
| native | substring dense miss n=4096 | 7.86 us | 181.08 ns | 43.4x |
| native | long dense miss n=4096 | 48.44 us | 4.57 us | 10.6x |
| native | substring miss n=32 | 63.92 ns | 14.52 ns | 4.4x |

The full run also covers byte hit, rare substring hit, offset-view hit, and byte miss n=32.

### DIFF
--- a/builtin/bytes_find.mbt
+++ b/builtin/bytes_find.mbt
@@ -13,28 +13,347 @@
 // limitations under the License.
 
 ///|
-/// Returns the offset of the first occurrence of the given
-/// bytes substring. If the substring is not found, `None` is returned.
+let find_max_short_pattern : Int = 32
+
+///|
+let find_max_bruteforce : Int = 64
+
+///|
+let find_prime_rk : UInt = 16777619U
+
+///|
+/// Returns the offset of the first occurrence of the given bytes substring.
+///
+/// If the substring is not found, `None` is returned.
 pub fn BytesView::find(target : BytesView, pattern : BytesView) -> Int? {
-  // TODO: more efficient algorithm
+  find_index(target, pattern)
+}
+
+///|
+/// Returns the offset of the first occurrence of the given bytes substring.
+///
+/// If the substring is not found, `None` is returned.
+pub fn Bytes::find(target : Bytes, pattern : BytesView) -> Int? {
+  target[:].find(pattern)
+}
+
+///|
+// Handle trivial cases first, use the short-pattern scanner when the pattern is
+// small, then use the guarded long-pattern scanner.
+fn find_index(target : BytesView, pattern : BytesView) -> Int? {
   let target_len = target.length()
   let pattern_len = pattern.length()
-  for i in 0..<=(target_len - pattern_len) {
-    for j in 0..<pattern_len {
-      guard target.unsafe_get(i + j) == pattern.unsafe_get(j) else { break }
-    } nobreak {
-      return Some(i)
+  if pattern_len == 0 {
+    return Some(0)
+  }
+  if pattern_len > target_len {
+    return None
+  }
+  if pattern_len == 1 {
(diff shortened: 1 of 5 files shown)


mizchi:
Reviewed from the perf side — this looks great, and it's a clear improvement over my #3702. Both of the changes you flagged are the right calls:

1. **Native v128 too** — agreed, there was no reason to gate the SIMD path to wasm; native lowers `@v128.*` to real vector ops as well. Good to have both linear-memory backends share `find_byte_from_bytes` / `find_short_candidate_from_bytes`.

2. **Rabin-Karp fallback** — this fixes the worst case my version didn't handle: a dense first-byte-but-no-full-match input (e.g. `aaaa…` searching `aaa…b`) was O(n·m) in the plain first-byte scanner. The `find_cutover = 4 + found/16` switchover keeps it linear while preserving the fast common-case scan. Mirroring Go's `strings.Index` heuristic is a sensible choice.

A couple of things I checked and they're fine:
- The first+last-byte SIMD prefilter loads `v128_load(pos + last_offset)` near the end — I worked through the bounds and `candidate_end` is set so the last vector load lands exactly on the final valid byte, no OOB. Nicely done.
- Rabin-Karp verifies every hash hit with `find_match_range`, so collisions can't produce false positives.

LGTM. Closing #3702 in favor of this.

PR #3711, opened 2026-06-26
perf(builtin): use unchecked array access in sort inner loops

mizchi:
## Motivation

Profiling `Array::sort` (native, Time Profiler) showed **~53% of the time in
`MutArrayView::at` / `MutArrayView::set`** — pure element-access overhead — with
zero allocation. The sort helpers operate on a `MutArrayView` and index it with
the checked `arr[i]` accessor, which on every comparison and swap performs a
bounds check *and* goes through the non-inlined `buf()` / `start()` / `len()`
accessors.

## Change

Every index used inside the sort helpers is derived from the view's own length
and is provably in bounds, so the hot reads/writes are switched to
`unsafe_get` / `unsafe_set`. This covers both `sort` (Compare) and `sort_by`:
`swap`, `rev_in_place`, partition, bubble, insertion, merge, heap sift-down and
choose-pivot.

No behavioral change — same algorithm, same results, just without the redundant
per-access bounds check.

## Benchmark

`builtin/sort_bench_test.mbt` (native, distinct `Int` keys):

| case | before | after | |
|---|---|---|---|
| `sort` n=1000 | 52.8 µs | **8.6 µs** | 6.1x |
| `sort` n=100000 | 8.65 ms | **3.88 ms** | 2.2x |
| `sort_by` n=10000 | 801 µs | **280 µs** | 2.9x |

Gains are data-dependent — duplicate-heavy inputs see closer to 2x — but the
bounds-check overhead is removed in all cases (profile: `MutArrayView::at`
disappears from the top entirely).

## Validation

`moon test -p builtin` passes on native (2835), wasm-gc (2877), wasm (2877) and
js (2865). The existing sort correctness tests (`fixed_test_sort`, sort/sort_by/
heap_sort/bubble_sort, `sort_by_fallback`) all still pass.

### DIFF
--- a/builtin/array_sort_by_impl.mbt
+++ b/builtin/array_sort_by_impl.mbt
@@ -48,8 +48,8 @@ fn[T] fixed_quick_sort_by(
     if pred is Some(p) {
       // pred is less than all elements in arr
       // If pivot equals to pred, then we can skip all elements that are equal to pred.
-      if cmp(p, arr[pivot]) == 0 {
-        let i = for i = pivot; i < len && cmp(p, arr[i]) == 0; {
+      if cmp(p, arr.unsafe_get(pivot)) == 0 {
+        let i = for i = pivot; i < len && cmp(p, arr.unsafe_get(i)) == 0; {
           continue i + 1
         } nobreak {
           i
@@ -62,9 +62,13 @@ fn[T] fixed_quick_sort_by(
     // Reduce the stack depth by only call quick_sort on the smaller partition.
     if left.length() < right.length() {
       fixed_quick_sort_by(left, cmp, pred, limit)
-      continue limit, right, Some(arr[pivot]), was_partitioned, balanced
+      continue limit,
+        right,
+        Some(arr.unsafe_get(pivot)),
+        was_partitioned,
+        balanced
     } else {
-      fixed_quick_sort_by(right, cmp, Some(arr[pivot]), limit)
+      fixed_quick_sort_by(right, cmp, Some(arr.unsafe_get(pivot)), limit)
       continue limit, left, pred, was_partitioned, balanced
     }
   }
@@ -82,7 +86,8 @@ fn[T] fixed_try_bubble_sort_by(
 ) -> Bool {
   let max_tries = 8
   for i in 1..<arr.length(); tries = 0 {
-    let sorted = for j = i, sorted = true; j > 0 && cmp(arr[j - 1], arr[j]) > 0; {
+    let sorted = for j = i, sorted = true; j > 0 &&
+                    cmp(arr.unsafe_get(j - 1), arr.unsafe_get(j)) > 0; {
       arr.swap(j, j - 1)
       continue j - 1, false
     } nobreak {
@@ -110,7 +115,9 @@ fn[T] fixed_try_bubble_sort_by(
 /// Returns whether the array is sorted.
 fn[T] fixed_bubble_sort_by(arr : MutArrayView[T], cmp : (T, T) -> Int) -> Unit {
   for i in 1..<arr.length() {
-    for j = i; j > 0 && cmp(arr[j - 1], arr[j]) > 0; j = j - 1 {
+    for j = i
+        j > 0 && cmp(arr.unsafe_get(j - 1), arr.unsafe_get(j)) > 0
+        j = j - 1 {
       arr.swap(j, j - 1)
     }
   }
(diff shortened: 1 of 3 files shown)


bobzhang:
@Guest0x0 it is a bit surprising that bounds checking overhead is too large, could it trigger GC ops?

bobzhang:
I did a follow-up analysis in separate worktrees:

- base: `3fa43b6037bf23ecfcb7b84f898fcf234f8fef28`
- PR head: `31a5b233cd81499d8358c7d8b295e38abdbfa05b`
- toolchain: `moon 0.1.20260626 (345c81b 2026-06-26)`

Short version: I do **not** think the big result here is caused by GC / incref / decref for the benchmarked `Array[Int]` path. It is mainly the checked `MutArrayView::at/set` fast path being much more expensive and less optimizable than `unsafe_get/unsafe_set`, multiplied by the number of element accesses in sort's inner loops.

### What the native code shows

For `MutArrayView::at[Int]`, the generated native C fast path does roughly:

```c
if (index >= 0) {
  len = self.end - self.start;
  ok = index < len;
} else {
  ok = false;
}
if (ok) {
  buf = self.buf;
  physical_index = self.start + index;
  if (physical_index < 0 || physical_index >= Moonbit_array_length(buf)) moonbit_panic();
  return buf[physical_index];
} else {
  // Build StringBuilder error message, abort, decref temporaries.
}
```

For `MutArrayView::unsafe_get[Int]`, generated native C does roughly:

```c
buf = self.buf;
physical_index = self.start + index;
if (physical_index < 0 || physical_index >= Moonbit_array_length(buf)) moonbit_panic();
return buf[physical_index];
```

So `unsafe_get` on native is not completely unchecked at the backing-buffer level. It removes the **view-level** bounds check and the cold formatted-abort path. In the hot loops here, all indices are already derived from the view length, so that view-level check is redundant.

For the `Int` versions I inspected, neither `at` nor `unsafe_get` contains `moonbit_incref` / `moonbit_decref`. There are some `moonbit_decref` calls around view slicing / recursive loop state, but those are not the access-site delta and are present in both versions.

For object element types, reads do have refcount traffic, but both variants share it. For example, `MutArrayView::at[String]` and `MutArrayView::unsafe_get[String]` both load the string and `moonbit_incref` it before returning. Similarly, checked and unsafe `set` both do the expected value `incref` and old-value `decref`; the checked version just adds the view bounds check and formatted-abort cold path.

### Reproduced sort benchmark

Command:

```sh
moon bench --target native -p builtin --file sort_bench_test.mbt --no-parallelize --no-render
```

I copied the PR's benchmark file into the base worktree to compare the same benchmark.

| case | base | PR head | speedup |
|---|---:|---:|---:|
| `Array::sort n=1000` | 49.00 us | 9.02 us | 5.4x |
| `Array::sort n=100000` | 8.56 ms | 4.12 ms | 2.1x |
| `Array::sort_by n=10000` | 744.61 us | 312.88 us | 2.4x |

`--release` was not materially different on my run:

| case | base release | PR head release | speedup |
|---|---:|---:|---:|
| `Array::sort n=1000` | 48.69 us | 8.76 us | 5.6x |
| `Array::sort n=100000` | 9.16 ms | 4.32 ms | 2.1x |
| `Array::sort_by n=10000` | 748.05 us | 318.18 us | 2.4x |

### Access-only microbenchmarks

I also added a temporary benchmark that only loops over a `MutArrayView`, independent of sorting. Results were essentially the same on base/head because these methods themselves did not change.

Command:

```sh
moon bench --target native -p builtin --file zz_pr3711_access_bench_test.mbt --no-parallelize --no-render
```

Representative base-worktree results:

| case | checked | unsafe | ratio |
|---|---:|---:|---:|
| read 100k `Int`s | 101.36 us | 3.74 us | 27.1x |
| write 100k `Int`s | 101.86 us | 8.46 us | 12.0x |
| read 20k `String`s | 22.91 us | 6.43 us | 3.6x |

The `String` read case is useful for the GC/refcount question: it includes object-return refcount traffic on both paths, yet the unsafe path is still faster and the ratio is not larger because of refcounting. This supports the conclusion that refcounting is shared absolute cost, not the main delta removed by this PR.

### Why the sort speedup can be this large

A single checked view access is not just one branch in generated native code. It is a public function-shaped operation with:

1. `index >= 0`
2. `index < self.len()` where `len` is `end - start`
3. `start + index`
4. the backend's physical backing-array check
5. a large formatted-abort cold path in the same generated function

Then sort multiplies that by a lot:

- every comparison usually performs two element reads,
- swaps perform multiple reads/writes,
- small arrays and insertion/bubble phases have very cheap `Int` comparisons, so access overhead dominates,
- the checked accessor shape appears much harder for the native optimizer to inline/vectorize away than the unsafe accessor.

So the surprising part is not GC. The surprising part is that `MutArrayView` checked indexing currently carries two layers of bounds logic on native, and the sort inner loops call it often enough that it becomes a major fraction of total work.

PR #3713, opened 2026-06-26
perf(builtin): SWAR BytesView::lexical_compare on native/wasm

mizchi:
## Change

`BytesView::lexical_compare` compared a byte at a time. On backends with cheap
64-bit integers (native, wasm), reading 8 bytes at a time **big-endian** lets a
single unsigned `UInt64` comparison settle a whole 8-byte run — a big-endian
read makes the `UInt64` ordering coincide with the lexical (unsigned, left-to-
right) byte ordering. Falls back to a scalar byte loop for the `< 8`-byte tail
and then compares lengths.

No public API change (same `BytesView::lexical_compare`; `Bytes::lexical_compare`
delegates to it). It stays in `builtin` — this is plain `UInt64` arithmetic, no
`v128` dependency.

## Benchmark (n=256)

| backend | scalar | this PR | |
|---|---|---|---|
| native | ~92–130 ns | **~28 ns** | 3.3–4.6× |
| wasm | ~172 ns | **~104 ns** | 1.7× |
| wasm-gc | ~157 ns | ~157 ns (scalar) | unchanged |
| js | ~137 ns | ~137 ns (scalar) | unchanged |

Early-exit case (inputs differ at byte 1): no regression on any backend
(~6–8 ns, same as scalar).

## ⚠️ Concern worth discussing: large per-runtime divergence

The reason this is **gated to native/wasm via `moon.pkg`** rather than applied
everywhere is that the SWAR path is a *regression* on the GC backends, because
`UInt64` is emulated there:

- **js**: applying SWAR unconditionally is **~37× slower** (5.1 µs vs 137 ns).
- **wasm-gc**: ~1.3× slower.

So the same source-level "optimization" swings from a 4× win to a 37× loss
depending purely on the runtime's `UInt64` cost. A couple of implications I'd
like input on:

1. **Is per-backend gating like this the pattern we want** for this kind of
   integer-width-sensitive optimization, or is the maintenance/readability cost
   (two impls + a `moon.pkg` split per op) not worth a native-only win?
2. **Benchmarking is noisy and runtime-dependent here** — the scalar baseline's
   "equal" case had high variance (σ up to ~50 ns), and the win only exists on
   2 of 4 backends. If we take more of these, it probably needs a more
   disciplined harness (fixed inputs per backend, regression gating that is
   itself per-backend) so a "win" on one runtime doesn't hide a "loss" on
   another. Happy to align on a convention before adding more SWAR ops.

I'm comfortable dropping this if the gating tradeoff isn't worth it; mainly
raising it as a concrete data point on the per-runtime divergence.

### DIFF
--- a/builtin/bytesview.mbt
+++ b/builtin/bytesview.mbt
@@ -728,41 +728,10 @@ pub impl Compare for BytesView with fn compare(self, other) -> Int {
 }
 
 ///|
-/// Performs a lexicographical comparison of two byte views.
-///
-/// This method compares the views byte by byte until a difference is found
-/// or one view is exhausted. Unlike the `Compare` trait implementation which
-/// uses shortlex order (shorter views come first), this method compares based
-/// purely on byte values until a difference is found.
-///
-/// # Returns
-///
-/// - A negative integer if `self` is lexicographically less than `other`
-/// - Zero if `self` is lexicographically equal to `other`
-/// - A positive integer if `self` is lexicographically greater than `other`
-///
-/// # Example
-///
-/// ```mbt check
-/// test {
-///   inspect(b"\x01\x02"[:].lexical_compare(b"\x01\x02\x03"), content="-1")
-///   inspect(b"\x01\x02\x03"[:].lexical_compare(b"\x01\x02"), content="1")
-///   inspect(b"\x01\x02\x03"[:].lexical_compare(b"\x01\x02\x03"), content="0")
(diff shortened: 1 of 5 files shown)


Yu-zh:
Thanks for finding this optimization point. I go all the way to use V128 for comparison in #3790

mizchi:
Closing this in favor of #3790, which carries the same optimization point further with the V128 implementation.

PR #3728, opened 2026-06-27
refactor(builtin): deprecate `at` in favor of index operator `[i]`

bobzhang:
## What

Deprecate the `at` method in favor of the index operator `[i]` for `Array`, `ArrayView`, `Bytes`, and `String`.

All four are declared with `#alias("_[_]")`, so `xs[i]` and `xs.at(i)` are the *same* function. This adds `#deprecated` to nudge users toward the shorter `xs[i]` spelling.

## Why it's safe

`#deprecated` on an aliased method fires **only** on the explicit `.at(i)` method call — verified against the toolchain:

| call | warns? |
|------|--------|
| `xs.at(i)` | ✅ deprecated |
| `xs[i]` (operator alias) | — no warning |
| `str.code_unit_at(i)` (named alias on `String::at`) | — no warning |

So `xs[i]` keeps working untouched, and `String`'s descriptive `code_unit_at` alias stays available.

No non-test call sites in core use `.at()` on these four types, so nothing internal needs migrating. `moon check --deny-warn` is clean and the builtin suite (2877 tests) passes.

## Note for reviewers

For `String`, `str[i]` returns a `UInt16` code unit (not a `Char`); the deprecation message points to both `str[i]` and `str.code_unit_at(i)`.

### DIFF
--- a/builtin/array.mbt
+++ b/builtin/array.mbt
@@ -180,6 +180,7 @@ pub fn[T] Array::unsafe_get(self : Array[T], idx : Int) -> T {
 ///
 #intrinsic("%array.get")
 #alias("_[_]")
+#deprecated("use `arr[i]` instead of `arr.at(i)`")
 pub fn[T] Array::at(self : Array[T], index : Int) -> T {
   let len = self.length()
   guard index >= 0 && index < len
--- a/builtin/arrayview.mbt
+++ b/builtin/arrayview.mbt
@@ -133,6 +133,7 @@ pub fn[T] ArrayView::start_offset(self : Self[T]) -> Int {
 /// }
 /// ```
 #alias("_[_]")
+#deprecated("use `view[i]` instead of `view.at(i)`")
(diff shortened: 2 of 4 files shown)


bobzhang:
Closing — on reflection, keeping the named `at` method is worth it.

While `xs[i]` is the everyday spelling, the `at` *name* serves cases the index operator can't:

- **Trait implementations / generic code** — an operation has to be referred to by a real method name (e.g. when satisfying or dispatching through a trait). The `_[_]` operator is sugar, not a first-class name, so deprecating `at` would push that code toward `unsafe_get` + manual bounds checks or awkward workarounds.
- **First-class function references** — `Array::at` can be passed around as a `(Self[T], Int) -> T` value; the `[_]` operator can't be handed to a higher-order function directly.

So deprecating `at` removes a genuinely useful named accessor for only a cosmetic gain. Keeping both `xs[i]` and `xs.at(i)`.

(The related cleanups #3729 / #3730 — which improve the accessor *bodies* — are already merged and stand on their own.)

PR #3731, opened 2026-06-27
fix(immut/vector): relax partial tail leaves folded in by concat

### LINKED ISSUE #3721: immut/vector: concat can build a radix node with a partial leaf, breaking get/at/set
## Summary

`@immut/vector.Vector::concat` can produce a tree whose `iter()`/`to_array()` view is correct but whose **random access** (`get`/`at`/`set`) is broken. The tail-normalization path of `concat` builds a `Node(_, None)` — which the getter treats as a full, radix-indexable node — that actually contains a **partial (non-full) leaf**, violating the tree invariant in `types.mbt`:

> For `Node`, the sizes array is `None` if the tree is full, i.e., we can use radix indexing.

## Minimal reproduction

```moonbit
test "concat breaks random access" {
  let mut v = @vector.new()
  let mut total = 0
  for _ in 0..<8 {
    v = v.concat(@vector.from_iter((total).until(total + 17)))
    total += 17
  }
  // iter()/to_array() are correct (all 136 elements, 0..135):
  assert_eq(v.iter().to_array(), Array::makei(136, i => i))
  // but random access throws "array element access out of bounds":
  for j in 0..<total {
    assert_eq(v.get(j), Some(j)) // RuntimeError: array element access out of bounds
  }
}
```

- `iter()` / `to_array()` / `each` / `fold`: **correct** (they traverse children directly).
- `get(j)` / `at(j)` / `set(j, _)`: **throw** `array element access out of bounds` inside `Tree::get`'s radix fast path (`tree.mbt:76`, `leaf[index & BITMASK]`), for an interior index.

A single `concat` does not trigger it; it takes a few normalizations to accumulate a partial leaf in a non-final position under a `None` node. Eight 17-element chunks (136 elements) is the smallest case I found.

## Root cause

The bug is on the small-right-operand path of `concat` (`other.tree is Empty`), which folds the left tail into the tree via `Vector::normalize_tree` → `Tree::push_end` / `Tree::append_right_leaf`. When a **partial** branch is appended to a node that is currently `None` (radix/full), these helpers keep `sizes = None`:

- `tree.mbt` `push_end`: the "create a new node" arms produce `Node([self, new_branch], None)` (e.g. `tree.mbt:233`, `tree.mbt:239`) and `push_sizes_last(None) => None` (`tree.mbt:177`), even though `new_branch` wraps a partial leaf.
- `tree_append.mbt` `append_right_leaf`: same pattern — `Node([self, new_branch], None)` (`tree_append.mbt:98`, `:103`) and `push_sizes_last(None) => None` (`:34`) when the appended leaf is shorter than `BRANCHING_FACTOR`.

Because the resulting node is marked `None`, `Tree::get`/`set` use `radix_indexing`, which assumes every leaf has exactly `BRANCHING_FACTOR` elements. The partial leaf then gets indexed out of range.

These code paths were previously uncovered by the test suite, which is how the bug went unnoticed.

## Expected

After any sequence of `concat`s, `get(j) == Some(v.to_array()[j])` for all valid `j` (random access must agree with iteration).

## Environment

- Found on `main` (`immut/vector`), all backends.

bobzhang:
## Summary

Fixes #3721 — `@immut/vector.concat` could build a radix node (`Node(_, None)`) containing a **partial leaf**, so `get`/`at`/`set` read out of bounds while `iter`/`to_array` stayed correct.

## Root cause

When the right operand is small (`other.tree` is `Empty`) and the tails don't fit in one leaf, `concat` folds the left vector's tail into its tree via `Vector::normalize_tree`, which used `Tree::push_end` / `Tree::append_right_leaf`. When the tail was shorter than `BRANCHING_FACTOR`, those helpers appended a **non-full** leaf but kept the enclosing node marked `None` — i.e. a full, radix-indexable node.

A single partial leaf at the right edge is still radix-navigable, so the corruption was latent. But a *later* concat pushed that leaf into an **interior** position, where radix indexing assumes every preceding child is full and therefore computes the wrong offsets — reading out of bounds. Minimal trigger: concat eight 17-element vectors (136 elements).

## Fix

Fold the tail through the existing **relaxing** `Tree::concat`, which computes proper subtree sizes (`compute_sizes`) and marks the node relaxed (`Some(sizes)`) whenever it contains a non-full leaf. This is the same path the general concat already uses, so it's well exercised.

- `push_end` is now unused and is removed.
- `append_right_leaf` is unchanged and remains on the **push hot path**, where the tail is always full (`BRANCHING_FACTOR`), so push builds the same left-packed radix trees as before — no read-path perf regression.

## Tests

Adds `concat_regression_test.mbt` asserting that random access (`get`/`at`) agrees with iteration and an `Array` reference after arbitrary operation sequences:
- the minimal repro (eight 17-element concats),
- a long chain of odd-sized concats (partial/full/over-full leaves) plus `set`,
- a deterministic randomized mix of concat/push/slice/pop across several seeds.

Verified the new tests **fail** on the pre-fix code and pass after the fix.

- `moon test -p moonbitlang/core/immut/vector` → 127 passed (124 existing + 3 new)
- `moon fmt` clean, no new warnings

Closes #3721


bobzhang:
in #3732

PR #3732, opened 2026-06-27
fix(immut/vector): concat no longer builds a radix node with a partial leaf

### LINKED ISSUE #3721: immut/vector: concat can build a radix node with a partial leaf, breaking get/at/set

bobzhang:
## Summary

Fixes #3721.

`@immut/vector.Vector::concat` could produce a tree whose `iter()`/`to_array()`
view was correct but whose **random access** (`get`/`at`/`set`) threw `array
element access out of bounds`.

The small-right-operand path of `concat` (`other.tree is Empty`) folds the left
tail into the tree via `normalize_tree` → `Tree::push_end` /
`Tree::append_right_leaf`. Those right-spine helpers propagated the radix marker
(`sizes = None`) even when the appended chunk was a **partial** leaf, yielding a
`Node(_, None)` that radix indexing wrongly treats as full and then indexes out
of range for an interior index.

## Fix

A right-spine append keeps a node radix only when every leaf stays full, i.e. the
appended leaf is itself full (`delta == BRANCHING_FACTOR`). Otherwise the touched
nodes carry an explicit sizes array (via `compute_sizes`). This covers both the
new-branch path and the spine-absorbed path (a partial leaf can land in a short,
left-packed subtree). `push_end` always appends a single value, so it always
relaxes; `append_right_leaf` keeps the radix fast path for the full-leaf case
used by `push`, so `push` is not affected.

## Commits

1. `test` — failing black-box regression test (minimal 8×17 repro + a sweep over
   many chunk sizes/counts).
2. `fix` — keep sizes when folding a partial leaf into the tree.
3. `refactor` — dedupe the radix-vs-relaxed sizing decision (no behavior change).

## Validation

- New regression tests fail before the fix, pass after.
- All existing `immut/vector` tests pass on **wasm, wasm-gc, js, native**.
- A local randomized stress test (3000 trials of mixed `push`/`concat`/`set`/`pop`)
  confirms random access always agrees with a reference model and a precise
  left-packed radix invariant holds (not committed).

### DIFF
--- a/immut/vector/concat_random_access_test.mbt
+++ b/immut/vector/concat_random_access_test.mbt
@@ -1,0 +1,58 @@
+// Copyright 2026 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+///|
+/// Regression test for https://github.com/moonbitlang/core/issues/3721.
+///
+/// Repeatedly concatenating partial chunks used to fold a partial leaf into a
+/// node that stayed marked as full (`sizes = None`), so iteration was correct
+/// but radix random access (`get`/`at`/`set`) went out of bounds.
+test "concat preserves random access (issue #3721)" {
+  let mut v = @vector.new()
+  let mut total = 0
+  for _ in 0..<8 {
+    v = v.concat(@vector.from_iter(total.until(total + 17)))
+    total += 17
+  }
+  let expected = Array::makei(total, i => i)
+  // Iteration was always correct ...
+  @debug.assert_eq(v.iter().to_array(), expected)
+  // ... and random access must agree with it.
+  @debug.assert_eq(Array::makei(total, j => v[j]), expected)
+  @debug.assert_eq(Array::makei(total, j => v.get(j).unwrap()), expected)
+  // `set` walks the same path, so it must agree too.
+  @debug.assert_eq(
+    Array::makei(total, j => v.set(j, -j)[j]),
+    Array::makei(total, j => -j),
+  )
+}
+
+///|
+/// Random access must match iteration for arbitrary `concat` chunkings, not just
+/// the minimal reproduction above.
+test "concat random access matches iteration for many chunkings (issue #3721)" {
+  for chunk in [1, 5, 17, 31, 32, 33, 50] {
+    for count in [2, 3, 5, 8, 16] {
+      let mut v = @vector.new()
+      let mut total = 0
+      for _ in 0..<count {
+        v = v.concat(@vector.from_iter(total.until(total + chunk)))
+        total += chunk
+      }
+      let by_iter = v.iter().to_array()
+      let by_index = Array::makei(total, j => v[j])
+      @debug.assert_eq(by_index, by_iter)
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


bobzhang:
### Codex CLI review

Ran `codex review` on the branch. It flagged a real follow-up gap in the first
version of the fix:

> When a full tail is appended to a radix node whose existing right spine starts
> with a partial leaf, `delta == BRANCHING_FACTOR` still leaves the node marked
> as radix even though earlier children are not full. … after normalizing a
> vector with a 31-element tail and then pushing enough elements to flush a full
> tail, random access can radix-index into the short leaf and abort.

Confirmed against the branch — `concat` then a run of `push`es flushing a full
tail did abort. Root cause: `new_branch_left` wrapped a **partial** leaf in
radix (`None`) nodes. On its own that node is fine (the partial leaf is at the
right edge), but a later full-leaf `push` turns it into an interior non-full
leaf under a radix node.

Fixed by marking `new_branch_left` branches relaxed whenever the leaf is partial
(folded into the fix commit). Re-verified:

- codex's exact scenario now passes;
- a randomized stress test (4000 trials of `concat` + long `push` runs + `pop`)
  confirms a precise left-packed radix invariant holds and random access always
  matches a reference model;
- all `immut/vector` tests pass on wasm, wasm-gc, js, native.

Rebased onto latest `main`.

bobzhang:
Re-ran `codex review` after the follow-up fix and rebase — clean, no findings:

> The changes consistently switch partial right-spine branches to explicit sizes and preserve radix indexing for full-leaf appends. The vector tests, including target-all runs, pass without regressions.

bobzhang:
Another `codex review` round (rebased onto latest `main`, already up to date) — clean, no findings:

> The changes correctly preserve explicit size metadata for non-radix vector tree nodes, and the added regression tests cover the random-access failure mode without introducing evident regressions.

CAIMEOX (approved):

PR #3738, opened 2026-06-28
perf(builtin): speed up Bytes::equal

### LINKED ISSUE #3589: imrpove bytes equal perf
> to be improved later cc @Yu-zh  

 _Originally posted by @bobzhang in [#3588](https://github.com/moonbitlang/core/pull/3588/changes#r3238450881)_

mizchi:
## Summary

- Split `Bytes::equal` into target-specific implementations.
- Use `Bytes::unsafe_range_equal` with a `physical_equal` fast path on non-wasm-gc targets.
- Keep wasm-gc on the byte loop plus the same alias fast path, because `unsafe_range_equal` regressed distinct-buffer cases there in local benchmarks.
- Add direct `Bytes::equal` benchmarks for equal, aliased, and end-mismatch cases.

Fixes #3589.

## Benchmarks

Benchmarks call `Bytes::equal(...)` directly with 100,000-byte inputs. Lower is better.

Before: `upstream/main` at `183fee1c` with the same benchmark file applied locally.
After: this PR at `a428940d`.

| target | case | before | after |
| --- | --- | ---: | ---: |
| native | equal | 23.69 us +/- 183.73 ns | 1.84 us +/- 32.15 ns |
| native | aliased object | 23.79 us +/- 124.79 ns | 5.34 ns +/- 0.13 ns |
| native | mismatch at end | 25.50 us +/- 212.08 ns | 1.85 us +/- 62.76 ns |
| wasm | equal | 122.26 us +/- 974.52 ns | 53.52 us +/- 7.72 us |
| wasm | aliased object | 121.63 us +/- 427.71 ns | 15.52 ns +/- 3.05 ns |
| wasm | mismatch at end | 123.85 us +/- 3.96 us | 47.87 us +/- 587.48 ns |
| js | equal | 467.67 us +/- 11.18 us | 50.43 us +/- 656.58 ns |
| js | aliased object | 119.06 us +/- 1.46 us | 7.37 ns +/- 0.25 ns |
| js | mismatch at end | 118.66 us +/- 650.21 ns | 36.47 us +/- 10.07 us |
| wasm-gc | equal | 35.47 us +/- 141.79 ns | 40.09 us +/- 7.59 us |
| wasm-gc | aliased object | 35.90 us +/- 486.91 ns | 5.22 ns +/- 0.22 ns |
| wasm-gc | mismatch at end | 35.54 us +/- 88.56 ns | 37.19 us +/- 1.88 us |

The wasm-gc distinct-buffer cases stay close to the old loop and are noisy; the alias fast path is the intended wasm-gc improvement.

I also checked `moon test --profile` after pointing `DEVELOPER_DIR` at Xcode so `xctrace` is available. It ran successfully, but the sample count was too low to use as the primary benchmark table.

## Validation

- `moon fmt`
- `moon test -p builtin --target wasm`
- `moon test -p builtin --target wasm-gc`
- `moon test -p builtin --target native`
- `moon test -p builtin --target js`
- `moon check -p builtin --target all`
- `moon info`
- `git diff --check`
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer moon test -p builtin --target native --profile`

### DIFF
--- a/builtin/bytes.mbt
+++ b/builtin/bytes.mbt
@@ -335,43 +335,6 @@ pub fn FixedArray::set_utf16be_char(
 }
 
 ///|
-/// Compares two byte sequences for equality. Returns true only if both sequences
-/// have the same length and contain identical bytes in the same order.
-///
-/// Parameters:
-///
-/// * `self` : The first byte sequence to compare.
-/// * `other` : The second byte sequence to compare.
-///
-/// Returns `true` if the byte sequences are equal, `false` otherwise.
-///
-/// Example:
-///
-/// ```mbt check
-/// test {
-///   let bytes1 = b"\x01\x02\x03"
-///   let bytes2 = b"\x01\x02\x03"
-///   let bytes3 = b"\x01\x02\x04"
-///   inspect(bytes1 == bytes2, content="true")
-///   inspect(bytes1 == bytes3, content="false")
-/// }
-/// ```
-pub impl Eq for Bytes with fn equal(self : Bytes, other : Bytes) -> Bool {
-  if self.length() != other.length() {
-    false
-  } else {
-    let len = self.length()
-    for i in 0..<len {
-      if self[i] != other[i] {
-        break false
-      }
-    } nobreak {
-      true
-    }
-  }
-}
-
-///|
 /// Compares two byte sequences based on shortlex order. First compares the lengths of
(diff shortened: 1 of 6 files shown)


Yu-zh:
I just realized that we have a primitive for it so I will close this PR in favor of https://github.com/moonbitlang/core/pull/3753

PR #3739, opened 2026-06-28
perf(immut): speed up sorted array builders

mizchi:
## Summary

- add unchecked `SortedMap::from_sorted_array` and `SortedSet::from_sorted_array` builders for strictly ascending duplicate-free arrays
- speed up existing `from_array` without changing its public API
  - monotonic inputs are detected and built in O(n)
  - large non-monotonic inputs are copied, sorted with original indices, compacted, then bulk-built
- speed up `from_iter` by collecting once and reusing the same bulk-build paths
  - `SortedMap::from_iter` preserves first equal key representation and last value
  - `SortedSet::from_iter` keeps its existing first-representative semantics, distinct from `from_array` reverse-insertion semantics
- keep small non-monotonic inputs on the old `add` path, preserving existing whitebox tree-shape tests
- add build benchmarks for ordered, reverse, almost-ordered, and random inputs

## Benchmark

Command: `moon bench -p immut/sorted_map --target native`

| case | mean |
| --- | ---: |
| `SortedMap::from_sorted_array n=10000` | 86.89 us |
| `SortedMap::from_array reverse n=10000` | 91.59 us |
| `SortedMap::from_array ordered n=10000` | 98.99 us |
| `SortedMap::from_array almost ordered n=10000` | 246.71 us |
| `SortedMap::from_array random n=10000` | 686.09 us |
| `SortedMap::from_iter ordered n=10000` | 124.54 us |
| `SortedMap::from_iter almost ordered n=10000` | 271.86 us |
| `SortedMap::from_iter random n=10000` | 731.08 us |

Command: `moon bench -p immut/sorted_set --target native`

| case | mean |
| --- | ---: |
| `SortedSet::from_sorted_array n=10000` | 77.83 us |
| `SortedSet::from_array reverse n=10000` | 80.29 us |
| `SortedSet::from_array ordered n=10000` | 85.37 us |
| `SortedSet::from_array almost ordered n=10000` | 170.76 us |
| `SortedSet::from_array random n=10000` | 585.87 us |
| `SortedSet::from_iter ordered n=10000` | 97.88 us |
| `SortedSet::from_iter almost ordered n=10000` | 179.65 us |
| `SortedSet::from_iter rev almost ordered n=10000` | 217.83 us |
| `SortedSet::from_iter random n=10000` | 612.88 us |

## Validation

- `moon fmt`
- `moon info`
- `moon test -p immut/sorted_map --target all`
- `moon test -p immut/sorted_set --target all`
- `moon check -p immut/sorted_map --target all`
- `moon check -p immut/sorted_set --target all`
- `moon bench -p immut/sorted_map --target native`
- `moon bench -p immut/sorted_set --target native`
- `git diff --check`

`moon check` still reports pre-existing missing-doc warnings in unrelated bench/bigint files, with 0 errors.

### DIFF
--- a/immut/sorted_map/map_test.mbt
+++ b/immut/sorted_map/map_test.mbt
@@ -13,6 +13,17 @@
 // limitations under the License.
 
 ///|
+priv struct SameRankKey {
+  rank : Int
+  label : String
+} derive(Eq)
+
+///|
+impl Compare for SameRankKey with fn compare(self, other) {
+  self.rank.compare(other.rank)
+}
+
+///|
 test "lookup" {
   let m = @sorted_map.SortedMap([
     (3, "three"),
@@ -46,6 +57,61 @@ test "is_empty" {
   inspect(m.is_empty(), content="true")
   let m = m.add(1, 1)
   inspect(m.is_empty(), content="false")
+}
+
+///|
+test "SortedMap ordered inputs" {
+  let ascending = @sorted_map.SortedMap([(1, "a"), (2, "b"), (3, "c")])
+  let descending = @sorted_map.SortedMap([(3, "c"), (2, "b"), (1, "a")])
+  @test.assert_eq(ascending.to_array(), [(1, "a"), (2, "b"), (3, "c")])
+  @test.assert_eq(descending.to_array(), [(1, "a"), (2, "b"), (3, "c")])
+}
+
+///|
+test "SortedMap ordered duplicate keeps last value" {
+  let m = @sorted_map.SortedMap([(1, "a"), (1, "b"), (2, "c")])
+  @test.assert_eq(m.to_array(), [(1, "b"), (2, "c")])
+}
+
+///|
+test "SortedMap ordered duplicate keeps first key and last value" {
+  let m = @sorted_map.SortedMap([
+    ({ rank: 1, label: "first" }, "a"),
+    ({ rank: 1, label: "second" }, "b"),
+    ({ rank: 2, label: "third" }, "c"),
+  ])
+  @test.assert_eq(m.to_array().map(kv => (kv.0.label, kv.1)), [
+    ("first", "b"),
+    ("third", "c"),
+  ])
+}
+
+///|
+test "SortedMap large unordered duplicate keeps first key and last value" {
+  let items = Array::makei(70, i => {
+    if i == 5 {
+      ({ rank: 10, label: "first" }, "a")
+    } else if i == 50 {
+      ({ rank: 10, label: "second" }, "b")
+    } else {
+      ({ rank: 1000 - i, label: "other-\{i}" }, "other")
+    }
+  })
+  let m = @sorted_map.SortedMap(items)
+  let matched = []
+  for k, v in m {
+    if k.rank == 10 {
+      matched.push((k.label, v))
+    }
+  }
+  @test.assert_eq(matched, [("first", "b")])
+}
+
+///|
+test "from_sorted_array" {
+  let m = @sorted_map.from_sorted_array([(1, "a"), (2, "b"), (3, "c")])
+  @test.assert_eq(m.to_array(), [(1, "a"), (2, "b"), (3, "c")])
+  @test.assert_eq(m.get(2), Some("b"))
 }
 
 ///|
--- a/immut/sorted_map/map_test.mbt
+++ b/immut/sorted_map/map_test.mbt
@@ -367,6 +367,37 @@ test "from_iter empty iter" {
 }
 
 ///|
+test "from_iter descending keys build ascending" {
+  debug_inspect(
+    @sorted_map.from_iter([(3, 3), (2, 2), (1, 1)].iter()),
+    content=(
+      #|<SortedMap: { 1: 1, 2: 2, 3: 3 }>
+    ),
+  )
+}
+
+///|
+test "from_iter descending duplicate keeps first key and last value" {
+  let m = @sorted_map.from_iter(
+    [
+      ({ rank: 2, label: "top" }, "a"),
+      ({ rank: 1, label: "first" }, "b"),
+      ({ rank: 1, label: "second" }, "c"),
+    ].iter(),
+  )
+  @test.assert_eq(m.to_array().map(kv => (kv.0.label, kv.1)), [
+    ("first", "c"),
+    ("top", "a"),
+  ])
+}
+
+///|
+test "from_iter small unsorted duplicate keeps last value" {
+  let m = @sorted_map.from_iter([(3, "a"), (1, "b"), (3, "c"), (2, "d")].iter())
+  @test.assert_eq(m.to_array(), [(1, "b"), (2, "d"), (3, "c")])
+}
+
+///|
 test "hash" {
   let map1 = @sorted_map.SortedMap([("one", 1), ("two", 2)])
   let map2 = @sorted_map.SortedMap([("one", 1), ("two", 2)])
--- a/immut/sorted_map/map_test.mbt
+++ b/immut/sorted_map/map_test.mbt
@@ -310,6 +310,42 @@ test "from_iter multiple elements iter" {
 }
 
 ///|
+test "from_iter duplicate keeps first key and last value" {
+  let m = @sorted_map.from_iter(
+    [
+      ({ rank: 1, label: "first" }, "a"),
+      ({ rank: 1, label: "second" }, "b"),
+      ({ rank: 2, label: "third" }, "c"),
+    ].iter(),
+  )
+  @test.assert_eq(m.to_array().map(kv => (kv.0.label, kv.1)), [
+    ("first", "b"),
+    ("third", "c"),
+  ])
+}
+
+///|
+test "from_iter large unordered duplicate keeps first key and last value" {
+  let items = Array::makei(70, i => {
+    if i == 5 {
+      ({ rank: 10, label: "first" }, "a")
+    } else if i == 50 {
+      ({ rank: 10, label: "second" }, "b")
+    } else {
+      ({ rank: 1000 - i, label: "other-\{i}" }, "other")
+    }
+  })
+  let m = @sorted_map.from_iter(items.iter())
+  let matched = []
+  for k, v in m {
+    if k.rank == 10 {
+      matched.push((k.label, v))
+    }
+  }
+  @test.assert_eq(matched, [("first", "b")])
+}
+
+///|
 test "from_iter single element iter" {
   debug_inspect(
     @sorted_map.from_iter([|(1, 1)|]),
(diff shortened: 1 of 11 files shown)


Yu-zh:
Pushed a few follow-up commits on top of your builders, @mizchi — nice base to work from. Summary of what they do, plus the full benchmark set.

## Commits

1. **sort build entries via `Compare`, not a closure** (`a7e4bffe`) — the unsorted-array build sorted through a `sort_by` comparator closure; a small private `SortEntry` type implementing `Compare` lets it use statically-dispatched `sort` instead. Ordering by `(key, index)` keeps duplicate keys in insertion order under the unstable sort, preserving last-writer-wins.
2. **specialize ascending/descending builders** (`f70cf6ac`) — the ascending monotonic case is identical to the sorted-array builder, so it routes there directly (`from_ascending_array`), paired with a dedicated `from_descending_array`. Drops the per-node `ascending` flag the combined builder re-tested at every level of the O(n) build.
3. **build unsorted maps/sets in place** (`a0f95388`) — dedupe the sorted working buffer in place and build the tree from its unique prefix, dropping the separate compacted array. Sets (elements are `Compare`) `stable_sort` the buffer directly with no wrapper, and `from_iter` reuses `to_array()`'s buffer in place (the borrowed constructor copies once via `to_owned()`); maps sort/dedupe a `FixedArray` scratch buffer in place.

## Benchmarks

Native backend, `n = 10 000`, median of 5 interleaved A/B rounds vs `9cdba99e`. This machine has ~15% run-to-run variance, so anything within a few % is noise.

**SortedMap**

| input | before | after | Δ |
|---|---:|---:|---:|
| `from_sorted_array` | 168 µs | 167 µs | −1% |
| constructor, ordered | 177 µs | 174 µs | −2% |
| constructor, reverse | 152 µs | 149 µs | −2% |
| constructor, almost-ordered | 393 µs | 339 µs | **−14%** |
| constructor, random | 917 µs | 729 µs | **−21%** |
| `from_iter`, ordered | 212 µs | 208 µs | −2% |
| `from_iter`, almost-ordered | 421 µs | 375 µs | **−11%** |
| `from_iter`, random | 949 µs | 766 µs | **−19%** |

**SortedSet**

| input | before | after | Δ |
|---|---:|---:|---:|
| `from_sorted_array` | 180 µs | 179 µs | −1% |
| constructor, ordered | 191 µs | 187 µs | −2% |
| constructor, reverse | 155 µs | 154 µs | −1% |
| constructor, almost-ordered | 315 µs | 204 µs | **−35%** |
| constructor, random | 834 µs | 363 µs | **−56%** |
| `from_iter`, ordered | 207 µs | 203 µs | −2% |
| `from_iter`, almost-ordered | 320 µs | 218 µs | **−32%** |
| `from_iter`, reverse-almost-ordered | 368 µs | 214 µs | **−42%** |
| `from_iter`, random | 831 µs | 379 µs | **−54%** |

Already-sorted / reverse / monotonic inputs are unchanged within noise — the wins land on the unsorted paths, where the in-place buffer work removes the extra allocation. The set gains more because its element is `Compare`: the unsorted path sorts the buffer directly and `from_iter` reuses it in place, while the map keeps an index-tagged `SortEntry` scratch buffer.

bobzhang:
Thanks — the bulk-build work here is a big win, and I'd like to land it. I measured it against
current `main` by porting your benches back (native, n=10000):

| case | main | this PR | |
| --- | ---: | ---: | ---: |
| `SortedMap` ordered | 4.02 ms | 178.69 µs | 22x |
| `SortedMap` reverse | 5.39 ms | 175.28 µs | 31x |
| `SortedMap` random | 3.12 ms | 994.69 µs | 3.1x |
| `SortedSet` ordered | 5.20 ms | 198.16 µs | 26x |
| `SortedSet` `from_iter` rev-almost | 5.23 ms | 243.72 µs | 21x |
| `SortedSet` random | 3.06 ms | 395.17 µs | 7.7x |

One change requested, plus a test addition.

## Please drop `SortedMap::from_sorted_array` and `SortedSet::from_sorted_array`

Violating their precondition silently produces a value that breaks the type's invariant:

```
unsorted [(3,c),(1,a),(2,b)]   -> to_array=[(3,c),(1,a),(2,b)]  length=3
                                  get(1)=Some(a) get(2)=Some(b) get(3)=None  contains(1)=true
duplicates [(1,a),(1,b),(2,c)] -> to_array=[(1,a),(1,b),(2,c)]  length=3  get(1)=Some(b)
   after remove(1)             -> to_array=[(1,a),(2,c)]  get(1)=Some(a)
bad.add(0,"z").add(4,"w")      -> [(0,z),(3,c),(1,a),(2,b),(4,w)]  get(1)=Some(a)
```

Key 3 is in `to_array()` but `get(3)` is `None`; `remove(1)` returns a map that still contains 1.
No abort, no error — and because these are persistent structures the malformed value gets shared
and propagates into everything derived from it. That is a different class of hazard from core's
existing `unsafe_*` helpers, whose failure mode is a local out-of-bounds read.

The cost side is small: `from_array` already detects ascending input in O(n) via
`sorted_map_array_order` and dispatches to the *same* `sorted_map_from_ascending_array`, so
`from_sorted_array` only skips one comparison scan. I measured 170.33 vs 178.69 µs (map) and
185.82 vs 198.16 µs (set) — about 5%, on top of the 20-30x above.

Dropping them also makes both `pkg.generated.mbti` files byte-identical to `main`, so the PR becomes
a pure performance change with no public API delta — much easier to land.

If an unchecked builder is wanted later, it should be a separate PR named `unsafe_from_sorted_array`
per core's `unsafe_` convention.

## Two fixes/additions on top

1. **`SameRankKey` / `SameRankValue` derive `Eq` over both fields while `Compare` orders on `rank`
   alone.** That violates the `Compare : Eq` contract (`builtin/traits.mbt:28`). Changed to a
   rank-only `Eq`. Your assertions keep their strength — they extract `label` and compare it as a
   plain `String`, never through `==` on the fixture.

2. **A quickcheck property for the set's bulk-build dispatch.** Your example tests cover the map
   well, but the set threads `keep_first` through a shared compaction and the combination
   *descending input x duplicates x which representative* was covered by nothing. Flipping that
   representative passes the entire existing suite; the new property is the only thing that fails
   (1/147).

   Note for anyone extending it: quickcheck sizes element `i` by `i`, so a generated `Array[Int]` of
   length >= 2 always starts `[0, 0, ..]` — generated arrays can never reach the duplicate-free
   monotonic builders. The shapes are named explicitly for that reason.

   I checked the property also passes unchanged on `main`, so it pins the pre-existing contract
   rather than blessing the new code.

## Two trade-offs worth noting explicitly (not blockers)

- `from_iter` now materializes the whole iterator (`utils.mbt:634`, `generic.mbt:93`), so live
  auxiliary memory goes from O(unique elements) to O(input length).
- Midpoint construction changes tree shape, so unspecified set collision representatives can differ
  downstream in `union`/`map` (e.g. `[1a,2a] union [1b..5b]` keeps `2b` on `main`, `2a` here). This
  bias was already builder-dependent, so it isn't a regression — but the change is not strictly
  observationally identical.

## Validation

`moon test --target all`: 120/120/120/118 sorted_map, 150/150/150/147 sorted_set — all passing.
`moon check --target all` clean with zero warnings on both. `moon info` shows no `.mbti` change vs
`main`. `moon fmt` clean.

Reviewed by Codex CLI at max reasoning. It cleared the production bulk-build code — no BST, size,
compaction, duplicate-value, or lookup defect in either builder — and its verdict on the follow-up:

> Verdict: approve the delta. No blocking issues remain. All six dispatch leaves are explicitly
> covered [...] the model property runs every shape through both `SortedSet(..)` and `from_iter`.
> [...] Both stated trade-offs are non-blocking.

Happy to open the follow-up as a PR against your branch if that's easier than applying it yourself.

bobzhang:
@mizchi I've pushed the two commits from my review above onto this branch directly (maintainer edits
were enabled) rather than asking you to apply them — they sit on top of your work, no rebase or
force-push, so your commit and authorship are untouched:

- `a0f0b675` drops the two unchecked `from_sorted_array` builders
- `227d234c` adds the sorted_set bulk-build property test and fixes the `Compare : Eq` violation in
  the `SameRank*` fixtures

Both `pkg.generated.mbti` files have now dropped out of the diff entirely, so this is a pure
performance change with no public API delta.

Validation on the pushed head: `moon test --target all` 120/120/120/118 sorted_map and
150/150/150/147 sorted_set, `moon check --target all` clean with zero warnings on both, `moon info`
showing no interface change, `moon fmt` clean.

Please shout if you'd rather I had left these as a suggestion — happy to revert and hand them over
as a patch instead. And thanks again for the bulk-build work; the 20-30x on ordered input is a
substantial win.

bobzhang:
Pushed one more commit (`0cfb3de6`), again on top of your work with no rebase or force-push.

## `from_iter` goes back to streaming

Bulk-building from an iterator needed `iter.to_array()` first, which changed live auxiliary memory
from O(distinct elements) to O(input length). Deduplicating a large stream into a small set — say
10M items with 10 distinct values — went from holding a 10-node tree to holding a 10M-element array.
That's a normal enough pattern that we'd rather not regress it, so both `SortedMap::from_iter` and
`SortedSet::from_iter` fold `add` again, exactly as on `main`.

The array constructors keep the full bulk build. A caller who wants that path and can afford the
buffer writes `SortedSet(iter.to_array())` and opts into the memory explicitly, which seems like the
right place for that decision.

With the iterator entry points folding again, the set's `keep_first` parameter had only one
reachable value, so it's gone from `sorted_set_compact_ordered_array`,
`sorted_set_from_unsorted_buffer` and `sorted_set_dedupe_in_place`, along with the two now
unreachable `sorted_set_from_iter_array*` helpers — about 80 lines lighter.

## Measured (native, n=10000)

| case | main | this branch |
| --- | ---: | ---: |
| `SortedMap` constructor ordered | 4.02 ms | 182.61 µs |
| `SortedMap` constructor random | 3.12 ms | 985.59 µs |
| `SortedSet` constructor ordered | 5.20 ms | 197.63 µs |
| `SortedSet` constructor random | 3.06 ms | 391.65 µs |
| `from_iter`, all shapes | 3.01–5.23 ms | 3.04–5.20 ms |

So the constructor speedup — the bulk of what this PR delivers — is untouched, and `from_iter`
returns to baseline by design.

## Validation

`moon test --target all`: 150/150/150/147 sorted_set, 120/120/120/118 sorted_map. `moon check
--target all` clean with zero warnings on both. `.mbti` still byte-identical to `main`, so the PR
remains a pure performance change. `moon fmt` clean.

Reviewed by Codex CLI at max reasoning, which confirmed the restoration matches `main` exactly, the
`keep_first` removal is complete with no representative flipped, and `stable_sort`/`has_duplicates`
are both still load-bearing for the surviving path:

> Both blockers are resolved. [...] All benchmark figures, package associations, extrema, and ranges
> match the supplied measurements. [...] No remaining issues. Approved.

@mizchi as before, happy to revert any of this if you'd rather take it in a different direction —
the constructor work is yours and it's a strong result.

PR #3744, opened 2026-06-29
[codex] refactor(json): thread parse depth explicitly

bobzhang:
## Summary

Refactor the `json` parser so nesting depth is threaded through the recursive parse calls instead of stored as mutable state on `ParseContext`.

This keeps `ParseContext` focused on lexer/cursor state (`offset`, `input`, `end_offset`) and makes the depth budget explicit in `parse_value`, `parse_value2`, `parse_object`, and `parse_array`.

## JSON package focus

- Removed `remaining_available_depth` from the private `ParseContext` struct.
- Changed `ParseContext::make` to only initialize cursor state.
- Passed `remaining_available_depth` as a labeled private parameter through recursive parser calls.
- Preserved the existing `max_nesting_depth` public API and `DepthLimitExceeded` behavior.

## Validation

- `moon check json`
- `moon test json` (175/175 passed)
- `moon fmt`
- `moon info`
- Confirmed `json/pkg.generated.mbti` is unchanged.
- Ran a fresh `codex review --uncommitted`; no actionable correctness findings were reported.

### DIFF
--- a/json/internal_types.mbt
+++ b/json/internal_types.mbt
@@ -17,20 +17,11 @@ priv struct ParseContext {
   mut offset : Int
   input : StringView
   end_offset : Int
-  mut remaining_available_depth : Int
 }
 
 ///|
-fn ParseContext::make(
-  input : StringView,
-  max_nesting_depth : Int,
-) -> ParseContext {
-  {
-    offset: 0,
-    input,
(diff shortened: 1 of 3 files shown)


bobzhang:
cc @hackwaly

PR #3746, opened 2026-06-29
perf(immut): bulk build hash collections

mizchi:
## What changed

- Add a bulk construction path for immutable `HashMap::from_array` / `from_iter` and `HashSet::from_array` / `from_iter`.
- The builder partitions entries by HAMT path segment and constructs branches from contiguous ranges, avoiding repeated persistent `add` work for larger inputs.
- Avoid the hot `SparseArray::add` path during bulk construction by adding `@immut/internal/sparse_array.from_sorted_fixed_array` and building branch children in one allocation.
- Preallocate `BuildEntry` arrays for `from_array`, and use iterator `size_hint` as capacity for `from_iter`.
- Keep the old `add` loop for small inputs (`<= 64`) to avoid bulk-builder overhead.
- Preserve existing duplicate semantics:
  - `HashMap::from_array` remains first value wins.
  - `HashMap::from_iter` remains last value wins.
  - `HashSet::from_array` / `from_iter` keep their existing duplicate representative behavior.
- Add build benchmarks for HashMap / HashSet.
- Add missing docs for public `bench` clock helpers and JS `BigInt` APIs that were surfaced by `moon check`.

## Profile findings

`moon test --profile` requires `xcrun xctrace`; on this machine I ran it with:

```sh
DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer moon test -p immut/hashmap --target native --profile --filter "profile HashMap*"
DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer moon test -p immut/hashset --target native --profile --filter "profile HashSet*"
```

The temporary profile drivers repeatedly built a 10k-element collection and were not committed.

Before the `SparseArray` change, the profile showed `SparseArray::add` as a real hotspot:

- HashMap: `SparseArray::add` was 17.5% inclusive, mostly `moonbit_make_ref_array_with_blit` / `moonbit_unsafe_ref_array_blit`.
- HashSet: `SparseArray::add` was 18.1% inclusive with the same copy/blit cost.

After constructing sparse arrays from sorted child indices, `SparseArray::add` disappeared from the top profile entries. Samples for the same profile driver dropped:

- HashMap: 1768 -> 1162 samples
- HashSet: 1799 -> 1171 samples

The remaining top cost is now `build_*_node_range` plus allocation/drop overhead while partitioning each HAMT level.

## Benchmarks

`moon bench -p immut/hashmap --target native`

| benchmark | before | after |
|---|---:|---:|
| `HashMap::from_array ordered n=10000` | 1.59 ms | 608.22 us |
| `HashMap::from_array random n=10000` | 1.59 ms | 609.36 us |
| `HashMap::from_iter random n=10000` | 1.61 ms | 624.88 us |

`moon bench -p immut/hashset --target native`

| benchmark | before | after |
|---|---:|---:|
| `HashSet::from_array ordered n=10000` | 1.57 ms | 549.90 us |
| `HashSet::from_array random n=10000` | 1.57 ms | 551.03 us |
| `HashSet::from_iter random n=10000` | 1.58 ms | 562.71 us |

## Validation

- `moon fmt`
- `moon info`
- `moon check -p bench --target all`
- `moon check -p bigint --target all`
- `moon check -p immut/internal/sparse_array --target all`
- `moon check -p immut/hashmap --target all`
- `moon check -p immut/hashset --target all`
- `moon test -p bench --target all`
- `moon test -p bigint --target all`
- `moon test -p immut/internal/sparse_array --target all`
- `moon test -p immut/hashmap --target all`
- `moon test -p immut/hashset --target all`
- `moon bench -p immut/hashmap --target native`
- `moon bench -p immut/hashset --target native`
- `git diff --check`

The missing-doc warnings seen earlier are fixed.

### DIFF
--- a/immut/hashmap/HAMT.mbt
+++ b/immut/hashmap/HAMT.mbt
@@ -228,11 +228,11 @@ fn[K : Eq, V] build_hashmap_node_range(
     let entry = entries[start]
     return Flat(entry.key, entry.value, entry.path.advance(depth))
   }
-  if entries[start].path.is_last_at(depth) {
+  if depth == 5 {
     return build_hashmap_node_by_add(entries, start, end, depth, reverse)
   }
   let starts = FixedArray::make(32, 0)
-  for i = start; i < end; i = i + 1 {
+  for i in start..<end {
     starts[entries[i].path.idx_at(depth)] += 1
   }
   let nexts = FixedArray::make(32, 0)
@@ -244,22 +244,27 @@ fn[K : Eq, V] build_hashmap_node_range(
   }
   let first = entries[start]
   let partitioned = Array::make(end - start, first)
-  for i = start; i < end; i = i + 1 {
+  for i in start..<end {
     let entry = entries[i]
     let idx = entry.path.idx_at(depth)
     let offset = nexts[idx]
     partitioned[offset] = entry
     nexts[idx] = offset + 1
   }
-  let mut child_count = 0
-  for idx = 0; idx < 32; idx = idx + 1 {
-    if nexts[idx] != starts[idx] {
-      child_count += 1
+  let child_count = for idx, next in nexts; count = 0 {
+    if next != starts[idx] {
+      continue count + 1
     }
+    continue count
+  } nobreak {
+    count
   }
-  let mut first_idx = 0
-  while nexts[first_idx] == starts[first_idx] {
-    first_idx += 1
+  let first_idx = for idx, next in nexts {
+    if next != starts[idx] {
+      break idx
+    }
+  } nobreak {
+    abort("unreachable")
   }
   let first_child = build_hashmap_node_range(
     partitioned,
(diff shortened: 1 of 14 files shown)


Yu-zh:
@mizchi thank you for the PR. I pushed the branch rebased onto current `main`, plus a follow-up that keeps each `BuildEntry`'s full path and passes the depth recursively, so partitioning no longer constructs a new entry/path at every HAMT level. In same-session native A/B benchmarks at n=10000 against the rebased initial bulk-builder commit, HashMap ordered/random/from_iter moved from 1.44/1.45/1.49 ms to 1.20/1.22/1.24 ms (about 16–17% faster), while HashSet moved from 1.33/1.35/1.37 ms to 1.14/1.15/1.18 ms (about 14–15% faster).

A possible follow-up is to reuse the per-node partition/count workspaces with ping-pong buffers or range-limited views, removing more temporary allocations. I left that out here because it introduces a recursive interval-confinement invariant and deserves a separate, focused change. Local all-target checks, 181 focused tests, and the full suite pass on the rebased head; the next step is for the newly triggered CI checks to complete.

Yu-zh (approved):

PR #3754, opened 2026-06-30
hongbo/warning as error

bobzhang:
- **Add BigInt empty octets tests**
- **enable warnings as errors**

### DIFF
--- a/moon.mod
+++ b/moon.mod
@@ -10,4 +10,4 @@ license = "Apache-2.0"
 
 keywords = [ "core", "standard library" ]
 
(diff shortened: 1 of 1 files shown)


bobzhang:
cc @peter-jerry-ye 

the warning looks like from test code? shall we always disable warnings there?

peter-jerry-ye:
missing-doc is applied to the test template. Currently we do not distinguish the template code from others, which makes it a bit awkward when new warnings are added and enabled.

bobzhang:
we will revisit if `@all` makes sense

PR #3759, opened 2026-06-30
feat(immut/hashmap): add HashMap constructor, deprecate from_array

bobzhang:
## Summary

Provide a `@hashmap.HashMap([...])` constructor for the immutable hash map (matching `@list.List([...])` / `@sorted_set.SortedSet([...])`), and deprecate `from_array`.

`HashMap` was a positional newtype `struct HashMap(Node?)`, whose **tuple constructor reserved the `HashMap` name** — a `HashMap::HashMap` function collided with it. Give the field a name instead:

```moonbit
struct HashMap[K, V] {
  data : Node[K, V]?
}
```

- Frees the `HashMap` identifier (named structs construct via `{ data: .. }`, not `HashMap(..)`), so `HashMap::HashMap(...)` is allowed.
- Keeps the **exact same `Node?` representation** — internal logic is unchanged apart from `self.0` → `self.data` and wrapping results in `{ data: .. }`. No enum, no `SparseArray` changes, no per-node conversions.
- The public type stays **abstract** (the field is private), so the `.mbti` is unchanged apart from the new constructor.

Then: add `HashMap::HashMap(ArrayView[(K, V)]) -> HashMap[K, V]`, deprecate `from_array` (+ its `of` alias), and migrate all usages across impl/tests/README. The `Show`/`output` format and its snapshot tests intentionally keep emitting `@immut/hashmap.from_array([...])`.

## Verification

- `moon check --deny-warn --target all` — 0 warnings
- `moon test` — all passing (default target)
- `immut/hashmap` green on wasm / wasm-gc / js / native (110/110 each)
- `moon fmt` and `moon info` produce no additional diff

### DIFF
--- a/immut/hashmap/HAMT.mbt
+++ b/immut/hashmap/HAMT.mbt
@@ -37,14 +37,14 @@ priv enum CurrNode[K, V] {
 /// Create a new instance.
 #as_free_fn
 pub fn[K, V] HashMap::new() -> HashMap[K, V] {
-  None
+  { data: None }
 }
 
 ///|
 /// Create a map with a single key-value pair.
 #as_free_fn
 pub fn[K : Hash, V] HashMap::singleton(key : K, value : V) -> HashMap[K, V] {
-  Some(Flat(key, value, @path.of(key)))
+  { data: Some(Flat(key, value, @path.of(key))) }
 }
 
 ///|
@@ -60,7 +60,7 @@ pub fn[K : Eq + Hash, V] HashMap::contains(
 /// Lookup a key from a hash map
 #alias(find, deprecated)
 pub fn[K : Eq + Hash, V] HashMap::get(self : HashMap[K, V], key : K) -> V? {
-  match self.0 {
+  match self.data {
     None => None
     Some(node) => node.get_with_path(key, @path.of(key))
   }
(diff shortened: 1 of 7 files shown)


immut/hashmap/HAMT.mbt
```diff
@@ -200,9 +200,13 @@ pub fn[K, V] HashMap::filter(
     }
   }
 
-  match self.0 {
-    None => None
-    Some(node) => go(node)
+  match self {
+    MapEmpty => MapEmpty
+    MapSome(node) =>
```
bobzhang: change `go` to return MapEmpty directly?

immut/hashmap/HAMT.mbt
```diff
@@ -275,9 +279,13 @@ pub fn[K : Eq + Hash, V] HashMap::remove(
   self : HashMap[K, V],
   key : K,
 ) -> HashMap[K, V] {
-  match self.0 {
-    None => None
-    Some(node) => node.remove_with_path(key, @path.of(key))
+  match self {
+    MapEmpty => MapEmpty
+    MapSome(node) =>
+      match node.remove_with_path(key, @path.of(key)) {
```
bobzhang: remove_with_path return MapEmpty?

immut/hashmap/HAMT.mbt
```diff
@@ -491,9 +499,13 @@ pub fn[K : Eq, V] HashMap::intersection(
     }
   }
 
-  match (self.0, other.0) {
-    (None, _) | (_, None) => None
-    (Some(a), Some(b)) => go(a, b)
+  match (self, other) {
+    (MapEmpty, _) | (_, MapEmpty) => MapEmpty
+    (MapSome(a), MapSome(b)) =>
+      match go(a, b) {
```
bobzhang: go return MapEmpty?

PR #3772, opened 2026-07-03
Optimize Bytes rev_find

mizchi:
## Summary
- handle `BytesView::rev_find` edge cases before entering the scan
- check the first and last bytes before comparing the remaining range
- add tests and native benchmarks for worst-case reverse search

## Benchmarks
Measured with native release benches. Before is `upstream/main@1c53c811` with this PR's benchmark file applied; after is this PR head `a023cece`.

| --- | ---: | ---: | ---: |
| Bytes::rev_find short needle at front | 52.63 us | 30.79 us | 1.71x faster (-41.5%) |
| Bytes::rev_find long needle at front | 52.49 us | 30.55 us | 1.72x faster (-41.8%) |

Time Profiler traces from the generated native bench executable showed the workload was almost entirely in `BytesView::rev_find`.

## Validation
- moon fmt
- moon info
- moon test -p moonbitlang/core/builtin --target all
- moon check --target all
- moon bench -p moonbitlang/core/builtin -f bytes_find_bench_test.mbt --target native --release

### DIFF
--- a/builtin/bytes_find.mbt
+++ b/builtin/bytes_find.mbt
@@ -360,14 +360,34 @@ fn find_long_by_byte_scanner(target : BytesView, pattern : BytesView) -> Int? {
 /// Returns the offset of the last occurrence of the given
 /// bytes substring. If the substring is not found, `None` is returned.
 pub fn BytesView::rev_find(target : BytesView, pattern : BytesView) -> Int? {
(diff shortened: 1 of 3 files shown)


mizchi:
Superseded by #3775 after splitting the original combined performance PR into focused changes.

PR #3777, opened 2026-07-03
Optimize aligned bitstring integer extraction

mizchi:
## Summary
- add exact-width byte-aligned fast paths to generic bitstring integer extraction
- cover `ArrayView[Byte]` and `BytesView` for 16/32-bit UInt and 64-bit UInt64 extraction
- add focused native benchmarks for aligned generic `unsafe_extract_uint*` calls

## Benchmarks
Measured with native release benches. Before is `upstream/main@1c53c811` with an equivalent benchmark file applied; after is this PR head `7d8649c9`.

| --- | ---: | ---: | ---: |
| `BytesView::unsafe_extract_uint_le aligned len=32` | 3.34 us | 2.07 us | 1.61x faster (-38.0%) |
| `ArrayView::unsafe_extract_uint_le aligned len=32` | 2.77 us | 2.25 us | 1.23x faster (-18.8%) |
| `BytesView::unsafe_extract_uint_be aligned len=32` | 3.18 us | 2.41 us | 1.32x faster (-24.2%) |
| `ArrayView::unsafe_extract_uint_be aligned len=32` | 3.71 us | 2.11 us | 1.76x faster (-43.1%) |
| `BytesView::unsafe_extract_uint64_le aligned len=64` | 1.89 us | 1.20 us | 1.57x faster (-36.5%) |
| `BytesView::unsafe_extract_uint64_be aligned len=64` | 2.17 us | 1.51 us | 1.44x faster (-30.4%) |
| `ArrayView::unsafe_extract_uint64_be aligned len=64` | 1.99 us | 1.45 us | 1.37x faster (-27.1%) |
| `ArrayView::unsafe_extract_uint64_le aligned len=64` | 1.93 us | 1.47 us | 1.31x faster (-23.8%) |

## Validation
- `moon fmt`
- `moon info`
- `moon test -p moonbitlang/core/builtin --target all`
- `moon check --target all`
- `moon bench -p moonbitlang/core/builtin -f bitstring_generic_aligned_bench_test.mbt --target native --release`

### DIFF
--- a/builtin/bitstring.mbt
+++ b/builtin/bitstring.mbt
@@ -108,8 +108,16 @@ pub fn ArrayView::unsafe_extract_uint_le(
   offset : Int,
   len : Int,
 ) -> UInt {
+  if (offset & 7) == 0 {
+    let byte_offset = offset >> 3
+    if len == 16 {
+      return bs.unsafe_extract_uint16_le_aligned(byte_offset)
+    }
+    if len == 32 {
+      return bs.unsafe_extract_uint_le_aligned(byte_offset)
+    }
+  }
   let bytes_needed = (len + 7) / 8
-  // TODO: add fast path for aligned case
   // non-aligned case: extract bytes using unsafe_extract_byte
   let b0 = bs.unsafe_extract_byte(offset, 8)
   match bytes_needed {
@@ -146,8 +154,16 @@ pub fn ArrayView::unsafe_extract_uint_be(
   offset : Int,
   len : Int,
 ) -> UInt {
+  if (offset & 7) == 0 {
+    let byte_offset = offset >> 3
+    if len == 16 {
+      return bs.unsafe_extract_uint16_be_aligned(byte_offset)
+    }
+    if len == 32 {
+      return bs.unsafe_extract_uint_be_aligned(byte_offset)
+    }
+  }
   let bytes_needed = (len + 7) / 8
-  // TODO: add fast path for aligned case
   // non-aligned case: extract bytes using unsafe_extract_byte
   let b0 = bs.unsafe_extract_byte(offset, 8)
   match bytes_needed {
@@ -191,8 +207,10 @@ pub fn ArrayView::unsafe_extract_uint64_le(
   offset : Int,
   len : Int,
 ) -> UInt64 {
+  if (offset & 7) == 0 && len == 64 {
+    return bs.unsafe_extract_uint64_le_aligned(offset >> 3)
+  }
   let bytes_needed = (len + 7) / 8
-  // TODO: add fast path for aligned case
   // non-aligned case: extract bytes using unsafe_extract_byte
   let b0 = bs.unsafe_extract_byte(offset, 8).to_uint64()
   let b1 = bs.unsafe_extract_byte(offset + 8, 8).to_uint64()
@@ -253,8 +271,10 @@ pub fn ArrayView::unsafe_extract_uint64_be(
   offset : Int,
   len : Int,
 ) -> UInt64 {
+  if (offset & 7) == 0 && len == 64 {
+    return bs.unsafe_extract_uint64_be_aligned(offset >> 3)
+  }
   let bytes_needed = (len + 7) / 8
-  // TODO: add fast path for aligned case
   // non-aligned case: extract bytes using unsafe_extract_byte
   let b0 = bs.unsafe_extract_byte(offset, 8).to_uint64()
   let b1 = bs.unsafe_extract_byte(offset + 8, 8).to_uint64()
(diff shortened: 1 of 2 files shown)


bobzhang:
## Review: right direction, but the guard shape regresses native u32 on current main — an entry/slow split fixes both

Thanks — the direction is right and the js/wasm-gc payoff is real and large. But the PR's benchmark table was measured against `upstream/main@1c53c811`, which is now stale (main has since gained the Bytes aligned-load optimizations, and toolchain codegen has moved). Re-measured on **current main** with this PR's own bench file (same machine, `moon bench --release`, `BytesView` LE rows shown; all rows behave alike):

| | current main | this PR | entry/slow split (see below) |
|---|---:|---:|---:|
| **native** u32 | 4.42 µs | **5.27 µs (+19% regression)** | **1.90 µs** (2.3x vs main) |
| **native** u64 | 3.07 µs | 2.69 µs (−12%) | **0.96 µs** (3.2x vs main) |
| **js** u32 | 24.08 µs | 4.14 µs (5.8x) | **3.99 µs** |
| **js** u64 | 115.8 µs | 93.9 µs | **92.2 µs** |
| **wasm-gc** u32 | 14.57 µs | 5.90 µs (2.5x) | **3.94 µs** |
| **wasm-gc** u64 | 12.43 µs | 4.18 µs (3.0x) | **3.78 µs** |

The native u32 regression is reproducible across repeat runs (σ ≈ 2%, delta 17–26% on all four u32 functions).

### Diagnosis: an inlining cliff, not the guards' arithmetic

Three experiments isolate it:
- Calling the aligned helper **unconditionally** runs at 1.94 µs — the helper is fast, and so is the guard arithmetic.
- **Reordering** the guards (`len == 32` first) doesn't help (5.4 µs).
- **Outlining the generic tail** into a separate slow function — leaving a tiny entry with just the fast-path guard — runs at 1.90 µs with all 25 extract tests passing.

Adding guards on top of the large generic body pushes the function past the native inliner's threshold; every call then pays a real call into the whole function. On current main the guard-free generic body still inlines (and LLVM fuses the byte-combine), which is why the PR's shape manages to be slower than no fast path at all on native u32.

### Recommended shape (measured strictly better than both main and this PR on all three backends)

```moonbit
pub fn BytesView::unsafe_extract_uint_le(
  bs : BytesView,
  offset : Int,
  len : Int,
) -> UInt {
  if len == 32 && (offset & 7) == 0 {
    return bs.unsafe_extract_uint_le_aligned(offset >> 3)
  }
  bitstring_bytesview_extract_uint_le_slow(bs, offset, len)
}

///|
fn bitstring_bytesview_extract_uint_le_slow(
  bs : BytesView,
  offset : Int,
  len : Int,
) -> UInt {
  if len == 16 && (offset & 7) == 0 {
    return bs.unsafe_extract_uint16_le_aligned(offset >> 3)
  }
  // ... existing generic byte-by-byte tail, unchanged ...
}
```

Same split for all eight functions (the u64 entries keep their single guard). The unaligned case is not pessimized relative to this PR — with the guards added, it already paid a non-inlined call; the split just makes that explicit while restoring the fast path's inlining.

### Codex CLI review (`xhigh`)

Correctness of the fast paths is clean: guard selection, `offset >> 3` conversion, endianness against the former byte assembly, `bs.start()` handling on `ArrayView`, unchanged fall-through for non-special widths/unaligned offsets, and the compiler-lowering claim (fixed-position patterns already use aligned reads; dynamic-offset patterns hit these generic entries) were all verified. Verdict:

> **[P1]** The submitted shape should not merge because the added guards make the four u32 generic functions too large to inline on native, producing a repeatable 17–26% regression. [...] Please outline the generic tails and re-baseline against current `main`.
>
> Re-baselining should also cover an unaligned offset, an aligned non-special width, and aligned 16-bit extraction. The committed benchmark contains only 32- and 64-bit cases, so those slow-path tradeoffs remain unmeasured.
>
> One diagnostic wording correction: on current JS the aligned intrinsic uses the straight-line MoonBit fallback, not literally a word load. The JS gain is therefore from bypassing repeated generic extraction/BigInt machinery; the word-load explanation applies to backends that lower the intrinsic accordingly.
>
> Review-by: Codex CLI <codex@openai.com>

### Summary of asks

1. Outline the generic tails (entry/slow split, shape above) for all eight functions.
2. Re-baseline the benchmark table against current main.
3. Extend the bench file with unaligned, aligned-16-bit, and non-special-width cases so the slow path's cost is visible.

With the split shape this PR becomes a clear win on every backend — native 2.3–3.2x, js up to 6x, wasm-gc 3–3.7x over current main.

*Benchmarks and review by Claude Code on behalf of @bobzhang; adversarial verification by Codex CLI.*

bobzhang:
Following up on the review above: the entry/slow-split shape is now implemented, benchmarked on all four backends, and Codex-verified in #4114 (superseding this PR, with credit — the aligned-delegation direction and workload shape are yours). Headline vs current main: native u32 2.7x / u64 3.2x, wasm 2–3.3x, wasm-gc 2.8–3.2x, js up to 12x on aligned extraction, with unaligned paths at par. If #4114 lands, this PR can be closed in its favor — or feel free to fold the split shape back into this branch instead.

bobzhang:
landed in #4114

PR #3781, opened 2026-07-03
Optimize StringView code_units for small slices

mizchi:
## Summary
- optimize `StringView::code_units` for small views over large backing strings
- allocate only the view length when the view is less than half of the backing string
- keep the existing whole-backing copy path for large views to preserve current throughput
- add focused `StringView::code_units` benchmarks

## Context
This follows the dependency hotspot analysis:
https://gist.github.com/mizchi/28f382dd0d9c7c1cb0ef42b202f7f24e

`StringView` is in the high-priority `builtin` area and appears frequently in string, regex, JSON, and parsing paths. The previous fallback implementation copied the entire backing string and then sliced it, which is disproportionately expensive for a small view into a large string.

## Benchmarks
Measured with native release benches. Before is `upstream/main@607c2745`; after is this PR head `74774ede`.

| --- | ---: | ---: | ---: |
| `StringView::code_units small slice of large backing n=16/100000` | 5.09 us | 17.02 ns | 299x faster (-99.7%) |
| `StringView::code_units full slice n=100000` | 4.94 us | 5.03 us | roughly unchanged |

I also tried caching `str()` / `start()` in `StringView` comparison loops, but dropped it because it regressed `compare` and ASCII folded comparisons on native release benches.

## Validation
- `moon fmt`
- `moon info`
- `moon test -p moonbitlang/core/builtin --target all`
- `moon check --target all`
- `moon bench -p moonbitlang/core/builtin -f stringview_code_units_bench_test.mbt --target native --release`

### DIFF
--- a/builtin/stringview.mbt
+++ b/builtin/stringview.mbt
@@ -137,13 +137,20 @@ pub fn StringView::unsafe_get(self : StringView, index : Int) -> UInt16 {
 /// This method yields code units, not Unicode characters. Surrogate pairs are
 /// represented as two `UInt16` values.
 ///
-/// Note: the default implementation copies the entire underlying string into a
-/// fresh `FixedArray` and then slices it, so the returned view does not alias
-/// the original string's storage. Some backends lower the intrinsic to a
-/// zero-copy view instead.
+/// Note: the default implementation copies code units into a fresh
+/// `FixedArray`, so the returned view does not alias the original string's
+/// storage. Some backends lower the intrinsic to a zero-copy view instead.
 #intrinsic("%stringview.code_units")
 pub fn StringView::code_units(self : StringView) -> ArrayView[UInt16] {
-  FixedArray::makei(self.str().length(), i => self.str().unsafe_get(i))[self.start():self.end()]
+  let len = self.length()
+  let str_len = self.str().length()
+  if len * 2 < str_len {
+    let str = self.str()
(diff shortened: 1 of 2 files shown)


builtin/stringview.mbt
```diff
@@ -137,13 +137,20 @@ pub fn StringView::unsafe_get(self : StringView, index : Int) -> UInt16 {
...
+/// storage. Some backends lower the intrinsic to a zero-copy view instead.
 #intrinsic("%stringview.code_units")
 pub fn StringView::code_units(self : StringView) -> ArrayView[UInt16] {
-  FixedArray::makei(self.str().length(), i => self.str().unsafe_get(i))[self.start():self.end()]
+  let len = self.length()
+  let str_len = self.str().length()
+  if len * 2 < str_len {
+    let str = self.str()
+    let start = self.start()
+    FixedArray::makei(len, i => str.unsafe_get(start + i))
```
Yu-zh, line 150: The start offset of arrayview created this way will be 0, which is different from the intrinsic implementation in the compiler.

Yu-zh (commented):
The intrinsic implementation in the compiler maintains the start and end offset of the original `StringView`

Yu-zh:
Closing this PR because of unexpected behavioral change. Feel free to reopen it if I misunderstood it.

PR #3794, opened 2026-07-05
perf(deque): use blit for to_array and contiguous-run loops for Eq/Compare

choplin:
This PR speeds up `Deque::to_array`, `Eq`, and `Compare` by avoiding per-element bounds-checked access through `Deque::at`.

## Changes

- **`to_array`**: previously filled the whole result with the front element via `Array::make`, then overwrote every slot through the bounds-checked `self[i]` (two writes per slot, plus a range check and a wrap-around branch per read). It now builds the result from the two ring-buffer segments returned by `as_views()` using `Array::append` (blit).
- **`Eq` / `Compare`**: previously walked both deques through `self[i]` / `other[i]`, paying a range check and a wrap-around branch per element per deque. They now walk the underlying buffers in contiguous runs — at most three per comparison — so the inner loops carry no wrap-around branches.

`moon info` reports no `.mbti` changes, so the external interface is unchanged.

## Benchmarks

`deque/conv_bench_test.mbt` (included in this PR), n=10000, Apple Silicon (M-series). "wrapped" means the deque's elements wrap around the internal ring buffer boundary.

### wasm-gc

| bench | before | after | speedup |
|---|---|---|---|
| to_array contiguous | 19.96 µs | 1.33 µs | 15.0x |
| to_array wrapped | 19.11 µs | 2.02 µs | 9.5x |
| op_equal contiguous | 29.61 µs | 4.73 µs | 6.3x |
| op_equal wrapped | 28.08 µs | 4.71 µs | 6.0x |
| compare wrapped | 19.33 µs | 5.87 µs | 3.3x |

### native

| bench | before | after | speedup |
|---|---|---|---|
| to_array contiguous | 7.55 µs | 0.49 µs | 15.4x |
| to_array wrapped | 7.64 µs | 0.71 µs | 10.8x |
| op_equal contiguous | 6.91 µs | 4.71 µs | 1.5x |
| op_equal wrapped | 6.72 µs | 4.95 µs | 1.4x |
| compare wrapped | 7.15 µs | 6.16 µs | 1.2x |

A note on the Eq/Compare approach: a simpler variant that keeps two cursors and wraps each with a per-element branch was ~2.9x faster on wasm-gc but 12–40% *slower* than the current code on native, so this PR uses the contiguous-run formulation, which is faster than the baseline on both targets.

## Checks

- `moon test` (wasm-gc, all packages), `moon test --target native/js -p moonbitlang/core/deque`: all pass
- `moon check` / `moon fmt` / `moon info`: clean, no `.mbti` diff

### DIFF
--- a/deque/conv_bench_test.mbt
+++ b/deque/conv_bench_test.mbt
@@ -1,0 +1,75 @@
+// Copyright 2026 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+///|
+let conv_bench_size = 10000
+
+///|
+fn make_contiguous_deque(len : Int) -> @deque.Deque[Int] {
+  let dq = @deque.Deque([], capacity=len)
+  for i in 0..<len {
+    dq.push_back(i)
+  }
+  dq
+}
+
+///|
+/// A deque holding the same elements as `make_contiguous_deque`, but whose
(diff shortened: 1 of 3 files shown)


choplin:
Sorry. I'm closing this PR for now since  I'm going to split it into two focused PRs for easier review — one for the `to_array` blit optimization, and one for the `Eq`/`Compare` contiguous-run loops.

PR #3797, opened 2026-07-07
test(argparse): split rejected config cases

peter-jerry-ye:
## Why

The argparse tests mixed command-definition validation failures with blackbox argv parsing behavior. That makes it harder to tell which cases are parser runtime errors and which are configuration invariants that protect the command model from ambiguity.

## What

This PR adds `rejected_config_test.mbt` for command definitions rejected before meaningful argv parsing. The moved cases now include comments explaining the invariant behind each rejection and the ambiguity or broken behavior it prevents. Runtime argv parse failures remain in the existing blackbox and coverage tests.

It also adds explicit coverage for a hidden `default_subcommand` target, matching the existing rule that a default subcommand must name a visible child.

## Correctness

The parser and public API are unchanged. Existing validation assertions are preserved under the new test file, and the new hidden-default case exercises the same validation path as the visible-target rule.

## Minimality

This is intentionally test-only: it reorganizes validation coverage without changing argparse behavior or generated interfaces.

### DIFF
--- a/argparse/argparse_blackbox_test.mbt
+++ b/argparse/argparse_blackbox_test.mbt
@@ -494,104 +494,6 @@ test "default subcommand help annotation and child error context" {
 }
 
 ///|
-test "default subcommand validation rejects ambiguous root configuration" {
-  try
-    @argparse.Command(
-      "demo",
-      subcommands=[Command("run")],
-      default_subcommand="missing",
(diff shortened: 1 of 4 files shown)


bobzhang:
@peter-jerry-ye ping?

PR #3806, opened 2026-07-13
perf(string): avoid iterator allocation in to_lower

Yu-zh:
## Summary

- keep the allocation-free early return when a string contains no ASCII uppercase code unit
- scan eight UTF-16 code units at a time with SIMD on native and linear-memory Wasm
- copy changed inputs once into an exact-size UTF-16 buffer, then lowercase them in place
- lowercase eight code units at a time on native/Wasm, use scalar buffer patching on wasm-gc, and retain the pre-optimization character-loop fallback on JavaScript
- preserve non-ASCII code units, non-BMP characters, unpaired surrogate halves, and `StringView` boundaries exactly
- add boundary, sparse-uppercase, one-unit-run, all-uppercase, late-uppercase, and no-change benchmarks

> [!NOTE]
> The branch is rebased onto `0b89deea` and uses the released Moon `0.1.20260807` / `moonc v0.10.7+bc794d341` toolchain. Commit `0682655d` restores the pre-optimization JavaScript path after verification showed that chunking one-code-unit unchanged runs allocated a tiny substring per run and caused the reported regression.

## Implementation

`find_ascii_uppercase_code_unit` keeps strings shorter than 24 code units on the scalar path. For longer inputs it first checks the initial code unit scalarly, preserving the common capitalized-string fast path, then scans complete eight-code-unit blocks with `i16x8` operations. If a block matches, a scalar pass over that block locates the first uppercase code unit; any remaining tail is also scalar.

Once a change is required:

- **native and linear Wasm** copy the entire view into an exact-size `FixedArray[UInt16]`, lowercase full blocks with SIMD, and patch the scalar tail
- **wasm-gc and other non-linear-memory backends** use the same exact-size copy with scalar patching
- **JavaScript** retains the original `find_by` plus character-loop `StringBuilder` fallback; JavaScript strings cannot use the in-place UTF-16 buffer path

If the pre-scan finds no uppercase code unit, `to_lower` returns the original string or view without allocating.

## Why

PR #3635 removes the per-call `Iter` allocation and batches unchanged runs, but conversion still crosses the builder API for each changed unit/run. Since `to_lower` already works at the UTF-16 level, copying the input once and patching that buffer removes this per-run overhead and exposes a straightforward SIMD loop.

The pre-scan is also worth retaining: lowercase inputs return `self` without allocating. SIMD makes that extra pass inexpensive for long strings on native and linear Wasm.

## Impact

Measured on an Apple M3 Max running macOS 26.2 with Moon `0.1.20260724` and `moonc v0.10.5+5e7afb0c0`.

Both revisions used the same `builtin/string_to_lower_bench_test.mbt` and were run with:

```sh
moon bench --release --target native builtin/string_to_lower_bench_test.mbt
moon bench --release --target wasm builtin/string_to_lower_bench_test.mbt
moon bench --release --target wasm-gc builtin/string_to_lower_bench_test.mbt
moon bench --release --target js builtin/string_to_lower_bench_test.mbt
```

The protocol used three measured invocations per revision and target, alternating which revision ran first so drift affects both sides evenly. Each table reports the median of each benchmark's reported mean. Lower is better; Δ is `HEAD / baseline - 1`. These historical tables were captured at `e62403e5`; their JavaScript rows document the regression that is superseded by the fallback verification below.

### 1. HEAD against main

Baseline is this PR's merge base, `8720b4b3`.

| target | case | main | HEAD | Δ |
|---|---|---:|---:|---:|
| native | short | 60.80 ns | 19.67 ns | −67.6% faster |
| native | mixed long | 1.47 µs | 85.65 ns | −94.2% faster |
| native | late uppercase | 851.20 ns | 95.32 ns | −88.8% faster |
| native | one-unit runs | 1.20 µs | 85.42 ns | −92.9% faster |
| native | no-change long | 777.93 ns | 58.23 ns | −92.5% faster |
| wasm | short | 175.55 ns | 59.92 ns | −65.9% faster |
| wasm | mixed long | 3.11 µs | 492.34 ns | −84.2% faster |
| wasm | late uppercase | 2.68 µs | 510.36 ns | −81.0% faster |
| wasm | one-unit runs | 3.33 µs | 496.23 ns | −85.1% faster |
| wasm | no-change long | 2.54 µs | 95.51 ns | −96.2% faster |
| wasm-gc | short | 62.51 ns | 28.41 ns | −54.6% faster |
| wasm-gc | mixed long | 2.06 µs | 1.10 µs | −46.6% faster |
| wasm-gc | late uppercase | 1.81 µs | 882.16 ns | −51.3% faster |
| wasm-gc | one-unit runs | 1.94 µs | 1.01 µs | −47.9% faster |
| wasm-gc | no-change long | 1.44 µs | 448.06 ns | −68.9% faster |
| js | short | 111.19 ns | 64.89 ns | −41.6% faster |
| js | mixed long | 5.57 µs | 4.54 µs | −18.5% faster |
| js | late uppercase | 1.76 µs | 1.56 µs | −11.4% faster |
| js | one-unit runs | 5.15 µs | 6.52 µs | +26.6% slower |
| js | no-change long | 1.87 µs | 1.71 µs | −8.6% faster |

Native, linear Wasm, and wasm-gc improve on every case. The historical JavaScript one-unit-run regression came from allocating a `StringView` and owned substring for each one-code-unit unchanged run. Commit `0682655d` removes JavaScript from that chunked path and restores the original character-loop implementation.

Fresh alternating runs on the current toolchain put JavaScript `one-unit runs` at 3.88–4.05 µs on the fallback versus 4.82 µs on the chunked PR revision, recovering about 18–20%. Direct comparison with current `main` (3.64–4.00 µs) is within run-to-run noise, as expected from using the same implementation.

### 2. HEAD against #3635

PR #3635 targets the same function. To keep the comparison to the two implementations, its branch (`b1f5585d`) was rebased onto the same `8720b4b3` base used above; it rebased cleanly and its production diff is unchanged at 82 insertions / 24 deletions.

One test-only edit was required to make that baseline runnable: #3635 asserts on `"x🤣ABy"[1:-1]`, and `String::sub` on current `main` rejects negative indices outright (`guard start >= 0 && start <= end && end <= len`), so the expression panics before `to_lower` is reached. Changing it to `[1:5]` — the same form this PR uses — makes #3635 pass 2881/2881 on native. No production code was touched.

| target | case | #3635 | this PR | Δ |
|---|---|---:|---:|---:|
| native | short | 60.21 ns | 19.83 ns | −67.1% faster |
| native | mixed long | 1.41 µs | 85.94 ns | −93.9% faster |
| native | late uppercase | 310.67 ns | 97.88 ns | −68.5% faster |
| native | one-unit runs | 2.09 µs | 86.26 ns | −95.9% faster |
| native | no-change long | 212.65 ns | 58.56 ns | −72.5% faster |
| wasm | short | 155.37 ns | 59.90 ns | −61.4% faster |
| wasm | mixed long | 2.91 µs | 500.06 ns | −82.8% faster |
| wasm | late uppercase | 976.17 ns | 514.89 ns | −47.3% faster |
| wasm | one-unit runs | 5.16 µs | 496.41 ns | −90.4% faster |
| wasm | no-change long | 477.06 ns | 95.56 ns | −80.0% faster |
| wasm-gc | short | 60.11 ns | 29.02 ns | −51.7% faster |
| wasm-gc | mixed long | 2.19 µs | 1.09 µs | −50.2% faster |
| wasm-gc | late uppercase | 911.22 ns | 884.10 ns | −3.0% faster |
| wasm-gc | one-unit runs | 3.48 µs | 1.03 µs | −70.4% faster |
| wasm-gc | no-change long | 452.06 ns | 461.15 ns | +2.0% slower |
| js | short | 72.71 ns | 66.55 ns | −8.5% faster |
| js | mixed long | 5.45 µs | 4.64 µs | −14.9% faster |
| js | late uppercase | 1.58 µs | 1.60 µs | +1.3% slower |
| js | one-unit runs | 8.44 µs | 6.57 µs | −22.2% faster |
| js | no-change long | 1.72 µs | 1.75 µs | +1.7% slower |

Across the full 11-benchmark matrix on all four targets, this PR is faster on 40 of 44 measurements. The four exceptions — wasm-gc `no-change long` (+2.0%) and `StringView no-change long` (+2.5%), and JavaScript `late uppercase` (+1.3%) and `no-change long` (+1.7%) — all have per-round ranges that overlap the baseline's, so they are ties rather than regressions. Every case where the two implementations genuinely diverge favours this PR.

The JavaScript rows in this historical #3635 table predate `0682655d`; JavaScript now uses the same fallback as `main`, so they are not claims about the current head.

PR #3635's published end-to-end native profile reported 12.3% fewer allocations and 3.0% fewer allocated bytes than the pre-change implementation. Those allocation measurements were not rerun here, so they remain context from #3635 rather than fresh measurements for this PR.

## Validation

The latest checks were run on `0682655d` with Moon `0.1.20260807` and `moonc v0.10.7+bc794d341`:

- `moon check --target all --warn-list +73` — clean
- targeted `builtin/string_methods.mbt` tests on native, Wasm, Wasm-GC, and JavaScript — 63/63 on each target
- `moon test` — 7018/7018
- `moon info && moon fmt` — no interface changes

The earlier release-toolchain validation below was run before the JavaScript-only fallback commit; its native/Wasm/Wasm-GC results remain representative.

- `moon check --deny-warn --target all` — clean
- `moon test -p moonbitlang/core/builtin --target <t>`
  - native: 2882/2882
  - wasm: 2924/2924
  - wasm-gc: 2924/2924
  - JavaScript: 2909/2909
- `moon test` (full repository suite, default target): 6897/6897
- `moon info --target wasm,wasm-gc,js,native` — no `.mbti` changes
- `moon fmt` — no diff
- native codegen inspection of `ascii_lowercase_copy`: `ldr q3` / `add.8h` / `cmhi.8h` / `and.16b` / `add.8h` / `str q3`, and of `find_ascii_uppercase_code_unit_simd`: `ldr q2` / `add.8h` / `cmhi.8h` / `umaxv.16b`
- Wasm codegen inspection: `v128.load`, `v128.store`, `i16x8.splat`, `i16x8.sub`, `i16x8.add`, `i16x8.le_u`, `v128.and`, and `v128.any_true` all present in the emitted module

### Rebase note

The rebase onto `8720b4b3` produced one conflict, in the `builtin/simd.mbt` header comment. `main` had independently un-gated `v128_make` and `v128_load` for aligned bitstring extraction and rewritten that paragraph; this branch had rewritten the same paragraph to cover string routines as well as byte scanners. The resolution keeps both facts. No executable code conflicted.

### DIFF
--- a/builtin/string_methods.mbt
+++ b/builtin/string_methods.mbt
@@ -1738,20 +1738,21 @@ test "View::replace_all boundary cases" {
 
 ///|
 /// Helpers for ASCII-only lowercase conversion over UTF-16 code units.
-/// Non-ASCII code units, including surrogate halves, are copied with
-/// `write_substring` and are never converted through `Char`.
+/// Non-ASCII code units, including surrogate halves, are copied unchanged and
+/// are never converted through `Char`.
 fn is_ascii_uppercase_code_unit(code_unit : Int) -> Bool {
   code_unit >= 0x41 && code_unit <= 0x5A
 }
 
 ///|
+#cfg(not(any(target="native", target="wasm")))
 fn find_ascii_uppercase_code_unit(
   data : String,
   start : Int,
   len : Int,
 ) -> Int? {
   let end = start + len
-  for i = start; i < end; i = i + 1 {
+  for i in start..<end {
     // SAFETY: i stays within the caller-provided UTF-16 slice bounds.
     let code_unit = data.unsafe_get(i).to_int()
     if is_ascii_uppercase_code_unit(code_unit) {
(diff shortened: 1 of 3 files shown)


builtin/string_methods.mbt
```diff
@@ -1736,53 +1736,218 @@ test "View::replace_all boundary cases" {
   inspect("Testing boundary condition", content="Testing boundary condition")
 }
 
+///|
+/// Helpers for ASCII-only lowercase conversion over UTF-16 code units.
+/// Non-ASCII code units, including surrogate halves, are copied unchanged and
+/// are never converted through `Char`.
+fn is_ascii_uppercase_code_unit(code_unit : Int) -> Bool {
```
bobzhang, line 1746: take UInt16 as argument?

builtin/string_methods.mbt
```diff
@@ -1736,53 +1736,218 @@ test "View::replace_all boundary cases" {
...
+  code_unit >= 0x41 && code_unit <= 0x5A
```
bobzhang, line 1747: code_unit is 'A'..='Z'

PR #3818, opened 2026-07-16
Optimize String find and rev_find with SIMD

Yu-zh:
## Summary

- add forward and reverse SIMD scanners for single UTF-16 code units
- route one-code-unit `StringView::find` and `StringView::rev_find` through eight-lane batch comparison on native and wasm
- replace the brute-force/Boyer-Moore-Horspool multi-code-unit paths with a first/last-code-unit two-anchor prefilter mirroring Bytes search
- verify only the unchecked middle range after an anchor hit
- cut over to a middle-first scalar scan when anchor matches become dense
- retain scalar implementations under `#cfg(not(any(target="native", target="wasm")))`
- cover SIMD boundaries, sliced views, isolated surrogate code units, dense-anchor fallback, and exhaustive view ranges

## Motivation

String search previously compared one UTF-16 code unit at a time for one-unit patterns, then used either brute force or a freshly allocated 256-entry Boyer-Moore-Horspool skip table for longer patterns. Bytes search already avoids those costs by batching candidate comparisons and using first/last anchors.

This change mirrors that structure for String search using the V128 String load intrinsic and `i16x8` comparisons. It preserves raw UTF-16 code-unit semantics, including sliced views and isolated surrogate code units.

## Benchmark results

Native ARM64 release build. Times are per operation.

### Single-code-unit search

Compared with the exact `origin/main` base (`8ab300fb`).

| Case | Scalar | SIMD | Speedup |
| --- | ---: | ---: | ---: |
| find absent, 64 code units | 26.53 ns | 11.15 ns | 2.38x |
| find absent, 256 code units | 90.79 ns | 21.42 ns | 4.24x |
| find at end, 256 code units | 98.49 ns | 20.46 ns | 4.81x |
| rev_find absent, 256 code units | 94.30 ns | 19.06 ns | 4.95x |
| rev_find at start, 256 code units | 95.43 ns | 18.70 ns | 5.10x |
| sliced StringView find absent, 256 code units | 89.59 ns | 20.16 ns | 4.44x |
| sliced StringView rev_find absent, 256 code units | 91.10 ns | 18.46 ns | 4.93x |

### Two-anchor multi-code-unit search

Compared with the exact pre-two-anchor PR commit (`f231dde4`) using the identical benchmark file.

| Case | Previous | Two-anchor | Speedup |
| --- | ---: | ---: | ---: |
| find `az` absent, 256 code units | 179.79 ns | 31.72 ns | 5.67x |
| find `az` at end, 256 code units | 179.78 ns | 32.50 ns | 5.53x |
| find long pattern absent, 256 code units | 190.62 ns | 28.31 ns | 6.73x |
| rev_find `za` absent, 256 code units | 90.09 ns | 32.38 ns | 2.78x |
| rev_find `za` at start, 256 code units | 94.04 ns | 31.53 ns | 2.98x |
| sliced StringView find absent, 256 code units | 178.92 ns | 31.30 ns | 5.72x |
| find dense false positives, 256 code units | 177.16 ns | 163.01 ns | 1.09x |
| rev_find dense false positives, 256 code units | 184.99 ns | 168.28 ns | 1.10x |

## Validation

- `moon fmt`
- `moon info` (no interface changes)
- `moon check --target all -p builtin --warn-list +73`
- `moon check --target llvm -p builtin --warn-list +73`
- `moon test --target all -p builtin`
  - wasm: 2889/2889
  - wasm-gc: 2889/2889
  - js: 2877/2877
  - native: 2847/2847
- `moon test` (6719/6719 tests passed)
- `git diff --check`

### DIFF
--- a/builtin/simd.mbt
+++ b/builtin/simd.mbt
@@ -12,9 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// The 128-bit vector operations used by the SIMD byte scanners in
-// `bytes_find.mbt`. `builtin` cannot import `moonbitlang/core/v128` (that would
-// be a cycle), so the few ops the scanners need are hosted here directly.
+// The 128-bit vector operations used by the SIMD byte and UTF-16 scanners in
+// `bytes_find.mbt` and `string_find_code_unit.mbt`. `builtin` cannot import
+// `moonbitlang/core/v128` (that would be a cycle), so the few ops the scanners
+// need are hosted here directly.
 //
 // Scanner-only operations are gated to the linear-memory backends that use
 // them. `v128_make` and `v128_load` also support aligned bitstring extraction,
@@ -40,6 +41,14 @@ fn i8x16_splat(value : Byte) -> V128 {
 }
 
 ///|
+#cfg(any(target="native", target="wasm"))
+#intrinsic("%v128.i16x8_splat")
+fn i16x8_splat(value : UInt16) -> V128 {
+  let lane = value.to_uint64() * (0x0001000100010001 : UInt64)
+  v128_make(lane, lane)
+}
+
+///|
 #borrow(bytes)
 #intrinsic("%v128.v128_load")
 fn v128_load(bytes : FixedArray[Byte], offset : Int) -> V128 {
@@ -47,6 +56,26 @@ fn v128_load(bytes : FixedArray[Byte], offset : Int) -> V128 {
     fixedarray_read_uint64_le(bytes, offset),
     fixedarray_read_uint64_le(bytes, offset + 8),
   )
+}
+
+///|
+// Loads eight consecutive UTF-16 code units beginning at `offset`. The caller
+// must ensure `offset + 8 <= str.length()`.
+#borrow(str)
+#cfg(any(target="native", target="wasm"))
+#intrinsic("%v128.v128_load_i16x8")
+fn v128_load_i16x8(str : String, offset : Int) -> V128 {
+  let lo = for i in 0..<4; word = (0 : UInt64) {
+    continue word | (str.unsafe_get(offset + i).to_uint64() << (16 * i))
+  } nobreak {
+    word
+  }
+  let hi = for i in 0..<4; word = (0 : UInt64) {
+    continue word | (str.unsafe_get(offset + 4 + i).to_uint64() << (16 * i))
+  } nobreak {
+    word
+  }
+  v128_make(lo, hi)
 }
 
 ///|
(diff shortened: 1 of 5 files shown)


bobzhang:
## Review sign-off (Claude)

Reviewed the full diff on the rebased tip (branch is 2 ahead / 0 behind current `main`, which now includes the merged #3918). **LGTM — sign-off.**

### Correctness

- **Intrinsics are real**: all four new names — `%v128.v128_load_i16x8`, `%v128.i16x8_splat`, `%v128.i16x8_eq`, `%v128.i16x8_bitmask` — are registered in the compiler's intrinsic table, so native/wasm get true vector code; the pure-MoonBit fallback bodies define matching lane semantics (LE 16-bit lanes packed into two `UInt64`s, consistent across `splat`/`eq_mask`/`high_bits`).
- **Mask bit math**: forward scans return `pos + mask.ctz()` (lowest matching lane); reverse scans return `chunk_start + 31 - mask.clz()` — correct highest-set-bit index for an 8-bit mask in a 32-bit `Int`.
- **Load bounds**: the two-anchor vector loop's worst-case load index is `candidate_end - 1 + last_offset + 7 = target_start + target_len - 1`, within the backing string for every view; single-unit loops are bounded by `pos + 8 <= end` with `end <= data.length()`. No out-of-bounds loads on any path.
- **Anchor enumeration is exhaustive and ordered**: every true match is an anchor candidate; forward scans candidates in increasing order (leftmost match wins), reverse in decreasing order (rightmost wins). Dense-anchor cutover resumes at `scanned + 1` (forward) / exclusive `candidate` (reverse), skipping only already-rejected positions.
- **Semantics preserved**: empty pattern returns `Some(0)` / `Some(len)` as before; raw UTF-16 code-unit matching (isolated surrogates included) is unchanged; view-relative results are mapped back correctly from backing-string offsets.
- The deleted BMH edge-case tests are subsumed by the new dispatch arms plus the exhaustive tests.

### Observations (non-blocking)

- Bytes search falls back to Rabin–Karp on dense anchors; this PR falls back to middle-first scalar brute force, so the worst case stays O(n·m) — the same class as the replaced BMH (also O(n·m) worst case), and the Two-Way TODO is retained. Fine as-is, worth remembering if pathological inputs ever matter.
- The middle-first comparison order in the fallback is well-motivated by the cutover precondition and documented in comments.

### Test coverage

Strong: exhaustive agree-with-scalar tests over every view range of a 31-unit string (single- and multi-unit patterns), SIMD block boundaries, sliced views, isolated `0xD800` in both directions, and dense-false-positive fallback paths for both `find` and `rev_find` — running the SIMD paths on native/wasm and the `#cfg` scalar paths on js/wasm-gc.

### Verification (rebased tip, moon 0.1.20260731 nightly)

- `moon check`: clean; `moon fmt` + `moon info`: zero drift
- `moon test` (wasm-gc): 7006/7006
- `moon test --target all -p builtin`: wasm 2936/2936, wasm-gc 2936/2936, js 2913/2913, native 2894/2894

No rebase was needed — the branch already sits on the current main tip and GitHub reports it mergeable.

Issue #3824, opened 2026-07-16
sorted_set: union defensively copies both trees and recounts the result size


bobzhang:
`sorted_set/set.mbt` `SortedSet::union` (the `TODO: optimize this. Avoid counting the size of the set` at ~line 175):

1. `copy_tree` is called on **both** operands up front. The copies are required by the current mutable design (`add_node`/`delete_node`/rotations mutate nodes in place, so subtree sharing would corrupt the source sets) — but it means union is O(n+m) allocations before any merging starts.
2. After the split/join merge, the result size is recomputed by a full `each` traversal because nodes store only `height`, not subtree size.

Possible directions:
- Store subtree `size` in `Node` (as the immutable `immut/sorted_set` does) — kills the recount and gives O(1) `length` for subtrees; costs one word per node and bookkeeping in every rotation.
- Count duplicates during `split` (a duplicate is discovered exactly when `split` hits `comp == 0`), then `size = size1 + size2 - dups` — no node layout change, but `split` needs a counting variant.
- For the copy overhead: iterate the smaller set and `add` into a copy of the larger — O(n + m log n) with one copy instead of two and no recount (`add` maintains size). Loses the asymptotic edge of split/join union for balanced sizes.

Related: `symmetric_difference` had the same shape and is addressed by #3819 with a merge-walk; `intersection` (each + contains) may also benefit from a merge-walk.

PR #3829, opened 2026-07-16
perf(sorted_set): compute union size from split drops instead of recounting

bobzhang:
Partially addresses #3824 — resolves the `TODO: optimize this. Avoid counting the size of the set` in `union`.

## Before

After building the merged tree, `union` recomputed the result size with a full `each()` traversal (an extra O(n+m) pass over the freshly built tree).

## After

An element common to both sets is dropped exactly once, at the found pivot of the split. `union` counts those drops and derives the size arithmetically: `|self| + |src| − dups`. The recount pass is gone.

## Rebased reimplementation (2026-08-20)

When this PR was first written, `union`'s `split` was the only splitter, and it gained an `Int` drop-count component. main has since grown `split_member` (the split-based `difference`/`intersection`/`symmetric_difference` rewrite), whose `found` flag carries exactly the same information — so the rebase re-implements the intent on today's machinery:

- `union` uses `split_member` and counts `found` pivots.
- The now-unused `split` is **deleted**; its whitebox test is replaced by an equivalent `split_member` test pinning `found` for present/absent/below-range/above-range/empty splits.
- **`split_member` returns a `#valtype` struct** (`SplitResult { left; found; right }`) instead of a 3-tuple, keeping the per-node result stack-allocated on native — same pattern as `JsonNumberScan` in `json/lex_number.mbt`. This benefits every split-based operation, not just union.

## Benchmarks (`moon bench --release`, n=10K sets)

`union`, main → this branch (recount removal + valtype):

| case | native | js | wasm-gc |
|---|---|---|---|
| 50% overlap | 518 → **429 µs** (−17%) | 297 → **222 µs** (−25%) | 197 → **158 µs** (−20%) |
| disjoint | 382 → **304 µs** (−20%) | 261 → **153 µs** (−41%) | 154 → **96 µs** (−38%) |
| identical | 592 → **520 µs** (−12%) | 323 → **277 µs** (−14%) | 232 → **193 µs** (−17%) |

`#valtype` struct vs tuple, isolated (affects union/difference/intersection/symmetric_difference): **native −5% to −9%**, js flat, wasm-gc within ±2% noise.

## Tests

Union exact-size assertions for overlapping/disjoint/identical operands cross-checked against an element count; the `split_member` whitebox test; the existing invariant and quickcheck suites all exercise the new path. Mutation-verified: double-counting `dups` fails 8 tests.

## Review

Reviewed by **Codex CLI** (`xhigh`) on the rebased reimplementation, first-round sign-off: the dups induction re-verified on `split_member` (disjoint fragments, pivot counted exactly once, empty-side arms cannot hide common elements), sizes read from unmutated copies, `SplitResult` preserves tuple semantics at all four call sites, no `split` callers remain, generated interface byte-identical. (The original pre-rebase version carried its own sign-off with the equivalent induction.)

## Validation

- Full repo `moon test`: 7463/7463 (sorted_set 71/71)
- `moon check --warn-list +unnecessary_annotation` clean, `moon fmt` applied, no `.mbti` changes

### DIFF
--- a/sorted_set/set.mbt
+++ b/sorted_set/set.mbt
@@ -177,12 +177,19 @@ pub fn[V : Compare] SortedSet::union(
   self : SortedSet[V],
   src : SortedSet[V],
 ) -> SortedSet[V] {
+  // An element common to both sets is dropped exactly once, at the found
+  // pivot of `split_member`; counting those drops gives the result size
+  // arithmetically instead of re-traversing the merged tree.
+  let mut dups = 0
   fn aux(a : Node[V]?, b : Node[V]?) -> Node[V]? {
     match (a, b) {
       (Some(_), None) => a
       (None, Some(_)) => b
       (Some({ value: va, left: la, right: ra, .. }), Some(_)) => {
-        let (l, r) = split(b, va)
+        let { left: l, found, right: r } = split_member(b, va)
+        if found {
+          dups += 1
+        }
         Some(join(aux(la, l), va, aux(ra, r)))
       }
       (None, None) => None
@@ -191,15 +198,8 @@ pub fn[V : Compare] SortedSet::union(
 
   match (self.root, src.root) {
     (Some(_), Some(_)) => {
-      let t1 = copy_tree(self.root)
-      let t2 = copy_tree(src.root)
-      let t = aux(t1, t2)
-      let mut ct = 0
-      let ret = { root: t, size: 0 }
-      // TODO: optimize this. Avoid counting the size of the set.
-      ret.each(_x => ct += 1)
-      ret.size = ct
-      ret
+      let t = aux(copy_tree(self.root), copy_tree(src.root))
+      { root: t, size: self.size + src.size - dups }
     }
     (Some(_), None) => { root: copy_tree(self.root), size: self.size }
     (None, Some(_)) => { root: copy_tree(src.root), size: src.size }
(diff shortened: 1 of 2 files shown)


bobzhang:
## Codex CLI review — post-rebase (`xhigh`)

The branch was reimplemented on today's main (the original targeted a base without `split_member`; the rebase moves union onto it, deletes the dead `split`, and adds the `#valtype` `SplitResult` struct). Codex CLI reviewed the reimplementation fresh, briefed to re-verify the duplicate-count induction on the new recursion shape, the size-at-entry arithmetic, left/right mapping at all four `SplitResult` call sites, `split`'s deletion, and the benchmark claims.

### Verdict — approved, first round

> No blocking findings.
>
> The duplicate-count induction holds: `split_member` partitions `b` into disjoint `< va` and `> va` fragments, while `found` accounts for the pivot exactly once; empty-side arms cannot contain common elements. Both inputs are copied before the merge and their stored sizes remain unchanged.
>
> `SplitResult` preserves the tuple semantics, and all callers retain the correct left/right mapping. No old `split` callers remain, and the generated interface is byte-identical.
>
> Dropping the 101-element splitter case is not blocking: the deeper randomized invariant scenarios already validate AVL structure, node count, isolation, and subsequent mutation.
>
> The commit-message benchmark percentages match the supplied measurements.
>

---
*Posted on behalf of Codex CLI by Claude Code. Validation outside Codex's read-only sandbox: full repo `moon test` 7463/7463; sorted_set 71/71; benchmarks on native/js/wasm-gc as tabled in the PR body; `dups += 2` mutation fails 8 tests.*

PR #3837, opened 2026-07-20
fix(stringbuilder): guard capacity growth against Int overflow

mizchi:
## Summary

- Guard `StringBuilder` capacity growth against `Int` overflow.
- Add allocation-free whitebox coverage for normal growth and overflow boundaries.
- Add a focused native growth benchmark.

## Root cause

`StringBuilder::grow_if_necessary` repeatedly doubled the backing capacity. Once the capacity exceeded `2^30`, `Int` overflow could make the loop cycle through non-positive values and never reach the requested length. The new private helper falls back to the exact requested capacity after an overflowing double.

This is the `StringBuilder` counterpart to #3822, but it changes the separate UTF-16 builder implementation.

## Benchmark

`StringBuilder growth from minimum capacity n=4096`, lower is better.

- Environment: Apple M5 (arm64), macOS Darwin 25.5.0, MoonBit `0.1.20260824`
- Command: `moon bench --release --target native -p moonbitlang/core/builtin -f stringbuilder_bench_test.mbt -i 5`
- Three sequential runs: 15.17 µs, 14.56 µs, 14.82 µs
- Mean across runs: **14.85 µs**

Each run contains 10 benchmark samples.

## Validation

- `moon info && moon fmt`
- `moon test`: 7546 passed, 0 failed
- `moon check`
- No generated interface or formatting diff

### DIFF
--- a/builtin/moon.pkg
+++ b/builtin/moon.pkg
@@ -55,5 +55,6 @@ options(
     "stringbuilder_buffer.mbt": [ "not", "js" ],
     "stringbuilder_buffer_wbtest.mbt": [ "not", "js" ],
     "stringbuilder_concat.mbt": [ "js" ],
+    "stringbuilder_grow_wbtest.mbt": [ "not", "js" ],
   },
 )
--- a/builtin/stringbuilder_bench_test.mbt
+++ b/builtin/stringbuilder_bench_test.mbt
@@ -89,6 +89,18 @@ test "bench StringBuilder::write_char non-BMP n=4096" (it : @bench.T) {
 }
 
 ///|
+test "bench StringBuilder growth from minimum capacity n=4096" (it : @bench.T) {
+  let piece = stringbuilder_bench_piece
+  it.bench(fn() {
+    let builder = StringBuilder(size_hint=1)
+    for _ in 0..<stringbuilder_bench_repeats {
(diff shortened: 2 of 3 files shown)


bobzhang:
Thanks for this, and sorry it sat so long.

The overflow guard itself has since landed in b903d495e ("fix: make buffer growth overflow-safe"), which introduced `stringbuilder_growth_capacity` in `builtin/stringbuilder_buffer.mbt` with the same `let next = space * 2; if next <= space { break required }` shape, plus an `abort` for the case where a caller's `len + n` wraps. Regression tests came with it in `builtin/stringbuilder_buffer_wbtest.mbt`. The same commit superseded the `buffer` counterpart in #3317.

After your 2026-09-01 rebase this PR is down to `stringbuilder_grow_wbtest.mbt`, which duplicates that existing whitebox coverage, and one benchmark — so I'm closing it as superseded.

The benchmark is worth having on its own, though: every current StringBuilder bench passes a `size_hint` covering the whole workload, so none of them ever exercises the doubling path. A small PR with just the `bench StringBuilder growth from minimum capacity n=4096` hunk would be welcome.

Heads-up that #3822 is in the same position — it still edits `buffer/buffer.mbt`, which b903d495e already rewrote.

PR #3840, opened 2026-07-20
perf(list): avoid allocations in is_suffix

mizchi:
## Summary

- avoid allocating reversed copies in `List::is_suffix`
- add boundary coverage for empty and longer suffixes, plus ArrayView construction
- add native benchmarks for core `List` operations, including construction, `is_suffix`, and `scan_right`

## Rationale

`is_suffix` previously reversed both input lists before comparing their prefixes. The new implementation compares lengths, advances to the candidate suffix, and compares directly. This preserves the API and behavior while avoiding temporary list allocations.

The construction benchmarks cover both large array conversion and repeated `cons`, so future runtime or representation work can measure allocation-related changes without modifying the public `List` enum.

## Benchmark

Native release, 100,000-element list with a 50,000-element suffix:

| implementation | mean |
| --- | ---: |
| before (reverse both lists) | 890.66 us |
| after (direct comparison) | 303.47 us |

This is approximately a 2.9x speedup on the measured case.

Construction baselines (native release):

| operation | mean |
| --- | ---: |
| `List(ArrayView)`, 100,000 elements | 473.23 us |
| `List(ArrayView)`, 8 elements | 49.48 ns |
| repeated `cons`, 100,000 elements | 498.34 us |

## Memory

Native release, 1,000,000-element list with a 500,000-element suffix, measured with `/usr/bin/time -l` and mimalloc statistics:

| implementation | maximum RSS | peak memory footprint |
| --- | ---: | ---: |
| before (reverse both lists) | 82.1 MB | 81.6 MB |
| after (direct comparison) | 38.0 MB | 37.4 MB |

The direct comparison reduces peak RSS by approximately 54% in this case.

## Validation

- `moon check list --target all`
- `moon test list/list_test.mbt --target wasm-gc --quiet`
- `moon bench list/list_bench_test.mbt --target native --release --index 0`
- `moon bench list/list_bench_test.mbt --target native --release --index 1`
- `moon bench list/list_bench_test.mbt --target native --release --index 12`
- `moon bench list/list_bench_test.mbt --target native --release --index 13`
- `moon bench list/list_bench_test.mbt --target native --release --index 14`

### DIFF
--- a/list/list.mbt
+++ b/list/list.mbt
@@ -1758,7 +1758,10 @@ pub fn[A : Eq] List::is_prefix(self : List[A], prefix : List[A]) -> Bool {
 /// }
 /// ```
 pub fn[A : Eq] List::is_suffix(self : List[A], suffix : List[A]) -> Bool {
-  self.rev().is_prefix(suffix.rev())
+  let self_length = self.length()
+  let suffix_length = suffix.length()
+  suffix_length <= self_length &&
+  self.drop(self_length - suffix_length).is_prefix(suffix)
 }
 
 ///|
--- a/list/list_bench_test.mbt
+++ b/list/list_bench_test.mbt
@@ -27,6 +27,12 @@ test "bench List::from_array n=100000" (it : @bench.T) {
 }
 
 ///|
+test "bench List::from_array n=8" (it : @bench.T) {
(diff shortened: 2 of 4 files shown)


bobzhang:
closed by #4109

PR #3851, opened 2026-07-22
feat(queue): make trait promotions explicit via extend

bobzhang:
## Summary

Part of the **per-package series** migrating `moonbitlang/core` off implicit trait-method promotion (see #3849 for `deque`). This one covers **`queue`** only.

It enables `implicit_impl_as_method` (E0079) in `queue/moon.pkg` and makes every trait-impl promotion explicit via `extend`, in a dedicated `queue/extends.mbt` shim.

Queue has **no methods worth keeping promoted** — the compiler flags only two, both deprecated:

| Deprecate — `#deprecated(…, skip_current_package=true)` + `#doc(hidden)` |
|---|
| `Show::{to_string, output}`, `@quickcheck.Arbitrary::{arbitrary}` |

- **`Show` is deprecated for collections (Rust-style).** Queue's `Show` impl is already `#deprecated` on `main` (steering to `@debug.Debug`), so both `to_string` and `output` are deprecated rather than promoted.
- **`Arbitrary`** → use the `Arbitrary` trait instead.
- Both carry `#doc(hidden)`, so they stay callable but are absent from `pkg.generated.mbti` — the **generated interface is unchanged**. (Queue's `Eq` impl isn't promoted anywhere, so it needs no extend.)

Just 2 files (`moon.pkg` + `extends.mbt`); no other package is touched.

## Verification
- `moon check --deny-warn --target all` — clean (zero warnings).
- `moon test --target all` — green on all four backends.

---

### DIFF
--- a/queue/extends.mbt
+++ b/queue/extends.mbt
@@ -1,0 +1,27 @@
+// Copyright 2026 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Queue has no promoted trait methods; every promotion below is deprecated.
+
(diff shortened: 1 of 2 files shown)


bobzhang:
Addressed the Copilot review: deprecation messages now name the concrete replacement — `@quickcheck.Arbitrary::arbitrary` for the `arbitrary` promotion, and the `Show` message now mirrors the existing impl-level deprecation (rationale + link to the debug README). Re-verified: `moon check --deny-warn --target all` clean, `.mbti` unchanged, scoped to `queue/`. (This concrete-replacement wording is now the standard for the rest of the migration series.)

PR #3859, opened 2026-07-22
feat(set): make trait promotions explicit via extend

bobzhang:
## Summary

Part of the **per-package series** migrating `moonbitlang/core` off implicit trait-method promotion (see #3849 for `deque`). This one covers **`set`** only.

| Promote — plain `pub extend` | Deprecate — `#deprecated(…, skip_current_package=true)` + `#doc(hidden)` |
|---|---|
| `BitAnd` (`&`), `BitOr` (`\|`), `BitXOr` (`^`), `Sub` (`-`), `Default` | `@debug.Debug`, `Eq`, the whole **`Show`** trait, `ToJson` |

- **Set-algebra operators** (`&`/`\|`/`^`/`-` = intersection/union/symmetric-difference/difference) are promoted per the arithmetic-operator policy (`.land()`/`.lor()`/… stay callable), unlike `Compare`'s `op_*`.
- **`Show`** is deprecated (Set is a collection; its `Show` impl is already `#deprecated` on `main`); **`ToJson`** → the free `@json.to_json`. Each message names the concrete replacement.
- The blackbox-test-only `Key` struct (`derive(Eq)`) gets an inline `pub extend Key with Eq` in the test file (test types can't be named from the source shim).
- `#doc(hidden)` on all deprecations, so `pkg.generated.mbti` gains only `Set::{land, lor, lxor, sub, default}`.

Migrates set's README doc examples off the now-deprecated `.to_json()` → `@json.to_json(...)`. Scoped to `set/`.

## Verification
- `moon test -p moonbitlang/core/set --target all` — green (65×4).

---

### DIFF
--- a/set/README.mbt.md
+++ b/set/README.mbt.md
@@ -215,14 +215,14 @@ Sets can be serialized to JSON as arrays:
 ///|
 test "json serialization" {
   let set = @set.Set([1, 2, 3])
-  let json = set.to_json()
+  let json = @json.to_json(set)
 
   // JSON representation is an array
   @debug.debug_inspect(json, content="Array([Number(1), Number(2), Number(3)])")
 
   // String set
   let string_set = @set.Set(["a", "b", "c"])
-  let string_json = string_set.to_json()
+  let string_json = @json.to_json(string_set)
   @debug.debug_inspect(
     string_json,
     content="Array([String(\"a\"), String(\"b\"), String(\"c\")])",
--- a/set/extends.mbt
+++ b/set/extends.mbt
@@ -1,0 +1,52 @@
+// Copyright 2026 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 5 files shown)


bobzhang:
Addressed the Copilot review: dropped the long GitHub URL from the `Show` message (now the short `Use `@debug.Debug` instead of `Show` for collections`, matching the other shims), and fixed the `Debug` replacement to `Debug::to_repr` (the suggested `@debug.to_repr` free function doesn't exist — debug exposes `@debug.to_string`/`repr` → String or the trait method `Debug::to_repr`). The concrete `Hash`/`Arbitrary` wording (`Hash::hash`, `@quickcheck.Arbitrary::arbitrary`) is the agreed standard for the series (per the earlier queue review); a follow-up PR aligns the 5 already-merged packages (deque/list/priority_queue/hashmap/sorted_set) to the same concrete wording so everything is consistent.

PR #3870, opened 2026-07-22
feat(cmp): make trait promotions explicit via extend

bobzhang:
## Summary

Part of the **per-package series** migrating `moonbitlang/core` off implicit trait-method promotion (see #3849 for `deque`). This one covers **`cmp`** only — the first **`Show` split** in the series.

| Promote — plain `pub extend` | Deprecate — `#deprecated(…, skip_current_package=true)` + `#doc(hidden)` |
|---|---|
| `Reverse` `Compare::{compare}`, `Show::{to_string}` | `Reverse` `@debug.Debug`, `Compare::{op_ge, op_gt, op_le, op_lt}`, `Eq`, `Hash`, `Show::{output}` |

- **`Reverse` derives a genuine (non-deprecated) `Show`**, so unlike the collections this is a **split**: `to_string` stays promoted, only the Logger-writing `output` is deprecated.
- `#doc(hidden)` on deprecations, so `pkg.generated.mbti` gains only `Reverse::{compare, to_string}`.
- The README doc example's local `Person` struct (`derive(Debug)`) gets an inline `pub extend Person with Debug::{to_repr}`. Scoped to `cmp/`.

## Verification
- `moon test -p moonbitlang/core/cmp --target all` — green.

---

### DIFF
--- a/cmp/extends.mbt
+++ b/cmp/extends.mbt
@@ -1,0 +1,45 @@
+// Copyright 2026 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// --- promoted: kept as regular methods ---
+
+///|
+pub extend Reverse with Compare::{compare}
+
+// --- deprecated: hidden from the generated interface ---
+
+///|
+#deprecated("Use `Debug::to_repr` instead", skip_current_package=true)
+#doc(hidden)
+pub extend Reverse with @debug.Debug::{to_repr}
+
+///|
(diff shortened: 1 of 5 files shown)


cmp/extends.mbt
```diff
@@ -0,0 +1,48 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// --- promoted: kept as regular methods ---
+
+///|
+pub extend Reverse with Compare::{compare}
+
+///|
+pub extend Reverse with Show::{to_string}
```
bobzhang: we should deprecate `impl Show for Reverse`? and #deprecate pub extend here

bobzhang:
Good call — applied it: `Reverse`'s Show is now **fully deprecated** (both `to_string` and `output`) rather than split, since `Reverse[T]` is an ordering wrapper (used like `sort_by_key(x => Reverse(x))`) rather than a display value — use `@debug.Debug` to inspect it. `Reverse` now promotes only `compare`; the doc example's `a.to_string()` moved to `@debug.debug_inspect(a, …)`. `moon check --deny-warn --target all` clean, tests green.

bobzhang:
Done — went all the way per your call: dropped `Show` from Reverse's `derive` and added a `#deprecated pub impl Show for Reverse` (reproducing the `Reverse(inner)` rendering), so the **whole Show is deprecated** (even `Show::to_string`/`inspect` on a Reverse warns), plus the promotion. Reverse promotes only `compare`; all Show usages (doc + 3 README examples) moved to `@debug.debug_inspect`. Clean check, tests green.

PR #3875, opened 2026-07-22
feat(encoding/utf8): make trait promotions explicit via extend

bobzhang:
## Summary

Part of the per-package series migrating `moonbitlang/core` off implicit trait-method promotion (see #3849 for `deque`). Covers **`encoding/utf8`** only.

`Malformed`'s only compiler-flagged promotion is `@debug.Debug`, deprecated (`#doc(hidden)`, → `Debug::to_repr`). No promotions; `pkg.generated.mbti` unchanged.

## Verification
- `moon test -p moonbitlang/core/encoding/utf8 --target all` — green.

---

### DIFF
--- a/encoding/utf8/extends.mbt
+++ b/encoding/utf8/extends.mbt
@@ -1,0 +1,22 @@
+// Copyright 2026 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


bobzhang:
Addressed the Copilot review: reworded the header comment to remove the contradiction (it no longer says "no promoted trait methods" while listing deprecated promotions) — now: "Every trait-method promotion below is deprecated (none are kept as regular methods)."

PR #3878, opened 2026-07-22
feat(immut/internal/path): make trait promotions explicit via extend

bobzhang:
## Summary

Part of the per-package series migrating `moonbitlang/core` off implicit trait-method promotion (see #3849 for `deque`). Covers **`immut/internal/path`** only.

The compiler flags only Eq for the Path type, which is deprecated in favour of the `==` / `!=` operators (no promotions).

Deprecations carry `#doc(hidden)`, so `pkg.generated.mbti` gains only the promoted methods. Scoped to `immut/internal/path/`.

## Verification
- `moon test -p moonbitlang/core/immut/internal/path --target all` — green.

---

### DIFF
--- a/immut/internal/path/extends.mbt
+++ b/immut/internal/path/extends.mbt
@@ -1,0 +1,22 @@
+// Copyright 2026 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


bobzhang:
Addressed the Copilot review: reworded the header comment to remove the contradiction (it no longer says "no promoted trait methods" while listing deprecated promotions) — now: "Every trait-method promotion below is deprecated (none are kept as regular methods)."

PR #3883, opened 2026-07-22
feat(bytes/internal/regex_engine/ast): make trait promotions explicit via extend

bobzhang:
## Summary

Part of the per-package series migrating `moonbitlang/core` off implicit trait-method promotion (see #3849 for `deque`). Covers **`bytes/internal/regex_engine/ast`** only.

The compiler flags only @debug.Debug for the Assertion, Pattern, PatternDesc and Quantifier types, all deprecated (no promotions).

Deprecations carry `#doc(hidden)`, so `pkg.generated.mbti` gains only the promoted methods. Scoped to `bytes/internal/regex_engine/ast/`.

## Verification
- `moon test -p moonbitlang/core/bytes/internal/regex_engine/ast --target all` — green.

---

### DIFF
--- a/bytes/internal/regex_engine/ast/extends.mbt
+++ b/bytes/internal/regex_engine/ast/extends.mbt
@@ -1,0 +1,38 @@
+// Copyright 2026 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


bobzhang:
Addressed the Copilot review: reworded the header comment to remove the contradiction — it now reads "None of ... keeps a promoted trait method — every trait-method promotion below is deprecated."

PR #3891, opened 2026-07-22
feat(strconv): make trait promotions explicit via extend

bobzhang:
## Summary

Part of the per-package series migrating `moonbitlang/core` off implicit trait-method promotion (see #3849 for `deque`). Covers **`strconv`** only.

Two types with genuine (non-deprecated) Show: StrConvError gets the standard Show split (Show::to_string promoted via extend, output deprecated). Decimal already had a private manual 'fn Decimal::to_string' delegating to Show::to_string, which collided with an extend (E4056); per the established remedy it is made 'pub fn Decimal::to_string' (with a doc comment) instead of an extend, so Show::to_string is promoted via that method. @debug.Debug and Show::output are deprecated for both types.

Deprecations carry `#doc(hidden)`, so `pkg.generated.mbti` gains only the promoted methods. Scoped to `strconv/`.

## Verification
- `moon test -p moonbitlang/core/strconv --target all` — green.

---

### DIFF
--- a/strconv/decimal.mbt
+++ b/strconv/decimal.mbt
@@ -584,11 +584,6 @@ pub impl Show for Decimal with fn output(self, logger) {
 }
 
 ///|
-fn Decimal::to_string(self : Decimal) -> String {
-  Show::to_string(self)
-}
-
-///|
 test "new" {
   let hpd = Decimal::from_int64_priv(1L)
   inspect(hpd.digits.length(), content="800")
--- a/strconv/extends.mbt
+++ b/strconv/extends.mbt
@@ -1,0 +1,43 @@
+// Copyright 2026 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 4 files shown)


strconv/decimal.mbt
```diff
@@ -584,7 +584,8 @@ pub impl Show for Decimal with fn output(self, logger) {
 }
 
 ///|
-fn Decimal::to_string(self : Decimal) -> String {
+/// Returns the string representation of this decimal.
+pub fn Decimal::to_string(self : Decimal) -> String {
```
bobzhang: pub extend

bobzhang:
Applied your suggestion: removed the redundant hand-written `fn Decimal::to_string { Show::to_string(self) }` and promoted it via `pub extend Decimal with Show::{to_string}` like the other types. Behavior-preserving (`Decimal::to_string` is still public), no more E4056, and it drops the special-case `pub fn` + doc. `moon check --deny-warn --target all` clean.

PR #3893, opened 2026-07-22
feat(bench): make trait promotions explicit via extend

bobzhang:
## Summary

Part of the per-package series migrating `moonbitlang/core` off implicit trait-method promotion (see #3849 for `deque`). Covers **`bench`** only.

Three types (Bench, Summary, Timestamp), all deprecated: @debug.Debug for all three, plus Summary's ToJson. No promotions. The README doc example's Summary.to_json() was migrated to the trait-qualified ToJson::to_json(summary) (matching bench_test.mbt; @json is not imported in bench).

Deprecations carry `#doc(hidden)`, so `pkg.generated.mbti` gains only the promoted methods. Scoped to `bench/`.

## Verification
- `moon test -p moonbitlang/core/bench --target all` — green.

---

### DIFF
--- a/bench/README.mbt.md
+++ b/bench/README.mbt.md
@@ -19,7 +19,7 @@ test "basic benchmarking" {
   })
 
   // The benchmark ran successfully (we can't inspect exact timing)
-  inspect(summary.to_json().stringify().length() > 0, content="true")
+  inspect(@json.to_json(summary).stringify().length() > 0, content="true")
 }
 ```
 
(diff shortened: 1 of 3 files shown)


bobzhang:
Addressed the Copilot review: switched the README example to `@json.to_json(summary)` (bench already depends on `@json` via moon.pkg), matching the deprecation message's recommended API. Re-verified `moon check --deny-warn --target all` clean.

PR #3918, opened 2026-07-24
refactor(queue): back Queue by Deque instead of a linked list

Yu-zh:
## What

Reimplements `@queue.Queue` as a thin wrapper around `@deque.Deque`, replacing the cons-cell linked list. Elements now live in one contiguous circular buffer instead of a heap cell per element.

## How

- Most `Queue` methods are one-line delegations: `push` → `push_back`, `pop`/`peek` → `pop_front`/`front`, plus `clear`, `length`, `is_empty`, `copy`, `iter`, `from_iter`, and the `Queue(arr)` constructor. `transfer` becomes `Deque::append` + `clear`, keeping the self-transfer no-op (guarded by physical equality of the inner deques).
- `each`/`eachi`/`fold` keep their `#locals(f)` guarantee by iterating with `for`-in loops instead of passing `f` to deque's methods — locality does not propagate across the package boundary.
- The `cell_equal` test poked at the removed `Cons` internals; its intent (empty ≠ non-empty) is folded into the `equal` test.

## API impact

`queue/pkg.generated.mbti` has a zero diff — the public queue API is unchanged, including the `abort("Queue is empty")` messages of `unsafe_peek`/`unsafe_pop`. The deque package is untouched.

## Deferred: `#valtype` unboxing

Marking the wrapper `#valtype` (so `Queue[A]` is represented as the bare deque reference, no wrapper allocation) was tried and works, but requires `Deque` to become a non-abstract `pub struct { // private fields }`, because `#valtype` fields cannot be abstract types. That's a public representation commitment on the deque package, so it's left for a follow-up discussion. The extra cost of the boxed wrapper is one small heap object per queue instance, not per element.

## Verification

- `moon check`: clean, whole repo
- `moon test`: 6775/6775 pass (wasm-gc)
- `moon test --target native -p queue -p deque -p debug`: 434/434 pass
- `moon test --target js -p queue -p deque -p debug`: 448/448 pass

### DIFF
--- a/deque/pkg.generated.mbti
+++ b/deque/pkg.generated.mbti
@@ -11,9 +11,7 @@ import {
 // Errors
 
 // Types and methods
-pub struct Deque[A] {
-  // private fields
-}
+type Deque[A]
 #alias(from_array)
 #alias(of, deprecated)
 #as_free_fn(of, deprecated)
--- a/deque/pkg.generated.mbti
+++ b/deque/pkg.generated.mbti
@@ -11,7 +11,9 @@ import {
 // Errors
 
 // Types and methods
-type Deque[A]
+pub struct Deque[A] {
+  // private fields
+}
 #alias(from_array)
 #alias(of, deprecated)
 #as_free_fn(of, deprecated)
--- a/deque/types.mbt
+++ b/deque/types.mbt
@@ -46,11 +46,11 @@
 /// - Element at index `i` is at `buf[(head + i) % buf.length()]`
 /// - When `len > 0`: front is `buf[head]`, back is `buf[(head + len - 1) % cap]`
 /// - When `len == 0`: no valid element, `head` can be any valid index
-struct Deque[A] {
+pub struct Deque[A] {
   /// Circular buffer storing elements. May contain uninitialized slots.
-  mut buf : UninitializedArray[A]
+  priv mut buf : UninitializedArray[A]
   /// Number of elements currently in the deque.
-  mut len : Int
+  priv mut len : Int
   /// Index of the first element (front). Valid range: 0 <= head < buf.length().
-  mut head : Int
+  priv mut head : Int
 }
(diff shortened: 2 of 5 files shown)


bobzhang:
## Review sign-off (Claude)

Reviewed the full diff, rebased onto latest `main`, and re-verified. **LGTM — sign-off.**

### Semantics checked against `@deque`'s public API

- **`Eq`**: `Deque`'s `Eq` is structural (length check + elementwise compare) — identical semantics to the old `length && first` chain comparison. The removed `cell_equal` test's intent (empty ≠ non-empty) is preserved in the `equal` test.
- **`iter`**: `Deque::iter` carries `size_hint=len`, matching the old hand-rolled `Iter::new(..., size_hint=len)`.
- **`unsafe_peek` / `unsafe_pop`**: `abort("Queue is empty")` messages preserved; `queue/pkg.generated.mbti` has a zero diff.
- **`transfer`**: `Deque::append` + `clear` is correct for distinct deques, and the `physical_equal(self.inner, dst.inner)` guard covers self-transfer — `inner` is never shared between distinct `Queue` wrappers (constructors and `copy` always allocate a fresh deque), so the guard is equivalent to the old wrapper-level check. Note: `transfer` goes from O(1) list splicing to an O(n) element copy; no documented complexity contract is broken, and it's the expected trade-off of a contiguous buffer.
- **`#locals(f)`**: preserved by iterating with `for`-in loops and calling `f` in the loop body instead of passing `f` across the package boundary — `each`/`eachi`/`fold` all follow this pattern.
- **`Arbitrary`**: `core/quickcheck` (which now owns the impl after #3946-era move on main) builds queues via public `Queue::from_iter`, so it composes with the new representation unchanged.

### Rebase

Conflicts came from main's removal of the in-package `Arbitrary` impl + `quickcheck` import (moved to `core/quickcheck`) and doc-comment sweeps. Resolutions:

- `queue/moon.pkg`: kept main's `quickcheck` import removal, added the `deque` import.
- `queue/queue.mbt`: took the deque-backed bodies; the tail conflict deleted **both** sides (main's removed `Arbitrary` impl and this branch's removed `cell_equal` test).
- `deque/` is net-untouched by the branch, as intended.

### Verification (after rebase, moon 0.1.20260731 nightly)

- `moon check`: clean, whole repo
- `moon test` (wasm-gc): 7002/7002
- `moon test --target native -p queue -p deque -p debug`: 355/355
- `moon test --target js -p queue -p deque -p debug`: 367/367
- `moon fmt` + `moon info`: zero drift; `queue/pkg.generated.mbti` unchanged

PR #3957, opened 2026-07-30
Add Eisel-Lemire decimal parsing fast path

Yu-zh:
## Summary

- add a correctly rounded Eisel-Lemire fast path to `internal/strconv`
- pass JSON scanner mantissa/exponent data directly to the fast path
- retain the existing Clinger path and exact Decimal fallback for ambiguous, truncated, subnormal, and out-of-range inputs
- add differential correctness tests, Canada-style coordinate cases, and parsing/JSON benchmarks
- record the Go 1.26.2 source adaptation in `NOTICE`

## Motivation

Canada coordinate values such as `-65.613616999999977` have 17-19 digit mantissas that exceed the exact Clinger limit. Nearly every coordinate therefore fell back to arbitrary-precision Decimal conversion, making decimal parsing the dominant JSON decode cost.

The new path handles those mantissas with cached powers and explicit ambiguity checks. Inputs that cannot be certified by the fast path continue through the existing exact conversion.

## Performance

Re-measured after the rebase onto `main`, per backend (Apple Silicon macOS, `moon bench`, mean of 10×N runs):

`parse_double`, long mantissas (4,096 parses):

| Target | Before | After | Improvement |
| --- | ---: | ---: | ---: |
| native | 779.48 µs | 196.98 µs | 3.96x |
| wasm-gc | 1.07 ms | 425.94 µs | 2.51x |
| js | 8.79 ms | 2.26 ms | 3.89x |

JSON long-mantissa array (10,000 numbers):

| Target | Before | After | Improvement |
| --- | ---: | ---: | ---: |
| native | 2.46 ms | 832.45 µs | 2.95x |
| wasm-gc | 3.86 ms | 1.39 ms | 2.78x |
| js | 23.72 ms | 6.35 ms | 3.73x |

Existing integer and short-float paths remained within benchmark noise on all three backends (fast-digit, underscore, JSON int-array, and JSON float-array benches all within a few percent).

## Validation

- `moon test internal/strconv json --target all`: 228/228 passed on native, JS, Wasm, and Wasm-GC
- full native suite: 6,835/6,835 passed
- full JS suite: 6,878/6,878 passed
- full Wasm-GC suite: 6,926/6,926 passed
- `moon check --target all --warn-list +73`
- `moon info` and `moon fmt`
- `git diff --check`
- no generated `.mbti` changes

A full `moon test --target all` run reached an unrelated host-environment failure in the builtin Wasm tests because `wasi_snapshot_preview1.random_get` was not callable. The affected `internal/strconv` and `json` packages pass independently on Wasm.


bobzhang:
Rebased onto `main` (was 110 commits behind; clean rebase, no conflicts) and added property-based tests in 87dcf241:

- **internal/strconv**: two quickcheck properties asserting that every value `try_eisel_lemire64` accepts is bit-for-bit equal to the exact Decimal conversion (reached through `parse_double` by zero-padding the spelling past the 19-digit fast-path limit). One samples mantissas across all magnitudes and exponents overshooting the table range on both sides; the other clusters mantissas around powers of two, where rounding carries and halfway ambiguity live.
- **json**: a spelling-independence property — plain, fraction-padded, and pointed spellings of the same value must parse to identical bits, with the padded spelling forcing the exact fallback that anchors the fast routes.

Validation: `moon check`, `moon fmt`, `moon info` (no `.mbti` drift), and the affected packages pass on native, js, and wasm-gc (266/266).

### Codex CLI review (ultra reasoning effort) — sign-off, no findings

> The Eisel–Lemire implementation matches its Go reference, preserves exact fallbacks, and integrates correctly with both parsers. All-target checks and the full 7,473-test suite pass.

Codex also ran an independent differential fuzz of the ported algorithm against CPython's float parsing — 1,000,000 random (mantissa, exponent, sign) cases biased across bit lengths and the full exponent range, 834,771 accepted by the fast path, **zero mismatches** — and confirmed the parser state handed to the fast path is exact whenever `many_digits` is false, with the JSON scanner preserving the same invariant.

PR #3958, opened 2026-07-31
fix: convert intended-panic guards to guard! (warning 0087)

bobzhang:
## Summary

Recent toolchains added **warning 0087 (`guard_inexhaustive`)** for `guard` statements without an `else` clause whose condition/pattern the compiler cannot prove total — such guards panic when they fail. `moon check --target all` currently reports 141 such warnings in core.

All flagged sites (library invariant/precondition checks, test extractions, docstring and `.mbt.md` examples) rely on the intended panic-on-failure behavior, so this PR mechanically converts them to the new `guard!` form: identical runtime semantics, but the panic is declared intentional and the warning is silenced.

- 141 sites across 44 files, every changed line is exactly `guard` → `guard!`
- `moon check --target all`: 0 warnings, 0 errors
- `moon test --target all`: all passing (js 6873, native 6838, wasm/wasm-gc ok)
- `moon fmt` clean; no public API change (generated `.mbti` unchanged)
- verified the `latest` release toolchain (moonc v0.10.5+5e7afb0c0, 2026-07-27) already parses `guard!`, so stable-check CI is safe

Codex review: "All 141 changes are exact `guard` → `guard!` rewrites; no other text changed. Every panic remains appropriate as a precondition, test assertion, or internal invariant."

### DIFF
--- a/json/from_json.mbt
+++ b/json/from_json.mbt
@@ -177,7 +177,7 @@ pub impl[X : FromJson] FromJson for Array[X] with fn from_json(json, path) {
   guard json is Array(a) else {
     decode_error(path, "Array::from_json: expected array")
   }
-  guard JsonPath::Index(path, index=0) is (Index(_) as new_path)
+  guard! JsonPath::Index(path, index=0) is (Index(_) as new_path)
   a.mapi((i, x) => {
     new_path.index = i
     FromJson::from_json(x, new_path)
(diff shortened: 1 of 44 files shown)


json/from_json.mbt
```diff
@@ -189,7 +189,7 @@ pub impl[X : FromJson] FromJson for ArrayView[X] with fn from_json(json, path) {
   guard json is Array(a) else {
     decode_error(path, "ArrayView::from_json: expected array")
   }
-  guard JsonPath::Index(path, index=0) is (Index(_) as new_path)
```
bobzhang, line 192: weird guard path

Issue #2195, opened 2025-05-31
[BUG] Behavior of Array and ArrayView when elements are removed during iteration


BigOrangeQWQ:
**Describe the bug**
This is an extension and additional example for issue #2081 
When iterating over an array and removing elements inside the loop, the output is different between backends.

**Code**
```moonbit
test "panic case js" {
  let arr = [1,2,3]
  for item in arr {
    println(item)
    let _ = arr.pop().unwrap()
  }
}

test "panic case js index/out_of_bounds" {
  let arr = [1,2,3]
  let view = arr[:]
  for item in view {
    println(item)
    let _ = arr.pop().unwrap()
  }
}
```

**Expected behavior**
Case 1 and Case 2 both behave differently on the JS backend compared to other backends.
For Case 1, other backends output 1, 2, 3, but the JS backend outputs 1, 2, 0.
For Case 2, it runs normally on other backends, but on the JS backend it causes an "index out of bounds" error.

peter-jerry-ye:
`Array` and `ArrayView` are not safe for concurrent editing; they are desynced if the `Array` is modified.

BigOrangeQWQ:
> `Array` and `ArrayView` are not safe for concurrent editing; they are desynced if the `Array` is modified.

For the first example, it doesn't use ArrayView. The for item in collections form shouldn't produce iteration results of 1, 2, 0. 
Moreover, this is a very common syntax pitfall - dismissing it as undefined behavior is irresponsible.

peter-jerry-ye:
I understand your frustration. And I understand that it would be better if we could throw an `ConcurrentError` or things like that.

However, this is not very likely to happen as this would impose performance penalty on everyone that writes the code correctly. The fact that we chose internal iterator in the first place was to have better performance. Also we have aggressive optimizations to drop the bound checks at runtime. Adding runtime check is clearly the opposite of what we want.

I think what we can do is:

- provide runtime check for debug mode, so that at least they can be detected during testing (throwing an exception in this case. In fact you'll find `moon test --release --target js` would give you another behavior).
- as @hackwaly suggested, add linters.

BigOrangeQWQ:
Alright, I understand the trade-offs.

bungcip:
IMHO, I support adding runtime check in debug mode. I think that a good compromise. just like rust adding check for integer overflow in debug mode, but in release mode it just wrap around.

PR #3972, opened 2026-08-04
[draft] fix(array): give Array::iter/rev_iter/iter2 live semantics

bobzhang:
## Problem

`Array::iter` delegates to `self[:].iter()`, which freezes `(buf, len)` in a view at iterator creation. Shrink operations (`remove`, `pop`, `truncate`, `clear`, `drain`) go through `unsafe_truncate_to_length`, which nulls the vacated slots for GC. A later iterator step then reads a nulled slot through `unsafe_get` against the stale bounds and hands out an invalid value of type `T` — a null deref/trap on native and wasm-gc, `undefined` leaking in on JS. See #2195.

Growth has the dual problem: a `push` that reallocates detaches the iterator onto the old buffer.

## Fix

`Array::iter`, `Array::rev_iter`, and `Array::iter2` now close over the array itself and re-read the **live** length and buffer at every step, with the element access adjacent to the bounds test. `unsafe_get` remains sound under any structural mutation because the bounds evidence never outlives the step that produced it:

- elements appended during iteration are visited;
- removing an element shifts its successors left, so the element moving into the current slot is skipped (the classic live-iteration wart — but type-safe and deterministic);
- iteration stops as soon as the array shrinks to the cursor; `rev_iter` clamps its cursor to the live length.

Cost: two dependent loads per step (`self.len`, `self.buf`) instead of two registers. When the loop body provably cannot mutate the array, both loads are loop-invariant and hoisting recovers the old codegen exactly; when the body is opaque, the loads are precisely the price of soundness.

**Intentionally unchanged:** `ArrayView::iter` (a view's contract *is* a frozen `(buf, start, len)` snapshot — `xs[:].iter()` stays the explicit opt-in for snapshot iteration with the hoisted codegen) and `FixedArray::iter` (buffer can never move or resize, so the hoisted form is already sound).

## Caveat: `for x in xs` is compiler-specialized

`for x in xs` over an `Array` does **not** go through `Array::iter` — the compiler specializes the lowering with the same frozen-snapshot pattern, so it keeps the old behavior until the compiler is updated to match (tests in this PR drive the iterators explicitly for exactly this reason; internal issue compiler/ideas#2130 tracks the compiler side). Helpers written with `for .. in` (`each`, `eachi`, `fold`) therefore also keep snapshot behavior for now; their docs now say so explicitly instead of claiming UB territory.

No `.mbti` changes — the public interface is untouched. Tests pass on wasm-gc, native, and js targets.

## Follow-ups (out of scope here)

- Compiler: switch the `for .. in` Array specialization to the same check-adjacent live loop, at which point `each`/`eachi`/`fold` inherit live semantics for free.
- Optional fail-fast: compare the live length against the creation-time length each step and abort on mismatch ("array modified during iteration"). Catches the common accident cheaply; full precision needs a mod-count word in the Array header — separate discussion.
- Views + shrink is a remaining (narrower) soundness hole independent of iteration: `let b = a[:]; a.pop() |> ignore; b[2]` reads a nulled slot through the view's stale bounds.

### DIFF
--- a/builtin/array.mbt
+++ b/builtin/array.mbt
@@ -415,10 +415,13 @@ pub impl[T] Add for Array[T] with fn add(self, other) {
 ///   inspect(sum, content="6")
 /// }
 /// ```
-/// This method uses the array iterator. Structural mutations during traversal
-/// are unsupported: appended elements are not visited, and shrinking the array
-/// with operations such as `remove`, `truncate`, `clear`, or `drain` may cause
-/// later iterator steps to fail.
+/// This method traverses the array via `for .. in`, whose current compiler
+/// lowering fixes the traversal bounds when iteration starts. Structural
+/// mutations during traversal are therefore unsupported here (unlike
+/// `Array::iter`, which has live semantics): appended elements are not
+/// visited, and shrinking the array with operations such as `remove`,
+/// `truncate`, `clear`, or `drain` may cause later steps to read invalid
+/// slots. Use `retain()` or `retain_map()` to delete elements by predicate.
 #locals(f)
 pub fn[T] Array::each(self : Array[T], f : (T) -> Unit raise?) -> Unit raise? {
   for v in self {
@@ -1306,13 +1309,12 @@ pub fn[T] Array::repeat(self : Array[T], times : Int) -> Array[T] {
 
 ///|
 /// Fold out values from an array according to certain rules.
-/// This method traverses the array through `self.iter()`, so the traversal
-/// bounds are fixed when folding starts.
-///
-/// Structural mutations to `self` inside `f` are unsupported. Appended
-/// elements are not visited, and shrinking the array with operations such as
-/// `remove`, `truncate`, `clear`, or `drain` may cause later fold steps to
-/// fail.
+/// This method traverses the array via `for .. in`, whose current compiler
+/// lowering fixes the traversal bounds when folding starts. Structural
+/// mutations to `self` inside `f` are therefore unsupported here (unlike
+/// `Array::iter`, which has live semantics): appended elements are not
+/// visited, and shrinking the array may cause later steps to read invalid
+/// slots.
 ///
 /// Example:
 ///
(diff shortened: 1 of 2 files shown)


bobzhang:
Closing this: upstream settled the question the other way, and the soundness motivation is separately gone.

## Snapshot semantics is the decided direction

`compiler/ideas!8676` (merged 2026-08-24, in nightly moonc from 2026-09-02) changed the `for x in arr` lowering to bind **both** `arr.len` and `arr.buf` before the loop — previously only `len` was frozen and `arr.buf` was reloaded each step. The stated rationale is to make `for a in arr` behave identically to `let it = arr.iter(); while it.next() is Some(a) { ... }` — i.e. it took the *current* `Array::iter` (`self[:].iter()`, a frozen view) as the reference, not the live version proposed here.

So this PR's follow-up bullet — "switch the `for .. in` Array specialization to the same check-adjacent live loop" — is not going to happen, and landing this PR would re-introduce the `for..in` / `.iter()` divergence with the polarity flipped.

## The safety motivation is gone

#4190 and #4189 stopped nulling vacated slots: `unsafe_truncate_to_length` now only lowers `self.len`, so a stale snapshot index reads a stale-but-valid `T` instead of trapping. The null-deref in this PR's problem statement no longer reproduces on native or wasm-gc.

Worth recording that live iteration was never buying correctness anyway — as this PR's own description notes, a removal still shifts successors left and skips an element. It swaps one silently-wrong answer for another and charges two dependent loads plus an unhoistable bounds test per step. The chosen semantics also matches Go's `range` over a slice, which evaluates the range expression once and copies the slice header (pointer *and* length), and which stays safe for exactly the reason MoonBit now does: reslicing doesn't zero the tail.

`Array::iter` therefore stays `self[:].iter()`. `ArrayView` is the semantics-defining type — it *is* a frozen `(buf, start, len)`, it has no other possible contract — and `Array::iter` is the `start = 0` instance of it.

## Still open: the JS backend

Measured on nightly 2026-09-02. Shrink = `truncate(2)` at step 1 of a 5-element array; grow = 32 pushes then a write to `arr[2]` mid-loop:

| case | native / wasm-gc | js |
|---|---|---|
| grow, both forms | `[1, 2, 3]` — old buffer, the later write is invisible | `[1, 2, 999]` — the captured buffer *is* the array |
| shrink, `for..in` | 5 steps, stale elements | 5 steps, **`undefined` leaks in as `T`** (an `Int` sum becomes `NaN\|0 == 0`) |
| shrink, `.iter()` | 5 steps, stale elements | stops after 2 steps |

On JS, `Array[T]` is a JS array — `JSArray::ofAnyArray` is `%identity`, `Array::length` is `arr.length` — and that representation is staying. So `arr.length = n` genuinely destroys the elements and there is no logical length to fall back on. Two notes for whoever picks up the fix:

- Core's `.iter()` is already accidentally safe on JS: `Option[X]` compiles to a nullable, so the `undefined` read from a hole past `.length` is indistinguishable from `None` and the `Iter` closure just terminates. That is why it stops at 2 steps. The real hole is the compiler's `for..in` lowering, which does not go through `Option`.
- The fix cannot be a live length read. On JS the captured buffer is the array object itself, so testing `i < arr.length` alone turns `for x in arr { arr.push(...) }` into an infinite loop. It has to be `i < snapshot_len && i < arr.length` — clamped to the snapshot, not live.

That clamp leaves JS live-clamped on shrink while native and wasm-gc keep handing out the stale elements, and the grow row already differs. Both are safe, but they are not the same observable behavior, so the documented contract should be "mutation during iteration is safe but unspecified" rather than a promise of snapshot semantics.

## Docs follow-up

`builtin/array.mbt` still says that shrinking during iteration "may cause later iterator steps to fail", which now describes none of the four behaviors in the table above.

#2195 stays open — the behavior it asks about is now defined on native and wasm-gc, but the JS `for..in` hole above is exactly what it reports.

PR #3977, opened 2026-08-05
feat(quickcheck): add char_range and zip/zip_with/zip_with3 generator combinators

bobzhang:
Follow-up to #3955, first of three scoped PRs from the feedback in https://github.com/moonbitlang/core/pull/3955#issuecomment-5189971358 (migrating bobzhang/toml's property tests to this package — moonbit-community/toml-parser#122).

- `char_range(lower, upper)`: characters in an inclusive range. Bounds are validated once at construction (non-empty, no surrogate code points), so the per-sample conversion uses safe `to_char()` — no `unsafe_to_char`.
- `zip` / `zip_with` / `zip_with3`: applicative combination of two or three generators, named after `Array`'s `zip_with`. Implemented via `flat_map`/`map` so they follow the existing state-splitting discipline.

Pure addition — no existing behavior or snapshot changes.

Test plan: new bounds/panic/snapshot tests in `generator_test.mbt`; `moon test quickcheck` 60/60; `moon check --deny-warn`, `moon fmt`, `moon info` clean.


quickcheck/generator.mbt
```diff
@@ -124,6 +154,24 @@ pub fn int_range(lower : Int, upper : Int) -> Generator[Int] {
   })
 }
 
+///|
+/// Generates a character in the inclusive range `[lower, upper]`.
+///
+/// The bounds must form a non-empty range of Unicode scalar values: this
+/// function aborts if `lower` exceeds `upper` or if the range contains
+/// surrogate code points, so every generated code point is a valid `Char`.
+pub fn char_range(lower : Char, upper : Char) -> Generator[Char] {
```
bobzhang, line 165: can we make char_range smarter to handle cases
```
lower .. 0xD800 0xDFFF..uper
```
instead of failing

quickcheck/generator.mbt
```diff
@@ -124,6 +154,24 @@ pub fn int_range(lower : Int, upper : Int) -> Generator[Int] {
...
+  guard lower <= upper else {
+    abort("char_range: lower bound exceeds upper bound")
+  }
+  guard upper.to_int() < 0xD800 || lower.to_int() > 0xDFFF else {
+    abort("char_range: range contains surrogate code points")
+  }
+  int_range(lower.to_int(), upper.to_int() + 1).map(code => {
```
bobzhang: not reusing int_range?

quickcheck/generator.mbt
bobzhang: Done in 4318770e — a range spanning the surrogate block now draws from the contiguous index space of its valid scalars and shifts past the block, so `char_range('\u{D000}', '\u{E800}')` works and never emits a surrogate (covered by a new test asserting both sides of the block are reached).

quickcheck/generator.mbt
bobzhang: It does reuse `int_range` — the draw is a single `int_range` over the collapsed index space (range width minus the surrogate gap), then a shift past the block in `map`. Kept that shape in the fix.

quickcheck/generator.mbt
```diff
@@ -124,6 +154,24 @@ pub fn int_range(lower : Int, upper : Int) -> Generator[Int] {
...
+    code.to_char().unwrap()
+  })
```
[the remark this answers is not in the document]
bobzhang: Good catch — fixed in 4318770e: bounds are now validated as Unicode scalar values (0..=0x10FFFF, non-surrogate), so an invalid `Char` produced via `unsafe_to_char` aborts at construction instead of panicking per sample.

CAIMEOX (approved):

PR #3983, opened 2026-08-06
perf(hashmap,hashset,set): annotate consuming parameters with #owned

Yu-zh:
## What

Adds `#owned(...)` to parameters consumed on every path — 11 annotations plus one enabling reorder:

- `HashMap`: `set`/`set_with_hash` **value** (stored on both the insert and the update path), and the `push_away`/`rehash_place_entry` entry plumbing
- `HashSet`: `push_away`/`set_entry`/`rehash_place_entry` entry plumbing
- `Set` (insertion-ordered): `push_away`/`set_entry`/`add_entry_to_tail`/`rehash_place_entry` entry plumbing, mirroring the pre-existing `#owned` annotations on builtin `Map`
- **Keys stay borrowed** everywhere: they are dropped on the key-already-exists path, where `#owned` would cost non-last-use callers an extra incref/decref pair (same policy as builtin `Map`)

The second commit reorders `Set::set_entry` to fix up neighbor links *before* the consuming store. Reading `entry.next` after the store forced the compiler to keep the owned reference alive across it; the naive annotation made the body worse (1 incref/1 decref → 2/4), while store-last makes it better than baseline (0/1). Worth remembering when annotating: field reads of an owned param belong before its consuming store.

## Generated code (native C)

Static incref/decref counts per emitted body, baseline → this branch. Every diffed change is an instruction removal; nothing is added:

| function | baseline | owned |
|---|---|---|
| HashMap::set_with_hash | 6 / 6 | 3 / 5 |
| HashMap::push_away | 2 / 2 | 1 / 2 |
| HashMap::rehash_place_entry | 3 / 3 | 1 / 2 |
| HashSet::add_with_hash | 2 / 3 | 2 / 2 |
| HashSet::push_away | 2 / 2 | 1 / 2 |
| HashSet::rehash_place_entry | 3 / 3 | 1 / 2 |
| Set::add_with_hash | 2 / 3 | 2 / 1 |
| Set::push_away | 2 / 2 | 1 / 0 |
| Set::set_entry | 1 / 1 | 0 / 1 |
| Set::add_entry_to_tail | 4 / 4 | 3 / 4 |
| Set::rehash_place_entry | 1 / 1 | 1 / 0 |

## Benchmarks

Native, 10k pre-built `String` keys, interleaved A/B binaries, medians over 10-round sessions. Hashing + Robin Hood probing dominates these paths (~50-90ns/op), so wins are small by construction:

- `Set::add` with growth: **−4~7%**, consistent across three sessions and under forced code alignment
- `HashMap::set` (pre-sized and growing): ±1%, within this rig's noise floor
- `HashSet::add` with growth: measured +4% at first — investigated, and it is a code-placement artifact, not the annotation: the executed hashset path contains strictly fewer instructions (table above; the bench loop itself is byte-identical), the +4% reproduces with a hashset-only binary and under `-falign-functions=64`, and collapses to **+0.5%** when every basic block is force-aligned (`-mllvm -align-all-blocks=5`), which removes placement luck symmetrically.

Control benches over untouched packages stayed within ±1% in the same runs.

### DIFF
--- a/hashmap/hashmap.mbt
+++ b/hashmap/hashmap.mbt
@@ -101,6 +101,7 @@ pub fn[K : Hash + Eq, V] HashMap::HashMap(
 /// }
 /// ```
 #alias("_[_]=_")
+#owned(value)
 pub fn[K : Hash + Eq, V] HashMap::set(
   self : HashMap[K, V],
   key : K,
@@ -110,6 +111,7 @@ pub fn[K : Hash + Eq, V] HashMap::set(
 }
 
 ///|
+#owned(value)
 fn[K : Eq, V] HashMap::set_with_hash(
   self : HashMap[K, V],
   key : K,
@@ -157,6 +159,7 @@ fn[K : Eq, V] HashMap::set_with_hash(
 }
 
 ///|
+#owned(entry)
 fn[K, V] HashMap::push_away(
   self : HashMap[K, V],
   idx : Int,
@@ -733,6 +736,7 @@ fn[K, V] HashMap::grow(self : HashMap[K, V]) -> Unit {
 }
 
 ///|
+#owned(entry)
 fn[K, V] HashMap::rehash_place_entry(
   self : HashMap[K, V],
   entry : Entry[K, V],
--- a/hashset/hashset.mbt
+++ b/hashset/hashset.mbt
@@ -114,6 +114,7 @@ fn[K : Eq] HashSet::add_with_hash(
 }
 
 ///|
+#owned(entry)
 fn[K] HashSet::push_away(
   self : HashSet[K],
   idx : Int,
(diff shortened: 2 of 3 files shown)


bobzhang:
## Review sign-off (Claude)

**LGTM — sign-off.** The Robin Hood entry helpers (`set_with_hash`, `push_away`, `set_entry`, `rehash_place_entry`) consume their entry on every branch — stored directly or handed to the next owned callee — and `HashMap::set` forwards into the owned `set_with_hash` with no boundary cost. The one code motion, moving `set_entry`'s neighbor-link fixup before the consuming store, is a verified-safe reorder: the link fixup touches `entry.next`/`self.tail`/neighbor `prev` while the store touches only `self.entries[new_idx]` — no overlap — and the comment explains the keep-alive rationale well.

PR #3986, opened 2026-08-06
perf(list,ref,lazy,lazy_list): annotate consuming parameters with #owned

Yu-zh:
## What

Adds `#owned(...)` to wrap-and-return constructors that store their argument in the new allocation on every path — 12 annotations:

- `List`: `cons`, `prepend`, `singleton`, `concat`/`rev_concat` (`other` is either returned or linked into the result)
- `Ref`: `Ref`/`new`
- `Lazy`: `Lazy` (the thunk is stored in the cell), `ready`
- `LazyList`: `cons`, `lazy_cons`, `concat`

## Generated code (native C)

Every annotated standalone body becomes a pure move: `ref::new` drops from 1 incref to 0 (all six monomorphic instances in the test binary), likewise `Lazy::ready`, `Lazy::Lazy`, `lazy_cons`; `LazyList::from_iter` goes 1 incref/3 decref → 1/1.

Two findings worth recording:

1. **Fully-inlined call sites already get this for free.** The hot `l = l.prepend(s)` loop compiles to byte-identical C with and without the annotation — once `prepend` is inlined, the RC optimizer already fuses the element-read incref into the node store. For tiny constructors like these, `#owned` only changes non-inlined boundaries: cross-package calls the inliner skips, function values, and generic instantiations kept out of line.
2. **The relocation equivalence is visible in the emitted code.** `Bench::keep` (which holds its argument borrowed and passes it to `ref::new`) gained exactly the incref that `ref::new` lost — the pair moves from callee to caller for non-last-use callers, it does not multiply.

## Benchmarks

Native, 10k pre-built `String` elements, interleaved A/B binaries, 12-round session on a quiet machine. As the codegen predicts, the workloads are neutral: `list_prepend` +0.3%, `ref_new` −0.1%, `lazy_ready` −0.2%, `lazy_list` build+force −0.0%, with controls over untouched packages at ±0.3% in the same run.

This branch is therefore a consistency/enablement change rather than a measured win: it completes the constructor family so non-inlined callers (and future callers in other packages) get the move semantics, at no cost to inlined ones.

### DIFF
--- a/lazy/lazy.mbt
+++ b/lazy/lazy.mbt
@@ -84,6 +84,7 @@ struct Lazy[A] {
 ///   @test.assert_eq(runs, 1)
 /// }
 /// ```
+#owned(thunk)
 pub fn[A] Lazy::Lazy(thunk : () -> A) -> Lazy[A] {
   { state: Unforced(thunk) }
 }
@@ -98,6 +99,7 @@ pub fn[A] Lazy::Lazy(thunk : () -> A) -> Lazy[A] {
 ///   @test.assert_eq(cell.force(), 7)
 /// }
 /// ```
+#owned(value)
 pub fn[A] Lazy::ready(value : A) -> Lazy[A] {
   { state: Forced(value) }
 }
--- a/lazy_list/lazy_list.mbt
+++ b/lazy_list/lazy_list.mbt
@@ -19,6 +19,7 @@
 ///|
 /// Strict-tail cons. Private — exposed indirectly through `lazy_cons`
 /// and as the building block for the transform combinators.
+#owned(head, tail)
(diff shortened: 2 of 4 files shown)


bobzhang:
## Review sign-off (Claude)

**LGTM — sign-off.** These are the cleanest cases in the campaign — pure single-store constructors (`List::cons`/`prepend`/`singleton`, `Ref::Ref`/`new`, `Lazy::Lazy`/`ready`, `lazy_cons`/`cons`) where the parameter goes straight into the built value, plus the two `concat`s, where `other` is consumed on every path: returned as-is on the `Empty` fast path, embedded as the rebuilt tail otherwise (and captured by the deferred thunk in the lazy case).

PR #3987, opened 2026-08-06
perf(sorted_map,sorted_set): annotate consuming parameters with #owned

Yu-zh:
## What

Adds `#owned(...)` to parameters consumed on every non-panic path in the two mutable AVL trees — 21 annotations plus one enabling reorder:

- `SortedMap`: `set`/`add_node` **value** (stored on both insert and replace paths; `key` stays borrowed, dropped on the duplicate path), `new_node` key+value, and the `balance`/`rotate_*` plumbing (the node is stored into the rotated tree or returned on every path)
- `SortedSet`: `add`/`add_node` value (duplicates are *replaced*, so the value is stored on every path), `singleton`, `new_node`/`new_node_update_height`, `join`/`join_left`/`join_right` (only the parameters each actually consumes — `join_left` keeps `r` borrowed since it destructures and drops that node, symmetrically `join_right` keeps `l`), `balance`/`rotate_*`

The second commit hoists `n.update_height()` above the store in `rotate_l`/`rotate_r` (both packages). With `#owned(n)`, the store consumes the reference, and touching `n` afterwards forced the compiler to keep it alive across the store — the naive annotation regressed those bodies (3 incref/2 decref → 3/3); store-last turns them into a net win (1/1 in sorted_map, 2/2 in sorted_set).

## Generated code (native C)

Total RC ops across all emitted `sorted_map`/`sorted_set` bodies: **416 → 368 (−12%)**. The heavy consumers carry it: `join_left`/`join_right` 22 ops → 11 each, `split` 22 → 16, the `new_node` family becomes pure moves (2-3 increfs → 0). A few borrowed callers (`add_node`, `delete_node`) gain +1-2 increfs — that is the relocation for calling now-owning callees (`balance`, `rotate_*`), paid once and saved more inside.

## Benchmarks

Native, 10k pre-built `String` elements, interleaved A/B binaries, two 10/8-round sessions on a quiet machine (spreads 1-3%, controls over untouched packages ±0.3-1.6%):

| bench | session 1 | session 2 |
|---|---|---|
| sorted_set union (5k ∪ 5k) | **−5.3%** | **−4.8%** |
| sorted_set add ×10k + remove ×10k | −2.2% | −2.4% |
| sorted_set add ×10k | −1.8% | −2.4% |
| sorted_map set ×10k | −1.3% | −1.1% |

`union` benefits most because it runs almost entirely inside the annotated `split`/`join`/`join_left`/`join_right` plumbing; the insert paths are diluted by string comparisons and node allocation.

### DIFF
--- a/sorted_map/map.mbt
+++ b/sorted_map/map.mbt
@@ -67,6 +67,7 @@ pub fn[K : Compare, V] SortedMap::SortedMap(
 /// ```
 #alias("_[_]=_")
 #alias(add, deprecated="Use set instead")
+#owned(value)
 pub fn[K : Compare, V] SortedMap::set(
   self : SortedMap[K, V],
   key : K,
@@ -567,6 +568,7 @@ fn[K, V] height_ge(x1 : Node[K, V]?, x2 : Node[K, V]?) -> Bool {
 }
 
 ///|
+#owned(root)
 fn[K, V] balance(root : Node[K, V]) -> Node[K, V] {
   let (l, r) = (root.left, root.right)
   let (hl, hr) = (height(l), height(r))
@@ -592,6 +594,7 @@ fn[K, V] balance(root : Node[K, V]) -> Node[K, V] {
 }
 
 ///|
+#owned(n)
 fn[K, V] rotate_l(n : Node[K, V]) -> Node[K, V] {
   let r = n.right.unwrap()
   n.right = r.left
@@ -602,6 +605,7 @@ fn[K, V] rotate_l(n : Node[K, V]) -> Node[K, V] {
 }
 
 ///|
+#owned(n)
 fn[K, V] rotate_r(n : Node[K, V]) -> Node[K, V] {
   let l = n.left.unwrap()
   n.left = l.right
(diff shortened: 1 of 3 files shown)


bobzhang:
## Review sign-off (Claude)

**LGTM — sign-off.** The most careful PR of the batch, and it holds up:

- The `rotate_l`/`rotate_r` reorderings (in both packages) are semantics-preserving: `n.update_height()` reads only `n`'s own fields, whose state is identical before and after the `r.left = Some(n)` store, and `r.update_height()` still runs after its new subtree is attached. The comments explain the last-use motivation clearly.
- The `join_left`/`join_right` annotations are deliberately asymmetric — the side that gets destructured for its fields (`r` in `join_left`, `l` in `join_right`) correctly stays borrowed since the node itself is never stored whole. That per-parameter precision is exactly right.
- `balance` and the rotations consume `root`/`n` on every branch (returned or handed to the next owned rotation); `add_node` consumes `value` exactly once per path (new node, replacement store, or recursion).

PR #3989, opened 2026-08-06
perf(builtin): annotate consuming Array/FixedArray parameters with #owned

Yu-zh:
## What

Adds `#owned(...)` to the value parameter of the builtin array store family — 17 annotations:

- `Array`: `push`, `insert`, `set`, `unsafe_set`, `make`, `fill`, `unsafe_resize_with_default`
- `FixedArray`: `fill`, `unchecked_fill`, `make_and_blit` (src, init)
- `UninitializedArray`: `unchecked_fill`, `make_and_blit` (src)
- `MutArrayView`: `set`, `unsafe_set`
- The wrappers forwarding unconditionally into the already-`#owned` blit intrinsics: `Array::unsafe_make_and_blit`, `Array::unsafe_make_and_blit_from_fixed`, `Array::from_fixed_array`

## Generated code (native C)

- `Array::push`: 1 incref/1 decref → 0/1 in every monomorphic instance — the store consumes the caller's reference
- The blit chain loses its `moonbit_incref(src)` before `moonbit_make_ref_array_with_blit`, which always consumes `src`. With the incref gone, a freshly built source arrives with refcount 1 and the runtime takes its steal path instead of copy-with-increfs — this is what makes `from_fixed_array` fast (below)
- Downstream bodies improve for free: `Array::resize_buffer` 1/2 → 1/1, `StringBuilder::grow`/`to_string`, `Json::stringify` internals — their array stores ride the same annotations
- `fill`/`make` gain +1-2 decrefs on their empty-range guard paths only: the owned surplus reference costs a constant ±1 op per call (against N per-element increfs in the loop, unchanged), which is the accepted trade of annotating this family. Benchmarks confirm neutrality
- No post-store-read hazards on this branch (scanned per-body; every hot-path change is a removal)

## Benchmarks

Native, 10k pre-built `String` elements, interleaved A/B binaries, two sessions plus an alignment-forced rebuild:

| bench | session 1 | session 2 | blocks 32B-aligned |
|---|---|---|---|
| `Array::push` (pre-sized) | **−8.6%** | **−8.9%** | −21.6% |
| `Array::push` (growing) | −3.8% | −4.5% | — |
| `FixedArray` → `Array::from_fixed_array` | **−41.0%** | **−40.4%** | −30.6% |
| `Array::make` / `fill` | ±0.3% | ±0.5% | — |

`from_fixed_array` converts a freshly built 16-element `FixedArray` per iteration: with the source uniquely owned, the runtime moves the elements instead of incref-ing each on copy and decref-ing each on free, eliminating the per-element RC traffic entirely.

One control note, for the record: a `sorted_set` control bench (no array involvement) read +5.3% against this branch's binary pair, and collapses to +0.2% when every basic block is force-aligned (`-mllvm -align-all-blocks=5`) — fixed-binary-pair code-placement luck, same phenomenon documented in #3983. The `Map` control stayed within ±0.4% throughout.

### DIFF
--- a/builtin/array.mbt
+++ b/builtin/array.mbt
@@ -32,6 +32,7 @@
 ///   debug_inspect(dynamic, content="[42, 42, 42]")
 /// }
 /// ```
+#owned(arr)
 pub fn[T] Array::from_fixed_array(arr : FixedArray[T]) -> Array[T] {
   let len = arr.length()
   Array::unsafe_make_and_blit_from_fixed(arr, allocate_len=len, len~)
@@ -71,6 +72,7 @@ pub fn[T] Array::from_fixed_array(arr : FixedArray[T]) -> Array[T] {
 /// ```
 /// This is because all the cells reference to the same object (the Array[Int] in this case).
 /// One should use makei() instead which creates an object for each index.
+#owned(elem)
 pub fn[T] Array::make(len : Int, elem : T) -> Array[T] {
   let arr = Array::make_uninit(len)
   for i in 0..<len {
@@ -228,6 +230,7 @@ pub fn[T] Array::get(self : Array[T], index : Int) -> T? {
 /// }
 /// ```
 #intrinsic("%array.unsafe_set")
+#owned(val)
 pub fn[T] Array::unsafe_set(self : Array[T], idx : Int, val : T) -> Unit {
   self.buffer().unsafe_set(idx, val)
 }
@@ -257,6 +260,7 @@ pub fn[T] Array::unsafe_set(self : Array[T], idx : Int, val : T) -> Unit {
 ///
 #intrinsic("%array.set")
 #alias("_[_]=_")
+#owned(value)
 pub fn[T] Array::set(self : Array[T], index : Int, value : T) -> Unit {
   let len = self.length()
   guard! index >= 0 && index < len
--- a/builtin/arraycore_nonjs.mbt
+++ b/builtin/arraycore_nonjs.mbt
@@ -29,6 +29,7 @@ fn[T] Array::make_uninit(len : Int) -> Array[T] {
 }
 
 ///|
+#owned(src)
 fn[T] Array::unsafe_make_and_blit(
   src : UninitializedArray[T],
   allocate_len~ : Int,
@@ -49,6 +50,7 @@ fn[T] Array::unsafe_make_and_blit(
 }
 
 ///|
+#owned(src)
 fn[T] Array::unsafe_make_and_blit_from_fixed(
   src : FixedArray[T],
   allocate_len~ : Int,
@@ -65,6 +67,7 @@ fn[T] Array::unsafe_make_and_blit_from_fixed(
 }
 
 ///|
+#owned(value)
 fn[T] Array::unsafe_resize_with_default(
   self : Array[T],
   new_len : Int,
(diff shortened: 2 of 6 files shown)


bobzhang:
## Review sign-off + independent perf verification (Claude)

Reviewed all 17 annotation lines, rebased onto current `main` (`d7c68d07`, no conflicts, pushed), and independently verified the performance claims. **LGTM — sign-off.**

### Review

- Every annotated parameter is genuinely consumed on the hot path (stored or forwarded to an already-`#owned` callee), and I found no post-store reads of an owned parameter in any annotated body — consistent with the PR's per-body scan.
- The `fill`/`make` guard-path trade (+1 decref on empty ranges, N per-element ops unchanged) is real but constant; benchmarks confirm neutrality within noise.
- `#owned` does not appear in generated interfaces — `moon info` produces zero drift.
- Full test matrix on the rebased tip: wasm-gc 7021/7021 (whole repo), **native 6931/6931 whole-repo** (the target where the convention generates code), builtin on all four targets green.
- Leak check: peak RSS is byte-identical between this branch and `main` across millions of `push`/`from_fixed_array`/`fill` bench iterations — no refcount-leak signal from the annotation flips.

### Independent benchmarks (different machine from the PR's)

Native ARM64, `moon bench --release`, two interleaved A/B sessions, identical bench file on both sides:

| bench | main | branch | session 1 | session 2 |
| `Array::push` pre-sized, 10k strings | 19.52 µs | 16.87 µs | **−13.6%** | **−13.7%** |
| `Array::push` growing, 10k strings | 44.59 µs | 41.67 µs | **−6.5%** | **−6.9%** |
| `Array::from_fixed_array`, fresh 16-elem | 73.21 ns | 59.74 ns | **−18.4%** | **−19.5%** |
| `Array::make` + `fill`, 64 strings | 157.83 ns | 161.04 ns | +2.0% | +0.9% (noise) |

Every claim reproduces in direction and stability; `push` actually does better here than the PR's −8.6/−8.9%. My `from_fixed_array` shows −19% rather than −41% because my loop builds the source via `FixedArray::makei` inside the measured region, diluting the ratio — the absolute delta is ~13.5 ns for 16 elements ≈ **0.85 ns/element, i.e. exactly one incref+decref pair removed per element**, which is the steal-path fingerprint the PR describes. Mechanism confirmed.

### Anything missing? (audit of the store family)

Two candidates, no blockers:

1. **`Array::resize(self, new_len, f : T)`** — forwards `f` into the now-`#owned` `unsafe_resize_with_default`, but only on the grow path, so callers still incref at that inner boundary. Under the same policy already accepted for `fill`/`make` (constant ±1 on the non-consuming path), `#owned(f)` here would extend the win to `resize` callers. Clearest follow-up.
2. **The bare-primitive definitions** — `FixedArray::set`/`unsafe_set`/`make` and `UninitializedArray::set`/`unsafe_set` (`= "%fixedarray.set"` style, no MoonBit body) carry no annotations. I probed that `#owned` *attaches* to such declarations without any compiler complaint; whether it affects codegen (vs. the prim table's built-in operand conventions) needs a compiler-side check. Worth either annotating for consistency or a comment documenting why they're excluded, so the next reader doesn't re-derive it.

Nothing else in the family qualifies: `MutArrayView::fill` doesn't exist, and `Array::append`/`blit_to` are element-wise copies where an owned container parameter wouldn't remove per-element traffic.

PR #3992, opened 2026-08-07
fix(quickcheck): omit empty categories from ObservationSummary debug rendering

bobzhang:
In the snapshot-first testing workflow, a report's `Debug` rendering is effectively the API surface — and every classes-only coverage snapshot (the common case when using `classify`) currently carries a noisy `labels: {}` field, e.g. in a downstream project (moonbit-community/toml-parser#122):

```
observations={ labels: {}, classes: { "contains-table": 570, ... } }
```

`ObservationSummary` now renders via a custom `Debug` impl that omits empty categories, so the above reads `observations={ classes: {...} }` (symmetrically, labels-only summaries drop `classes: {}`). The text `failure_report` path already skipped empty categories; this aligns the Debug path.

Test plan: new classes-only rendering test; one existing labels-only snapshot updated (drops `, classes: {}`); `moon test quickcheck` 68/68; `moon check --deny-warn`, `moon fmt`, `moon info` clean (no interface change — the type is private).

### DIFF
--- a/quickcheck/driver_test.mbt
+++ b/quickcheck/driver_test.mbt
@@ -474,14 +474,14 @@ test "report accepts InspectError while shrinking Raised" {
 test "classes-only summary omits the empty labels field" {
   let report = @quickcheck.report(
     (_ : Int) => true,
-    observe=input => [@quickcheck.classify(input >= 0, "non-negative")],
+    observe=_ => [@quickcheck.classify(true, "observed")],
     count=3,
     max_size=2,
   )
   debug_inspect(
     report,
     content=(
-      #|Passed(tests=3, observations={ classes: { "non-negative": 3 } })
+      #|Passed(tests=3, observations={ classes: { "observed": 3 } })
     ),
   )
 }
--- a/quickcheck/driver_test.mbt
+++ b/quickcheck/driver_test.mbt
@@ -135,7 +135,7 @@ test "discarded cases do not contribute observations" {
     content=(
       #|Passed(
       #|  tests=2,
-      #|  observations={ labels: { <List: ["1"]>: 1, <List: ["2"]>: 1 }, classes: {} },
+      #|  observations={ labels: { <List: ["1"]>: 1, <List: ["2"]>: 1 } },
       #|)
     ),
   )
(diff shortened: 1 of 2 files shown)


quickcheck/driver_test.mbt
```diff
@@ -469,3 +469,19 @@ test "report accepts InspectError while shrinking Raised" {
     ),
   )
 }
+
+///|
+test "classes-only summary omits the empty labels field" {
+  let report = @quickcheck.report(
+    (_ : Int) => true,
+    observe=input => [@quickcheck.classify(input >= 0, "non-negative")],
```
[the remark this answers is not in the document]
bobzhang: Not flaky as-is (the driver's default seed=37 makes the run deterministic), but the robustness point stands — the count would drift if Int generation or the default seed ever changed, and this test only exercises the rendering shape. Fixed in 1f9fa57a: the classification is now unconditional (`classify(true, "observed")`), so the count always equals the test count.

quickcheck/observation.mbt
```diff
@@ -33,7 +33,22 @@ type Observations = Array[Observation]
...
+/// `labels: {}` field (and vice versa).
+impl @debug.Debug for ObservationSummary with fn to_repr(self) {
+  let fields : Map[String, @debug.Repr] = Map([])
+  if !self.labels.is_empty() {
+    fields["labels"] = Repr(self.labels)
+  }
+  if !self.classes.is_empty() {
+    fields["classes"] = Repr(self.classes)
+  }
+  @debug.Repr::record(fields)
```
bobzhang:
```
@debug.Repr(Map[
         .. if !self.labels.is_empty() { [ ("labels", Repr(self.labels)]},
          .. if !self.class.is_empty() { ...}
         ])
```

bobzhang: Done — the record is now built with conditional spreads in the literal. One deliberate deviation from the snippet: kept `Repr::record(...)` instead of `Repr(Map(...))`, since rendering the summary as a Map would quote the keys (`{ "labels": ... }`) and churn every observation snapshot; record keeps the unquoted field style.

PR #4000, opened 2026-08-07
Add uniform Float and Double range sampling

CAIMEOX:
## Summary

Add `Rand::float_range` and `Rand::double_range` for uniformly sampling representable floating-point values from a half-open interval `[min, max)`.

The implementation follows the corrected gamma-section algorithm from Frédéric Goualard, [“Drawing Random Floating-point Numbers from an Interval” (v5)](https://hal.science/hal-03282794v5/document).

## Why

The naive implementation:

```text
min + (max - min) * random()
```

is not generally uniform over the representable values in the interval. Intermediate floating-point rounding can give some values more probability than others and can even return the excluded upper bound. The gamma-section algorithm instead divides the interval using an endpoint-derived spacing and selects those sections uniformly.


random/pkg.generated.mbti
```diff
@@ -21,7 +21,9 @@ pub fn Rand::bigint(Self, Int) -> @bigint.BigInt
 pub fn Rand::boolean(Self) -> Bool
 pub fn Rand::chacha8(seed? : Bytes) -> Self
 pub fn Rand::double(Self) -> Double
+pub fn Rand::double_range(Self, min~ : Double, max~ : Double) -> Double
```
bobzhang: a single API?

```
double(Self, min? : Double =0., max? : Double=1.) 
```

random/pkg.generated.mbti
```diff
@@ -21,7 +21,9 @@ pub fn Rand::bigint(Self, Int) -> @bigint.BigInt
...
 pub fn Rand::float(Self) -> Float
+pub fn Rand::float_range(Self, min~ : Float, max~ : Float) -> Float
```
bobzhang: same

PR #4012, opened 2026-08-12
refactor(builtin): default `end` to `self.length()` on view constructors

Yu-zh:
## What

The view/slice constructors in `builtin` all declared `end? : Int` with no default and then re-derived the missing case inside the body, each with a slightly different idiom:

```moonbit
let end = match end { Some(end) | (None with end = len) => end }
let end_offset = if end_offset is Some(o) { o } else { self.length() }
match end { None => fixed[start:], Some(e) => fixed[start:e] }
```

A default-argument expression can reference an earlier parameter, so `end? : Int = self.length()` states the same thing in the signature. The body then works with a plain `Int` and the unwrapping goes away — 26 functions, −62 lines net.

Covered types: `ArrayView`, `MutArrayView`, `BytesView`, `StringView`, `ReadOnlyArray`, `UninitializedArray` (plus the deprecated `String::substring`).

## Why it is behavior-preserving

- **`len`-based cases** — the default is literally the `len` the body already computed.
- **`StringView::{get_view,sub}`** used `self.end()` in the `None` branch. `self.start() + self.length()` is the same value, since `StringView::length()` is defined as `end() - start()`.
- **`{String,StringView}::clamped_view`** clamped the `Some` branch into `[0, len]`. `self.length()` is already in range, so it clamps to itself.
- **`ReadOnlyArray::view`** fell back to `fixed[start:]`, whose own default is the fixed array's length — and `ReadOnlyArray::length()` is defined as exactly that.

`moon info` produces **no `.mbti` diff**: the generated interface renders both forms as `end? : Int`, so this is not a source- or interface-breaking change for downstream users.

## Not converted

- `Iter::view` genuinely needs the `Option`: an iterator has no length, and `None` selects a `drop`-only fast path that `Some` cannot express.
- `{Array,FixedArray}::fill` take a range but are not view constructors, so they are left alone here.

## Verification

```
moon check --deny-warn --target all          # clean
moon info --target wasm,wasm-gc,js,native    # no diff
moon fmt                                     # no diff
moon bundle --all                            # clean
moon test --target all                       # wasm 6966 / wasm-gc 6967 / js 6911 / native 6885, 0 failed
moon test --release --target js,wasm,wasm-gc # 0 failed
moon test --release --target native          # 6885 passed, 0 failed
MOONC_INTERNAL_PARAMS='use_js_builtin_string = 1 |' \
  moon test --target wasm-gc [--release]     # 6967 passed, 0 failed
```

### DIFF
--- a/builtin/arrayview.mbt
+++ b/builtin/arrayview.mbt
@@ -254,12 +254,9 @@ pub fn[T] ArrayView::unsafe_get(self : ArrayView[T], index : Int) -> T {
 pub fn[T] Array::view(
   self : Array[T],
   start? : Int = 0,
-  end? : Int,
+  end? : Int = self.length(),
 ) -> ArrayView[T] {
   let len = self.length()
-  let end = match end {
-    Some(end) | (None with end = len) => end
-  }
   guard start >= 0 && start <= end && end <= len else {
     abort("View index out of bounds")
   }
@@ -301,12 +298,9 @@ pub fn[T] Array::view(
 pub fn[T] Array::get_view(
   self : Array[T],
   start? : Int = 0,
-  end? : Int,
+  end? : Int = self.length(),
 ) -> ArrayView[T]? {
   let len = self.length()
-  let end = match end {
-    Some(end) | (None with end = len) => end
-  }
   guard start >= 0 && start <= end && end <= len else { None }
   Some(ArrayView::make(self.buffer(), start, end - start))
 }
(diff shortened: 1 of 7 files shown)


builtin/arrayview.mbt
```diff
@@ -254,12 +254,9 @@ pub fn[T] ArrayView::unsafe_get(self : ArrayView[T], index : Int) -> T {
...
   self : Array[T],
   start? : Int = 0,
-  end? : Int,
+  end? : Int = self.length(),
 ) -> ArrayView[T] {
   let len = self.length()
-  let end = match end {
-    Some(end) | (None with end = len) => end
-  }
   guard start >= 0 && start <= end && end <= len else {
```
bobzhang, line 260: what's the motivation? for easy inlining?
Note this could introduce some duplicated computation

Yu-zh: duplicated computations are caused by the wrappers and I plan to inline them as well (if the size of default values are reasonably small)

PR #4016, opened 2026-08-12
fix(bigint): make bitwise ops depend on value, not construction history

### LINKED ISSUE #3810: [BUG] BigInt XOR result depends on construction history
**Describe the bug**
Two BigInt values that compare equal can produce different results when used in the same XOR expression.

A value parsed from hexadecimal behaves differently from the equivalent BigInt literal, indicating that bitwise XOR depends on internal storage capacity or construction history rather than only on numeric value.

**To Reproduce**
Steps to reproduce the behavior, add the following test:

```moonbit
test "xor depends on value, not construction history" {
  let from_hex = @bigint.BigInt::from_string("80000000", radix=16)
  let literal = 2147483648N

  assert_eq(from_hex, literal)
  assert_eq(from_hex ^ -from_hex, literal ^ -literal)
}
```

Run it on a non-JavaScript backend:

```bash
moon test --target wasm
```

1. The first assertion passes, confirming that both operands are equal.
2. The second assertion fails:

```
0 != -4294967296
```

The observed results are:

```
from_hex ^ -from_hex = 0
literal ^ -literal   = -4294967296
```

**Expected behavior**
Equal BigInt values should be substitutable in all arithmetic and bitwise operations.

For `x = 2147483648`, using arbitrary-precision two’s-complement semantics:

```
x ^ -x = -4294967296
```

Both construction paths should therefore produce `-4294967296`.

ChenJiasen-123:
# Fix: Make `BigInt` bitwise operations depend on the value, not on construction history

> close #3810

## Summary

The `BitAnd`, `BitOr` and `BitXOr` implementations of `@bigint.BigInt` computed their two's-complement working width from `limbs.length()` (the allocated `FixedArray` capacity) instead of the logical number of significant limbs `len`. Because different construction paths allocate different capacities for numerically equal values, the result of a bitwise operation depended on **how** the operands were constructed rather than on their value.

> *Two `BigInt` values that compare equal must be substitutable in every arithmetic and bitwise operation.*

## Reproduction

On the non-JavaScript backend:

```moonbit
test "xor depends on value, not construction history" {
  let from_hex = @bigint.BigInt::from_string("80000000", radix=16)
  let literal  = 2147483648N            // == from_hex

  assert_eq(from_hex, literal)                         // passes
  assert_eq(from_hex ^ -from_hex, literal ^ -literal)  // FAILS
}
```

Observed before the fix:

- `from_hex ^ -from_hex` = `0`
- `literal   ^ -literal`   = `-4294967296` = `-2^32`

## Root cause

The `BigInt` struct keeps a magnitude array `limbs` (`FixedArray[UInt]`, base `2^32`) alongside a `len` field. By invariant, `limbs` may contain leading zeros, so the struct comment explicitly forbids using `limbs.length()`; one must use `len`. The bitwise implementations violated this in two places.

### 1. Working width used capacity instead of length

```moonbit
let max_length = if self.limbs.length() < other.limbs.length() {
  other.limbs.length() + 1
} else {
  self.limbs.length() + 1
}
```

`from_string("80000000", radix=16)` goes through `from_string_radix_pow2`, which allocates exactly 1 limb, while the literal `2147483648N` goes through `from_int64` → `from_uint64`, which allocates 2 limbs. Equal values therefore received different working widths.

### 2. Result dropped the "sign extension" limb

The result kept only `max_length - 1` limbs:

```moonbit
let limbs = FixedArray::make(max_length - 1, 0U)
limbs.unsafe_blit(0, x_limbs, 0, max_length - 1)
```

For `x = 2^31` we have `x ^ (-x) = -2^32`, whose magnitude is `[0, 1]` in base `2^32` — i.e. it needs **two** limbs. With `max_length = 2` only 1 limb survives, the top limb `1` is truncated, and the result collapses to `0`.

Because the two construction paths chose different `max_length`, one path "happened" to be wide enough and the other was not.

### Affected file and functions

| File | Trait method | Lines (approx.) |
|------|-------------|-----------------|
| `bigint/bigint_nonjs.mbt` | `BitAnd::land` | 1569–1634 |
| | `BitOr::lor` | 1659–1724 |
| | `BitXOr::lxor` | 1749–1814 |

All three implementations share the same structure and received the same three changes. Below, each change is shown as a before/after diff.

### Change 1: working width uses `len` instead of `limbs.length()`

**Before:**
```moonbit
let max_length = if self.limbs.length() < other.limbs.length() {
  other.limbs.length() + 1
} else {
  self.limbs.length() + 1
}
```

**After:**
```moonbit
let max_length = if self.len < other.len {
  other.len + 1
} else {
  self.len + 1
}
```

**Why:** `limbs.length()` is the allocated `FixedArray` capacity, which varies with the construction path even for equal values. `len` is the number of significant limbs and depends only on the value, as required by the struct invariant.

### Change 2: copy only the significant limbs

**Before:**
```moonbit
x_limbs.unsafe_blit(0, self.limbs, 0, self.limbs.length())
y_limbs.unsafe_blit(0, other.limbs, 0, other.limbs.length())
```

**After:**
```moonbit
x_limbs.unsafe_blit(0, self.limbs, 0, self.len)
y_limbs.unsafe_blit(0, other.limbs, 0, other.len)
```

**Why:** Copying the full capacity could include leading-zero limbs beyond `len`. Although those zeros are harmless for positive numbers, they are redundant; copying only `len` limbs is cleaner and consistent with the new width calculation. The remaining slots in `x_limbs`/`y_limbs` stay zero-initialized by `FixedArray::make`.

### Change 3: keep all `max_length` limbs in the result

**Before:**
```moonbit
let limbs = FixedArray::make(max_length - 1, 0U)
limbs.unsafe_blit(0, x_limbs, 0, max_length - 1)
{ limbs, sign: new_sign, len: normalize_len(limbs, max_length - 1) }
```

**After:**
```moonbit
let limbs = FixedArray::make(max_length, 0U)
limbs.unsafe_blit(0, x_limbs, 0, max_length)
{ limbs, sign: new_sign, len: normalize_len(limbs, max_length) }
```

**Why:** The old code discarded the top limb (the "sign extension" slot). For negative results whose magnitude needs one more limb than the operands (e.g. `x ^ (-x) = -2^32` with magnitude `[0, 1]`), that top limb carries real data and must not be dropped. Keeping all `max_length` limbs and letting `normalize_len` strip trailing zeros is always safe: positive results have a zero top limb that gets trimmed, while negative results keep the limb they need.

### Correct values after the fix

For `x = 2^31`:

| Expression | Result |
|-----------|--------|
| `x & (-x)` | `2^31` = `2147483648` |
| `x \| (-x)` | `-2^31` = `-2147483648` |
| `x ^ (-x)` | `-2^32` = `-4294967296` |

## Regression test

The new test compares both construction paths and asserts the exact expected values (not only mutual equality, which would also pass when both paths are consistently wrong):

```moonbit
test "bitwise ops depend on value, not construction history" {
  let from_hex = @bigint.BigInt::from_string("80000000", radix=16)
  let literal = 2147483648N
  assert_eq(from_hex, literal)

  assert_eq(from_hex & -from_hex, literal & -literal)
  assert_eq(from_hex & -from_hex, 2147483648N)

  assert_eq(from_hex | -from_hex, literal | -literal)
  assert_eq(from_hex | -from_hex, -2147483648N)

  assert_eq(from_hex ^ -from_hex, literal ^ -literal)
  assert_eq(from_hex ^ -from_hex, -4294967296N)
}
```

## Validation

| Check | Result |
|-------|--------|
| `moon test --target wasm -p bigint` | 528 / 528 passed |
| `moon info` | interface (`.mbti`) unchanged |
| `moon fmt` | formatting clean |

The generated public interface (`bigint/pkg.generated.mbti`) is unchanged: this is a pure internal implementation fix with no public API impact.

### DIFF
--- a/bigint/bigint_nonjs.mbt
+++ b/bigint/bigint_nonjs.mbt
@@ -1567,16 +1567,16 @@ pub fn BigInt::to_octets(self : BigInt, length? : Int) -> Bytes {
 /// }
 /// ```
 pub impl BitAnd for BigInt with fn land(self : BigInt, other : BigInt) -> BigInt {
-  let max_length = if self.limbs.length() < other.limbs.length() {
-    other.limbs.length() + 1
+  let max_length = if self.len < other.len {
+    other.len + 1
   } else {
-    self.limbs.length() + 1
+    self.len + 1
   }
   // Extend the limbs to store the sign bits
   let x_limbs = FixedArray::make(max_length, 0U)
-  x_limbs.unsafe_blit(0, self.limbs, 0, self.limbs.length())
+  x_limbs.unsafe_blit(0, self.limbs, 0, self.len)
   let y_limbs = FixedArray::make(max_length, 0U)
-  y_limbs.unsafe_blit(0, other.limbs, 0, other.limbs.length())
+  y_limbs.unsafe_blit(0, other.limbs, 0, other.len)
 
   // Calculate the complement code per 2
   if self.sign == Negative {
@@ -1628,9 +1628,9 @@ pub impl BitAnd for BigInt with fn land(self : BigInt, other : BigInt) -> BigInt
   }
 
   // return the result
-  let limbs = FixedArray::make(max_length - 1, 0U)
-  limbs.unsafe_blit(0, x_limbs, 0, max_length - 1)
-  { limbs, sign: new_sign, len: normalize_len(limbs, max_length - 1) }
+  let limbs = FixedArray::make(max_length, 0U)
+  limbs.unsafe_blit(0, x_limbs, 0, max_length)
+  { limbs, sign: new_sign, len: normalize_len(limbs, max_length) }
 }
 
 ///|
@@ -1657,16 +1657,16 @@ pub impl BitAnd for BigInt with fn land(self : BigInt, other : BigInt) -> BigInt
 /// }
 /// ```
 pub impl BitOr for BigInt with fn lor(self : BigInt, other : BigInt) -> BigInt {
-  let max_length = if self.limbs.length() < other.limbs.length() {
-    other.limbs.length() + 1
+  let max_length = if self.len < other.len {
+    other.len + 1
   } else {
-    self.limbs.length() + 1
+    self.len + 1
   }
   // Extend the limbs to store the sign bits
   let x_limbs = FixedArray::make(max_length, 0U)
-  x_limbs.unsafe_blit(0, self.limbs, 0, self.limbs.length())
+  x_limbs.unsafe_blit(0, self.limbs, 0, self.len)
   let y_limbs = FixedArray::make(max_length, 0U)
-  y_limbs.unsafe_blit(0, other.limbs, 0, other.limbs.length())
+  y_limbs.unsafe_blit(0, other.limbs, 0, other.len)
 
   // Calculate the complement code per 2
   if self.sign == Negative {
@@ -1718,9 +1718,9 @@ pub impl BitOr for BigInt with fn lor(self : BigInt, other : BigInt) -> BigInt {
   }
 
   // return the result
-  let limbs = FixedArray::make(max_length - 1, 0U)
-  limbs.unsafe_blit(0, x_limbs, 0, max_length - 1)
-  { limbs, sign: new_sign, len: normalize_len(limbs, max_length - 1) }
+  let limbs = FixedArray::make(max_length, 0U)
+  limbs.unsafe_blit(0, x_limbs, 0, max_length)
+  { limbs, sign: new_sign, len: normalize_len(limbs, max_length) }
 }
 
 ///|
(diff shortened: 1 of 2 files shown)


bobzhang:
Reviewed and verified — merging. Summary of the verification performed:

- **Diff audit**: the three changes are applied uniformly to `land`/`lor`/`lxor`; the unchanged two's-complement machinery (the `top == 0xFFFFFFFF` negativity test, the carry/borrow loops) remains sound under the new widths precisely because `max(len) + 1` guarantees a pure sign limb above both operands — and a negative result's magnitude is bounded by `2^(32W−1)`, so keeping all `max_length` limbs always suffices.
- **Completeness**: the only remaining `limbs.length()` read on a `BigInt` field (division) is benign under the `limbs[len..] == 0` invariant; all other uses are on local working buffers.
- **Oracle test**: ~2,700 assertions at this head — boundary pairs (±2^31, ±2^32, limb-straddling, Int64 extremes) plus 599 pseudo-random pairs — checking all three operators against the `Int64` ground truth through two construction paths: all exact.
- **Negative control**: with main's pre-fix implementation swapped back in, the regression test here fails with exactly the reported symptom (`0 != -4294967296`).

One observation: the regression inputs are well chosen — decimal `from_string` and `from_int64` allocate slack capacity that masks the truncation, so the `radix=16` path used here is the discriminating one (the other independent trigger is `Int64::min_value`, where `x ^ -x = -2^64` needs the extra limb). A follow-up PR will add a QuickCheck property suite for bigint (construction-path invariance × Int64 oracle × algebraic laws) in the same style as the recent `immut/hashmap`/`hashset` suites.

PR #4020, opened 2026-08-13
test(bigint): quickcheck property suite; fix(bigint): from_octets normalization

bobzhang:
## What

Follow-up to #4016: a QuickCheck property suite for `BigInt` in the style of the recent `immut/hashmap`/`hashset` suites — and it **found and fixed a real bug**.

### The suite

1. **Int64 oracle** — equality, `compare`, bitwise, and (widened) arithmetic must agree with the machine-integer result.
2. **Construction-path invariance** — one value, three internal representations: `from_string(radix=16)` (tight limbs), decimal `from_string` (slack), and **zero-padded `from_octets`** (swept over every padding amount up to four extra limbs). Every operation, plus `Hash`/`compare`/`is_zero`/renderings, must agree across all of them.
3. **Algebraic & two's-complement laws** at arbitrary precision — ring laws, truncated-division laws, bitwise identities/absorption/De Morgan, `x ^ -1 == -x - 1`, shift composition/cancellation/floor decomposition, round trips, and the negation identities `x & -x` / `x | -x` / `x ^ -x == -2·(x & -x)`.
4. **Deterministic negation-identity sweep** over single-bit values at positions `0..=199` through both string paths — pinning the #3810 trigger class (`32k − 1`), verified red on the pre-#4016 implementation.

### The bug it caught (red/green in this PR)

The deep review of #4016 had flagged by audit that `from_octets` looked suspicious. Adding it as a third invariance path made QuickCheck falsify **after 3 cases, shrinking to the minimal counterexample `([0], [0], false, false)`**: `from_octets` stripped at most *one* leading-zero limb (`len: max(1, limbs_len − 1)`), so zero-padded input — the normal shape for fixed-width crypto-style buffers — produced non-normalized values that print correctly but compare unequal to equal values, hash differently, and (for a padded zero) report `is_zero()` as false.

Fix: normalize with `normalize_len`, like every other constructor. The JS backend's independent `from_octets` was already value-correct; the suite now pins both.

A distribution lesson worth recording: the first version derived the padding from a generated integer and sailed green — the trigger needs pad ≥ 8 and the generator rarely produced it. Sweeping all pads deterministically inside the property removed the sensitivity. Path coverage *and* parameter coverage both matter.

## Validation

- Full repo suite: 7017/7017; `moon check --target all` clean; wasm-gc 169/169 and JS 144/144 for bigint; `moon fmt` stable
- Red verified at the test-only commit (`4ec25138`): the invariance property falsifies on the unfixed implementation

### DIFF
--- a/bigint/bigint_nonjs.mbt
+++ b/bigint/bigint_nonjs.mbt
@@ -1475,11 +1475,7 @@ pub fn BigInt::from_octets(input : BytesView, signum? : Int = 1) -> BigInt {
       limbs[i] = (limbs[i] << 8) | input[bytes_idx].to_uint()
     }
   }
-  if limbs[limbs_len - 1] == 0 {
-    { limbs, sign: Positive, len: max(1, limbs_len - 1) }
-  } else {
-    { limbs, sign: Positive, len: limbs_len }
-  }
+  { limbs, sign: Positive, len: normalize_len(limbs, limbs_len) }
 }
 
 ///|
--- a/bigint/moon.pkg
+++ b/bigint/moon.pkg
@@ -11,6 +11,7 @@ import {
 
 import {
   "moonbitlang/core/bench",
+  "moonbitlang/core/quickcheck",
   "moonbitlang/core/test",
 } for "test"
 
--- a/bigint/quickcheck_test.mbt
+++ b/bigint/quickcheck_test.mbt
@@ -38,7 +38,7 @@ fn hex_of(seeds : ArrayView[Int]) -> String {
 
 ///|
 /// A value built through the exactly-allocating hex path.
-fn via_hex(seeds : ArrayView[Int], negative : Bool) -> @bigint.BigInt raise {
+fn via_hex(seeds : ArrayView[Int], negative : Bool) -> @bigint.BigInt {
   let magnitude = @bigint.BigInt::from_string(hex_of(seeds), radix=16)
   if negative {
     -magnitude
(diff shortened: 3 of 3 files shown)


bobzhang:
Codex xhigh review round complete. Verdict on the substance: **the `from_octets` fix is correct for all representable inputs** (empty/all-zero/exact-multiple/partial-head-limb inputs, `signum` interaction, negative-zero avoidance, and no caller depends on the old shape), and **every stated property is mathematically valid** for the implemented semantics, including the negative-operand shift floor-decomposition and negation identities. It also executed the prebuilt non-JS and JS artifacts: all new tests pass on both.

It found two Medium generator degeneracies in the suite itself, both fixed in `d84f4578`:

- `Array::arbitrary` draws element `i` with size `i`, so `xs[0]` is structurally **always zero** — every "random" shift amount was 0, making the shift laws vacuous. Shift amounts now come from deterministic sweeps crossing every limb boundary.
- `Arbitrary<Int64>` yields `random % size` with size ≤ 100, so the Int64 oracle never saw a real 64-bit value. Each generated pair is now checked both raw and spread across the full 64-bit range via a fixed affine map.

Two Low pre-existing observations recorded for future reference (out of scope here): `from_octets` computes `len * 8` in 32-bit `Int`, so a ≥ 512 MB view would wrap to an invalid zero-limb value; and `0N.to_octets(length=0)` aborts on non-JS but returns empty bytes on JS.

PR #4056, opened 2026-08-14
fix(json): reject raw unpaired surrogates in string lexing instead of aborting

### LINKED ISSUE #4049: json: parser aborts on strings containing lone surrogates combined with escapes
## Summary

`@json.parse` crashes the process (an abort from a `guard!`, not a catchable `ParseError`) when a JSON string contains a **raw lone trailing surrogate** together with **any escape sequence**.

## Minimal repro

A 5-code-unit JSON text: `"` U+DC00 `\n` `"`

```moonbit
test {
  let lone_low = String::from_array([(0xDC00).unsafe_to_char()])
  // aborts in StringView::sub (builtin/stringview.mbt guard!) instead of parsing:
  ignore(@json.parse("\"" + lone_low + "\\n\""))
}
```

Escape-free strings with the same lone surrogate parse fine (the fast path accepts them), so this is an internal inconsistency as well as a totality violation: a parser should either succeed or raise its documented `ParseError`, never abort.

## Root cause

`json/lex_string.mbt`, `ParseContext::lex_string_slow` — the local `flush` helper slices the pending run with the **checked** `ctx.input[start:end]` (`StringView::sub`), which `guard!`-aborts when the code unit at a slice boundary is a trailing surrogate. The fast path in `lex_string` uses the bounds-check-only `view(start_offset~, end_offset~)` and therefore tolerates raw lone surrogates; only the slow path (taken whenever the string contains a backslash escape) crashes.

## Affected targets

All backends — wasm-gc, js, and native (the abort is in shared library code; reproduced on all three).

## Severity

High for robustness: any consumer parsing untrusted JSON text can be crashed (denial of service) by a short crafted input. Found by adversarial QuickCheck testing (the default `Arbitrary String` never generates lone surrogates, which is why existing property tests missed it).

bobzhang:
Fixes #4049

## The bug

`@json.parse` crashed the process (a `guard!` abort, not a catchable `ParseError`) on strings mixing a **raw lone trailing surrogate** with **any escape sequence** — minimal repro: the 5-code-unit text `"` U+DC00 `\n` `"`. Root cause: `lex_string_slow`'s `flush` sliced the pending run with the checked `ctx.input[start:end]`, which aborts when a slice boundary lands on a trailing surrogate. Meanwhile the escape-free fast path silently **accepted** raw lone surrogates, producing ill-formed strings.

## The fix: reject raw unpaired surrogates

Per the project's design direction — *"we should maintain String unicode safe, so `unsafe_to_char` is indeed unsafe"* — this is robustness/defense-in-depth: under the well-formedness invariant such input should not exist (a well-formed `String` cannot contain a raw lone surrogate), but unsafe code can manufacture one, and a clean, documented error beats a process abort or an ill-formed parse result. The string lexer now rejects raw unpaired surrogates with `ParseError` (`InvalidChar`, sensible position) on **both** the fast path (previous silent acceptance removed) and the slow path, and `flush` slices with the bounds-check-only `view(start_offset~, end_offset~)`, so **no abort path remains**.

Well-formed pairs — raw astral characters, including next to escapes — parse exactly as before.

## Scope

Raw (unescaped) surrogates only. Unpaired `\uXXXX` **escape** sequences (which currently manufacture ill-formed strings from valid ASCII input) are tracked in #4062 and fixed in a separate PR; rejection there aligns with serde_json (Python and JS accept lone surrogates, Go substitutes U+FFFD).

## Tests

Deterministic regressions in `json/lex_string_test.mbt`: escape-free raw lone surrogates (leading, trailing, reversed pair), raw lone surrogates next to escapes (the former abort), the exact `ParseError` shape, and still-accepted well-formed pairs with a stringify/parse roundtrip.

Verified: `moon check` clean; json suite green on **wasm-gc, js, native**; no `.mbti` changes. The adversarial QuickCheck suite that found the bug is #4045.

### DIFF
--- a/json/lex_string.mbt
+++ b/json/lex_string.mbt
@@ -17,7 +17,10 @@ fn ParseContext::lex_string(ctx : ParseContext) -> String raise ParseError {
   let string_start = ctx.offset
   // Fast path for ordinary strings: scan raw UTF-16 code units and materialize
   // the slice directly when there are no escapes or control characters.
-  for i in string_start..<ctx.end_offset {
+  for i = string_start {
+    if i >= ctx.end_offset {
+      break
+    }
     let c = ctx.input.unsafe_get(i)
     if c == '"' {
       ctx.offset = i + 1
@@ -29,7 +32,23 @@ fn ParseContext::lex_string(ctx : ParseContext) -> String raise ParseError {
       // \t) are invalid inside a JSON string.
       ctx.offset = i + 1
       ctx.invalid_char(shift=-1)
+    } else if c.is_leading_surrogate() {
+      if i + 1 < ctx.end_offset &&
+        ctx.input.unsafe_get(i + 1).is_trailing_surrogate() {
+        continue i + 2
+      }
+      // MoonBit strings stay Unicode well-formed, so a raw unpaired
+      // surrogate must be rejected rather than smuggled into the parsed
+      // string. (Well-formed input cannot contain one, but unsafe code can
+      // manufacture such a String; a clean error beats undefined behavior.)
+      ctx.offset = i + 1
+      ctx.invalid_char(shift=-1)
+    } else if c.is_trailing_surrogate() {
+      // A bare trailing surrogate can never start a surrogate pair.
+      ctx.offset = i + 1
+      ctx.invalid_char(shift=-1)
     }
+    continue i + 1
   }
   raise InvalidEof
 }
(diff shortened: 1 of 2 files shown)


bobzhang:
**Independent review by OpenAI Codex CLI** (codex-cli 0.147.0, read-only session; posted on its behalf):

---

No blocking findings.

- The string fix is correct: `view` preserves raw UTF-16 code units while still checking bounds. Lexer offsets always fall after the opening quote/escape or before an ASCII quote/backslash, so this cannot split a valid raw surrogate pair. Control-character and escape validation remain unchanged.
- Coverage includes both surrogate kinds before and after escapes, including the previously crashing lone-low-surrogate boundaries.
- Applying the sign after exact `Int64`→`Double` conversion correctly produces `-0.0` and is equivalent for every nonzero integer in the safe range. Larger integers and decimal/exponent paths remain untouched.
- Negative-zero tests inspect the sign bit directly and cover integer, fractional, exponent, and underflow spellings. Existing tests cover the `2^53` slow-path boundary; adding explicit `±(2^53-1)` cases would be optional rather than required.
- The changes are minimal, idiomatic, and introduce no API or representation changes.

Verdict: LGTM

PR #4057, opened 2026-08-14
fix(diff): stop emitting zero-length Equal edits from group(context=0)

### LINKED ISSUE #4051: diff: group(context=0) emits zero-length Equal edits, violating the documented Edit invariant
## Description

`Diff::group(context=0)` produces hunks whose `edits()` contain zero-length `Equal` edits whenever an interior equal run is split. The `Edit` enum documents (diff/edit.mbt): "edits never have zero length, and adjacent edits of the same kind are coalesced" — so every consumer of `Hunk::edits()` may rely on positive lengths, and this output violates that invariant.

## Minimal repro

```moonbit
test "context=0 zero-length Equal edits" {
  let old = [1, 9, 2, 2, 2, 3, 9, 4][:]
  let new = [1, 8, 2, 2, 2, 3, 7, 4][:]
  let hunks = @diff.Diff(old~, new~).group(context=0)
  for h in hunks {
    debug_inspect(h.edits(), content="")
  }
}
```

Actual output — the first hunk ends with a zero-length `Equal`, the second begins with one:

```
[
  Delete(old_index=1, old_len=1, new_index=1),
  Insert(old_index=2, new_index=1, new_len=1),
  Equal(old_index=2, new_index=2, len=0),
]
[
  Equal(old_index=6, new_index=6, len=0),
  Delete(old_index=6, old_len=1, new_index=6),
  Insert(old_index=7, new_index=6, new_len=1),
]
```

Rendered output (`header()` / `render()`) happens to be unaffected because a zero-length edit contributes no lines, so the bug is only visible through the structured `Hunk::edits()` API — e.g. custom renderers that branch on the first/last edit kind, or any consumer trusting the documented invariant.

## Root cause

In `group_edits` (diff/edit.mbt), the split branch for a large interior equal run unconditionally pushes a context-sized `Equal` trailer into the closing hunk and seeds the next hunk with a context-sized `Equal` leader:

```moonbit
if len > context * 2 {
  pending.push(Equal(old_index~, new_index~, len=context))       // len=0 when context=0
  result.push({ edits: pending, old, new })
  let offset = len.saturating_sub(context)
  pending = [Equal(old_index=old_index + offset, ..., len=len - offset)] // len=0 when context=0
}
```

With `context=0` both pushed edits have `len=0`. The fix is to skip both pushes when `context == 0` (they carry no elements).

Found by a QuickCheck property that applies grouped hunks back onto `old` as a patch while enforcing the documented `Edit` invariants.

bobzhang:
Fixes #4051

`group(context=0)` emitted zero-length `Equal` edits around every split of an interior equal run: the split branch of `group_edits` unconditionally pushes a context-sized `Equal` trailer into the closing hunk and seeds the next hunk with a context-sized `Equal` leader, and with `context=0` both have `len=0`. That violates the documented "edits never have zero length" invariant of `Edit` as exposed through `Hunk::edits()` (rendered output was unaffected since empty edits contribute no lines).

Minimal repro:

```moonbit
let old = [1, 9, 2, 2, 2, 3, 9, 4][:]
let new = [1, 8, 2, 2, 2, 3, 7, 4][:]
@diff.Diff(old~, new~).group(context=0)
// hunk 0 ended with  Equal(old_index=2, new_index=2, len=0)
// hunk 1 began with  Equal(old_index=6, new_index=6, len=0)
```

The fix skips both context edits when `context == 0` (they carry no elements); a deterministic regression test in `diff_test.mbt` pins the corrected hunk shapes. Found by the QuickCheck property suite in #4047, which stacks on this branch.

Testing: `moon check` clean, `moon test -p moonbitlang/core/diff` passes, no `.mbti` changes.

### DIFF
--- a/diff/diff_test.mbt
+++ b/diff/diff_test.mbt
@@ -655,3 +655,28 @@ test "edits() with the views agrees with render()" {
     assert_eq(rebuilt.to_string(), h.render(show=l => l))
   }
 }
+
+///|
+test "group without context splits on every equal run and emits no empty edits" {
+  // Regression: `group(context=0)` used to emit zero-length `Equal` edits
+  // around each split of an interior equal run, violating the documented
+  // "edits never have zero length" invariant of `Edit`.
+  let old = [1, 9, 2, 2, 2, 3, 9, 4][:]
+  let new = [1, 8, 2, 2, 2, 3, 7, 4][:]
+  let hunks = @diff.Diff(old~, new~).group(context=0)
+  assert_eq(hunks.length(), 2)
+  assert_true(
+    hunks[0].edits()
+    is [
+      Delete(old_index=1, old_len=1, new_index=1),
+      Insert(old_index=2, new_index=1, new_len=1),
+    ],
+  )
+  assert_true(
+    hunks[1].edits()
+    is [
+      Delete(old_index=6, old_len=1, new_index=6),
+      Insert(old_index=7, new_index=6, new_len=1),
+    ],
+  )
+}
--- a/diff/edit.mbt
+++ b/diff/edit.mbt
@@ -116,16 +116,22 @@ fn[T] group_edits(
         }
         // Split if this equal range is large enough
         if len > context * 2 {
-          pending.push(Equal(old_index~, new_index~, len=context))
+          if context > 0 {
+            pending.push(Equal(old_index~, new_index~, len=context))
+          }
           result.push({ edits: pending, old, new })
           let offset = len.saturating_sub(context)
-          pending = [
-            Equal(
-              old_index=old_index + offset,
-              new_index=new_index + offset,
-              len=len - offset,
-            ),
-          ]
+          pending = if context > 0 {
+            [
+              Equal(
+                old_index=old_index + offset,
+                new_index=new_index + offset,
+                len=len - offset,
+              ),
+            ]
+          } else {
+            []
+          }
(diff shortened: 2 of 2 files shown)


bobzhang:
**Independent review by OpenAI Codex CLI** (codex-cli 0.147.0, read-only session; posted on its behalf):

---

No blocking issues found.

- The pre-fix claim is correct: with `context=0`, every positive interior `Equal` enters the split branch, which previously appended and prepended `Equal(..., len=0)`.
- Guarding both synthesized context edits is correct and minimal. Leading/trailing equal runs still trim away; changes at either endpoint work; adjacent change operations remain in one hunk; and every positive interior equal run separates change clusters.
- `context>0` behavior is unchanged because both new guards always take their original branches.
- Rendering and headers remain identical. Empty edits rendered no lines; trailing empty edits were ignored when computing hunk ends; and a removed leading empty edit has the same cursor coordinates as the following change.
- No consumer should rely on these empty edits: they violate the documented invariant, and the hunk boundary already conveys the split. Repository consumers only iterate the structured edits for rendering or change-pair detection.
- The regression test is adequate: one interior split exercises both the trailing and leading empty-edit sites, while asserting exact public `Hunk::edits()` shapes and coordinates. Existing tests cover zero-context single-hunk rendering and positive-context grouping.

Verdict: LGTM

bobzhang:
cc @myfreess

PR #4061, opened 2026-08-15
fix(json): preserve the sign of -0 in the integer fast path

### LINKED ISSUE #4053: json: parse("-0") loses the sign of negative zero
## Summary

`@json.parse("-0")` returns **+0.0**, while `parse("-0.0")` and `parse("-0e0")` return **-0.0**. The same mathematical literal parses to different IEEE-754 values depending on its spelling.

## Minimal repro

```moonbit
test {
  guard @json.parse("-0") is Number(a, ..)
  inspect(a.reinterpret_as_int64() < 0L, content="false") // sign lost (should be true)
  guard @json.parse("-0.0") is Number(b, ..)
  inspect(b.reinterpret_as_int64() < 0L, content="true")  // sign kept
}
```

RFC 8259 numbers are IEEE-754 doubles and every mainstream implementation (e.g. JS `JSON.parse("-0")`) yields -0. The bug is invisible to `Json` equality because `-0.0 == 0.0` under `Double` comparison, so it silently corrupts sign-sensitive data (`1.0 / x`, `atan2`, ...).

Negative values that *underflow* to zero (`-1e-400`, `-4.9e-325`) already preserve the sign correctly — only the plain integer spelling `-0` is affected.

## Root cause

`json/lex_number.mbt`, `ParseContext::lex_number_end` — the safe-integer fast path negates the mantissa as an `Int64` before converting to `Double`:

```moonbit
let v = scan.mantissa.reinterpret_as_int64()
let signed = if scan.negative { -v } else { v }   // -0L == 0L: sign lost
return { value: signed.to_double(), repr: None }
```

Negating after the `to_double()` conversion preserves `-0.0` and is exact for the whole safe-integer range.

## Affected targets

All backends — wasm-gc, js, and native (shared code; confirmed on all three).

## Severity

Low-to-medium: silent data corruption limited to the sign of zero for the exact literal `-0`; internally inconsistent with the `-0.0` / `-0e0` paths of the same lexer. Found by adversarial QuickCheck testing (random 64-bit patterns essentially never produce exactly `-0.0`, which is why existing property tests missed it).

bobzhang:
Fixes #4053

`parse("-0")` returned **+0.0** while `parse("-0.0")` / `parse("-0e0")` returned **-0.0** — the same mathematical literal parsed to different IEEE-754 values depending on spelling, contrary to RFC 8259 number semantics and `JSON.parse`. The bug is invisible to `Json` equality because `-0.0 == 0.0` under `Double` comparison.

Root cause: the safe-integer fast path in `lex_number_end` (`json/lex_number.mbt`) negated the mantissa as an `Int64` (`-0L == 0L`) before converting to `Double`. Fixed by negating after the conversion — exact for the whole safe-integer range.

Deterministic regression tests in `json/lex_number_test.mbt` cover every spelling of negative zero (`-0`, `-0.0`, `-0e0`, `-0.00E-7`) plus negative values that underflow to zero (`-1e-400`, `-4.9e-325`, huge negative exponents), which already preserved the sign.

Based directly on `main`; independent of #4056. The adversarial QuickCheck suite that found this bug is #4045.

Verified: `moon test -p moonbitlang/core/json` green on wasm-gc, js, and native; no `.mbti` changes.

### DIFF
--- a/json/lex_number.mbt
+++ b/json/lex_number.mbt
@@ -399,12 +399,16 @@ fn ParseContext::lex_number_end(
     // returned Double lossless. `reinterpret_as_uint64` / `reinterpret_as_int64`
     // are value-preserving here because both operands sit in [0, 2^53), well
     // inside the overlap of Int64+ and UInt64.
+    //
+    // The sign is applied after the Int64 -> Double conversion so that `-0`
+    // parses to the IEEE-754 negative zero (`-(0L)` is still `0L`, but
+    // `-(0.0)` is `-0.0`), matching the `-0.0` / `-0e0` paths below.
     if !scan.many_digits &&
       scan.exponent == 0L &&
       scan.mantissa <= SAFE_INTEGER_LIMIT.reinterpret_as_uint64() {
-      let v = scan.mantissa.reinterpret_as_int64()
-      let signed = if scan.negative { -v } else { v }
-      return { value: signed.to_double(), repr: None }
+      let v = scan.mantissa.reinterpret_as_int64().to_double()
+      let value = if scan.negative { -v } else { v }
+      return { value, repr: None }
     }
     return ctx.lex_integer_end(start, end)
   }
--- a/json/lex_number_test.mbt
+++ b/json/lex_number_test.mbt
@@ -170,3 +170,27 @@ test "parse number with huge exponent" {
     ),
   )
 }
+
+///|
+/// Every spelling of a negative zero — including values that underflow to
+/// zero — must keep the IEEE-754 sign bit. The integer spelling `-0` used to
+/// lose it: the integer fast path negated an `Int64` (where `-0 == 0`)
+/// before converting to `Double`.
+test "parse preserves the sign of negative zero" {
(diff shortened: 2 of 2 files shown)


json/lex_number.mbt
```diff
@@ -399,12 +399,16 @@ fn ParseContext::lex_number_end(
...
+    // parses to the IEEE-754 negative zero (`-(0L)` is still `0L`, but
+    // `-(0.0)` is `-0.0`), matching the `-0.0` / `-0e0` paths below.
     if !scan.many_digits &&
       scan.exponent == 0L &&
       scan.mantissa <= SAFE_INTEGER_LIMIT.reinterpret_as_uint64() {
-      let v = scan.mantissa.reinterpret_as_int64()
-      let signed = if scan.negative { -v } else { v }
-      return { value: signed.to_double(), repr: None }
+      let v = scan.mantissa.reinterpret_as_int64().to_double()
+      let value = if scan.negative { -v } else { v }
```
bobzhang, line 410: FTR： here v used to be int where -v (when v is 0) does not make sense

PR #4064, opened 2026-08-15
fix(json): reject unpaired \uXXXX surrogate escapes in strings

### LINKED ISSUE #4062: json: parser constructs ill-formed strings from unpaired \uXXXX escape sequences
## Summary

`@json.parse` decodes every `\uXXXX` escape by writing the hex value directly into the result (`buf.write_char(c.unsafe_to_char())` in `json/lex_string.mbt`, `lex_string_slow`). For surrogate-range values this **manufactures an ill-formed, lone-surrogate MoonBit string out of perfectly valid ASCII JSON input**:

```moonbit
test {
  let s = @json.parse("\"\\uD800\"")
  // s is String containing the unpaired leading surrogate U+D800 —
  // an ill-formed UTF-16 sequence no safe API should be able to produce.
}
```

This violates the String well-formedness invariant ("we should maintain String unicode safe, so `unsafe_to_char` is indeed unsafe"): user code that never touches unsafe APIs can end up holding ill-formed strings simply by parsing attacker-controlled ASCII JSON.

## Spec position

- **RFC 8259 §8.2** explicitly flags unpaired surrogate escapes as unpredictable-behavior territory: the escape *syntax* is grammatically valid, but "the behavior of software that receives JSON texts containing such values is unpredictable".
- **RFC 7493 (I-JSON)** forbids them outright: messages MUST NOT contain `\uXXXX` escapes representing unpaired surrogates.
- Ecosystem: **JS `JSON.parse` and Python `json` accept** them (producing ill-formed strings), **Go `encoding/json` substitutes U+FFFD**, **serde_json (Rust) rejects** with a parse error. Under the Rust-style well-formedness stance, we align with **serde_json**: reject with the documented `ParseError`.

## Proposed resolution

In `lex_string_slow`'s `\u` handling: a leading-surrogate escape (`\uD800`–`\uDBFF`) must be immediately followed by a trailing-surrogate escape (`\uDC00`–`\uDFFF`) and the pair combined into one Unicode scalar value; a bare trailing-surrogate escape, an unpaired leading-surrogate escape, or a non-escape following a leading-surrogate escape raise `InvalidChar`. Valid escaped pairs (e.g. `\uD83D\uDE00` → U+1F600) keep parsing exactly as before.

This is a **behavior change**: `"\uD800"` currently parses successfully and would become a parse error.

## Affected targets

All backends — wasm-gc, js, native (shared code).

Related: #4049 covers the raw (unescaped) lone-surrogate side, including the process abort. Found during adversarial QuickCheck testing (#4045).

bobzhang:
Fixes #4062

## The bug

The parser decodes every `\uXXXX` escape by writing the hex value into the result unchecked (`buf.write_char(c.unsafe_to_char())` in `lex_string_slow`), so `"\uD800"` **manufactures an ill-formed lone-surrogate string out of perfectly valid ASCII JSON input** — user code that never touches unsafe APIs ends up holding an ill-formed `String`, violating the well-formedness invariant (*"we should maintain String unicode safe, so `unsafe_to_char` is indeed unsafe"*).

## Spec position

Per **RFC 8259** the `\uXXXX` escape *syntax* is grammatically valid for any hex value, but §8.2 explicitly flags unpaired surrogate escapes as unpredictable-behavior territory; **RFC 7493 (I-JSON)** forbids them outright. Ecosystem: JS `JSON.parse` and Python accept them (producing ill-formed strings), Go substitutes U+FFFD, **serde_json (Rust) rejects — we align with serde_json**.

## The fix

An escaped leading surrogate (`\uD800`–`\uDBFF`) must be immediately followed by an escaped trailing surrogate (`\uDC00`–`\uDFFF`); the pair is combined into one Unicode scalar value. A bare trailing-surrogate escape, an unpaired leading-surrogate escape, and mixed escaped/raw halves raise the documented `ParseError` (`InvalidChar`). Valid escaped pairs (`\uD83D\uDE00` → 😀), any hex digit case, and all non-surrogate escapes parse exactly as before.

## ⚠️ Behavior change

`"\uD800"` (and every other unpaired surrogate escape) **previously parsed successfully** and now raises a parse error. No test or snapshot in the repository relied on the old acceptance.

## Scope and relation to #4056

This PR covers only the **escaped** spellings and is **based on `main`, independent of #4056** (the diff regions in `lex_string.mbt` do not textually overlap: #4056 touches the fast path, `flush`, and the raw-character branch; this PR touches only the `\u` escape branch). Raw (unescaped) lone surrogates — including the process abort — are #4049, fixed in #4056.

## Tests

Deterministic regressions in `json/lex_string_test.mbt`: every unpaired spelling (leading alone, bare trailing, double leading, reversed escaped pair, escaped-leading + raw-trailing), the exact `ParseError` shape, and still-accepted valid pairs.

Verified standalone on this branch: `moon check` clean; json suite green on **wasm-gc, js, native** (201/201 each); no `.mbti` changes. Found by the adversarial QuickCheck suite in #4045.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ changelog should follow [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 #### Changed
 
+- **BREAKING**: `@json.parse` now rejects unpaired `\uXXXX` surrogate escapes in strings, raising `ParseError::InvalidChar` at the backslash that opens the offending escape; an escaped leading surrogate must be followed immediately by an escaped trailing surrogate, and the pair decodes to the character it denotes. Previously such escapes were decoded unchecked and produced a `String` that was not well-formed Unicode. `@json.valid` reports the same documents as invalid. JSON emitted by `JSON.stringify` in JavaScript can contain these escapes, so input that JavaScript and Python accept may now be rejected — as it is by Rust's serde_json
 - `@json.inspect` has been migrated to `json_inspect`
 - `String::sub` and `StringView::sub` now panic on invalid indices instead of raising `CreatingViewError`. The `CreatingViewError` type has been removed.
 
--- a/json/README.mbt.md
+++ b/json/README.mbt.md
@@ -51,6 +51,33 @@ test "parse and validate jsons" {
 }
 ```
 
+#### What may appear inside a string
+
+Every string in a parsed document is well-formed Unicode, so each `\uXXXX`
+escape has to denote a Unicode scalar value. An escaped leading surrogate
+must be followed immediately by an escaped trailing surrogate, and the pair
+decodes to the single character it stands for; an escape that cannot pair up
+is a parse error.
+
+```mbt check
+///|
+test "surrogate escapes" {
+  // A surrogate pair decodes to the one character it denotes.
+  assert_true(@json.parse("\"\\uD83D\\uDE00\"") == Json::string("😀"))
+  // An escape that cannot pair up is rejected rather than producing a
+  // string that is not well-formed Unicode.
+  assert_false(@json.valid("\"\\uD800\""))
+}
+```
+
+RFC 8259 §9 leaves what a string may contain to the implementation, and this
+is where MoonBit draws that line: a `String` is required to be well-formed,
+so the alternatives would be to hand one back that is not, or to substitute
+U+FFFD and lose the difference between two distinct keys. Note that
+`JSON.stringify` in JavaScript does emit lone surrogates this way, so a
+document JavaScript and Python accept can be rejected here — as it is by
+Rust's serde_json; Go substitutes U+FFFD instead.
+
 ### Object Navigation
 
 ```mbt check
--- a/json/README.mbt.md
+++ b/json/README.mbt.md
@@ -54,10 +54,11 @@ test "parse and validate jsons" {
 #### What may appear inside a string
 
 Every string in a parsed document is well-formed Unicode, so each `\uXXXX`
-escape has to denote a Unicode scalar value. An escaped leading surrogate
-must be followed immediately by an escaped trailing surrogate, and the pair
-decodes to the single character it stands for; an escape that cannot pair up
-is a parse error.
+escape must denote a Unicode scalar value on its own or be one half of a
+correctly ordered surrogate pair. An escaped leading surrogate must be
+followed immediately by an escaped trailing surrogate, and the pair decodes
+to the single character it stands for; an escape that cannot pair up is a
+parse error, reported at the backslash that opens it.
 
 ```mbt check
 ///|
@@ -76,7 +77,8 @@ so the alternatives would be to hand one back that is not, or to substitute
 U+FFFD and lose the difference between two distinct keys. Note that
 `JSON.stringify` in JavaScript does emit lone surrogates this way, so a
 document JavaScript and Python accept can be rejected here — as it is by
-Rust's serde_json; Go substitutes U+FFFD instead.
+Rust's serde_json when parsing into `String` or `Value`; Go's
+`encoding/json` substitutes U+FFFD instead.
 
 ### Object Navigation
 
--- a/json/lex_string.mbt
+++ b/json/lex_string.mbt
@@ -63,7 +63,38 @@ fn ParseContext::lex_string_slow(ctx : ParseContext) -> String raise ParseError 
           Some('/') => buf.write_char('/')
           Some('u') => {
             let c = ctx.lex_hex_digits(4)
-            buf.write_char(c.unsafe_to_char())
+            if c is (0xD800..=0xDBFF) {
+              // A leading-surrogate escape is only meaningful as the first
+              // half of an escaped surrogate pair; combine it with the
+              // immediately following trailing-surrogate escape into one
+              // Unicode scalar value. Anything else would manufacture a
+              // string containing an unpaired surrogate, which MoonBit
+              // strings disallow (RFC 8259 calls the behavior for such
+              // escapes unpredictable; I-JSON forbids them).
+              match ctx.read_char() {
+                Some('\\') => ()
+                Some(_) => ctx.invalid_char(shift=-1)
+                None => raise InvalidEof
+              }
+              match ctx.read_char() {
+                Some('u') => ()
+                Some(_) => ctx.invalid_char(shift=-1)
+                None => raise InvalidEof
+              }
+              let c2 = ctx.lex_hex_digits(4)
+              if c2 is (0xDC00..=0xDFFF) {
+                let combined = (c << 10) + c2 - 0x35fdc00
+                buf.write_char(combined.unsafe_to_char())
+              } else {
+                ctx.invalid_char(shift=-1)
+              }
+            } else if c is (0xDC00..=0xDFFF) {
+              // A bare trailing-surrogate escape can never form a scalar
+              // value.
+              ctx.invalid_char(shift=-1)
+            } else {
+              buf.write_char(c.unsafe_to_char())
+            }
           }
           Some(_) => ctx.invalid_char(shift=-1)
           None => raise InvalidEof
--- a/json/lex_string.mbt
+++ b/json/lex_string.mbt
@@ -62,6 +62,9 @@ fn ParseContext::lex_string_slow(ctx : ParseContext) -> String raise ParseError 
           Some('\\') => buf.write_char('\\')
           Some('/') => buf.write_char('/')
           Some('u') => {
+            // The backslash that opened this escape. `ctx.offset` is just
+            // past the `u`, and `\` and `u` are one code unit each.
+            let escape_start = ctx.offset - 2
             let c = ctx.lex_hex_digits(4)
             if c is (0xD800..=0xDBFF) {
               // A leading-surrogate escape is only meaningful as the first
@@ -73,12 +76,12 @@ fn ParseContext::lex_string_slow(ctx : ParseContext) -> String raise ParseError 
               // escapes unpredictable; I-JSON forbids them).
               match ctx.read_char() {
                 Some('\\') => ()
-                Some(_) => ctx.invalid_char(shift=-1)
+                Some(_) => ctx.unpaired_surrogate(escape_start)
                 None => raise InvalidEof
               }
               match ctx.read_char() {
                 Some('u') => ()
-                Some(_) => ctx.invalid_char(shift=-1)
+                Some(_) => ctx.unpaired_surrogate(escape_start)
                 None => raise InvalidEof
               }
               let c2 = ctx.lex_hex_digits(4)
@@ -86,17 +89,17 @@ fn ParseContext::lex_string_slow(ctx : ParseContext) -> String raise ParseError 
                 let combined = (c << 10) + c2 - 0x35fdc00
                 buf.write_char(combined.unsafe_to_char())
               } else {
-                ctx.invalid_char(shift=-1)
+                ctx.unpaired_surrogate(escape_start)
               }
             } else if c is (0xDC00..=0xDFFF) {
               // A bare trailing-surrogate escape can never form a scalar
               // value.
-              ctx.invalid_char(shift=-1)
+              ctx.unpaired_surrogate(escape_start)
             } else {
               buf.write_char(c.unsafe_to_char())
             }
           }
-          Some(_) => ctx.invalid_char(shift=-1)
+          Some(c) => ctx.invalid_char(shift=-c.utf16_len())
           None => raise InvalidEof
         }
         start = ctx.offset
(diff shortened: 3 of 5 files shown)


bobzhang:
## Codex CLI review

Run with `codex exec` (`gpt-5.6-sol`, reasoning effort **ultra** — maximum depth with task delegation), against the branch checked out locally rather than the diff text, and asked for both an adversarial correctness pass and an explicit position on the compatibility break. Posted verbatim.

---

Thanks @bobzhang. My bottom line for reviewers: the surrogate-pair decoding, scalar reconstruction, buffer writes, and cursor bookkeeping are correct, and all four targets pass. I agree that strict rejection is a reasonable default because returning an ill-formed `String` or silently replacing object keys is worse. I would not merge this as-is, though: this is a user-visible compatibility break that is absent from the public contract/changelog, and the new `InvalidChar` diagnostics can report a valid hex digit—or even the wrong UTF-16 position—as the offending character.

**Correctness**

The new branch handles the requested cases as follows:

| Input condition | Outcome |
|---|---|
| Physical EOF after `"\uD800` | `InvalidEof` |
| Closing quote immediately after `"\uD800"` | `InvalidChar` on the quote at column 7 |
| `"\uD800x"` | `InvalidChar` on `x` |
| `"\uD800\n"` | `InvalidChar` on `n`, since the second escape is not `\u` |
| Short second escape | `InvalidEof` at physical EOF; otherwise `InvalidChar` on the closing quote/invalid hex digit |
| Bare `"\uDC00"` | Rejected, currently as `InvalidChar(... column: 6, '0')` |
| `"\uD800\uD800"` | Rejected after decoding the second leading surrogate |
| `"\uDE00\uD83D"` | Rejected at the first, bare trailing surrogate |
| Escaped leading + raw trailing surrogate | Clean `ParseError`, rather than the previous abort |
| Upper/lowercase hex | Correctly accepted through the shared `A`–`F`/`a`–`f` branches |

The reconstruction is algebraically correct:

```text
(hi << 10) + lo - 0x35fdc00
= 0x10000 + ((hi - 0xD800) << 10) + (lo - 0xDC00)
```

Thus `D800/DC00` maps to U+10000 and `DBFF/DFFF` maps to U+10FFFF. `StringBuilder::write_char` correctly emits the corresponding UTF-16 pair on non-JS targets, while JS uses the character-to-string intrinsic.

The fast/slow interaction is also sound. Every escaped spelling contains a backslash, so the fast path unconditionally falls back before materializing the string. After a valid pair, both six-code-unit escapes have been consumed and `start = ctx.offset` resumes at precisely the following source unit. The suffix/next-escape test exercises this successfully. Raw lone surrogates can still traverse the fast path in this branch, but that is the separately tracked #4049/#4056 issue, not an escaped-surrogate bypass.

All textual entry points converge here: values use `lex_string`, as do both first and subsequent object keys; `valid` delegates to `parse`. `FromJson` consumes an already-built `Json`, and `ToJson`/`Json::string` may preserve a pre-existing unsafely constructed string but do not decode JSON text.

One real error-offset defect remains: the new mismatch arms at lines 74–82 call `read_char()` and then use `shift=-1`. A non-BMP scalar consumes two UTF-16 units, so `"\uD800😀"` reports column 8 and U+FFFD instead of the emoji at column 7. Bind the returned character and use `shift=-ch.utf16_len()`, as `lex_main.mbt` already does.

Validation passed read-only:

- `moon check --deny-warn --target all`
- JSON tests: 201/201 on `wasm`, `wasm-gc`, `js`, and `native`
- Targeted lexer tests: 6/6 on native

Against `HEAD^`, all eight rejection assertions fail as intended: seven inputs are accepted, while escaped-high + raw-low aborts in the old slow-path flush. The exact `expect_parse_error` check also fails because the parent parses successfully. The four positive compatibility assertions already pass on the parent, appropriately—they are preservation checks, not regressions.

**Should this be rejected at all?**

Yes, I think strict rejection is the right default, but it must be presented as MoonBit’s policy rather than as required JSON behavior.

[RFC 8259 §8.2](https://www.rfc-editor.org/rfc/rfc8259.html#section-8.2) explicitly says its ABNF admits unpaired surrogate escapes, while [§9](https://www.rfc-editor.org/rfc/rfc8259.html#section-9) both requires grammar-conforming input to be accepted and permits implementations to limit string character contents. [I-JSON](https://www.rfc-editor.org/rfc/rfc7493.html#section-2.1) forbids surrogates, but it is a separate profile that also forbids noncharacters and duplicate member names; `@json.parse` is not currently advertised as an I-JSON parser.

The compatibility cost is real: ECMAScript’s [`QuoteJSONString`](https://tc39.es/ecma262/multipage/structured-data.html#sec-quotejsonstring) deliberately emits lone surrogates as `\uXXXX`, so valid output from JavaScript can become unparseable. Nevertheless:

- Preserving the old value lets a safe parser manufacture a `Char` through an operation documented as undefined for surrogate values.
- Unconditional U+FFFD replacement is lossy and can collapse distinct object keys such as `\uD800` and `\uD801`.
- A checked `ParseError` is recoverable and makes the loss explicit.

That makes fail-fast rejection preferable as the strict default. MoonBit already follows this pattern for UTF-8 with strict `@utf8.decode` and separate `decode_lossy`; an explicit JSON replacement mode would be a useful interoperability counterpart. I would not retain an exact-code-unit mode unless MoonBit formally supports WTF-16 strings and audits APIs that currently assume well-formed UTF-16.

**Findings**

1. **[blocking] Document the new accepted-language contract.** `parse` still only says “Parse a JSON input,” `valid` does not define validity, the README only gives generic `ParseError` examples, and the changelog is unchanged. Document the Unicode-scalar restriction, pairing rule, `valid` semantics, JavaScript/Python incompatibility, and this breaking behavior. Describe it as a deliberate character-content limit, not blanket RFC 8259 or I-JSON conformance.

2. **[should] Fix the error semantics and position before pinning them.** `InvalidChar(..., '0')` for `"\uDC00"` blames the final valid hex digit; invalid second code units do the same. Prefer a dedicated `InvalidUnicodeEscape(Position)`/surrogate error at the escape start. At minimum, fix the two-unit offset bug described above and choose a documented position convention.

3. **[should] Expand the regression matrix.** Add U+10000 (`\uD800\uDC00`), physical EOF and malformed/short hex in the second escape, and representative exact error assertions. One object-key success/failure case would pin the other parser route. The existing U+10FFFF and suffix tests already cover the upper arithmetic boundary and resume index.

4. **[nice-to-have] Split the monolithic regression test.** An early assertion currently masks later cases, including the old escaped-high/raw-low process abort.

5. **[nice-to-have] Consider an explicit lossy interoperability mode.** A `parse_lossy` or `surrogate_policy=Reject|Replace` API would let callers intentionally consume JavaScript-produced payloads while keeping strict parsing safe and lossless-by-default.

VERDICT: REQUEST CHANGES

---

I reproduced the error-offset defect from **Finding 2** independently before posting, since it is the one concrete code bug here:

```
"\uD800😀"  ->  Invalid character '�' at line 1, column 8
"\uD800x"   ->  Invalid character 'x' at line 1, column 7
"\uD800中"   ->  Invalid character '中' at line 1, column 7
```

The ASCII and BMP cases report the real character at the right column; the non-BMP one reports a broken half at column 8. `lex_main.mbt:75` already does `let shift = -c.utf16_len()` for exactly this reason, while the new arms in `lex_string.mbt` discard the character and hardcode `shift=-1`. Binding it and using `-ch.utf16_len()` fixes both the column and the reported character.

bobzhang:
## Codex CLI re-review

Second pass with `codex exec` (`gpt-5.6-sol`, reasoning effort **ultra**), after `cea6d3a0` addressed the first round. **REQUEST CHANGES → APPROVE WITH NITS.** Posted verbatim.

---

Thanks for the thorough follow-up. I re-reviewed both commits and found no remaining merge blocker.

1. **Finding 1 — resolved.** `parse`, `valid`, the package README, and the changelog now document the accepted-language change and interoperability cost. The standards framing is accurate: [RFC 8259 §9](https://www.rfc-editor.org/rfc/rfc8259.html#section-9) permits limits on string character content, while [§8.2](https://www.rfc-editor.org/rfc/rfc8259.html#section-8.2) explicitly notes that the grammar allows unpaired surrogates. Calling this something narrower than I-JSON is also correct; [RFC 7493](https://www.rfc-editor.org/rfc/rfc7493.html#section-2.1) additionally excludes noncharacters and duplicate names.

   Documentation nits:

   - “Each `\uXXXX` escape has to denote a Unicode scalar value” is technically imprecise because each half of an accepted surrogate pair is not independently a scalar. Say that an escape must denote a scalar directly or participate in a correctly ordered pair.
   - `parse` currently promises that an escape unable to pair always raises `InvalidChar` at its backslash. Physical EOF remains `InvalidEof`, while malformed/short second escapes retain their normal hex/EOF diagnostics. The documentation should qualify that promise.
   - The interoperability summary is right for ordinary serde_json `String`/`Value` parsing and Go’s `encoding/json` v1, but serde_json’s byte mode accepts WTF-8 and Go’s experimental v2 parser is stricter. Qualifying those names would be more exact.

2. **Finding 2 — resolved in the implementation.** At `escape_start`, `ctx.offset - 2` is correct: the matched ASCII backslash and `u` are exactly one UTF-16 code unit each. `shift = escape_start - ctx.offset` algebraically recovers that absolute offset regardless of how far scanning advanced, including both `read_char()` mismatch arms, a non-BMP mismatch, and a completed but non-trailing second escape.

   The two `-c.utf16_len()` changes are also correct. They affect only pre-existing invalid inputs whose offending character is non-BMP; BMP positions are unchanged. The old result landed inside a surrogate pair, so keeping these localized fixes in this PR is appropriate. They should ideally gain exact tests for an unknown non-BMP escape and a non-BMP character in `lex_hex_digits`.

   The exact snapshots are not over-pinned because `parse` now promises the opening-backslash convention. Positions are 1-based by line and 0-based by UTF-16 code-unit column. That follows from `offset_to_position`, although the code-unit basis should preferably be exposed in the public `Position` documentation.

3. **Finding 3 — resolved, with minor coverage gaps.** The requested boundary, EOF, malformed/short second escape, exact-error, adjacent-pair, and object-key cases are present. I ran the exact current test file against exported copies of all three revisions:

   | Test block | `HEAD` | `HEAD~1` | `origin/main` |
   |---|---:|---:|---:|
   | Leading-surrogate rejection | Pass | Pass | Fail |
   | Bare trailing-surrogate rejection | Pass | Pass | Fail |
   | Escaped/raw halves | Pass | Pass | Process abort |
   | Object keys | Pass | Pass | Fail |
   | Exact surrogate error | Pass | Fail | Fail |
   | Valid pairs | Pass | Pass | Pass |

   Thus, not all six fail on either baseline. That is expected: the first commit already implements the rejection behavior, and the valid-pair block is preservation coverage that should pass on `origin/main`. Exact assertions are still missing for the second mismatch arm, completed non-trailing second escape, and terminal-backslash EOF, but the centralized helper makes this non-blocking.

4. **Finding 4 — resolved in substance, with two nits.** Splitting the test improves isolation. However, the escaped/raw test remains third and aborts the old process, so a regression to the original behavior would still mask the final three blocks; moving it last would complete the isolation. Also, the comment saying the old parser “accepted all of the rejections below” is false because physical EOF and malformed/short escapes were already rejected.

5. **Finding 5 — not implemented, but remains non-blocking.** With the strict default and incompatibility now documented, a lossy mode is an additive API/design decision that can be considered separately.

Validation passed:

- `moon check --deny-warn --target all`
- `moon test -p moonbitlang/core/json` on `wasm`, `wasm-gc`, `js`, and `native`: 207/207 on each target
- `git diff --check`

The tracked worktree remained clean.

VERDICT: APPROVE WITH NITS

---

The remaining nits are applied in `701b9eff`, since three of them were prose that was simply wrong:

- **1a** — "each escape must denote a scalar value" is not true of either half of an accepted pair. Now: must denote a scalar on its own *or* be one half of a correctly ordered pair. Same wording fixed in the README.
- **1b** — the promise that a failed pairing always raises `InvalidChar` at the backslash overstated it. Running out of input still raises `InvalidEof`, and a malformed second escape is reported as the hex-digit error it is. Both qualified.
- **1c** — narrowed the comparisons: serde_json rejects when parsing into `String`/`Value` while its byte mode admits WTF-8, and it is Go's `encoding/json` that substitutes U+FFFD, its v2 parser being stricter.
- **2 / 3** — added the missing exact assertions: the second mismatch arm, a well-formed second escape that is not a trailing surrogate, a trailing backslash at EOF, plus the two arms whose `shift` was corrected (an unknown escape naming a non-BMP character, and a non-BMP character among hex digits) — the latter reported at columns 2 and 4 rather than as a broken half one column further on.
- **4** — the preamble's claim that the old parser accepted every rejection below it was false; the EOF and malformed-escape spellings were already errors. Corrected, and the escaped-leading + raw-trailing case moved last so a regression to the old aborting behaviour cannot mask the blocks after it.

Not taken: exposing the code-unit basis in `Position`'s own docs (finding 2's "preferably") — that documents a pre-existing public type shared by every parser error, so it seemed better as its own change than buried in a surrogate fix. Say the word if you'd rather have it here.

Full matrix green after the follow-up: `moon check --deny-warn --target all`, `moon test --target all` (7241/7242/7186/7160), `moon fmt`, `moon info`.

PR #4071, opened 2026-08-15
fix(lazy_list): make concat stack-safe under any nesting

### LINKED ISSUE #4066: lazy_list: LazyList::concat is not stack-safe under left-nesting
## Summary

`LazyList::concat` forces its result by recursing once per nesting level, so a left-nested chain of `concat`s overflows the stack when the chain gets long. The trigger is the idiomatic accumulator loop:

```moonbit
let mut acc : @lazy_list.LazyList[Int] = @lazy_list.empty()
for i in 0..<10000 {
  acc = acc.concat(@lazy_list.from_iter([i].iter()))
}
let _ = acc.take(2).to_array()   // RangeError: Maximum call stack size exceeded
```

Note that it does not take a full traversal to hit this — `acc.head()` is fine, but reaching the **second** element is already enough, because that is the first tail force that has to walk the whole nesting.

## Reproduction

```moonbit
///|
fn left_nested(depth : Int) -> @lazy_list.LazyList[Int] {
  let mut acc : @lazy_list.LazyList[Int] = @lazy_list.empty()
  for i in 0..<depth {
    acc = acc.concat(@lazy_list.from_iter([i].iter()))
  }
  acc
}

///|
test "left-nested concat" {
  inspect(left_nested(10000).take(2).to_array().length(), content="2")
}
```

| target | depth 2000 | depth 5000 | depth 10000 | depth 50000 |
| wasm | ok | ok | **overflow** | **overflow** |
| wasm-gc | ok | ok | **overflow** | **overflow** |
| js | ok | ok | **overflow** | **overflow** |
| native | ok | ok | ok | ok |

Stack trace (wasm-gc):

```
RangeError: Maximum call stack size exceeded
    at @moonbitlang/core/lazy.Lazy::force[...LazyList[Int]]
    at @moonbitlang/core/lazy_list.LazyList::concat[Int].anonymous
    at @moonbitlang/core/lazy.Lazy::force[...LazyList[Int]]
    at @moonbitlang/core/lazy_list.LazyList::concat[Int].anonymous
    ...
```

## Cause

`lazy_list/lazy_list.mbt`:

```moonbit
pub fn[A] LazyList::concat(self : LazyList[A], other : LazyList[A]) -> LazyList[A] {
  match self.cell {
    Empty => other
    Cons(x, tail~) => LazyList::lazy_cons(x, () => tail.force().concat(other))
  }
}
```

Construction is fine — `self.cell` is already materialized, so building the chain is O(1) per step. The problem is the thunk. For `((a ++ b) ++ c) ++ d`, forcing the outermost tail runs `tail.force().concat(other)`, whose `tail.force()` is the depth-(d-1) concat's thunk, whose `tail.force()` is the depth-(d-2) one, and so on: the recursion depth is the nesting depth. Because the result is re-nested to the same depth after each step, every subsequent element costs the same again, so the traversal is also quadratic in time.

`concat_lazy` (used by `flat_map`) has the same shape, though `flat_map` nests to the right and so does not hit it.

## Why this looks like a bug rather than a limitation

Stack safety is an explicit design goal elsewhere in this package — `flat_map` documents that it "Skips empty inner lists iteratively so a long run of empty mappings does not blow the stack", and `filter`, `drop`, `drop_while`, `each`, `fold` and `to_array` are all written as loops for the same reason. `concat` is the one combinator whose forcing path is recursive in a way the caller can drive without realising it, and `acc = acc.concat(...)` in a loop is a natural thing to write.

## Notes

Found while writing a property-based specification suite for the package. That PR is tests-only and deliberately does not include a case for this, so it stays green. Happy to add a regression test alongside whichever fix you prefer.

bobzhang:
Closes #4066.

## The bug

`LazyList::concat` built its result as

```moonbit
Cons(x, tail~) => LazyList::lazy_cons(x, () => tail.force().concat(other))
```

O(1) to construct, but impossible to force without one stack frame per `++`. For `((a ++ b) ++ c) ++ d`, the outermost tail thunk forces the one below it, which forces the one below that. The accumulator idiom

```moonbit
acc = acc.concat(xs)
```

builds exactly that shape, so a 10000-step loop overflowed the stack on wasm, wasm-gc and js as soon as the **second** element was reached — the first tail force is the one that has to walk the whole nesting. It was quadratic in time as well, because the chain is re-nested to the same depth after every element. `concat_lazy`, which `flat_map` chains through, had the same shape.

## The fix

The pending right-hand sides are reified as data on the cell's tail instead of being closed over by a recursive thunk. Forcing walks down to the innermost tail in a loop, collects what is still owed, and hands the remainder to the cell it produces:

- **stack** is constant in the nesting depth, whichever way the chain is associated;
- **a full traversal is linear**, because only the entry node is memoized rather than each level being re-nested;
- the pending sequence is a **catenable tree**, so joining two of them is O(1) and appending to a list while consuming it from the front stays linear too.

`Tail` (`lazy_list/types.mbt`, `lazy_list/tail.mbt`) is `@lazy.Lazy` — same states, same at-most-once guarantee, same reentrancy abort — plus one `Append` state. It replaces the `@lazy.Lazy` tail rather than wrapping one: a `LazyList` allocates a tail per element, so boxing a `@lazy.Lazy` inside a wrapper measured **~30% slower** on build-and-traverse (1.90s vs 1.51s for 20M cells, native release), while the merged representation measures at parity on both native and wasm-gc.

The public API is unchanged (`pkg.generated.mbti` is untouched); `moon.pkg` drops the now-unused `@lazy` import.

## Behaviour that is deliberately unchanged

Composing arbitrarily many lazy transforms (`xs.map(f).map(f)…` ten thousand deep, and likewise `flat_map`) still nests one thunk per transform. That is a property of composed transforms, not of `concat`'s associativity, and the README now says so explicitly.

## Tests

| what | where |
| --- | --- |
| left- and right-nested chains at depth 50000, forced to the 2nd element and fully traversed | `lazy_list_test.mbt` |
| 50000 appends of the empty list | `lazy_list_test.mbt` |
| `flat_map` over a deeply nested source (`concat_lazy`) | `lazy_list_test.mbt` |
| randomly shaped concat trees — ordering under arbitrary association | `quickcheck_test.mbt` |
| appending while consuming from the front | `quickcheck_test.mbt` |
| right-hand side untouched until the left ends; both sides memoized across traversals | `lazy_list_test.mbt` |
| reentrancy guard, on a thunk and on a suspended append | `panic_test.mbt` |

`moon check --deny-warn --target all`, `moon test --target all`, `moon test --release` (js/wasm/wasm-gc and native), `moon test --target llvm`, `moon fmt`, `moon info` and `moon bundle --all` all clean.

### DIFF
--- a/lazy_list/README.mbt.md
+++ b/lazy_list/README.mbt.md
@@ -1,9 +1,9 @@
 # LazyList
 
-A persistent, re-traversable linked list with memoized lazy tails. Built on
-top of `@lazy.Lazy`. Use it when you have a pull-based source (`Iter`,
-generator, recursive definition) that you want to consume *more than once*
-without re-running the underlying computation.
+A persistent, re-traversable linked list with memoized lazy tails. Use it
+when you have a pull-based source (`Iter`, generator, recursive definition)
+that you want to consume *more than once* without re-running the underlying
+computation.
 
 ## Table of Contents
 
@@ -44,6 +44,10 @@ so a second traversal walks the cached cells instead of re-evaluating.
 
 This shape mirrors Haskell's lazy lists / Scala's `LazyList` and is sometimes
 called "odd-style" laziness.
+
+Caching in place is a mutation, so a `LazyList` is not thread-safe even though
+it is persistent: traversing one from two threads needs external
+synchronization, the same contract `@lazy.Lazy` documents.
 
 ---
 
@@ -286,6 +290,52 @@ and never trigger user-supplied thunks. The alternative, even-style
 emptiness check itself a forcing operation. Odd-style is what Haskell's `[]`
 and Scala's `LazyList` use, and it's simpler to reason about.
 
+### Why `concat` is stack-safe at any nesting
+
+How a chain of `concat`s is associated is the caller's choice, and the most
+natural way to write one is also the hardest case:
+
+```mbt check
+///|
+test {
+  let mut acc : @lazy_list.LazyList[Int] = @lazy_list.empty()
+  for i in 0..<10000 {
+    acc = acc.concat(@lazy_list.from_iter([i].iter()))
+  }
+  debug_inspect(acc.take(3).to_array(), content="[0, 1, 2]")
+  inspect(acc.iter().count(), content="10000")
+}
+```
+
+Every step nests the previous result inside the *left* argument of the next
+`concat`, so the value is `((([] ++ a) ++ b) ++ c) ++ …`. The obvious
+implementation — `lazy_cons(x, () => tail.force().concat(other))` — builds
+that in O(1) per step but cannot force it: the outermost tail thunk forces
+the one below it, which forces the one below that, one stack frame per `++`.
+It also re-nests the chain to the same depth after every element, which makes
+a full traversal quadratic.
+
+So the pending right-hand sides live on the cell's tail as *data* instead of
+being closed over by a thunk. Forcing walks down to the innermost tail in a
+loop, collecting what is still owed, and hands the remainder to the cell it
+produces. Stack use is constant in the nesting depth, and a full traversal is
+linear — including when you append and consume at the same time, since
+joining two pending sequences is O(1). `flat_map` suspends its inner lists
+through the same representation.
+
+What this does *not* cover is composing arbitrarily many lazy transforms:
+`xs.map(f).map(f)...` ten thousand deep nests ten thousand thunks, and so
+does `flat_map`, so forcing one cell walks all of them. That is a property of
+composed transforms rather than of `concat`'s associativity, and it is
+unchanged here.
+
+That suspended-append state is the one reason a cell's tail is not simply a
+`@lazy.Lazy`. It is otherwise the same thing — the same states, the same
+at-most-once guarantee, the same reentrancy check — and the state lives
+*inside* the memoized cell rather than wrapping one because a `LazyList`
+allocates a tail per element, so an extra box per cell would cost every
+traversal, not just the ones that go through `concat`.
+
 ### Why is `drop_while` eager but `take_while` lazy?
 
 ```mbt check
@@ -324,11 +374,12 @@ dropWhile p xs@(x:_) | p x = dropWhile p xs'    -- eager recursion
 
 ### Why doesn't `map` / `filter` / `take_while` / `flat_map` accept `raise?`?
 
-Because their callbacks fire from inside `@lazy.Lazy` thunks. Thunks are
-non-raising by design (see `@lazy/lazy.mbt`): a raising thunk would force
-every consumer of every lazy cell to handle the effect, and memoizing
-failure forces an awkward choice between "cache the error and re-raise on
-every retry" and "retry on every force."
+Because their callbacks fire from inside the memoized tail thunks. Those are
+non-raising by design, for the reasons `@lazy` sets out (see
+`@lazy/lazy.mbt`): a raising thunk would force every consumer of every lazy
+cell to handle the effect, and memoizing failure forces an awkward choice
+between "cache the error and re-raise on every retry" and "retry on every
+force."
 
 If you need a fallible transform, wrap the result yourself:
 
--- a/lazy_list/concat_bench_test.mbt
+++ b/lazy_list/concat_bench_test.mbt
@@ -1,0 +1,79 @@
+// Copyright 2026 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// The three shapes that decided the tail representation, kept runnable so
+// the trade-off can be re-measured instead of re-argued:
+//
+//   * `build and traverse` is the ordinary path, one tail allocated per
+//     element. It is why `Tail` subsumes `@lazy.Lazy` rather than wrapping
+//     one — a wrapper adds a box and a dispatch to every cell here, which
+//     measured ~30% on this benchmark.
+//   * `left-nested concat` is the shape from #4066: with a recursive
+//     forcing path it overflows the stack rather than running slowly.
+//   * `concat while consuming` is why the pending sequence is a catenable
+//     tree — a flat list copies the whole backlog on every step.
+
+///|
+let concat_bench_len : Int = 2048
+
+///|
+fn bench_left_nested(depth : Int) -> @lazy_list.LazyList[Int] {
+  let mut acc : @lazy_list.LazyList[Int] = @lazy_list.empty()
+  for i in 0..<depth {
+    acc = acc.concat(@lazy_list.from_iter([i].iter()))
+  }
+  acc
+}
+
+///|
+test "bench LazyList build and traverse n=2048" (it : @bench.T) {
+  let source = Array::makei(concat_bench_len, i => i)
+  it.bench(fn() {
+    let xs = @lazy_list.from_iter(source.iter())
+    it.keep(xs.iter().count())
+  })
+}
+
+///|
+test "bench LazyList re-traverse a forced list n=2048" (it : @bench.T) {
+  let xs = @lazy_list.from_iter(Array::makei(concat_bench_len, i => i).iter())
+  it.keep(xs.iter().count())
+  // Every cell is cached now, so this measures the walk alone.
+  it.bench(fn() { it.keep(xs.iter().count()) })
+}
+
+///|
+test "bench LazyList left-nested concat, build only n=2048" (it : @bench.T) {
+  it.bench(fn() { it.keep(bench_left_nested(concat_bench_len).is_empty()) })
+}
+
+///|
+test "bench LazyList left-nested concat, build and traverse n=2048" (
+  it : @bench.T,
+) {
+  it.bench(fn() { it.keep(bench_left_nested(concat_bench_len).iter().count()) })
+}
+
+///|
+test "bench LazyList concat while consuming n=2048" (it : @bench.T) {
+  let source = Array::makei(concat_bench_len + 1, i => i)
+  it.bench(fn() {
+    let mut cur = @lazy_list.from_iter(source.iter())
+    for i in 0..<concat_bench_len {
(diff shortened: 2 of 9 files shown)


bobzhang:
## Codex CLI review sign-off

Reviewed with `codex exec` (`gpt-5.6-sol`, reasoning effort **xhigh**) over three rounds. Verbatim final verdict, on commit `d5f95127`:

> **VERDICT: APPROVE WITH NITS**
>
> The implementation is correct: ordering matches `xs ++ ys`; forcing is stack-safe across concat associations; one complete traversal is linear in elements plus pending nodes; laziness, memoization, and reentrancy protection remain intact; and the `#owned` annotations accurately describe transferred arguments. `a51bb025..HEAD` contains only the stated comment/test-name changes.
>
> Validation passed: deny-warning checks on all targets; `moon fmt` and `moon info` produced no diff; lazy-list tests passed 63/63 on wasm, wasm-gc, and JavaScript, and 61/61 on native.
>
> 1. [non-blocking] The shaped-concat wording still slightly overclaims coverage. The test name says "any nesting shape," although the LCG samples only a subset, and the generated reduction trees are not generally logarithmic in depth—the residual reduction stack can be much deeper. "Generated association shapes" and "far shallower than the 50,000-deep combs" would be precise.
> 2. [non-blocking] "Milliseconds" versus "tens of seconds" is machine/backend-dependent and not asserted by the test. Prefer saying the complexity regression manifests in runtime, unless the measured environment is specified.

Both remaining nits are applied in `d5f95127`: the property is renamed to *"a chain of concats forces at generated nesting shapes"*, `shaped_concat`'s doc now says the trees it draws nest far more shallowly than the 50000-deep combs rather than claiming they are logarithmic, and the interleaving test's comment says the quadratic variant shows up only as runtime instead of quoting wall-clock numbers.

### Earlier rounds

**Round 1** — `APPROVE WITH NITS`, three non-blocking findings, all addressed:

1. *Interleaving concatenation and forcing can still be quadratic.* `cur = cur.concat(singleton); cur = cur.tail().unwrap()` copied the whole pending list every step — measured ≈0.95s / 2.70s / 9.78s at 4k / 8k / 16k iterations. Fixed by making `Pending` a catenable tree (`Nil | More | Cat`): joining is O(1), and `Pending::uncons` rotates the left spine right for amortized O(1) per element. Covered by a new property at 20000 steps.
2. *Deep `flat_map` composition remains stack-proportional.* Correctly identified as pre-existing and shared with `map`, not introduced here. The README now states it explicitly and scopes the stack-safety claim to how a `concat` chain is associated.
3. *The quickcheck property did not really sweep nesting shapes.* Replaced the split-point sweep with `shaped_concat`, which builds a random binary concat tree through a merge stack.

**Round 2** — `APPROVE WITH NITS`, three non-blocking findings, all addressed by scoping the claims: the amortization on `Pending::uncons` is now stated for one linear consumption (re-consuming a retained remainder replays its rotations), `shaped_concat`'s doc no longer overclaims what its bit stream reaches, and the interleaving property is named for what it asserts (order) rather than for the cost it exercises.

No blocking findings were raised in any round.

bobzhang:
## Codex CLI design & impact review

Second review pass, this time asking for a judgement on whether the change is worth its complexity rather than another correctness check. Run with `codex exec` (`gpt-5.6-sol`, reasoning effort **ultra** — maximum depth with task delegation), on commit `d5f95127`. Posted verbatim.

---

Bottom line: this fix is worth making, and `Tail` plus a catenable `Pending` is the right design for the guarantees claimed. The implementation cost is substantial for a young package, but it buys O(1) construction, constant stack under arbitrary association, linear traversal—including append while consuming—and memoized re-traversability. No materially simpler design preserves all four. I would merge this without changing the public API.

**Cost/benefit**

The benefit is meaningful: a natural-looking `acc = acc.concat(xs)` loop currently overflows when the second element forces the accumulated tail, and full traversal is quadratic. That is too severe a trap for a public standard-library combinator, especially in a package that otherwise deliberately uses loops to avoid stack growth. A separate bulk API would not repair that contract.

The cost is real. Production `.mbt` source grows from 465 to 705 lines, about 52%. The raw count slightly exaggerates the mechanism—`tail.mbt` contains roughly 81 non-comment lines—but future maintainers must understand three private types, mutable memoization, the denotation of `Append`, and the amortization of `Cat`. The strongest argument against this PR is that no other core package currently uses `lazy_list`, repeated append is often discouraged for linked lists, and the package was originally attractive as a thin demonstration of `@lazy.Lazy`. For an example package, I would prefer the simpler abstraction and document a builder. For a standard-library collection with a public `concat`, I would not: this is exactly when changing the still-private representation is cheapest.

The complexity is well-contained. Existing combinators still see `Empty | Cons` and call `tail.force()`; the difficult invariant lives in `Tail`. Its denotation is small enough to review: `Append(t, p)` means `force(t)` followed by `p`, and `Cat(a, b)` means the pending sequence of `a` followed by `b`. The property suite covers ordering, arbitrary generated associations, laziness, memoization, empty segments, reentrancy, and deep combs. It cannot prove amortized complexity, but the rotation is standard and the 20,000-step interleaving test is a useful severe-regression canary.

`Pending::uncons` earns its keep. Comparing HEAD with the earlier flat-`Pending` revision on that existing 5×20,000 append/consume case, warm local runs were approximately 0.09s versus 24.7s on native. The flat version saved only about 15 source lines. Its hidden quadratic behavior is not a reasonable trade for that saving. The ephemeral amortization also fits normal use: traversal threads one remainder forward, and the resulting public tails are memoized, so re-traversing the same `LazyList` does not replay the rotations.

The `.mbti` is unchanged and all new types are private. I also found no imports or references to `lazy_list` outside its own package, so cross-package risk is negligible. Removing `@lazy` does lose a useful dogfooding example and introduces a semantic-drift risk if `Lazy` later changes its memoization or reentrancy policy. It does not materially weaken the user-facing story: `Lazy` remains prelude-exported, and retaining an otherwise unnecessary allocation layer merely to showcase the sibling package would be the wrong priority.

**Alternatives considered**

- **Keep `@lazy.Lazy` inside `Tail`: reject.** A realistic wrapper still needs `Tail`, `Append`, `Pending`, the append-spine evaluator, and an outer cache/reentrancy state for normalized appends. It delegates only the easy base-thunk branch while adding another box and dispatch to every ordinary cell. The reported 1.90s versus 1.51s build-and-traverse result is a substantial tax for little conceptual reduction.

- **Flat `Pending`: reject.** It fixes the reported stack overflow and ordinary left-nested traversal, but concatenating pending sequences copies the growing backlog when append and consumption are interleaved, making that pattern quadratic. `Cat` plus the right rotation removes a measured performance cliff for very little extra code.

- **Store pending work in `Cons`: reject.** A shape such as `Cons(head, base_lazy, pending)` recomputes normalization on repeated `tail()` calls unless it gains another memo field. Adding that cache or an `Appended` cell variant recreates the same state machine while spreading its coherence invariant through every `Cons` match, including transforms and `Debug`. Pending append work belongs on the continuation, so `Tail` is the cleaner boundary.

- **Leave binary `concat` recursive and add `concat_all`/`join`: reject as a substitute.** A bulk helper could essentially memoize `parts.flat_map(xs => xs.iter())` through `from_iter`, and it may be useful for discoverability. It would still leave existing binary `concat` association-sensitive and force callers to know about the trap. It also expands the API to work around an implementation defect.

- **Other representations: reject.** A top-level `Append` list node makes `head` and `is_empty` normalization operations, violating the head-strict O(1) model. A mutable deque breaks persistence; an immutable vector is a heavy dependency with weaker bounds; difference lists hide the same stack in closures; and a generic stack-safe bind added to `@lazy` merely moves this defunctionalized continuation queue into a broader package.

**Further improvement**

- [must] None. I see no design or impact blocker.

- [should] As a follow-up, retain a reproducible benchmark covering ordinary build/traverse, deep left-nested concat, and interleaved `concat`/`tail`. Performance is doing real design-selection work here; preserve the evidence without adding brittle wall-clock assertions to tests.

- [nice-to-have] Simplify `Pending` from `Nil | More(Tail, Pending) | Cat(...)` to `Nil | One(Tail) | Cat(...)`. Every construction currently uses `More(t, Nil)`, so `One` would remove impossible states and a transient `Cat(Nil, ...)` path. This is not worth delaying the fix.

- [nice-to-have] Add a public README/type-doc sentence that `LazyList` memoization is not thread-safe and externally synchronized sharing is required, matching `@lazy.Lazy`’s documented contract.

- [nice-to-have] Consider `concat_all` only if real call sites show a discoverability need; it is unnecessary for correctness after this change.

VERDICT: SHIP AS IS

bobzhang:
Applied everything actionable from the design review above, in `912105a4`. It raised no `[must]` items; these are the `[should]` and `[nice-to-have]` ones.

**Benchmark kept in-tree** — `lazy_list/concat_bench_test.mbt`, following the repo's existing `*_bench_test.mbt` / `@bench.T` convention, covering the three shapes that actually decided the representation. `moon bench ./lazy_list --target native`:

| bench (n=2048) | mean |
| --- | --- |
| build and traverse | 109.79 µs ± 0.52 µs |
| re-traverse a forced list | 26.62 µs ± 0.08 µs |
| left-nested concat, build only | 321.67 µs ± 3.14 µs |
| left-nested concat, build and traverse | 436.25 µs ± 2.33 µs |
| concat while consuming | 512.20 µs ± 2.86 µs |

The third and fourth rows are the useful pair: forcing the whole left-nested chain adds ~115 µs on top of ~322 µs of construction, i.e. traversal is not paying for the nesting depth. No wall-clock assertions are in the test suite — these exist to be re-run, not to fail CI on a slow machine.

**`Pending` lost an impossible state** — `Nil | More(Tail, Pending) | Cat(..)` → `Nil | One(Tail) | Cat(..)`. Every construction site was `More(t, Nil)`, so `More`'s tail could only ever be empty; `One` says that in the type. It also removes the transient `Cat(Nil, _)` that `uncons` used to build when a `More` ran out. The remaining `Cat(Nil, b)` arm is now genuinely unreachable — `Pending::concat` is the only outside constructor of a `Cat` and never gives one an empty side — and is commented as such rather than asserted, since an empty side means exactly "nothing to append here".

**Thread-safety documented** — this was a real gap the change introduced: the package used to inherit `@lazy.Lazy`'s "not thread-safe, synchronize externally" contract by construction, and now that tails memoize in place here it has to state it itself. Added to the `LazyList` type doc and to the README's mental-model section.

**Not taken:** `concat_all` / `join`. The review's own conclusion was that it is unnecessary for correctness after this change and should wait for real call sites, and I agree — adding public API to work around a fixed implementation defect is the wrong direction.

Full matrix re-run after the changes: `moon check --deny-warn --target all`, `moon test --target all`, `moon test --release` (js/wasm/wasm-gc and native), `--target llvm`, `moon fmt`, `moon info` — all clean, 7332/7333/7277/7249/7256 tests passing.

Issue #4074, opened 2026-08-16
string: Regex thread set grows exponentially when a repetition body has variable length (OOM at 31 characters)


bobzhang:
## Summary

A repetition whose body can consume a *variable* number of characters makes the thread set grow exponentially with the length of the subject. `^(?:.a?)*$` — "any number of (any character, optionally followed by `a`)" — exhausts a 1.4 GB heap on a **31-character** subject.

```moonbit
///|
test "blowup" {
  let re = @string.Regex("^(?:.a?)*$")
  // Fine up to about 30 characters, then the heap goes.
  ignore(re.execute("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")) // 31 'a's
}
```

On wasm-gc this ends the process:

```
<--- Last few GCs --->
[…] Scavenge (interleaved) 1369.4 (1397.9) -> 1363.3 (1414.4) MB […] allocation failure;
[…] Mark-Compact (reduce) 1363.3 (1414.4) -> 1363.2 (1383.6) MB […]
# Fatal JavaScript out of memory: Ineffective mark-compacts near heap limit
```

Nothing about the pattern is exotic — the same shape turns up as `^(?:.\s?)*$`, `(?:[[:alpha:]][[:digit:]]?)*`, `(?:\w-?)*` and so on. A 31-character subject is not an adversarial input.

## What the trigger is

The body has to be able to match more than one length. When it cannot, the engine is fine:

| pattern | body | result |
| --- | --- | --- |
| `^(?:.a)*$` | always 2 characters | fine at 200 characters (and beyond) |
| `^(?:.a?)*$` | 1 or 2 characters | out of memory at 31 |
| `^(?:.a?b?)*$` | 1, 2 or 3 characters | out of memory at 31 |

Greedy or lazy makes no difference — `^(?:.a?){2,}?$` dies at the same length — and neither does the quantifier: `*`, `+` and `{2,}` all reproduce it.

The `re""` literal path is affected as well, since an optional element cannot be folded into a character class the way an alternation of classes can:

```moonbit
///|
test "the literal path too" {
  let re = re"^(?:.a?)*$"
  ignore(re.execute("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"))
}
```

## Why it looks like a defect rather than a limit

[...]

PR #4075, opened 2026-08-16
fix(regex): stop applying a sequence's continuation twice

### LINKED ISSUE #4072: string: Regex mismatches when alternation branches accept the same character (missed matches, spans outside the language, repeat bounds exceeded)
## Summary

When an alternation's branches are single-character matchers whose character sets **overlap**, and the alternation is followed by anything else, `Regex::execute` fails to match. The characters that lie in exactly one branch match fine; the ones in the intersection are the ones that break.

```moonbit
///|
test "overlapping alternation branches" {
  let re = @string.Regex("(?:[ab]|[bc])x")
  inspect(re.execute("ax") is Some(_), content="true")  // 'a' — only branch 1
  inspect(re.execute("cx") is Some(_), content="true")  // 'c' — only branch 2
  inspect(re.execute("bx") is Some(_), content="true")  // 'b' — in BOTH: fails
}
```

The third assertion fails: `execute("bx")` returns `None`.

The smallest reproducer is an alternation of a character with itself:

```moonbit
///|
test "minimal" {
  // "ac" is plainly in the language of (?:a|a)c
  inspect(@string.Regex("(?:a|a)c").execute("ac") is Some(_), content="true")
}
```

## Scope

It is not backend-specific — wasm, wasm-gc, js and native all behave identically.

**The `re""` literal is mostly not affected**, because that path normalises an alternation of character classes into a single union class before it reaches the engine, so no `Alternation` node is ever built:

```
re"(?:[ab]|[bc])x"              =>  Sequence([Char([(97, 99)]), Char([(120, 120)])])
@string.Regex("(?:[ab]|[bc])x") =>  Sequence([Alternation([Char([(97, 98)]), Char([(98, 99)])]), Char([(120, 120)])])
```

The second one is what fails. So the exposed paths are the ones that build an `Alternation` of `Char` nodes and hand it to the engine:

- `Regex::Regex` / `Regex::compile` — i.e. any pattern that is not a compile-time literal;
- the **combinator API**, even when every piece is a literal:

```moonbit
///|
test "combinators are affected too" {
  let re = (re"[ab]" | re"[bc]") + re"x"
  inspect(re.execute("ax") is Some(_), content="true")  // ok
  inspect(re.execute("bx") is Some(_), content="true")  // fails
}
```

Everything downstream of `execute` inherits it — `find` yields no matches, `split` returns the subject unsplit, `replace_by` replaces nothing.

## What does and does not trigger it

| pattern | subject | result | |
| --- | --- | --- | --- |
| `(?:a\|b)c` | `ac` | `"ac"` | disjoint branches — fine |
| `(?:[ab]\|[cd])x` | `bx` | `"bx"` | disjoint — fine |
| `(?:a\|a)c` | `ac` | **none** | identical branches |
| `(?:a\|.)x` | `ax` | **none** | one branch subsumes the other |
| `(?:[ab]\|[bc])x` | `bx` | **none** | partial overlap, on the shared character |
| `(?:[ab]\|[bc]\|[cd])x` | `bx`, `cx` | **none** | both shared characters |
| `(?:ab\|ac)d` | `abd` | `"abd"` | multi-character branches — fine |
| `(?:a\|a)` | `a` | `"a"` | no continuation — fine |
| `c(?:a\|a)` | `ca` | `"ca"` | alternation last — fine |
| `(a\|.)c` | `ac` | `"ac"` | **capturing** group — fine |
| `(?:a\|.)*c` | `ac` | `"ac"` | quantified — fine |

Two things stand out. The alternation only breaks when something follows it, and wrapping the same alternation in a **capturing** group makes it work.

## Hypothesis

`internal/regex_engine/automata/delta.mbt` documents step 4 of the derivative loop as "Removes duplicate threads to avoid redundant computation". When a character is in the intersection of two branches, both branches produce a thread at the same position, and those threads are indistinguishable — so they are candidates for that dedup. A capture group gives the two threads distinct slots, which is consistent with the capturing form being the one that works, and with the failure needing a continuation to be observable at all. So the suspicion is that dedup is collapsing the two threads in a way that also drops the continuation they were supposed to carry, rather than keeping one thread *with* its continuation.

That is a hypothesis from the outside, not a diagnosis — I have not traced the thread set.

## Notes

Found by a property-based suite that compares `Regex` against the textbook definition of the language a regular expression denotes (all reachable end positions, computed inductively with a fixpoint for unbounded repetition). Four separate properties fail on this: anchored membership, leftmost match, `find`, and capture positions. Happy to open that suite as a PR alongside whichever fix you prefer; as it stands it would go red on this, so I have kept it back.

bobzhang:
Closes #4072.

## The bug

When more than one alternation branch accepts the same character, the alternation leaves more than one thread behind — two that have finished it where the branches are single characters, one finished and one still going where they differ in length. The text after the alternation then had to appear **twice** for the match to succeed.

```moonbit
Regex("(?:a|a)c").execute("ac")      // None
Regex("(?:a|a)c").execute("acc")     // matched "acc"
Regex("(?:a|a)bc").execute("abcbc")  // matched "abcbc"
```

Beyond missed matches it reported spans that are not in the language, and let a counted repetition run past its maximum:

```moonbit
Regex("(?:.a|a)(?:b|)").execute("abb")   // matched "abb" — the language admits ends 1 and 2 only
Regex("(?:.|ab){2,4}").execute("aaaaa")  // matched five characters
```

Both `Regex::compile` and the combinator API (`(re"[ab]" | re"[bc]") + re"x"`) were affected. The `re""` literal mostly escaped, because that path folds an alternation of character classes into a single class before the engine sees it — which is also why this went unnoticed.

## The cause

The engine is a Brzozowski-derivative simulation over a `ThreadSet`, where `Seq(pref, first, next)` means "the threads in `first`, then `next`".

`ThreadSet::remove_duplicates(self, next)` walks one set whose shared continuation is `next`. Its arm for a *finished* thread — one sitting at `Eps` — rewrote it to `Exp(marks, next)`, folding the continuation into the thread. The `Seq` arm of the same function then does

```moonbit
ts_seq(pref, first.remove_duplicates(next), next)
```

which appends `next` again. Hence the doubling. It only shows when the alternation leaves more than one thread alive, so disjoint branches were fine. Wrapping the same alternation in a capturing group also hid it — because the zero-width mark expressions a capture inserts change the continuation structure, not because slots distinguish the threads: dedup ignores marks and slots entirely.

## The fix

One line: leave the finished thread at `Eps` and emit it unchanged, keeping the `next.id` dedup key.

The key is load-bearing and the rewrite is not the same thing as the key. Keying a finished thread by `next.id` is what collapses it with a thread already sitting at `next`; without that collapse, a sequence whose first part can both finish and continue doubles its thread set every character. I tried deleting the special case outright first — keying finished threads by the `Eps` id instead — and `^(?:.a?)*$` exhausted the heap at ~31 characters. So the equivalence stays as a *key*, and only the emitted thread changes. As a bonus this restores `ts_seq`'s ability to collapse a sequence whose first part has come down to one finished thread, which the rewrite had been defeating. The resulting behaviour matches [ocaml-re](https://github.com/ocaml/ocaml-re/blob/master/lib/automata.ml), which this engine is a port of.

## Not fixed here

`^(?:.a?)*$` still exhausts the heap at 31 characters on `main` and on this branch alike — a repetition body of variable length makes the thread set grow like Fibonacci in the subject length. That is a separate, pre-existing defect, filed as #4074. This PR is deliberately no better and no worse on it.

## Tests

Five regression tests in `string/regex_test.mbt`, each of which fails without the fix: identical branches with a continuation of one and of several characters; overlapping character classes probed on both the exclusive and the shared character, plus the same alternation through the combinator API; branches of differing length in both orders; the two over-match shapes (a span outside the language, and a `{n,m}` exceeding `m`); and an anchored case plus greedy/lazy and left-branch capture preference, to pin down that the overlap does not disturb priority or which group is reported.

The property suite that found this — `Regex` against the textbook definition of the language a regular expression denotes — is not included: it also reaches #4074 and would exhaust the heap in CI. I will land it once #4074 is fixed, as a regression suite for both.

## Verification

`moon check --deny-warn --target all`, `moon test --target all`, `moon test --release` (js/wasm/wasm-gc and native), `moon test --target llvm`, `moon fmt`, `moon info`, `moon bundle --all` — all clean, no regressions across the 7000+ existing tests.

### DIFF
--- a/internal/regex_engine/automata/thread_set.mbt
+++ b/internal/regex_engine/automata/thread_set.mbt
@@ -217,15 +217,32 @@ fn ThreadSet::split_at_first_match(self : ThreadSet) -> (ThreadSet, ThreadSet) {
 }
 
 ///|
+/// Drops threads that cannot contribute anything an earlier thread does not
+/// already cover: two threads with the same future differ only in priority,
+/// so the later one can go.
+///
+/// Threads in `self` all continue with `next` once they finish, which is
+/// what makes a thread that has *already* finished — one sitting at `Eps` —
+/// equivalent to a thread sitting at `next` itself. Both are keyed by
+/// `next.id` so that the pair collapses to one; without that, a sequence
+/// whose first part can both finish and continue doubles its thread set on
+/// every character.
+///
+/// The equivalence is used as a dedup key only. Rewriting the finished
+/// thread to `Exp(marks, next)` would *look* like the same statement, but
+/// the caller wraps the result in `ts_seq(.., next)`, which appends `next`
+/// again — so the continuation would be matched twice over. Keeping the
+/// thread at `Eps` also leaves `ts_seq` able to collapse a sequence whose
+/// first part has come down to a single finished thread.
 fn ThreadSet::remove_duplicates(self : ThreadSet, next : Expr) -> ThreadSet {
   let seen = @hashset.HashSet([])
   for thread in self; result = ts_empty {
     match thread {
       End(_) => break result + ts_one(thread)
-      Exp(marks, { def: Eps, .. }) =>
+      Exp(_, { def: Eps, .. }) =>
         if !seen.contains(next.id) {
           seen.add(next.id)
-          continue result + ts_exp(marks, next)
+          continue result + ts_one(thread)
         } else {
           continue result
         }
--- a/string/regex_test.mbt
+++ b/string/regex_test.mbt
@@ -724,6 +724,129 @@ test "capture/email_with_named_groups" {
   )
 }
 
+// Alternation branches that accept the same character.
+//
+// When more than one branch can consume the same character, the alternation
+// leaves more than one thread behind — two that have finished it where the
+// branches are single characters, one finished and one still going where
+// they differ in length. A thread that has finished used to be rewritten to
+// carry the continuation, and the sequence it sat in then appended that
+// continuation a second time, so the text after the alternation had to
+// appear twice for the match to succeed. Every case below turns on that,
+// and each character is exercised on both sides of the overlap.
+
+///|
+test "execute/alternation with an overlapping branch" {
+  // The two branches are the same expression, so both accept 'a'.
+  guard @string.Regex("(?:a|a)c").execute("ac") is Some(m) else {
+    fail("expected a match")
+  }
+  inspect(m.content(), content="ac")
+  // The continuation must be consumed once, not twice.
+  guard @string.Regex("(?:a|a)c").execute("acc") is Some(m) else {
+    fail("expected a match")
+  }
+  inspect(m.content(), content="ac")
+  // ...however long it is.
+  guard @string.Regex("(?:a|a)bc").execute("abcbc") is Some(m) else {
+    fail("expected a match")
+  }
+  inspect(m.content(), content="abc")
+}
+
+///|
+test "execute/alternation of overlapping character classes" {
+  let regex = @string.Regex("(?:[ab]|[bc])x")
+  // 'a' comes only from the left branch, 'c' only from the right, and 'b'
+  // from both — it is the shared one that used to fail.
+  guard regex.execute("ax") is Some(m) else { fail("expected a match") }
+  inspect(m.content(), content="ax")
+  guard regex.execute("bx") is Some(m) else { fail("expected a match") }
+  inspect(m.content(), content="bx")
+  guard regex.execute("cx") is Some(m) else { fail("expected a match") }
+  inspect(m.content(), content="cx")
+  // The same alternation reached through the combinator API.
+  let combined = (re"[ab]" | re"[bc]") + re"x"
+  guard combined.execute("bx") is Some(m) else { fail("expected a match") }
+  inspect(m.content(), content="bx")
+}
+
+///|
+test "execute/alternation branches of different lengths" {
+  // Both branches start with 'a', and they finish at different points.
+  guard @string.Regex("(?:a|ab)c").execute("ac") is Some(m) else {
+    fail("expected a match")
+  }
+  inspect(m.content(), content="ac")
+  guard @string.Regex("(?:ab|a)c").execute("abc") is Some(m) else {
+    fail("expected a match")
+  }
+  inspect(m.content(), content="abc")
+  // A span that is not in the language must not be reported: from 0 only
+  // `a` applies, so the match can end at 1 or 2 but never at 3.
+  guard @string.Regex("(?:.a|a)(?:b|)").execute("abb") is Some(m) else {
+    fail("expected a match")
+  }
+  inspect(m.content(), content="ab")
+}
+
+///|
+test "execute/overlapping alternation under a counted repetition" {
+  // Each iteration can only take one character here, so the bounds are what
+  // decide the length.
+  guard @string.Regex("(?:.|ab){2}").execute("aaaaa") is Some(m) else {
+    fail("expected a match")
+  }
+  inspect(m.content(), content="aa")
+  guard @string.Regex("(?:.|ab){2,4}").execute("aaaaa") is Some(m) else {
+    fail("expected a match")
+  }
+  inspect(m.content(), content="aaaa")
+}
+
+///|
+test "execute/overlapping alternation keeps anchors and preference" {
+  // Anchored: the whole subject has to be consumed exactly once.
+  guard @string.Regex("^(?:a|a)c$").execute("ac") is Some(m) else {
+    fail("expected a match")
+  }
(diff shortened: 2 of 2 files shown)


bobzhang:
## Codex CLI review sign-off

Reviewed with `codex exec` (`gpt-5.6-sol`, reasoning effort **xhigh**), asked to be adversarial about the fix itself. Verbatim verdict, on commit `0358c91a`:

> **VERDICT: APPROVE WITH NITS**
>
> The implementation fix is correct, preserves priority/captures, and does not worsen the known growth issue. No blocking findings.
>
> 1. [non-blocking] The representation change is correct. `Seq` derives its `first` set before entering `next`; `Chr` produces `Exp(..., Eps)`, and `ts_seq` explicitly expects a singleton `Eps` so it can collapse it to `Exp(marks, next)`. Emitting the original thread at `remove_duplicates` therefore applies the continuation exactly once. The only callers are the recursive `Seq` arm and top-level `delta`; none requires a pre-folded continuation. This also restores the behavior of the [upstream OCaml `re` implementation](https://github.com/ocaml/ocaml-re/blob/02706daf9874ece336b84c4dbd0622cecb6c4c90/lib/automata.ml).
> 2. [non-blocking] Priority and marks are preserved. Both versions append one thread at the same ordered position and use the same dedup key. `ts_one(thread)` retains the exact `MarkSlotMap`; the singleton `ts_seq` collapse then produces the intended `Exp(marks, next)`. It cannot reorder competitors because collapse occurs only for a singleton. A scratch capture probe covering overlapping alternatives plus greedy/lazy named captures passed with the expected retained groups.
> 3. [non-blocking] The `next.id` collision is intentional and sound. Expression IDs are unique, so an unrelated expression cannot collide; the only cross-arm equality is the exact continuation node. A finished path's effective future is that continuation, so retaining the earlier path and its marks implements the ordered/Perl-style choice. One nuance relevant to #4074: recursive `remove_duplicates` calls allocate fresh `seen` sets rather than sharing one across nesting levels. That existing limitation is unchanged by this commit.
> 4. [non-blocking] The fix introduces no growth mechanism. The special arm emits no more threads than before, keeps the same key, and can now replace a singleton `Seq` wrapper with a flatter `Exp`. A local native probe of `^(?:.a?)*$` on 31 characters passed on both `origin/main` and the fix, taking 13.43s versus 13.46s—effectively identical. The known exponential behavior remains, but is no worse.
> 5. [non-blocking] The regression tests are effective. All four new test blocks fail against the unfixed implementation. After isolating assertions masked by earlier guards, the old engine independently produced `acc` instead of `ac`, `abcbc` instead of `abc`, `abb` instead of `ab`, `aaaaa` instead of `aaaa`, and failed the combinator case. The committed tests cover both construction APIs, partial overlap, branch length/order, false-positive spans, and exact/ranged repetition. A small additional anchored and capture-priority test would directly pin the invariants reviewed in items 2–3.
> 6. [non-blocking] The function doc comment is accurate, but the test preamble and commit message say overlap always leaves "one finished" and "one unfinished" thread. Identical branches and overlapping character classes actually leave multiple finished `Exp(Eps)` threads. The commit message's claim that capture slots keep threads apart is also misleading: dedup ignores marks and slots; capture groups mask these cases because their inserted zero-width mark expressions change the continuation structure.
>
> Validation passed: `moon fmt --check`, `moon info` with no source `.mbti` changes, `moon check --deny-warn --target all`, all 276 `string` tests on wasm/wasm-gc/JS/native, and the full all-target suite. The tracked worktree remains clean.

Both actionable nits are applied in `e4b607e6`.

**Nit 5** — added `execute/overlapping alternation keeps anchors and preference`, covering the anchored form (`^(?:a|a)c$` must match `ac` and must *not* match `acc`), greedy vs lazy inside the overlapping branch (`(?:(a+)|a)(a*)` → group 1 `aaa`; `(?:(a+?)|a)(a*)` → group 1 `a`), and left-branch preference (`(?:(a)|(a))b` → group 1 set, group 2 unset). Verified it fails against the unfixed engine.

**Nit 6** — both statements were wrong and are corrected in the test preamble, the commit message and the PR description. Overlap leaves *more than one* thread, and where the branches are single characters both of them have finished. And the reason a capturing group hides the bug is that the zero-width mark expressions a capture inserts change the continuation structure — not that slots distinguish the threads, since dedup ignores marks and slots entirely. That was my hypothesis in #4072 and it was incorrect; I have noted the correction there.

Item 1's observation that this restores the behaviour of ocaml-re, which the engine is a port of, is the strongest independent evidence the change is right, and I have cited it in the PR description.

hackwaly (approved):

PR #3315, opened 2026-03-19
perf(sorted_set): O(n+m) difference, intersection, symmetric_difference

bobzhang:
## Summary

Replace O(n·log m) set operations with split-based divide-and-conquer, matching the approach used in `immut/sorted_set`.

### Before
```moonbit
// O(n * log m) — iterates self, calls contains on src for each element
self.each(x => if !src.contains(x) { ret.add(x) })
```

### After
```moonbit
// split-based recursive merge on private copies of both trees
fn aux(a, b) {
  match (a, b) {
    (Some({ value: va, left: la, right: ra, .. }), _) => {
      let (lb, found, rb) = split_member(b, va)
      // recursively process left and right subtrees
    }
  }
}
```

The split/join merge is O(m·log(n/m + 1)) ⊆ O(n + m); the upfront `copy_tree` of both operands is Θ(n + m), so each operation is O(n + m) overall. For two 100K-element sets this is roughly 17x less work than the old O(n·log m) loops.

### Changes
- Add `split_member` — like `split` but also returns whether the pivot was found
- Add `concat` — joins two trees where all left elements < all right elements
- Add `remove_min` — helper for `concat`
- Rewrite `difference`, `intersection`, `symmetric_difference` using the split-based approach
- Result sizes are derived arithmetically from the number of shared elements discovered during the merge (each shared element is found exactly once as a split pivot), so no separate counting traversal of the result is needed
- Empty operands short-circuit before any tree is copied, mirroring `union`
- Remove the TODO comment about symmetric_difference optimization

### Update (2026-08-19 rebase)
Rebased onto current `main` (conflicts with the `#owned` annotations and the `new()` → `new_sorted_set()` rename resolved), and split into two commits:

1. **perf** — the rewrite above, now with the arithmetic size derivation and empty-operand fast paths.
2. **test** — new blackbox QuickCheck properties (filter-model equivalence for content/order/size, algebra laws, self/empty identities, result/input independence) and a new whitebox `invariant_wbtest.mbt` (mirroring `sorted_map`'s) that recursively verifies BST bounds, AVL balance, cached heights, and the `size` field of every operation result. All new tests are mutation-verified: a size-formula mutation fails 9 tests, a balance-only mutation fails exactly the 3 whitebox scenario tests (blackbox tests are provably blind to it), and skipping one `copy_tree` fails 6 tests including the dedicated no-sharing property.

## Test plan
- [x] Full repo `moon test`: 7454/7454 pass
[...]

PR #4108, opened 2026-08-19
perf(sorted_set): probe fallback for strongly asymmetric difference/intersection

bobzhang:
## Summary

Follow-up to #3315. Post-merge benchmarking of the split-based set operations against the old iterate+contains loops found one severe asymmetric regression: the split-based merge copies **both** operands up front (Θ(n + m)), so `small.difference(large)` went from 2.7 µs to 1.5 ms (native, 100 vs 100K elements) — the old probe loop was O(n log m) and sublinear in the larger set.

### Fix

- **`difference`**: falls back to the probe loop (`self.each` + `src.contains`) when `self.size < src.size / 16`.
- **`intersection`**: its key set is symmetric and fits in the smaller side, so either strongly smaller side probes: `self` smaller probes directly; `src` smaller probes with `src` but inserts the value **looked up in `self`**, preserving `self`'s stored representative for types whose `Compare` inspects only part of the value — the same semantics as the split path, which always emits `self`'s `va` (this subtlety was caught by Codex review, see below).
- **`symmetric_difference`**: deliberately unguarded — every element of both sides can land in the result, so the copies are inherent; measured old-vs-new wins at every shape tried (e.g. 100 vs 100K: 18.7 ms → 1.6 ms native).

### Why 1/16

The cutoff is tuned to the probe loop's **worst case** (disjoint sets: every probed element is inserted into the growing result), measured on all four backends at m = 100K:

| boundary shape | native | js | wasm-gc | wasm |
|---|---|---|---|---|
| probe loop, disjoint n=6000 | 0.88 ms | 0.66 ms | 0.55 ms | 3.6 ms |
| split merge just above cutoff | 1.60 ms | 0.78 ms | 0.68 ms | 4.0 ms |

The probe loop stays ahead of the merge in its worst case at the cutoff on every backend, while at n = m/8 the split merge already wins the same worst case (native 1.70 ms vs 1.93 ms) — so 1/16 is the conservative choice. Just below the cutoff the guarded API is at parity with a hand-written probe loop (266 µs vs 271 µs native at n=6000): the guard adds no measurable overhead.

### Measured effect (old loop vs guarded API; native / js / wasm-gc / wasm)

| case | before (unguarded merge) | after |
|---|---|---|
| `difference` 100 vs 100K | 1.5 ms / 0.87 ms / 0.69 ms / 3.8 ms | **2.2 µs / 2.3 µs / 1.8 µs / 11.5 µs** (parity with old) |
| `intersection` 100K vs 100 | old code itself took 1.25 ms / 2.3 ms / 1.2 ms / 3.7 ms | **10 µs / 7 µs / 6 µs / 34 µs** |
| equal-size 100K, 50% overlap | — | unchanged split path, 2.4–9.1x faster than old loops |

### Tests

- QuickCheck property forcing the asymmetric fallback in **both operand orders** against the filter model (content, order, size).
- Whitebox invariant scenarios (BST bounds, AVL balance, cached heights, size field, input intactness) at 30-vs-3000 sizes.
- A key-only-`Compare` payload test proving all three `intersection` paths and the `difference` fallback emit `self`'s stored representatives.
- Mutation-verified: corrupting the `difference` fallback fails 4 tests; corrupting the `intersection` fallback fails 6; re-introducing the representative bug fails the `Keyed` test.

## Test plan
- [x] `sorted_set`: 70/70 on wasm-gc, native, and js
- [x] `moon info`: no public API change
- [x] Codex CLI review, 4 rounds: rounds 1–2 at `xhigh` (representative-preservation blocker fixed and re-verified), rounds 3–4 at `ultra` (no code/API/test findings; one commit-message legend inconsistency fixed) — see review comments

### DIFF
--- a/sorted_set/invariant_wbtest.mbt
+++ b/sorted_set/invariant_wbtest.mbt
@@ -182,6 +182,16 @@ test "set operations keep invariants (asymmetric sizes)" {
 }
 
 ///|
+test "set operations keep invariants (extreme asymmetry, probe fallback)" {
+  // One side far below 1/16 of the other: difference and intersection take
+  // the O(n log m) probe fallback instead of the split-based merge.
+  for seed in 1..<=3 {
+    set_ops_scenario(0xACE1UL + seed.to_uint64(), 3000, 30, 5000)
+    set_ops_scenario(0xACE2UL + seed.to_uint64(), 30, 3000, 5000)
+  }
+}
+
+///|
 test "set operations with self and empty keep invariants" {
   let rng = Rand::{ state: 77UL }
   let a : SortedSet[Int] = new_sorted_set()
--- a/sorted_set/quickcheck_test.mbt
+++ b/sorted_set/quickcheck_test.mbt
@@ -65,6 +65,38 @@ test "quickcheck: symmetric_difference matches the filter model" {
 }
 
 ///|
+test "quickcheck: strongly asymmetric sizes match the filter model" {
+  // Forces the O(n log m) probe fallback in difference/intersection
+  // (taken when one side is under 1/16 of the other) and checks it against
+  // the same filter model as the split-based path.
+  @quickcheck.check((input : (Array[Int], Array[Int])) => {
+    let (xs, ys) = input
+    let k = if xs.length() < 10 { xs.length() } else { 10 }
+    let small = @sorted_set.from_array(xs[0:k])
+    let big = @sorted_set.from_array(ys)
+    for i in 0..<600 {
+      big.add(10000 + i)
+    }
+    let d_small_big = small.difference(big)
+    let e_d_small_big = small.to_array().filter(x => !big.contains(x))
+    let d_big_small = big.difference(small)
+    let e_d_big_small = big.to_array().filter(x => !small.contains(x))
+    let i_small_big = small.intersection(big)
+    let e_i_small_big = small.to_array().filter(x => big.contains(x))
+    let i_big_small = big.intersection(small)
+    let e_i_big_small = big.to_array().filter(x => small.contains(x))
+    d_small_big.to_array() == e_d_small_big &&
+    d_small_big.length() == e_d_small_big.length() &&
+    d_big_small.to_array() == e_d_big_small &&
+    d_big_small.length() == e_d_big_small.length() &&
+    i_small_big.to_array() == e_i_small_big &&
+    i_small_big.length() == e_i_small_big.length() &&
+    i_big_small.to_array() == e_i_big_small &&
+    i_big_small.length() == e_i_big_small.length()
+  })
+}
+
+///|
 test "quickcheck: set algebra laws relate the operations" {
   @quickcheck.check((input : (Array[Int], Array[Int])) => {
     let (xs, ys) = input
--- a/sorted_set/set.mbt
+++ b/sorted_set/set.mbt
@@ -152,6 +152,26 @@ pub fn[V : Compare] SortedSet::contains(self : SortedSet[V], value : V) -> Bool 
 }
 
 ///|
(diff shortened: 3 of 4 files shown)


bobzhang:
## Codex CLI review

Adversarial review by Codex CLI (`gpt-5.6-sol`, reasoning effort `xhigh`, read-only sandbox), briefed with the full benchmark matrix for all four backends (native / js / wasm-gc / wasm): the regression numbers, the cutoff-boundary worst-case measurements backing the 1/16 constant, and the guard-overhead parity numbers. Asked to attack the guard conditions (including tiny-set integer division), the intersection operand handling, the early returns, prose accuracy, and test adequacy.

### Round 1 — one blocking issue

> **Blocking:** `intersection` no longer consistently preserves `self`'s stored representative. When `src` is much smaller, it becomes `small` and its value is inserted directly; the split path instead inserts `va` from `self`. This is observable for a valid type whose `Eq`/`Compare` use a key but which carries additional payload — stored representatives matter because `add` explicitly replaces compare-equal values. For example, with 32 `self` elements and a one-element `src` containing the same key but different payload, the fallback returns `src`'s payload whereas the previous implementation returned `self`'s. The new property uses only `Int`, so it cannot detect this regression. Probe using the smaller side for keys, but retrieve and insert the matching representative from `self`; add a key-only `Compare` test with distinct payloads.

**Resolution:** the operand-swap tuple was replaced with two explicit guards; when `src` is the probing side, each match inserts `lookup(self.root, x)` — a new private helper returning `self`'s stored value that compares equal — so both fallbacks emit `self`'s representatives like the split path. A `Keyed` (key-only `Compare` + payload) regression test covers all three intersection paths and the difference fallback; re-introducing the bug fails it. Re-measured after the fix: intersection 100K-vs-100 is 10.2 µs native / 33.6 µs wasm — unchanged.

### Round 2 — focused re-review of the fix: approved

> No blocking findings. The fix preserves `self` representatives across both probe fallbacks and the split path; `lookup` and the `Keyed` regression test are adequate. No new issues found.
>

---
*Posted on behalf of Codex CLI by Claude Code. Validation outside Codex's read-only sandbox: `sorted_set` 70/70 on wasm-gc/native/js; `moon check` clean; no `.mbti` change; all three fallback branches mutation-verified.*

bobzhang:
## Codex CLI review — rounds 3–4 at reasoning effort `ultra`

The two rounds above ran at `xhigh`. It turns out the current Codex CLI (v0.145) has an `ultra` reasoning tier above `xhigh`, so the branch was re-reviewed at `ultra` as a final gate: a complete fresh pass plus a focus list (tiny-set integer-division edge cases in the guards, `lookup` correctness and `each`-ordering interaction in the src-probing fallback, early-return safety, every prose claim against the four-backend benchmark matrix, residual test gaps).

### Round 3 (`ultra`, full pass) — one blocking finding, prose-only

> The implementation and test audits are clean: the guards are conservative for tiny sizes, both intersection orders preserve `self` representatives, `lookup` mirrors `contains`, and the early returns skip only split-specific work.
>
> **Blocking finding:** Commit message lines 24–29 label benchmarks "old loop vs guarded API," but the intersection figures are ordered guarded API vs old code. Amend the message header or reorder those figures before merge.
>
> No code, API, or test correctness findings.

**Resolution:** the commit message legend now reads "guarded API vs old loop", matching the figure order (commit `4cf512c7`, message-only amendment — the tree is unchanged from the reviewed commit).

### Round 4 (`ultra`, focused confirmation) — approved

> No blocking findings. Commit `4cf512c7` now says "guarded API vs old loop," matching the figure order. Other claims are unchanged, and the review paragraph records all three rounds.
>

---
*Posted on behalf of Codex CLI by Claude Code.*

PR #4109, opened 2026-08-19
perf(list): single-pass two-pointer has_suffix; rename is_prefix/is_suffix with deprecated aliases

bobzhang:
## Summary

Supersedes the `is_suffix` optimization proposed in #3840 (thanks @mizchi for identifying the allocation problem and the benchmark methodology) with a strictly faster single-pass version, and renames the pair to `has_prefix`/`has_suffix` to match the `String`/`Bytes` naming convention.

### Three implementations compared

| | approach | cost |
|---|---|---|
| main | `self.rev().is_prefix(suffix.rev())` | 2 reversal allocations (O(n)+O(m) nodes) |
| #3840 | `len cmp` + `drop` + `is_prefix` | allocation-free, but **2n + m** steps (two `length()` walks, then drop, then compare) |
| this PR | two-pointer lead/lag | allocation-free, **n + m** steps, no `length()` calls |

The algorithm: advance a `lead` pointer one node per `suffix` node (proving along the way that the suffix isn't longer), then slide `lead` and a `lag` pointer together; when `lead` falls off the end, `lag` sits exactly `suffix.length()` nodes from the end — the only window where the suffix can match — and an element-wise comparison of that equal-length window decides. An empty suffix returns `true` in O(1).

### Benchmarks (`moon bench --release`, n=100K, m=50K unless noted; old / #3840 / this)

| case | native | js | wasm-gc |
|---|---|---|---|
| match | 1.33 ms / 418 µs / **287 µs** | 621 / 337 / **200 µs** | 783 / 312 / **221 µs** |
| mismatch at window start | 1.34 ms / 315 µs / **184 µs** | 610 / 255 / **124 µs** | 758 / 237 / **140 µs** |
| shared-tail suffix (`l.drop(50K)`) | 1.34 ms / 394 µs / **272 µs** | 469 / 332 / **194 µs** | 745 / 318 / **211 µs** |
| tiny suffix m=10 | 833 µs / 305 µs / **154 µs** | — | — |
| empty suffix | 817 µs / 302 µs / **6.7 ns** | 260 / 234 µs / **8.1 ns** | 263 / 237 µs / **7.5 ns** |

1.4–2x faster than #3840's version on every shape, 3–5x faster than main, and O(1) instead of O(n) for the empty suffix.

### Rename

`is_suffix` → `has_suffix`, `is_prefix` → `has_prefix`, each keeping the old name as `#alias(is_..., deprecated)` — the same convention as `String::has_suffix` (which itself carries `ends_with` as a deprecated alias) and `Bytes::has_prefix`/`has_suffix`. All in-repo call sites (tests, docstrings, README) are migrated; a repo-wide grep found no callers outside the `list` package.

### Semantics note (caught by Codex review)

An earlier draft compared the window with List `==`, whose physical-equality shortcut would have made a NaN-carrying list a suffix of itself — observably diverging from the old element-wise behavior. The window comparison is element-wise (`has_prefix` on the equal-length window), reproducing the old semantics exactly, and a NaN regression test pins this.

### Tests

- Boundary coverage: empty/whole/longer/same-length-different/window-start-mismatch/single-element/physically-shared-tail.
- NaN test pinning the element-wise window semantics.
- QuickCheck property: `has_suffix` against an Array-slice reference model, plus the law that every `drop(k)` is a suffix — both as the physically shared tail and as an unshared copy.
- Mutation-verified: skipping the window comparison fails 4 tests; accepting a longer suffix fails 2; the NaN test fails on the `==`-based window.

## Test plan
- [x] `list`: 226/226 on wasm-gc and js, 218/218 on native (`panic_test.mbt` is excluded on native)
- [x] `moon info`: `.mbti` change is exactly the rename + deprecated aliases
- [x] Codex CLI review (2 rounds, reasoning `xhigh`): round-1 blocking issue (NaN/physical-equality semantics) fixed and re-verified; round-2 sign-off — see review comment

### DIFF
--- a/list/README.mbt.md
+++ b/list/README.mbt.md
@@ -364,8 +364,8 @@ test {
 ///|
 test {
   let list = @list.List([1, 2, 3, 4, 5])
-  @test.assert_eq(list.is_prefix(List([1, 2, 3])), true)
-  @test.assert_eq(list.is_suffix(List([4, 5])), true)
+  @test.assert_eq(list.has_prefix(List([1, 2, 3])), true)
+  @test.assert_eq(list.has_suffix(List([4, 5])), true)
 }
 ```
 
--- a/list/list.mbt
+++ b/list/list.mbt
@@ -1733,10 +1733,11 @@ pub fn[A : Eq] List::remove(self : List[A], elem : A) -> List[A] {
 ///
 /// ```mbt check
 /// test {
-///   @test.assert_eq(@list.List([1, 2, 3, 4, 5]).is_prefix(List([1, 2, 3])), true)
+///   @test.assert_eq(@list.List([1, 2, 3, 4, 5]).has_prefix(List([1, 2, 3])), true)
 /// }
 /// ```
-pub fn[A : Eq] List::is_prefix(self : List[A], prefix : List[A]) -> Bool {
+#alias(is_prefix, deprecated)
+pub fn[A : Eq] List::has_prefix(self : List[A], prefix : List[A]) -> Bool {
   for a = self, b = prefix {
     match (a, b) {
       (_, Empty) => break true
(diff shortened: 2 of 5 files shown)


bobzhang:
## Codex CLI review

Adversarial review by Codex CLI (`gpt-5.6-sol`, reasoning effort `xhigh`, read-only sandbox), briefed with the two-pointer algorithm's invariants to verify, the three-way benchmark numbers (native/js/wasm-gc), the rename's mechanical-completeness checklist, and the test/mutation evidence. Asked to attack: off-by-ones in both phases, the unreachable-arm justification, the empty-suffix guard, Eq-based vs element-wise window comparison, `.mbti` surface, prose claims, and test adequacy.

### Round 1 — one blocking issue

> **Blocking:** `lag == suffix` changes behavior for physically shared tails. `List::equal` immediately returns `true` for physically identical nodes, bypassing element equality. Consequently, for `xs = List([@double.not_a_number])`, the old `xs.is_suffix(xs)` returned `false` because `NaN == NaN` is false; the new implementation returns `true`. The deprecated alias therefore changes observable behavior. Shared-tail tests use only reflexive `Int` values, and QuickCheck is likewise restricted to `Int`. Use element-wise comparison for the equal-length window, or explicitly specify and test this semantic change.

**Resolution:** the window comparison is now `lag.has_prefix(suffix)` — element-wise on the equal-length window, reproducing the old rev-based comparison set exactly — with a comment explaining why List `==` is deliberately not used, and a NaN regression test (`xs == xs` holds, `xs.has_suffix(xs)` and `xs.has_prefix(xs)` do not). Benchmarks were re-run after the fix; the single-pass structure, not the physical shortcut, was the performance story, so the numbers barely moved.

### Round 2 — focused re-review of the fix: approved

> Approved. No blocking issues in the ROUND 2 delta.
>
> The equal-length invariant makes `lag.has_prefix(suffix)` correct and element-wise. The NaN test directly catches the prior physical-equality bug, comments are accurate, and the commit message removes the invalid shared-tail O(1) claim. The delta only modifies the implementation and regression test; `git diff --check` is clean.
>

---
*Posted on behalf of Codex CLI by Claude Code. Validation outside Codex's read-only sandbox: `list` 226/226 on wasm-gc/js and 218/218 on native; `moon check` clean; `.mbti` diff is exactly the rename; mutations verified as described in the PR body.*

PR #4110, opened 2026-08-19
refactor: use `<+` template writing in library code

bobzhang:
Part 2 of 2. **Library code only** — the test-file half is #4111, which can
land independently.

Sweep replacing `write_string` / `write_char` / `write_object` sequences with
the `<+` template-writing operator, continuing what 1c47134d ("Use
interpolation for fixed output builders") started in `show.mbt`,
`json/types.mbt` and `ref.mbt`.

15 files, **-130 / +126** lines (+44 of which is the new benchmark, so the
library code itself is a net -48).

## Which form, and why

I benchmarked the three candidate spellings before sweeping, because they are
*not* equivalent. All numbers below are `moon bench`, 2000 writes per
iteration, native and js:

| spelling | native | js |
| --- | --- | --- |
| `write_string("(") ; write_object(a) ; …` (baseline) | 50.1 µs | 71.4 µs |
| `logger <+ "(\{a}, \{b})"` | 70.7 µs (**+41%**) | 100.3 µs (**+40%**) |
| `logger <+ "(\{l => l.write_object(a)}, …)"` | 49.1 µs (par) | 75.3 µs (par) |
| `write_string("x=\{a}, y=\{b}")` (baseline) | 155.5 µs | 95.1 µs |
| `logger <+ "x=\{a}, y=\{b}"` | 66.3 µs (**2.3x**) | 98.9 µs (par) |
| `logger <+ "x=\{l => …}, y=\{l => …}"` | 45.8 µs (**3.4x**) | 64.5 µs (**1.5x**) |
| `write_string("aaaa"); write_string("bbbb"); write_string("cccc")` | 28.9 µs | 27.3 µs |
| `logger <+ "aaaabbbbcccc"` | 13.6 µs (**2.1x**) | 6.7 µs (**4.1x**) |

The reason is in the desugaring. `\{x}` becomes
`write_string_interpolation(x)`, whose parameter is `&Show`, so every
interpolation coerces its argument into a trait object — one allocation per
hole, on every backend:

```js
logger.method_table.method_4(logger.self,
  { self: a, method_table: Int_as_Show });   // <- allocation
```

The inline-writer form `\{l => …}` hits the compiler's `make_inline_write`
path instead and desugars to `writer |> (l => …)`, which emits *byte
identical* code to the direct call it replaces. And adjacent literal chunks
are concatenated at compile time, so a run of literal writes collapses into
one `write_string`.

So the sweep uses:

* `buf.write_string("...\{x}...")` → `buf <+ "...\{x}..."` — the old form
  built a throwaway `StringBuilder` + `String` and then copied it in.
* runs of adjacent literal writes → one template.
* `write_object(x)` → `\{l => l.write_object(x)}`, **not** `\{x}`, so
  monomorphic `Show` dispatch is preserved. (#4111 uses plain `\{x}` in test
  bodies, where readability wins and one box per element in a three-element
  assertion is irrelevant.)

## Streaming instead of materializing

Three helpers grew writer variants so callers stop building intermediate
strings:

* `json`: `escape(str, escape_slash~) -> String` → `escape_to(buf, str,
  escape_slash~)`. `stringify` now escapes straight into the output builder
  rather than allocating one escaped `String` per key and per string value.
* `v128`: added `u64_hex_to(logger, value)`; `u64_hex` keeps its `String`
  signature for the `Debug` impl.
* `builtin`: `BytesView::escape_to` — this also de-duplicates the escaping
  loop, which `Show` and `ToJson` each had their own copy of. It is generic
  over the logger (`fn[L : Logger]`) rather than taking `&Logger`, so
  `ToJson`'s concrete `StringBuilder` keeps direct dispatch; taking `&Logger`
  there measured ~5% slower on `Bytes::to_json` (92.8 µs → 98.1 µs native,
  4 KiB input), and the generic version is back at 93.3 µs.

`json/stringify_bench_test.mbt` is added here so the claim is reproducible.
Escape-heavy document, n = 200:

|  | before | after | |
| --- | --- | --- | --- |
| native | 152.0 µs | 122.8 µs | **1.24x** |
| native, `indent=2` | 158.4 µs | 131.5 µs | **1.20x** |
| js | 193.6 µs | 175.8 µs | **1.10x** |
| js, `indent=2` | 206.3 µs | 186.2 µs | **1.11x** |

## Deliberately not touched

* **`builtin/tuple_show.mbt`** (150 `write_string` calls, the single biggest
  candidate). Keeping it allocation-free needs the inline-writer form, and
  the 16-tuple impl would become one ~450 character string literal that
  `moon fmt` cannot break. The five-line-per-impl status quo reads better.
* **Isolated `write_string("literal")` calls.** `x <+ "lit"` desugars to
  exactly `x.write_string("lit")`, so converting the ~250 standalone ones
  would be pure diff noise. They are converted only inside `Show` impls that
  this PR was already rewriting.
* **Isolated `write_string(x.to_string())`.** For types whose `Show` only
  implements `to_string` (`Int`, `Byte`, …), the default `output` *is*
  `write_string(self.to_string())` — so `<+ "\{x}"` would allocate the same
  string and add a box on top.

## Follow-up worth considering (not in this PR)

The pre-existing `<+ "\{x}"` sites in `builtin/show.mbt` (`Show for X?`,
`Show for Result`), `ref.mbt` and `cmp.mbt` pay that ~40% trait-object cost
today. Either switching them to the inline-writer form, or teaching the
desugarer to emit a monomorphic `write_object` when the interpolated
expression's `Show` instance is statically known, would recover it — the
latter would make plain `\{x}` the right answer everywhere and is the better
fix.

## Verification

* `moon check --deny-warn --target all`
* `moon test` on wasm / wasm-gc / js / native — 7458 / 7459 / 7403 / 7375
  passed, 0 failed (re-run after the split)
* `moon info --target wasm,wasm-gc,js,native` — no `.mbti` drift (every new
  helper is package-private)
* `moon fmt` clean

## Review

Reviewed by Codex CLI (`codex exec`, read-only sandbox) against the desugaring
in `parsing_util.ml`. Verdict: approve, no correctness or shipped-path
performance blocker. It independently confirmed output equivalence for
`Show for Json` (including the `repr=Some(...)` case), both `stringify` escape
sites, the `json_path` recursion moved into interpolation holes, the
`BytesView` quote split between `Show` and `ToJson`, `diff/hunk`'s
`write_string(x.to_string())` → `write_object(x)`, `V128`'s padding and word
order, and the literal-brace vs `\{` escaping in the `\u{...}` rewrites.

Its two findings are addressed: the `&Logger` dispatch on
`BytesView::to_json` is fixed by the generic signature, and the plain-`\{x}`
holes it flagged were all in test code, which now lives in #4111 with that
choice called out as deliberate.

Note that the review ran against the pre-split commit (all 33 files); the
library hunks here are unchanged from what it read.

### DIFF
--- a/builtin/bytesview.mbt
+++ b/builtin/bytesview.mbt
@@ -325,17 +325,25 @@ pub fn BytesView::iter2(self : BytesView) -> Iter2[Int, Byte] {
 }
 
 ///|
-pub impl Show for BytesView with fn output(self, logger) {
-  logger.write_string("b\"")
+/// Writes the bytes with printable ASCII kept as-is and everything else
+/// rendered as `\xHH`, without the surrounding `b"`/`"`. Shared by `Show`,
+/// which adds the quotes, and `ToJson`, which does not.
+///
+/// Generic over the logger rather than taking `&Logger` so that `ToJson`'s
+/// concrete `StringBuilder` keeps direct dispatch.
+fn[L : Logger] BytesView::escape_to(self : BytesView, logger : L) -> Unit {
   for byte in self {
     if byte is (b' '..=b'~') && byte != b'"' && byte != b'\\' {
       logger.write_char(byte.to_char())
     } else {
-      logger.write_string("\\x")
-      logger.write_string(byte.to_hex())
+      logger <+ "\\x\{l => l.write_string(byte.to_hex())}"
     }
   }
-  logger.write_string("\"")
+}
+
+///|
+pub impl Show for BytesView with fn output(self, logger) {
+  logger <+ "b\"\{l => self.escape_to(l)}\""
 }
 
 ///|
@@ -577,14 +585,7 @@ pub fn BytesView::to_owned(self : BytesView) -> Bytes {
 ///|
 pub impl ToJson for BytesView with fn to_json(self) -> Json {
   let sb = StringBuilder()
-  for byte in self {
-    if byte is (b' '..=b'~') && byte != b'"' && byte != b'\\' {
-      sb.write_char(byte.to_char())
-    } else {
-      sb.write_string("\\x")
-      sb.write_string(byte.to_hex())
-    }
-  }
+  self.escape_to(sb)
   Json::string(sb.to_string())
 }
 
--- a/builtin/char.mbt
+++ b/builtin/char.mbt
@@ -498,9 +498,7 @@ fn Char::escape_to(self : Char, logger : &Logger, quote? : Bool = true) -> Unit 
     ' '..='~' => logger.write_char(self)
     _ =>
       if !self.is_printable() {
-        logger.write_string("\\u{")
-        logger.write_string(self.to_hex())
-        logger.write_char('}')
+        logger <+ "\\u{\{l => l.write_string(self.to_hex())}}"
       } else {
         logger.write_char(self)
       }
--- a/builtin/console.mbt
+++ b/builtin/console.mbt
@@ -81,8 +81,7 @@ fn base64_encode(data : FixedArray[Byte]) -> String {
     let x1 = base64[(b0 & 0x03) << 4]
     buf.write_char(x0.to_char())
     buf.write_char(x1.to_char())
-    buf.write_char('=')
-    buf.write_char('=')
+    buf <+ "=="
   } else if rem == 2 {
     let b0 = data[len - 2].to_int()
     let b1 = data[len - 1].to_int()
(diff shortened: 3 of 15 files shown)


bobzhang:
Split: the test-file half moved to #4111. This PR is now library code only (15 files), which is the part that wants review. The Codex review quoted in the description ran against the pre-split commit; the library hunks here are byte-for-byte what it read.

PR #4111, opened 2026-08-19
refactor(test): use `<+` template writing in test files

bobzhang:
Part 1 of 2, split out of #4110. **Test and benchmark files only** — no
library code is touched, and no expected output changes.

18 files, **−85 / +52** lines.

## What changes

* `buf.write_string("...\{x}...")` → `buf <+ "...\{x}..."`. The old form built
  a throwaway `StringBuilder` + `String` for the interpolation and then copied
  the result into `buf`; the template writes the pieces straight through.

* Runs of writes collapse into one template:

  ```moonbit
  arr.rev_eachi((i, x) => {
    buf.write_object(i)
    buf.write_string(": ")
    buf.write_object(x)
    buf.write_string("\n")
  })
  // becomes
  arr.rev_eachi((i, x) => buf <+ "\{i}: \{x}\n")
  ```

  Adjacent literal chunks are concatenated by the desugarer at compile time,
  so the literal part of a run becomes a single `write_string`.

## Note on the hole form

These use the plain `\{x}` hole, which desugars to
`write_string_interpolation(x)` and coerces `x` to `&Show` — one box per hole.
That is the right trade in a three-element assertion, but not in library code,
which is why #4110 uses the inline-writer form `\{l => l.write_object(x)}`
there instead. The reasoning and benchmarks for that choice live in #4110.

## Scope boundary

Strictly `*_test.mbt` / `*_wbtest.mbt` files. `builtin/fixedarray.mbt` also has
a `<+` rewrite confined to an in-file `test {}` block, but it is a library file,
so it stays in #4110 rather than here.

## Verification

* `moon check --deny-warn --target all`
* `moon test` on wasm / wasm-gc / js / native — 7458 / 7459 / 7403 / 7375
  passed, 0 failed
* `moon fmt` clean

### DIFF
--- a/bigint/bigint_nonjs_wbtest.mbt
+++ b/bigint/bigint_nonjs_wbtest.mbt
@@ -17,9 +17,8 @@ struct MyBigInt(BigInt)
 
 ///|
 impl Show for MyBigInt with fn output(self, logger) {
-  logger.write_string(
-    "{limbs : \{@debug.Repr(self.limbs)}, sign : \{@debug.Repr(self.sign)}, len : \{self.len} }",
-  )
+  logger <+
+    "{limbs : \{@debug.Repr(self.limbs)}, sign : \{@debug.Repr(self.sign)}, len : \{self.len} }"
 }
 
 ///|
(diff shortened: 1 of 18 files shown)


bobzhang:
Split from #4110. This half is test-only and can land independently; #4110 now carries the library changes and is the one that needs review.

PR #4112, opened 2026-08-20
perf(builtin): pattern-based fast paths for char predicates

bobzhang:
## Summary

Pattern-based fast paths for `Char::is_ascii_whitespace`, `Char::is_whitespace`, and `Char::is_numeric` — the speed of #3703's integer rewrites where patterns can deliver it, with every check staying in the char pattern DSL. Follow-up to [the benchmark review posted on #3703](https://github.com/moonbitlang/core/pull/3703#issuecomment-5344270021) (thanks @mizchi for the bench methodology and for identifying the hot spots), which showed that most of that PR's win is early-exit *structure*, not integer arithmetic — and that the arithmetic actually loses to patterns on some backends.

### Changes

```moonbit
// six singletons -> one range + space
pub fn Char::is_ascii_whitespace(self : Self) -> Bool {
  self is ('\u{09}'..='\u{0D}' | ' ')
}

// ASCII fast path + below-table reject in front of the unchanged range table
pub fn Char::is_numeric(self : Self) -> Bool {
  if self is ('0'..='9') {
    return true
  }
  if self < '\u{B2}' {   // U+00B2 is the table minimum
    return false
  }
  self is (...)          // table unchanged, leading digit range moved to the guard
}

// one ASCII branch deciding via the same range pattern; non-ASCII table unchanged
pub fn Char::is_whitespace(self : Self) -> Bool {
  if self <= '\u{20}' {
    return self is ('\u{09}'..='\u{0D}' | ' ')
  }
  self is (...)          // table minus the two ASCII alternatives
}
```

### Benchmarks

#3703's `char_predicate_bench_test.mbt` methodology, `moon bench --release`, same-day interleaved baselines, branch cells are means of two runs (repeats within ~2%):

| bench | native | js | wasm-gc |
|---|---:|---:|---:|
| `is_numeric` ascii n=62000 | 886 µs → **79 µs (11.2x)** | 1.44 ms → **224 µs (6.4x)** | 2.65 ms → **164 µs (16.2x)** |
| `is_numeric` unicode n=22000 | 250 → **156 µs** | 758 → **415 µs** | 807 → **460 µs** |
| `is_ascii_whitespace` ascii | 13.7 → 13.9 µs (par) | 226 → **112 µs (2.0x)** | 77.8 → **70.5 µs** |
| `is_whitespace` ascii / unicode | par | par | 147/53.1 → **126/46.0 µs** |

Versus #3703's integer versions: its `is_ascii_whitespace` arithmetic measures 28.1 µs on native — **2.0x slower than this pattern** — and 1.9x slower on js; its `is_numeric` ties this version on all three backends. Only `is_whitespace` keeps an arithmetic edge on native (32 vs 68 µs): every pattern form that reaches it there regresses another backend (char-compare guards are 1.7x slower on wasm-gc than the arithmetic and slower than main; a single ordered `match` ties wasm-gc but is 1.8x slower on native). This PR takes the only readable form measured to regress **nowhere**, and treats the remaining gap as a moonc codegen issue (char patterns and comparisons should lower to the same decision tree regardless of surface syntax) rather than a reason to hand-write integer code in core.

### Tests

`char/predicate_reference_test.mbt` keeps the previous flat patterns **verbatim** as reference functions and checks all three predicates against them for **every Unicode scalar value** (1,112,064 code points; `Int::to_char` excludes exactly the surrogates). Mutation-verified: shifting the numeric reject bound by one fails at code point 178; shortening the ASCII whitespace range fails at code point 13.

## Test plan
- [x] Full repo `moon test`: 7462/7462
- [x] `char` suite 41/41 on wasm-gc, native, and js
- [x] `moon info`: no `.mbti` change (implementation-only)
- [x] Codex CLI review (`xhigh`): first-round sign-off — set-equivalence, table-minimum claim, byte-identical references, and surrogate coverage all independently verified; see review comment

### DIFF
--- a/builtin/char.mbt
+++ b/builtin/char.mbt
@@ -136,7 +136,7 @@ pub fn Char::is_ascii_uppercase(self : Self) -> Bool {
 /// Checks if the value is an ASCII whitespace character:
 /// U+0020 SPACE, U+0009 HORIZONTAL TAB, U+000A LINE FEED, U+000B VERTICAL TAB, U+000C FORM FEED, or U+000D CARRIAGE RETURN.
 pub fn Char::is_ascii_whitespace(self : Self) -> Bool {
-  self is ('\u{20}' | '\u{09}' | '\u{0A}' | '\u{0B}' | '\u{0C}' | '\u{0D}')
+  self is ('\u{09}'..='\u{0D}' | ' ')
 }
 
 ///|
@@ -164,10 +164,11 @@ pub fn Char::is_digit(self : Self, radix : UInt) -> Bool {
 ///|
 /// Returns true if this char has the White_Space property.
 pub fn Char::is_whitespace(self : Self) -> Bool {
+  if self <= '\u{20}' {
+    return self is ('\u{09}'..='\u{0D}' | ' ')
+  }
   self
-  is ('\u0009'..='\u000D'
-  | '\u0020'
-  | '\u0085'
+  is ('\u0085'
   | '\u00A0'
   | '\u1680'
   | '\u2000'..='\u200A'
@@ -181,9 +182,14 @@ pub fn Char::is_whitespace(self : Self) -> Bool {
 ///|
 /// Returns true if this char has one of the general categories for numbers.
 pub fn Char::is_numeric(self : Self) -> Bool {
+  if self is ('0'..='9') {
+    return true
+  }
+  if self < '\u{B2}' {
+    return false
+  }
   self
-  is ('\u0030'..='\u0039'
-  | '\u00B2'
+  is ('\u00B2'
   | '\u00B3'
   | '\u00B9'
   | '\u00BC'
--- a/char/predicate_reference_test.mbt
+++ b/char/predicate_reference_test.mbt
@@ -1,0 +1,204 @@
+// Copyright 2026 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Exhaustive equivalence check for the restructured char predicates: the
+// reference functions below are the previous single flat patterns, kept
+// verbatim, and every Unicode scalar value is compared against them. Any
+// off-by-one in the restructured fast paths (a shifted range bound, a
+// dropped table entry) fails here with the exact code point.
+
+///|
+fn reference_is_ascii_whitespace(c : Char) -> Bool {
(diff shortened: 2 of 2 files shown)


bobzhang:
## Codex CLI review

Adversarial review by Codex CLI (`gpt-5.6-sol`, reasoning effort `xhigh`, read-only sandbox), briefed with the full three-backend benchmark matrix (including the rejected alternatives and the deliberate `is_whitespace` trade), and asked to attack: set-equivalence of each restructure with the original pattern, the `U+00B2` table-minimum claim, the soundness of removing the ASCII alternatives from the `is_whitespace` tail, `Char` comparison semantics, blind spots in the exhaustive test itself (reference fidelity, surrogate exclusion), CI cost of the ~1.1M-scalar sweep, and every number in the commit message.

### Round 1 — approved, no blocking findings

> No blocking findings.
>
> - All three rewrites are set-equivalent; `B2` is the numeric table minimum, and the whitespace guard fully decides every value through `U+0020`.
> - Reference patterns match `origin/main` byte-for-byte. The inclusive sweep covers 1,112,064 scalar values; `Int::to_char` excludes exactly the surrogate range.
> - `Char` ordering is Unicode code-point ordering, so both guards are semantically sound.
> - The roughly 6.67 million predicate calls are reasonable for CI, and the commit-message ratios and claims agree with the brief.
>

---
*Posted on behalf of Codex CLI by Claude Code. Validation outside Codex's read-only sandbox: full repo `moon test` 7462/7462; `char` 41/41 on wasm-gc/native/js; no `.mbti` change; equivalence test mutation-verified at both guard boundaries.*

bobzhang:
## Addendum: why not `self is ' ' || …` (space-first)?

A natural follow-up question on these predicates: space is the most common whitespace character, so should it get a dedicated first check? Measured on all three backends (Codex-CLI-verified analysis, 4 rounds; the guard variant also lets the ASCII branch of `is_whitespace` drop its `| ' '` alternative):

**On mixed, realistic text** (code-like, spaces + `\n` hits, 27% whitespace), space-first is **worse on every backend for both functions** — e.g. `is_ascii_whitespace` js 112 → 127 µs, `is_whitespace` wasm-gc 126 → 137 µs. Reordering the pattern alternatives space-first (without a guard) is also uniformly 2–17% worse.

**A controlled sweep** (64-char arrays, k literal spaces interleaved with letters, k ∈ {0,16,32,48,64}) shows the trade precisely:
- At k=0 (pure rejects), five of six cells regress 5–13%; native `is_ascii_whitespace` is flat.
- The guard starts winning between k=0 and k=16 on wasm-gc (up to 1.8x at k=64), between k=32 and k=48 on js for `is_ascii_whitespace`; js `is_whitespace` showed no benefit at any sampled fraction.

**The structural account** (verified against the code shapes): the guard is a bet that the evaluated char is a literal space — a space then costs one comparison; every `'\t'..'\r'` hit costs exactly one comparison *more* than the current form; and in `is_whitespace`, every reject above U+0020 (letters, digits, punctuation — the common case) pays one extra comparison, while control-char rejects stay count-neutral (the guard is offset by the dropped `| ' '` alternative).

**Conclusion**: no variant dominates. Space-first only collects on hit streams that are overwhelmingly literal spaces (e.g. skipping space-only indentation) — and a call site that knows it has that shape can simply write `c is ' ' || c.is_whitespace()` itself. The library default keeps the current form, which is worse nowhere.

*Measurements and analysis by Claude Code; verified by Codex CLI (`xhigh`, sign-off after a 4-round adversarial audit of the claims against the data and code).*

PR #4114, opened 2026-08-20
perf(builtin): aligned bitstring extraction fast paths via entry/slow split

bobzhang:
## Summary

Supersedes #3777 (thanks @mizchi for identifying the aligned-delegation opportunity and the benchmark workload). That PR's inline guards, re-benchmarked against current main, **regressed native u32 by 17–26%**: the guards pushed the large generic bodies past the native inliner's threshold ([full analysis on #3777](https://github.com/moonbitlang/core/pull/3777#issuecomment-5352307912)). This PR lands the shape that keeps inlining: each of the eight generic extractors becomes a **small entry** (aligned exact-width checks delegating to the existing `*_aligned` intrinsic-backed helpers) that tail-calls the previous byte-by-byte body, **outlined verbatim into a private `*_slow` function**.

## Benchmarks (`moon bench --release`, 8192-byte views, XOR-fold loops; main → this PR, BytesView cells)

| case | native | js | wasm | wasm-gc |
|---|---:|---:|---:|---:|
| aligned u32 | 5.01 → **1.88 µs** (2.7x) | 30.5 → **2.5–4.6 µs**¹ | 16.8 → **8.2 µs** (2.0x) | 14.5 → **5.1 µs** (2.8x) |
| aligned u64 | 3.06 → **0.94 µs** (3.2x) | 111.8 → **93.3 µs** | 13.6 → **4.1 µs** (3.3x) | 12.4 → **3.8 µs** (3.2x) |
| unaligned u32/u64 | par | par | par | par |
| aligned len=16 | 0.48 → 3.76 µs² | 21.2 → **8.0 µs** | 19.2 → **16.6 µs** | 15.2 → **11.3 µs** |
| aligned len=24 | par | bimodal¹ | 17.7 → 23.7 µs³ | par |

¹ V8 tier-up bimodality: the same build measures both fast and slow steady states on these cells (u32-le swings 2.6–20 µs within a run; the be cell is stable at 2.5 µs).
² Main's trivially-inlined two-byte path gets **loop-vectorized** by LLVM in this synthetic XOR-fold (0.117 ns/op is sub-cycle — impossible without SIMD), which no call-based path can match; single-call latency is normal scalar speed (~0.9 ns/op), and every other backend improves on this case. Verified against native assembly during review.
³ A real, disclosed concession isolated to aligned widths that have no fixed-width helper — rare in bitstring patterns.

For comparison, #3777's inline-guard shape measured **5.27 µs** on native aligned u32 — slower than doing nothing.

## Tests

- **`bitstring_extract_equiv_test.mbt`**: sweeps every `(offset ∈ 0..=16, len ∈ 9..=32 and 33..=64)` combination — aligned and unaligned, exact and partial widths — on both view types, both endiannesses, and both zero-origin and nonzero-origin subviews, against a byte-composition reference built from `unsafe_extract_byte`. Mutation-verified: swapping one aligned helper's endianness fails the sweep.
- **`bitstring_extract_bench_test.mbt`**: the aligned exact widths plus the slow-path shapes (unaligned, aligned-16, aligned-24), so future changes see both sides of the guard.

## Review

Codex CLI (`xhigh`), first-round sign-off: all eight outlined slow bodies verified **verbatim** against origin/main; all twelve guard/helper pairings correct in view, width, endianness, and `offset >> 3` conversion; LE/BE reference semantics verified including partial-byte placement; the vectorization explanation corroborated against native assembly; generated interface unchanged. Its two wording nits and the nonzero-origin-view coverage suggestion are incorporated.

## Validation

- Full repo `moon test`: 7464/7464 (builtin 2943/2943)
- `moon check --warn-list +unnecessary_annotation` clean, `moon fmt` applied, no `.mbti` change (the `*_slow` functions are private)

### DIFF
--- a/builtin/bitstring.mbt
+++ b/builtin/bitstring.mbt
@@ -108,8 +108,26 @@ pub fn ArrayView::unsafe_extract_uint_le(
   offset : Int,
   len : Int,
 ) -> UInt {
+  if (offset & 7) == 0 {
+    if len == 32 {
+      return bs.unsafe_extract_uint_le_aligned(offset >> 3)
+    }
+    if len == 16 {
+      return bs.unsafe_extract_uint16_le_aligned(offset >> 3)
+    }
+  }
+  bs.unsafe_extract_uint_le_slow(offset, len)
+}
+
+///|
+/// Outlined slow path of `unsafe_extract_uint_le`, kept out of the entry
+/// function so the aligned fast paths above stay small enough to inline.
+fn ArrayView::unsafe_extract_uint_le_slow(
+  bs : ArrayView[Byte],
+  offset : Int,
+  len : Int,
+) -> UInt {
   let bytes_needed = (len + 7) / 8
-  // TODO: add fast path for aligned case
   // non-aligned case: extract bytes using unsafe_extract_byte
   let b0 = bs.unsafe_extract_byte(offset, 8)
   match bytes_needed {
@@ -146,8 +164,26 @@ pub fn ArrayView::unsafe_extract_uint_be(
   offset : Int,
   len : Int,
 ) -> UInt {
+  if (offset & 7) == 0 {
+    if len == 32 {
+      return bs.unsafe_extract_uint_be_aligned(offset >> 3)
+    }
+    if len == 16 {
+      return bs.unsafe_extract_uint16_be_aligned(offset >> 3)
+    }
+  }
+  bs.unsafe_extract_uint_be_slow(offset, len)
+}
+
+///|
+/// Outlined slow path of `unsafe_extract_uint_be`, kept out of the entry
+/// function so the aligned fast paths above stay small enough to inline.
+fn ArrayView::unsafe_extract_uint_be_slow(
+  bs : ArrayView[Byte],
+  offset : Int,
+  len : Int,
+) -> UInt {
   let bytes_needed = (len + 7) / 8
-  // TODO: add fast path for aligned case
   // non-aligned case: extract bytes using unsafe_extract_byte
   let b0 = bs.unsafe_extract_byte(offset, 8)
   match bytes_needed {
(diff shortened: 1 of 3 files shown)


bobzhang:
## Codex CLI review (`xhigh`)

Reviewed as the implementation of the P1 verdict Codex itself issued on #3777 (the inline-guard shape un-inlines the generic functions on native), briefed to verify the outlined bodies verbatim, all guard/helper pairings, the equivalence sweep's reference semantics, the four-backend benchmark matrix, and both documented concessions.

### Verdict — approved, first round

> Verdict: approve. No blocking findings.
>
> - All eight slow bodies are verbatim from `origin/main`; only the eight resolved TODO lines were removed. All twelve guard/helper pairings have the correct view, width, endianness, and `offset >> 3` conversion.
> - LE and BE references implement the intended partial-byte placement correctly. The inclusive sweeps cover every stated combination, both views, and both endiannesses.
> - The native aligned-16 arithmetic is consistent: 4096 operations make 481 ns ≈ 0.117 ns/op and 3.76 µs ≈ 0.918 ns/op. The former requires vectorization or equivalent strength reduction; cached native assembly corroborates scalar loads on this branch. The wasm aligned-24 regression is real and material (~34%) but explicitly disclosed and isolated to an outlined non-fast width.
> - `pkg.generated.mbti` is unchanged. `git diff --check` is clean.
>

Its two non-blocking wording nits (bench-file header phrasing) and its coverage suggestion (the sweep only used zero-origin views) were incorporated before push — the sweep now also exercises nonzero-origin subviews (`bytes[3:]`, `arr[3:]`), covering the aligned helpers' view-start handling.

---
*Posted on behalf of Codex CLI by Claude Code. Validation outside Codex's read-only sandbox: full repo `moon test` 7464/7464 across the sweep; benchmarks on native/js/wasm/wasm-gc as tabled in the PR body; endianness-swap mutation fails the sweep.*

bobzhang:
cc @Yu-zh

PR #4117, opened 2026-08-20
perf(builtin): count rev_foldi's index up instead of deriving it per iteration

bobzhang:
## Summary

Completes #3784 (thanks @mizchi) on current main. Three of that PR's four `FixedArray` reverse-iteration optimizations have already landed; the remaining delta is `rev_foldi`, which still derives its ascending index as `len - i - 1` on every iteration of the descending range loop. That PR proposed a C-style three-variable loop; this PR takes the same optimization while keeping the range form the neighboring functions already use — the index simply becomes a second loop state:

```moonbit
pub fn[A, B] FixedArray::rev_foldi(
  self : FixedArray[A],
  init~ : B,
  f : (Int, B, A) -> B raise?,
) -> B raise? {
  for i in self.length()>..0; index = 0, acc = init {
    continue index + 1, f(index, acc, self.unsafe_get(i))
  } nobreak {
    acc
  }
}
```

## Benchmarks (fresh, against current main — #3784's table predates the three already-merged changes)

`rev_foldi`, n=100K sum fold, `moon bench --release`:

| backend | main | this PR | |
|---|---:|---:|---|
| native | 14.88 µs | **11.19 µs** | 1.33x |
| js | 72.4 µs | **59.9 µs** | 1.21x |
| wasm-gc | 75.5 µs | **65.6 µs** | 1.15x |

The readable range form measures **identical** to #3784's C-style loop (11.19 vs 11.20 µs native) — the C-style rewrite buys nothing over `for i in … >..0; index = 0, acc = init`.

**Why it's faster**: per-element times (~0.1 ns) show the fold is fully inlined and auto-vectorized in this benchmark; `len - i - 1` adds a per-lane subtract dependency that vectorizes worse than an independently counting index. With an opaque, non-inlinable closure the difference shrinks accordingly.

## Tests

Adopts #3784's regression test pinning the reversed index/element pairing (`[10,20,30]` traced to `30120210`, i.e. pairs `(0,30),(1,20),(2,10)`), which a miscounted index would break.

## Review

Codex CLI (`xhigh`), first-round sign-off: index/element pairing verified against the documented reverse-range and simultaneous-update semantics (`f` receives the pre-increment index), the trace recomputed by hand, raise-polymorphism and `nobreak` return intact, benchmark ratios recomputed, claims microbenchmark-scoped. No `.mbti` change.

## Validation

- builtin 2944/2944 (includes the pairing test); `moon check --warn-list +unnecessary_annotation` clean; `moon fmt` applied; `moon info` unchanged

### DIFF
--- a/builtin/fixedarray.mbt
+++ b/builtin/fixedarray.mbt
@@ -774,9 +774,8 @@ pub fn[A, B] FixedArray::rev_foldi(
   init~ : B,
   f : (Int, B, A) -> B raise?,
 ) -> B raise? {
-  let len = self.length()
-  for i in len>..0; acc = init {
-    continue f(len - i - 1, acc, self.unsafe_get(i))
+  for i in self.length()>..0; index = 0, acc = init {
+    continue index + 1, f(index, acc, self.unsafe_get(i))
   } nobreak {
     acc
   }
@@ -795,6 +794,16 @@ test "rev_foldi" {
   }
   let sum = ([1, 2, 3, 4, 5] : FixedArray[_]).rev_foldi(init=0, f)
   inspect(sum, content="25")
+}
+
+///|
+test "rev_foldi preserves reversed index element pairs" {
+  let trace = ([10, 20, 30] : FixedArray[_]).rev_foldi(init=0, (
+    index,
+    acc,
+    elem,
+  ) => acc * 1000 + index * 100 + elem)
+  inspect(trace, content="30120210")
 }
 
 ///|


bobzhang:
## Codex CLIreview (`xhigh`) — approved, first round

> No blocking findings; approved.
>
> - `rev_foldi` preserves `(index, element)` as `(0,len−1)…(len−1,0)`. MoonBit's reverse-range and simultaneous-update semantics confirm `f` receives the pre-increment index.
> - The regression trace correctly evaluates to `30120210`.
> - Raise-polymorphism and `nobreak` accumulator return remain intact.
> - Benchmark ratios recompute to 1.3298×, 1.2087×, and 1.1509×; claims are appropriately microbenchmark-scoped.
> - No `.mbti` change.
>

*Posted on behalf of Codex CLI by Claude Code. Validation outside its read-only sandbox: builtin 2944/2944; benchmarks on native/js/wasm-gc as tabled; the readable range form measured identical to #3784's C-style loop.*

PR #4125, opened 2026-08-22
perf(hashset): probe with unchecked array access

bobzhang:
First of two PRs splitting #3712, which bundled a struct-of-arrays layout with unchecked probing. Split apart, they turn out to pay on **different backends** — which the combined PR could not show, because each masked the other's cost. This is the probing half; the layout half is stacked on top of it.

## The change

Every probe index in `HashSet` comes from `hash & capacity_mask` and is re-masked as `(idx + 1) & capacity_mask` on each step, so it is always in bounds for `entries`, whose length is `capacity`. The bounds check on those accesses is provably redundant.

Eight masked-index sites become `unsafe_get`/`unsafe_set`, each with the invariant written out: the probe loops in `add_with_hash`, `push_away`, `contains`, `remove`, `shift_back` and `rehash_place_entry`, plus `set_entry`'s store. No layout change, no API change, `pkg.generated.mbti` untouched.

Also adds `hashset/hashset_bench_test.mbt`, which the package did not have (from #3712, credited).

## Measurements

n=50000, this branch vs `origin/main`, interleaved on one machine in a single session:

| js | `add` | 3.08 ms | 2.51 ms | **18% faster** |
| js | `contains` | 1.21 ms | 1.01 ms | **17% faster** |
| native | `add` | 2.35 ms | 2.29 ms | ~2%, ranges overlap |
| native | `contains` | 875 µs | 857 µs | ~2%, ranges overlap |
| wasm-gc | `add` | 2.30 ms | 2.30 ms | unchanged |
| wasm-gc | `contains` | 1.09 ms | 1.11 ms | unchanged |

The win is js-only, and it is a real one. js emits an explicit length comparison per access; removing it pays. wasm-gc's array access traps inherently, so there is nothing to remove. On native the boxed entry's pointer chase and `Option` unwrap dominate and the check disappears into the noise.

That asymmetry is the argument for the split: this half is what makes **js** faster, and the layout half (#stacked) is what makes **native** faster — by 35% — while costing js 18%. Bundled, they averaged out to "js unchanged" and the trade was invisible.

## Safety

An out-of-bounds unchecked access is undefined behaviour, not a test failure, so a green suite is weak evidence here. The argument is the representation invariant:

- construction establishes `capacity >= 1`, a power of two, with `entries.length() == capacity` and `capacity_mask == capacity - 1` — note `next_power_of_two(0) == 1`, and a negative capacity panics before anything is built;
- `grow` is the only resizing path, and it installs the new array, capacity and mask together before any rehash probe;
- `add_with_hash` grows before computing its first index;
- `shift_back` gets an in-bounds `cur` by either route its callers take — masked from `remove`, or after a successful checked read from `retain`;
- therefore every `set_entry` caller supplies a proven in-bounds index.

Iteration sites (`each`, `eachi`, `retain`, `to_array`) stay checked: their indices are equally provable, but no benchmark shows a gain, so they keep the form that is cheaper to audit.

## Review

Reviewed by Codex CLI at `ultra` reasoning effort — approved, verdict in a comment below. Its two substantive notes are addressed: the `shift_back`/`set_entry` SAFETY comments now document both routes an index can arrive by, rather than claiming everything is mask-derived.

Two of its suggestions I have **not** acted on, for a follow-up: the benchmark covers insertion and successful lookup but not misses, removal, or wraparound displacement; and js uses a per-process random hash seed, so a deterministic key set would make the comparison sturdier.

### DIFF
--- a/hashset/hashset.mbt
+++ b/hashset/hashset.mbt
@@ -94,7 +94,9 @@ fn[K : Eq] HashSet::add_with_hash(
     self.grow()
   }
   let (idx, psl) = for psl = 0, idx = hash & self.capacity_mask {
-    match self.entries[idx] {
+    // SAFETY: `idx` starts masked by `capacity_mask` and every step re-masks
+    // it, so it is in bounds for `entries`, whose length is `capacity`.
+    match self.entries.unsafe_get(idx) {
       None => break (idx, psl)
       Some(curr_entry) => {
         if curr_entry.hash == hash && curr_entry.key == key {
@@ -121,7 +123,8 @@ fn[K] HashSet::push_away(
   entry : Entry[K],
 ) -> Unit {
   for psl = entry.psl + 1, idx = (idx + 1) & self.capacity_mask, entry = entry {
-    match self.entries[idx] {
+    // SAFETY: same masked-index invariant as `add_with_hash`.
+    match self.entries.unsafe_get(idx) {
       None => {
         entry.psl = psl
         self.set_entry(entry, idx)
@@ -149,7 +152,9 @@ fn[K] HashSet::set_entry(
   entry : Entry[K],
   new_idx : Int,
 ) -> Unit {
-  self.entries[new_idx] = Some(entry)
+  // SAFETY: every caller supplies an index already proven in bounds -- from
+  // a masked probe, or from `shift_back`'s `cur`.
+  self.entries.unsafe_set(new_idx, Some(entry))
 }
 
 ///|
@@ -158,7 +163,8 @@ pub fn[K : Hash + Eq] HashSet::contains(self : HashSet[K], key : K) -> Bool {
   // inline lookup to avoid unnecessary allocations
   let hash = Hash::hash(key)
   for i = 0, idx = hash & self.capacity_mask {
-    guard self.entries[idx] is Some(entry) else { break false }
+    // SAFETY: same masked-index invariant as `add_with_hash`.
+    guard self.entries.unsafe_get(idx) is Some(entry) else { break false }
     if entry.hash == hash && entry.key == key {
       break true
     }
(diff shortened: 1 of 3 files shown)


bobzhang:

Approved.

> **Approved for `37f2daa8..79b52f63`; no blocking correctness or API findings.**
>
> Non-blocking review notes:
>
> - The SAFETY comments at `hashset/hashset.mbt:155` and `:219` are too narrow. `retain` passes a checked linear index to `shift_back`, not a mask-derived index. The code is safe because `entries[i]` just succeeded; subsequent indices are masked. The comments should document both proven routes.
> - The brief says `iter` remains checked, but it already used `unsafe_get` on `origin/main`. That existing access is safe because `i < captured_len` and storage never shrinks.
> - The benchmark covers insertion and successful lookup, but not misses, removal, or wraparound displacement. JavaScript also uses a random per-process hash seed, so repeated-process intervals or a deterministic test key would strengthen the comparison.
>
> The bounds proof holds:
>
> - Construction establishes `capacity >= 1`, power-of-two capacity, `entries.length() == capacity`, and `capacity_mask == capacity - 1`. Notably, `next_power_of_two(0) == 1`.
> - `copy` preserves those invariants, the representation is externally opaque, and `grow` is the only resizing path.
> - `grow` finishes installing the new array, capacity, and mask before rehashing; there is no probe or callback while those fields are transiently updated.
> - For positive power-of-two capacity, `x & capacity_mask` is always in `[0, capacity)`. Every probe continuation remasks.
> - `add_with_hash` grows before computing its first index.
> - `shift_back` receives either a masked index from `remove` or a checked index from `retain`; later `cur` values come from masked `next`.
> - Every `set_entry` caller therefore supplies a proven in-bounds index.
>
> The zero-capacity branch in `grow` is currently unreachable. A useful missed mutation is changing construction to preserve an actual zero capacity: the existing test calls `add` first, which would grow and still pass, while `contains` or `remove` on the fresh zero-length table would be undefined. A deterministic collision test wrapping from `capacity - 1` to zero would also protect the remasking assumptions.
>
> Tests cannot establish absence of unchecked out-of-bounds access. The actual evidence is the closed representation-invariant and complete caller proof above; debug index guards, wraparound stress tests, and sanitizer runs would be supporting regression evidence.
>
> The JavaScript explanation is credible: checked access emits an explicit bounds comparison while unchecked access is direct. Wasm-GC's inherent array trap explains its neutrality; the precise native explanation remains an inference. A repeatable 17–18% JavaScript improvement with neutral native and Wasm-GC results is enough to justify this independently of the layout change.
>

The first note was a genuine inaccuracy and is fixed: `shift_back` and `set_entry` now document both routes an in-bounds index arrives by, instead of claiming every one is mask-derived. The second was an error in my brief, not the code — `iter` was already unchecked on `main`.

Its observation about the unreachable zero-capacity branch is worth recording: if construction ever stopped rounding capacity up, `add` would still grow and pass, while `contains`/`remove` on a zero-length table would be undefined. That branch in `grow` is dead today, but it is the shape of a future footgun.

PR #4126, opened 2026-08-22
perf(hashmap): probe with unchecked array access

bobzhang:
The `hashmap` counterpart of #4125. Same idea, same invariant, no layout change — `pkg.generated.mbti` is untouched.

## The change

Every probe index in `HashMap` comes from `hash & capacity_mask` and is re-masked as `(idx + 1) & capacity_mask` on each step, so it is always in bounds for `entries`, whose length is `capacity`. The bounds check on those accesses is provably redundant.

26 masked-index sites across ten probe loops become `unsafe_get`/`unsafe_set`: `set`, `get`, `get_from_bytes`, `get_from_string`, `at`, `get_or_init`, `update`, `update_or_default`, `get_or_default`, `contains`, `contains_kv`, `remove`, `push_away`, `shift_back` and `rehash_place_entry`. With ten loops, restating the justification at each one would be noise, so the invariant is written out once at the top of the file with a marker at each loop head. The two linear iteration sites keep the checked form.

Also adds `hashmap/hashmap_bench_test.mbt`, which the package did not have. Following review feedback on #4125, it covers misses and removal as well as insertion and hits.

## Measurements

n=50000, this branch vs `origin/main`, interleaved on one machine in a single session:

| js | `set` | 2.49 ms | 2.34 ms | 6% faster |
| js | `get` hit | 508 µs | 465 µs | 8% faster |
| js | `get` miss | 690 µs | 610 µs | **12% faster** |
| js | `remove` | 3.43 ms | 3.12 ms | 9% faster |
| native | `set` | 2.25 ms | 2.16 ms | 4% faster |
| native | `get` hit | 348 µs | 334 µs | 4% faster |
| native | `get` miss | 522 µs | 496 µs | 5% faster |
| native | `remove` | 2.65 ms | 2.55 ms | 4% faster |
| wasm-gc | all four | | | unchanged |

Two differences from the `hashset` result worth noting. **Native gains here**, where in `hashset` it was lost in the noise — small, but consistent across all four operations with non-overlapping ranges, which fits `HashMap` having longer and more numerous probe loops. And **misses gain most**, which also fits: a miss walks the probe sequence to its end, so it pays the most bounds checks per call.

wasm-gc is unchanged, as in `hashset` — its array access traps inherently, so there is no separate check to remove.

## Safety

An out-of-bounds unchecked access is undefined behaviour, not a test failure, so the argument is the representation invariant, not the green suite:

- `capacity` is always a power of two and at least 1, `capacity_mask == capacity - 1`, `entries.length() == capacity`;
- every probe index starts at `hash & capacity_mask` and re-masks each step;
- `grow` installs the new array, capacity and mask together before any rehash probe runs;
- `set_with_hash` re-masks from scratch after growing;
- `shift_back` has only two callers, both passing a masked index — unlike `hashset`, there is no `retain` route reaching it with a linear index.

## Review

Codex CLI review at `ultra` reasoning effort is running; its verdict will be posted as a comment.

### DIFF
--- a/hashmap/hashmap.mbt
+++ b/hashmap/hashmap.mbt
@@ -12,6 +12,19 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// SAFETY NOTE for the unchecked probe accesses in this file.
+//
+// `capacity` is always a power of two and at least 1, `capacity_mask` is
+// `capacity - 1`, and `entries.length()` is `capacity`. Every probe index is
+// therefore in bounds: it starts as `hash & capacity_mask` and each step
+// re-masks with `(idx + 1) & capacity_mask`. `grow` installs the new array,
+// capacity and mask together before any rehash probe runs, and `set_with_hash`
+// re-masks from scratch after growing. `shift_back` is the one entry point
+// that can be reached with an index which is not mask-derived -- `retain` in
+// utils.mbt passes a linear one -- so it carries its own proof at its
+// definition. On that basis the probe loops below use `unsafe_get` /
+// `unsafe_set`; the linear iteration sites keep the checked form.
+
 ///|
 let default_init_capacity = 8
 
@@ -119,8 +132,9 @@ fn[K : Eq, V] HashMap::set_with_hash(
   hash : Int,
 ) -> Unit {
   // Only grow when actually inserting a new entry, not when updating existing
+  // SAFETY: masked probe index; see the note at the top of this file.
   for psl = 0, idx = hash & self.capacity_mask {
-    match self.entries[idx] {
+    match self.entries.unsafe_get(idx) {
       None => {
         // Need to insert new entry - check if grow is needed first
         if self.size >= self.capacity / 2 {
@@ -129,7 +143,7 @@ fn[K : Eq, V] HashMap::set_with_hash(
           continue 0, hash & self.capacity_mask
         }
         let entry = { psl, key, value, hash }
-        self.entries[idx] = Some(entry)
+        self.entries.unsafe_set(idx, Some(entry))
         self.size += 1
         return
       }
@@ -148,7 +162,7 @@ fn[K : Eq, V] HashMap::set_with_hash(
           }
           self.push_away(idx, curr_entry)
           let entry = { psl, key, value, hash }
-          self.entries[idx] = Some(entry)
+          self.entries.unsafe_set(idx, Some(entry))
           self.size += 1
           return
         }
@@ -165,17 +179,18 @@ fn[K, V] HashMap::push_away(
   idx : Int,
   entry : Entry[K, V],
 ) -> Unit {
+  // SAFETY: masked probe index; see the note at the top of this file.
   for psl = entry.psl + 1, idx = (idx + 1) & self.capacity_mask, entry = entry {
-    match self.entries[idx] {
+    match self.entries.unsafe_get(idx) {
       None => {
         entry.psl = psl
-        self.entries[idx] = Some(entry)
+        self.entries.unsafe_set(idx, Some(entry))
         break
       }
       Some(curr_entry) =>
         if psl > curr_entry.psl {
           entry.psl = psl
-          self.entries[idx] = Some(entry)
+          self.entries.unsafe_set(idx, Some(entry))
           continue curr_entry.psl + 1,
             (idx + 1) & self.capacity_mask,
             curr_entry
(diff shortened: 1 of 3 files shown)


bobzhang:
## Codex CLI review (`ultra` reasoning effort) — round 1: changes requested

Two P2s, both correct, both now fixed in `24fbe4e0`.

> **Request changes. I found no current out-of-bounds access among the 26 conversions, but two issues should be corrected before approval.**
>
> **[P2] The `shift_back` safety premise is false.** `hashmap/hashmap.mbt:22` says `shift_back` is only called with a masked index. There are actually three callers: `update`, with masked `idx`; `remove_with_hash`, with masked `idx`; and `retain`, with linear `i` at `hashmap/utils.mbt:476`. The `retain` route is currently safe: the checked `self.entries[i]` access succeeded immediately beforehand, capacity never shrinks, and subsequent `cur` values are masked `next` values. Nevertheless, the committed justification and review brief are factually wrong.
>
> There are also 15 converted loops, not ten, and only ten local safety markers. Markers are missing for `push_away`, `get_or_init`, `update`, `update_or_default`, and `shift_back`.
>
> **[P2] The removal benchmark measures insertion plus removal.** The map is constructed and filled inside the timed closure before removing entries. The reported result therefore cannot be attributed to `remove`. On native, the combined improvement is about `0.10 ms`, while `set` alone improves by about `0.09 ms`, so this benchmark provides little evidence of a removal-specific gain.

It also corrected two claims of mine, and the correction to the second is worth keeping visible:

> `get_from_bytes` and `get_from_string` actually compute their hashes internally. […] I do not agree that HashMap's probes are longer than HashSet's: at 50,000 entries, HashMap is about 38% loaded while HashSet is about 76% loaded under their respective growth policies. The native difference needs probe-count, profiling, or generated-code evidence; the number of API loops alone does not explain an individual operation.

### What changed

- **`shift_back`**: the central note no longer claims every index reaching it is mask-derived. It is now identified as the one entry point that can be reached otherwise, and carries its own proof at its definition naming all three routes. I had grepped for callers only within `hashmap.mbt`, which is why I missed `retain` in `utils.mbt` — the same route Codex had already caught in `hashset`.
- **Markers** now cover all 15 loops, not 10.
- **Benchmark** renamed to `bench HashMap::set+remove n=50000`, with a comment recording why removal cannot be isolated: `@bench.T` has no per-iteration setup hook, so the map must be rebuilt inside the timed closure. The PR body and commit message report it under that name.
- **The native explanation is withdrawn.** The commit now says the cause is not established, and names the disproved guess, rather than asserting it. An 18% claim I cannot support is worse than an honest gap.

Round 2 review is running; its verdict will follow.

bobzhang:
## Codex CLI review — round 2: approved

> **Approved. No P0/P1/P2 findings.**
>
> - The `shift_back` proof is correct and complete. Its only callers are `update`, `remove_with_hash`, and `retain`. The first two use masked indices; `retain` performs a successful checked read first, and capacity cannot shrink — even through reentrant callbacks. Each subsequent `cur` is a previously masked `next`.
> - All 15 converted probe loops now have adjacent safety markers, covering exactly 26 converted accesses: 15 gets and 11 sets.
> - The round-two HashMap changes are comments only. The benchmark body is unchanged; only the intended `set+remove` label and explanation changed.
> - `git diff --check` is clean and `hashmap/pkg.generated.mbti` is identical to `origin/main`.
> - Performance remains "not established." A plausible hypothesis is that HashSet's higher collision workload dilutes the saved native check: the benchmark keys average about 1.307 probes per HashMap hit versus 2.647 for HashSet. A cheap follow-up is checked/unchecked `HashSet::contains` at capacity 131072, matching HashMap's load, plus a miss benchmark. I would not change the commit's wording without that measurement.
>

Its hypothesis for the native difference is a good one and worth recording even though I am not acting on it here: the benchmark keys average **1.307 probes per HashMap hit against 2.647 for HashSet**. Under that reading the saved bounds check is a fixed cost per probe, so it is a larger share of a short probe than a long one — which is the opposite of the explanation I originally offered. The commit keeps saying "not established" until someone measures `HashSet::contains` at capacity 131072 to match HashMap's load factor.

bobzhang:
## Copilot review addressed

Both comments were the same two findings Codex raised at `ultra`, and both are fixed.

**1. The `shift_back` call graph was wrong.** Correct — `HashMap::retain` in `utils.mbt` is a third caller passing a linear slot index, which my grep missed because it searched only `hashmap.mbt`. Fixed in `76cfd18b`: the central note no longer claims every index reaching `shift_back` is mask-derived, and `shift_back` carries its own proof at its definition naming all three routes, including the condition that makes the `retain` path safe — its own checked read succeeded immediately before, and capacity never shrinks. Markers now cover all 15 converted loops rather than 10.

**2. The removal benchmark measured insertion plus removal.** Also correct. I originally took the "rename/report as a combined workload" option, but the suggestion to prebuild and copy is better and it works — `3aa48c76` implements it.

Copying a prebuilt map is a single pass over the table rather than 50000 hash-and-probe insertions: **0.86 ms against the 2.4 ms a rebuild costs on js**. A companion benchmark times the copy alone, so removal reads off by difference:

| | main | this branch |
| --- | --- | --- |
| `copy+remove` | 1.73 ms | 1.58 ms |
| `copy` alone | 0.86 ms | 0.89 ms |
| **removal, by difference** | **0.875 ms** | **0.690 ms** |

That is **21% faster on removal** — an attributable number, where the combined benchmark could only say a mixed workload improved. The `set+remove` benchmark is kept alongside, named for what it measures.

Thanks — the copy suggestion produced a genuinely better measurement than the rename I had settled for.

Note the branch was rebased onto newer `main` in the meantime; `3aa48c76` is stacked on that rather than replacing it. `moon test` is 7541/7541.

PR #4129, opened 2026-08-22
perf(set): probe with unchecked array access

bobzhang:
Fourth and last in the wave applying unchecked probing to core's hash containers, after #4125 (`hashset`), #4126 (`hashmap`) and #4128 (`Map`).

## Measurements

New file `set/linked_hash_set_bench_test.mbt`. n=50000, this branch vs `origin/main`, interleaved on one machine in a single session:

| js | `add` | 3.41 ms | 2.87 ms | **16% faster** |
| js | `contains` hit | 1.20 ms | 1.02 ms | **15% faster** |
| js | `contains` miss | 1.40 ms | 1.16 ms | **17% faster** |
| js | `add+remove` | 5.05 ms | 4.14 ms | **18% faster** |
| native | all | | | within noise |
| wasm-gc | all | | | within noise |

Same shape as the other three: the saving is the explicit length comparison js emits per access, so js gains and the backends whose array access traps inherently do not. The percentages are empirical — js randomises its hash seed, so they move between runs.

## The change

Eleven masked-probe sites become `unsafe_get`/`unsafe_set`. The invariant is stated once at the top of the file with a marker at each loop head.

`Set` is a linked hash set, so as with `Map` its `entries` accesses take indices from two places: masked probe indices, and slot indices stored in the list itself (`prev`, `tail`, and the ones the copy path walks). **Only the first kind is converted.** Every access through a stored index keeps the checked form, since their being in bounds rests on the list being maintained correctly rather than on arithmetic alone.

`shift_back` is unchecked, but on a *local caller-based* proof rather than on that invariant: both of its callers pass masked probe indices, and `Set` has no `retain`, so unlike `Map` there is no route reaching it with a stored index at all. That proof is what makes the `set_entry` call inside it sound — a subtlety the review of #4128 caught in the `Map` version, where I had claimed `shift_back` stayed checked while it was in fact calling an already-converted `set_entry`.

## Review

Codex CLI review at `ultra` reasoning effort to follow. Its findings on #4128 were applied here pre-emptively, since the two containers share a structure.

### DIFF
--- a/set/linked_hash_set.mbt
+++ b/set/linked_hash_set.mbt
@@ -50,6 +50,25 @@ struct Set[K] {
 
 // Implementations
 
+// SAFETY NOTE for the unchecked probe accesses in this file.
+//
+// `capacity` is always a power of two and at least 1, `capacity_mask` is
+// `capacity - 1`, and `entries.length()` is `capacity`. A probe index is
+// therefore always in bounds: it starts as `hash & capacity_mask` and each
+// step re-masks with `(idx + 1) & capacity_mask`. `grow` installs the new
+// array, capacity and mask together before any rehash probe runs.
+//
+// Only those probe indices use `unsafe_get` / `unsafe_set`. `Set` also
+// stores slot indices in the list itself -- `prev` on each entry and `tail`
+// on the set -- and `copy` walks the whole list through them, from
+// `entries[self.tail]` back along each `entries[prev]`. Every access through
+// a stored index keeps the checked form. They are in bounds too, being
+// former probe indices in a table that never shrinks, but that argument
+// depends on the list being maintained correctly rather than on arithmetic
+// alone, so it is not one to build unchecked access on. `shift_back` is
+// unchecked, but on a local caller-based proof rather than on that
+// invariant: both of its callers pass masked probe indices.
+
 ///|
 let default_init_capacity = 8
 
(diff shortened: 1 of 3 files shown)


bobzhang:
## Copilot review addressed

Correct on both counts, fixed in `acfc1aed`.

The safety note described "the index `retain` destructures out of an entry" — but **`Set` has no `retain`**. I wrote the `Map` note first (#4128) and carried its wording across without re-checking it against this file. The actual stored-index traversal here is `copy`, which walks the whole list from `entries[self.tail]` back along each `entries[prev]`. The note now says that.

The second occurrence, in `shift_back`'s own proof, had a related error: it named the callers as `remove` and `remove_with_hash`, but they are `remove` and `remove_and_check`. Also corrected.

Neither affected the generated code — both are comments — but a safety note describing a call graph the file does not have is worse than no note, because the next person extending the unchecked accesses would reason from it. Preventing exactly that is what these notes are for.

## Why this only pays on js

Worth recording here, since it is the same for all four containers in this wave. The three backends compile a checked array access to three quite different things.

**js** — the check is a *function call* on every access:

```js
$bound_check(_tmp$3, idx);      // function $bound_check(arr, index) {
const _bind$2 = _tmp$3[idx];    //   if (index < 0 || index >= arr.length) throw ... }
```

against a bare `self.entries[idx]` when unchecked. There are 25 such call sites in the generated `hashset` bench module alone. That is what the 15–20% buys back.

**native** — the check is inlined straight-line code:

```c
if (idx < 0 || idx >= Moonbit_array_length(entries)) { moonbit_panic(); }
entries[idx]
```

A compare against a header word already in cache, plus a branch that is never taken and so predicts perfectly. The out-of-order engine hides it behind the dependent load, which is why removing it lands within noise on most of these containers.

**wasm-gc** — there is nothing to remove. The checked version compiles to a bare `array.get`, with no `array.len` comparison and no branch, because `array.get` traps on out-of-bounds as part of its own semantics. Checked and unchecked emit the *same instruction*. That is why wasm-gc comes out exactly flat every time, rather than merely close.

So this wave is best understood as a js optimization that happens to be free elsewhere, rather than a general one that only shows up on js.

PR #4130, opened 2026-08-22
fix(hashset): make `copy` independent of the original

bobzhang:
`HashSet::copy` allocates a fresh `entries` array and then blits the old one into it. `FixedArray[Entry[K]?]` holds *references*, so both sets come away sharing the same `Entry` objects — and `Entry::psl` is mutable, decremented by `shift_back` whenever a removal shifts a run back.

So **a removal on either set silently corrupts the other**:

```moonbit
let original = @hashset.HashSet([])
for i in 0..<32 { original.add(i * 8) }   // colliding keys
let duplicate = original.copy()
for i in 0..<32 { duplicate.remove(i * 8) }

original.length()        // 32  — `size` is a separate field per set
// but 6 of the 32 keys are no longer findable by `contains`
```

Insertions can disturb the original the same way, once one displaces a shared entry.

`HashSet` was the only container doing this. `HashMap::copy`, `Map::copy` and `Set::copy` all rebuild their entries already; this makes `HashSet` match them.

## How it surfaced

While benchmarking removal for the struct-of-arrays work (#4127), I added a copy-then-remove benchmark and the numbers made no sense — `main`'s copy was implausibly cheap and the removal figures were unstable. Codex CLI, reviewing that PR at `ultra` effort, worked out why: the benchmark's timed closure was mutating the shared template, so every iteration after the first ran against progressively corrupted state. That invalidated my benchmark, and the reason it was invalid turned out to be a genuine bug underneath.

The struct-of-arrays branch fixes this incidentally, because it rebuilds slots rather than blitting. This commit fixes it on `main` on its own, so the fix is not gated on that larger and still-contested change.

## Test

`hashset/copy_test.mbt` covers both directions — mutating the copy must not disturb the original, and vice versa — using colliding keys so removal actually reaches `shift_back`. Without the fix the first test fails with 26 of 32 keys still reachable.

`moon test` passes 7543/7543 (7484/7484 on js). `pkg.generated.mbti` unchanged.

### DIFF
--- a/hashset/copy_test.mbt
+++ b/hashset/copy_test.mbt
@@ -1,0 +1,96 @@
+// Copyright 2026 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+///|
+/// A key whose hash maps every value into one of only 8 buckets, so
+/// insertions collide heavily and deterministically -- independent of the
+/// runtime hash seed, which is randomized on some targets. Without this,
+/// whether the test reaches `shift_back` at all would depend on how `Int`
+/// hashing happens to mix, and the regression could go unnoticed.
+priv struct Collide(Int) derive(Eq)
+
+///|
+impl Hash for Collide with fn hash(self) {
+  let Collide(x) = self
+  x % 8
+}
+
+///|
+impl Hash for Collide with fn hash_combine(self, hasher) {
+  let Collide(x) = self
+  hasher.combine_int(x % 8)
+}
+
+///|
+/// `copy` used to blit the entry references, leaving both sets sharing
+/// `Entry` objects. Because `Entry::psl` is mutable and `shift_back`
+/// decrements it, a removal on either set silently corrupted the other's
+/// probe sequences.
+test "HashSet::copy is independent of the original" {
+  let original = @hashset.HashSet([])
+  for i in 0..<32 {
+    original.add(Collide(i))
+  }
+  let duplicate = original.copy()
+  for i in 0..<32 {
+    duplicate.remove(Collide(i))
+  }
+  inspect(duplicate.length(), content="0")
+  inspect(original.length(), content="32")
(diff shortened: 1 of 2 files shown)


bobzhang:
## Copilot review addressed

> This test assumes `i * 8` will reliably produce colliding hashes to exercise `shift_back`, but `Int` hashing is mixed, so this may not deterministically create a collision cluster and could miss the regression. Use a local key type with a controlled `Hash` implementation to guarantee collisions.

Correct, and the fix makes the test strictly better. The tests now use a `Collide` key type that maps every value into one of eight buckets — the same pattern `hashmap/hashmap_coverage_test.mbt` already uses, and for the same reason: the runtime hash seed is randomised on some targets, so a test that depends on `Int` mixing is not a reliable witness.

It also made the failure far sharper. Against the unfixed code:

| key type | original keys still reachable after emptying the copy |
| --- | --- |
| `Int` (`i * 8`) | 26 of 32 |
| `Collide` | **2 of 32** |

## Self-review

I tried to break this rather than confirm it. What I checked, and what it turned up:

**Only one of the three tests is a regression witness.** I verified each against the unfixed code: the removal test fails (2 of 32), but the insertion test and the empty-set test pass. Insertion-direction corruption is real — `push_away` mutates shared entries' `psl` — but `contains` on the original cannot observe it here, because the mutated PSL ends up no smaller than the true one and the probe still terminates correctly. I narrowed that test to stay under the growth threshold (growth reassigns every PSL and masks the effect) and it still does not catch the bug. It is kept as a property guard, not claimed as a regression test.

**The fix is complete for the type.** `Entry` has exactly `psl`, `hash`, `key`, and all three are rebuilt. Slots are copied position for position, so iteration order is preserved — worth stating, since `HashSet` iteration order is slot order.

**Nothing depended on the sharing.** No caller of `.copy()` inside `hashset` at all, so no internal behaviour changes.

**No sibling has the same defect.** I read all three: `HashMap::copy` rebuilds via `Some({ psl, key, value, hash })`, and `Map::copy` / `Set::copy` build a `new_entry` per slot while walking their lists. `HashSet` was the only container blitting, and a search for `blit_to` across the four containers finds nothing else.

**It costs something, and the PR should say so.** `copy` is now ~1.9x slower on native (241 µs → 455 µs) and ~1.5x on js (281 µs → 430 µs) at 50000 elements, because it now actually copies. That is the price of the operation being correct, and it brings `HashSet` in line with what the other three already pay. I do not think a faster-but-shared variant is worth offering: sharing mutable probe metadata has no safe use.

**Where I would still expect trouble:** nothing in this diff, but the same class of defect — a container sharing mutable internals through a shallow copy — is worth checking for elsewhere in core. Outside this PR's scope, and I have not swept for it.

`moon test`: 7544/7544 on native and wasm-gc, 7485/7485 on js. `pkg.generated.mbti` unchanged.

PR #4131, opened 2026-08-22
perf(hashset): store the table as two arrays, deriving each PSL

bobzhang:
Supersedes #4127, which I have marked draft. Same goal — stop allocating a boxed `Entry` per key — but without the js regression that made #4127 unlandable.

The design is Codex CLI's, proposed while explaining that regression. I implemented and measured it.

## What changed

#4127 stored three arrays: `psls`, `hashes`, `keys`. A probe step read all three, at the same index in three arrays 512 KB apart, so it touched **three cache lines where the boxed layout touched one** — the entry object holds psl, hash and key together. That cost js insertion 15% growing and 44% pre-sized.

A slot's probe-sequence length is not independent state. It is the distance between where the key sits and where its hash asked it to sit:

```
psl = (idx - (stored_hash & mask)) & mask
```

So `psls` goes away. Two arrays remain, and a probe step trades a random load in a 512 KB array for two arithmetic operations.

`hashes[i] == 0` marks an empty slot. Any `Int` can be a real hash, so `store_hash` remaps `0` onto `1`; the two then share a bucket, which costs an occasional extra key comparison and nothing else, since lookups compare keys as well as hashes.

`shift_back` no longer writes a PSL at all — moving an occupant one slot back *is* the decrement.

## Verifying the premise before building

`psl` was a `mut` field, which suggests independent state. I checked rather than assumed: I instrumented `set_slot` on the three-array branch to assert `psl == (idx - (hash & mask)) & mask` on **every** write, then ran the full suite on three backends plus a four-bucket collision stress with removals, shift-back cascades, growth and copy. It never fired. Every PSL write happened as a key landed in a slot, always equal to its displacement.

## Measurements

n=50000, against `main`, interleaved on one machine in a single session:

| native | `add` | 2.23 ms | 1.42 ms | **36% faster** |
| native | `contains` | 842 µs | 744 µs | 12% faster |
| native | `copy` | 484 µs | 65 µs | **86% faster** |
| native | `add` string | 3.51 ms | 3.22 ms | 8% faster |
| native | `contains` string | 2.16 ms | 1.84 ms | 15% faster |
| wasm-gc | `add` | 2.20 ms | 1.78 ms | **19% faster** |
| wasm-gc | `contains` | 1.06 ms | 987 µs | 7% faster |
| wasm-gc | `copy` | 373 µs | 168 µs | **55% faster** |
| wasm-gc | `add` string | 3.36 ms | 3.00 ms | 11% faster |
| wasm-gc | `contains` string | 2.61 ms | 2.41 ms | 8% faster |
| js | `add` | 2.37 ms | 2.33 ms | **unchanged** |
| js | `contains` | 962 µs | 904 µs | 6% faster |
| js | `copy` | 432 µs | 314 µs | 27% faster |
| js | `add` string | 3.41 ms | 3.58 ms | **5% slower** |
| js | `contains` string | 2.40 ms | 2.21 ms | 8% faster |

Faster on fourteen of fifteen. The two js insertion figures are the mean of two alternating samples per side, since they are what the design turns on: integer `add` came out 2.32/2.33 against 2.38/2.36, string `add` 3.57/3.59 against 3.42/3.39.

String keys are benchmarked because the boxed layout stores a pointer inside its entry where this one stores it directly in the key array — a difference integer benchmarks cannot show. That case is the single remaining regression, at 5%.

## Tests

The derivation is now load-bearing rather than merely true, so `derived_psl_test.mbt` covers the paths that used to write a PSL: a four-bucket collision stress with removal and reinsertion across six growth rounds, keys whose hash is the remapped sentinel, and a 20000-operation differential run against a bitmap model.

## Review

Verified by Codex CLI at `xhigh`. It found **no algorithmic defect** — `psl_at` correct across wraparound, `shift_back`'s implicit decrement exact, `push_away` preserving the Robin Hood recurrence and terminating, growth reinserting under the new mask, the sentinel remap sound, every `psl_at` call guarded by an occupied slot, and miss lookup retaining its early exit. Full verdict in a comment below.

It withheld a sign-off on methodology rather than correctness: its sandbox is read-only, so it could not run fresh interleaved benchmarks and would not certify numbers it had not reproduced. The alternating samples above are my response to that; the benchmark claims remain mine.

One gap it names and I have not closed: deleting `keys.set_null(cur)` in `shift_back` would leave every test passing while removed keys stayed reachable. That is the same liveness gap discussed on #3712 — no portable way to assert it across four backends — so the guard is a comment at the line explaining why it is not redundant.

### DIFF
--- a/hashset/derived_psl_test.mbt
+++ b/hashset/derived_psl_test.mbt
@@ -1,0 +1,131 @@
+// Copyright 2026 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// The probe-sequence length is derived from a slot's position rather than
+// stored, so these exercise every path that used to write a PSL: displacement,
+// shift-back cascades, growth, copying, and the remapped sentinel.
+
+///|
+priv struct Coll(Int) derive(Eq)
+
+///|
+impl Hash for Coll with fn hash(self) {
+  let Coll(x) = self
+  x % 4
+}
+
+///|
+impl Hash for Coll with fn hash_combine(self, hasher) {
+  let Coll(x) = self
+  hasher.combine_int(x % 4)
+}
+
+///|
+/// Heavy collisions: long probe runs, push_away displacement, shift_back
+/// cascades, growth, and copy -- every path the derived PSL has to get right.
+test "two-array stress" {
+  let s = @hashset.HashSet([])
+  for round in 0..<6 {
+    for i in 0..<200 {
+      s.add(Coll(i + round * 200))
+    }
+    for i in 0..<200 {
+      if i % 3 == 0 {
+        s.remove(Coll(i + round * 200))
+      }
+    }
+    for i in 0..<200 {
+      if i % 3 == 0 {
+        s.add(Coll(i + round * 200))
+      }
+    }
+  }
+  inspect(s.length(), content="1200")
+  let mut found = 0
+  for i in 0..<1200 {
+    if s.contains(Coll(i)) {
+      found += 1
+    }
+  }
+  inspect(found, content="1200")
+  // absent keys must stay absent
+  let mut absent = 0
+  for i in 1200..<1400 {
+    if !s.contains(Coll(i)) {
+      absent += 1
+    }
+  }
+  inspect(absent, content="200")
+  let dup = s.copy()
+  for i in 0..<600 {
+    dup.remove(Coll(i))
+  }
+  inspect(dup.length(), content="600")
+  inspect(s.length(), content="1200")
+}
+
+///|
+/// The sentinel path: keys whose hash is exactly `empty_hash` must round-trip,
+/// since their stored hash is remapped.
+test "keys hashing to the empty sentinel" {
+  let s = @hashset.HashSet([])
+  for i in 0..<400 {
+    s.add(Coll(i))
+  }
+  // Coll(0), Coll(4), ... hash to 0, which is the sentinel value.
+  inspect(s.contains(Coll(0)), content="true")
+  inspect(s.contains(Coll(4)), content="true")
+  inspect(s.length(), content="400")
+  s.remove(Coll(0))
+  inspect(s.contains(Coll(0)), content="false")
+  inspect(s.contains(Coll(4)), content="true")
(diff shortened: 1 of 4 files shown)


bobzhang:
## Codex CLI verification (`xhigh`)

No algorithmic defect found. It withheld sign-off on methodology — a read-only sandbox prevented it running fresh benchmarks — not on correctness.

> **I would not land it yet. I found no algorithmic correctness defect, but the required fresh benchmark comparison and full validation could not be completed because the workspace is read-only: Moon cannot acquire its `_build` lock.**
>
> Algorithm review:
>
> - `psl_at` correctly computes circular displacement modulo the power-of-two capacity.
> - In `shift_back`, moving `next` to `cur = next - 1 mod capacity` reduces its derived PSL by exactly one, including wraparound. Stopping at an empty slot or PSL zero matches the old implementation.
> - `push_away` preserves the previous Robin Hood recurrence exactly. It must terminate because insertion occurs with at least one empty slot; growth caps occupancy at 13/16.
> - Growth reinserts every occupied entry using the new mask, so no old-table displacement information is needed.
> - Mapping hashes `0 → 1` is sound. Real hashes zero and one share their home bucket and cached fingerprint, but equality comparisons distinguish their keys. Insertion, lookup, removal, and growth all use the remapped value consistently.
> - Every `psl_at` call is guarded by an occupied hash slot or receives an occupant directly from such a slot.
> - Miss lookup retains the same Robin Hood early exit. Independently, an empty slot must occur before a full-table scan because the table is never close to full.
>
> Test mutation not caught: remove `self.keys.set_null(cur)`. Membership tests would still pass because the hash sentinel controls occupancy, but removed reference keys would remain retained.
>
> - `git diff --check` passed. `hashset/pkg.generated.mbti` is unchanged from `origin/main`. Direct formatter comparison found all four changed files formatted. Cached native and JS executables passed the derived-PSL and internal wraparound tests.
>
> Benchmark conclusion: undetermined. […] there is no same-session `origin/main` baseline, and those binaries predate the commit by several minutes. Therefore they cannot establish either parity or a real string regression.

That last point is the right call and I have answered it with alternating samples rather than a single pass, since the two js insertion figures are what the whole design turns on:

| js, n=50000 | two-array (A1, A2) | main (B1, B2) |
| --- | --- | --- |
| `add` integer keys | 2.32, 2.33 ms | 2.38, 2.36 ms |
| `add` string keys | 3.57, 3.59 ms | 3.42, 3.39 ms |

Integer insertion is at parity — marginally ahead, with ranges overlapping. String insertion is a real 5% regression, which is tighter than the 7% I first reported from a single sample.

The uncaught mutation is the same liveness gap discussed on #3712: asserting that a removed key became collectable needs weak references or finalization, which is not portable across the four backends. The guard is a comment at the line, since what actually threatens that call is someone reading `shift_back` and deleting a line that looks redundant beside the `hashes` write above it.

`moon check --deny-warn` clean on native, js and wasm-gc; `moon test` 7549/7549 and 7490/7490 on js; `pkg.generated.mbti` unchanged.

bobzhang:
to focus on 1.0 api design first, perf will be reworked on after 1.0

PR #4135, opened 2026-08-24
fix(builtin, deque): stop nulling vacated slots so views cannot read invalid memory

Yu-zh:
Taking a view from an `Array` or a `Deque` shares the underlying buffer, and every shrinking operation cleared the slots it vacated with `%fixedarray.set_null`. A view created before such a mutation then read a null slot. `Array::view` and `Deque::as_views` are both plain `pub`, so this is undefined behaviour reachable from entirely safe code:

```moonbit
let arr = [Box::{ v: 11, .. }, Box::{ v: 22, .. }, Box::{ v: 33, .. }]
let view = arr[0:3]
let _ = arr.pop()
view[2].v            // SIGSEGV on native
```

## Approach

Vacated slots are no longer cleared, so a view can only ever observe valid values of `T`. Mutating a container while a view of it is alive stays a program error, but what the view yields is now merely unspecified rather than invalid.

The cost is that a removal retains what it removes. That applies uniformly -- `clear` empties an array the same way `pop` shortens it, and neither writes to the slots it gives up -- so there is no operation-by-operation rule to learn and no existing signature changes. The elements are released when a later push reuses the slot, when the buffer grows, or when the container is dropped, which means the clear-and-refill pattern self-heals: each push releases one old occupant, so repeated fill/drain cycles on one container hold flat.

| | cost | capacity |
| --- | --- | --- |
| `fill_unused(value)` — new | one pass over the unused capacity, no allocation | kept |

`fill_unused` overwrites every slot from `length()` to `capacity()` — or its wrapped equivalent for a `Deque` — which is exactly the region any removal leaves behind. `shrink_to_fit` was already releasing those elements by letting the old buffer go; that is now documented rather than incidental.

```
Array::fill_unused(Self[T], T)
Deque::fill_unused(Self[A], A)
```

Two new methods, nothing else in the public surface moves. `Array::resize` shrinks like `truncate` and no longer releases on that branch. `Deque::clear` and `Deque::truncate` both collapse to length assignments, and `Deque::truncate(0)` no longer routes through `Deque::clear`.

Writing into `[length(), capacity())` is safe without tracking a high-water mark: that region is always either NULL or a live reference, never garbage. `%fixedarray.make_uninit` NULL-fills for reference element types — it has to, since `moonbit_drop_object` walks a REF_ARRAY's full capacity and skips slots with `if (!obj) continue` — and `moonbit_make_ref_array_with_blit` NULL-fills everything outside the blitted range when a buffer grows. It is the same region `Array::push` writes into on every push past the previous high-water mark.

`%fixedarray.set_null` now has no users in core.

## Testing

- `moon test` on `--target native` (7479), `wasm` (7561), `wasm-gc` (7561) and `js` (7495), all passing
- 20 new regression tests covering: views after `pop`/`clear`/`truncate`/`drain`; `fill_unused` after each of them, including the wrapped case for `Deque` and a run of removals settled by one call; and that no shrinking operation replaces the buffer, pinned by pushing after emptying and observing through a view taken beforehand
- `moon check --deny-warn --target all` clean apart from a pre-existing `encoding/utf8/decode_js.mbt:21` FFI deprecation
- `moon info`, `moon fmt` — `.mbti` churn is the two added methods and nothing else


bobzhang:
Split into two independent PRs so each container can be reviewed on its own, both rebased onto current `main`:

- #4190 — the `Array` half (`builtin/`)
- #4189 — the `Deque` half (`deque/`)

Each PR body has a "Changes relative to #4135" section listing what differs from this branch (for `Array`: `resize_buffer` now copies only the live prefix, and `fill_unused` uses the `%fixedarray.fill` intrinsic; for `Deque`: docs only) and a "Review notes" section with the design consequences an independent review pass surfaced.

PR #4147, opened 2026-08-24
perf(quickcheck): execute zipped generators directly

mizchi:
Implement `Generator::zip_with` and `zip_with3` directly instead of composing `flat_map` and `map`. The previous composition allocated temporary generators and closures for every generated sample.

The state splitting and evaluation order are preserved.

Native benchmark (`zip_with`, 1,000,000 samples):

| before | after | change |
| ---: | ---: | ---: |
| 38.66 ms | 11.64 ms | 70% faster |

Validation: all 82 QuickCheck tests pass on wasm, wasm-gc, JS, and native; `moon check --target all` and `moon info` pass.

### DIFF
--- a/quickcheck/generator.mbt
+++ b/quickcheck/generator.mbt
@@ -125,7 +125,10 @@ pub fn[T, U, V] Generator::zip_with(
   other : Generator[U],
   combine : (T, U) -> V,
 ) -> Generator[V] {
-  self.flat_map(first => other.map(second => combine(first, second)))
+  Generator((size, state) => {
+    let other_state = state.split()
+    combine(self.run(size, state), other.run(size, other_state))
+  })
 }
 
 ///|
@@ -136,7 +139,15 @@ pub fn[T, U, V, W] Generator::zip_with3(
   third : Generator[V],
   combine : (T, U, V) -> W,
 ) -> Generator[W] {
-  self.flat_map(a => second.flat_map(b => third.map(c => combine(a, b, c))))
+  Generator((size, state) => {
+    let second_state = state.split()
+    let third_state = second_state.split()
+    combine(
+      self.run(size, state),
+      second.run(size, second_state),
+      third.run(size, third_state),
+    )
+  })
 }
 
 ///|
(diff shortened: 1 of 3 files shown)


CAIMEOX (approved):

bobzhang:
Reviewed and verified locally on `c511bfa1`. This is correct and the numbers hold up — queued for auto-merge once CI finishes.

**The equivalence claim checks out.** I diffed the new implementations against the old `flat_map` compositions over 40 seeds × 7 sizes × 25 samples for both `zip_with` and `zip_with3`, plus `Arbitrary`-backed generators and nested `zip_with`/`zip_with3` compositions: byte-identical throughout, on all four backends. The test is sensitive to exactly the thing that could break — changing `second_state.split()` to `state.split()` in `zip_with3` fails 3 of those cases.

The one real reordering is in `zip_with3`: the second `split` now happens before `self.run`, where the old nesting did it after. That is safe because `RandomState::split` advances the parent's seed and returns a *new* record (`{ seed: mix64(seed1), gamma: mix_gamma(self.seed) }`), so parent and child never alias and the two operations commute.

**One thing worth recording, since the value-level diff cannot see it:** `combine(self.run(..), other.run(..))` makes the run order depend on argument evaluation order, where the old nesting fixed it structurally. For pure generators this is unobservable, since the states are disjoint. For an effectful generator it is observable, so I checked it directly with generators that append to a shared trace — `["a", "b"]` and `["a", "b", "c"]` on wasm, wasm-gc, js and native. MoonBit evaluates arguments left to right, so effectful generators still fire in the same order as before.

**Numbers reproduce.** `bench Generator::zip_with n=1000000`, native release, on an M-series mac: 35.15 ms → 9.46 ms, a 73% reduction — the same shape as the 38.66 → 11.64 ms in the description. `zip` delegates to `zip_with`, so it inherits the win.

**Everything else is clean.** `moon test -p quickcheck --target all` 157/157 on wasm, wasm-gc, js and native; `moon check --target all`, `moon fmt --check` and `moon info` all clean with no `.mbti` change; and the new benchmark runs on all four targets, not just native.

Only nit, not worth holding the PR for: there is a benchmark for `zip_with` but none for `zip_with3`, which got the larger structural change.

PR #4152, opened 2026-08-24
Lazily construct collection shrink candidates

mizchi:
Defers construction of per-element shrink iterator chains until structural collection candidates have been exhausted. It also returns the empty Array/List candidate without traversing or copying the original collection.

For obtaining the first shrink candidate from a 10,000-element collection on native release:

| Collection | Before | After |
| --- | ---: | ---: |
| Array | 767.83 µs | 212.59 ns |
| List | 811.98 µs | 119.95 ns |

This is currently stacked on #4149 because that PR establishes empty-first List shrinking. Once #4149 merges, this PR’s diff will reduce to the lazy construction change and its benchmark.

Validation: `moon test quickcheck/shrink --target all`, `moon check --target all`, and `moon info`.


bobzhang:
Thanks for this — the lazy-construction idea holds up and is worth more now than when you filed it. I've opened #4208 with it rebased onto current `main`, with you as `Co-Authored-By`. Your branch is untouched.

It needed a re-cut rather than a conflict resolution, because the ground moved:

- **#4149 was closed**, so the empty-collection-first ordering never landed. #4208 therefore keeps only the changes that leave the candidate sequence exactly as it is: `deferred(..)` around the trailing `.concat(shr_sub_terms(..))` in both instances, and the `k == n` shortcut in `removes_array`. The `[|xs.take(0)|]` prepend in `Shrink for @list.List` is dropped.
- **`main` had already absorbed two pieces of this PR**: the `bench` test import in `quickcheck/moon.pkg`, and `Shrink for @list.List` already starts at `k = n / 2`.

One thing worth knowing about the List number in your table: it came from the empty-first prepend, not from the laziness. With `main`'s ordering the first `k` is 1, so the first candidate *is* an `n - 1` element list — `O(n)` to materialise no matter how lazily the chain is built. Only the second traversal goes away, so List lands at 287.79 µs → 147.79 µs at n = 1000 rather than dropping to nanoseconds. Array does get the full win: 121.52 µs → 236.84 ns.

Your benchmarks also turned up a real latent bug, which is arguably the most valuable thing here: at 10,000 elements the List case dies with `RangeError: Maximum call stack size exceeded` on js and wasm-gc, because `removes_list` recurses once per element. That is independent of this change — shrinking a large list will overflow the stack on those backends today. #4208 lowers the benchmark to 1,000 and calls the bug out for a separate issue.

Leaving this open for you to close, or for a maintainer to.

PR #4154, opened 2026-08-25
feat: improve scalar Arbitrary

CAIMEOX:
This PR improves QuickCheck’s default generation for fixed-width integer types (`Byte`, `Int16`, `UInt16`, `Int`, `UInt`, `Int64`, and `UInt64`). Instead of restricting values by the size hint, generators now use a weighted mix of full-width random values, canonical small values, binary and decimal boundaries, and type extrema. Integer shrinking now tries zero first and converges from coarse to fine, including overflow-safe handling of signed minimum values. `BigInt` remains unchanged, and there are no public API changes.


quickcheck/shrink/shrink.mbt
```diff
@@ -46,21 +46,18 @@ impl Shrink with fn shrink(_a) {
...
+  [|..if x != 0 { [0] }, ..[ for z = x / 2; z != 0; z = z / 2 => x - z ]|]
 }
 
 ///|
 pub impl Shrink for Int64 with fn shrink(x) {
   [|
-    ..[ for z = x / 2; (x - z).abs() < x.abs(); z = z / 2 => x - z ].rev_iter(),
     ..if x != 0 {
       [(0 : Int64)]
     },
```
bobzhang: Int64(0) or just `0`

quickcheck/shrink/shrink.mbt
```diff
@@ -103,20 +100,24 @@ pub impl Shrink for @bigint.BigInt with fn shrink(x) {
 ///|
 pub impl Shrink for UInt with fn shrink(x) {
   [|
-    ..[ for z = x / 2; z > 0; z = z / 2 => x - z ].rev_iter(),
     ..if x != 0 {
       [(0 : UInt)]
```
bobzhang: same

quickcheck/shrink/shrink.mbt
```diff
@@ -103,20 +100,24 @@ pub impl Shrink for @bigint.BigInt with fn shrink(x) {
...
+    ],
   |]
 }
 
 ///|
 pub impl Shrink for UInt64 with fn shrink(x) {
   [|
-    ..[ for z = x / 2; z > 0; z = z / 2 => x - z ].rev_iter(),
     ..if x != 0 {
       [(0 : UInt64)]
```
bobzhang: same

quickcheck/shrink/shrink.mbt
```diff
@@ -133,10 +134,12 @@ pub impl Shrink for Bool with fn shrink(b) {
 pub impl Shrink for Byte with fn shrink(x) {
   let xi = x.to_int()
   [|
-    ..[ for z = xi / 2; z > 0; z = z / 2 => (xi - z).to_byte() ].rev_iter(),
     ..if xi != 0 {
       [Int::to_byte(0)]
```
bobzhang: same

bobzhang:
Nice direction. Biasing generation toward canonical values, binary/decimal boundaries and type extrema is exactly what a mature QuickCheck should do — uniform random integers almost never hit the values that actually break code. I went through the generator, the shrink ladders, and the repo-wide fallout. One blocking issue, two test-strength regressions, and three design questions.

## Blocking: the default shrink budget can no longer minimize full-width integers

`max_shrinks` still defaults to `100` (`quickcheck/driver.mbt:295`, and again at `:412` for `check`), but counterexamples are now full-width. The ladder spends **two** attempts per accepted halving, because `0` is re-emitted and rejected at every rung:

```moonbit
[|..if x != 0 { [0] }, ..[ for z = x / 2; z != 0; z = z / 2 => x - z ]|]
```

So a 64-bit counterexample needs ~127 attempts to reach the minimum, and gets cut off at 100.

Reproduction — `(x : Int64) => x == 0L`, whose minimal counterexample is `1`:

| seed | default (`max_shrinks=100`) | `max_shrinks=100000` |
|------|------------------------------|----------------------|
| 3 | `counterexample=6154`, `attempts=100` (capped) | `counterexample=1`, `attempts=127` |
| 7 | `counterexample=6972`, `attempts=100` (capped) | `counterexample=1`, `attempts=127` |
| 8 | `counterexample=2426`, `attempts=100` (capped) | `counterexample=1`, `attempts=125` |

3 of 8 seeds fail to minimize. Before this PR the first counterexample was bounded by `size` (<= `max_size` = 100), so 100 attempts always sufficed — this is a direct consequence of dropping the size bound, and it degrades the single most user-visible feature of the library ("here is the *minimal* failing input").

The PR's own updated snapshots already show the symptom: `quickcheck/driver_test.mbt:505` and `quickcheck/README.mbt.md:117` both record a plain `Int` burning **63 of its 100** attempts.

Two ways out:

- Raise the default `max_shrinks` (>= 130 covers 64-bit; both call sites), or
- Stop re-probing `0` on every rung — emit it once — which halves the ladder cost and keeps the budget meaningful.

I'd lean toward the second plus a modest bump, with an `Int64`/`UInt64` regression test in the driver suite.

## Test strength

**`Arbitrary for HashSet` is now close to vacuous** (`quickcheck/arbitrary_type_test.mbt:480`). It previously asserted `samples[i].symmetric_difference(cases[i]).is_empty()` across all 20 samples. It now snapshots 4, and the recorded value is:

```
[[], [], [-1073741824, -2], []]
```

Three of the four are empty sets, so effectively one sample is checked. Same shape for `Arbitrary for HashMap` at `:125` (4 of 20, and `verify_hashmap_content` was deleted along with the only `contains_kv` coverage). The snapshots are deterministic, so re-recording the 20-entry tables is mechanical and would preserve the original strength.

**Vacuous assertion left behind** (`quickcheck/arbitrary_test.mbt:120`):

```moonbit
inspect("[0, 0, 0, 1, 2, -2]", content="[0, 0, 0, 1, 2, -2]")
```

This compares a string literal to itself and can never fail. The commented-out line just below shows it was meant to be `samples[9]`. Every other expectation in that function was re-recorded, so this one is now both dead and stale.

## Design questions (non-blocking)

**1. The canonical/boundary buckets are width-independent, so narrow types get squeezed.** `canonical_integer_bits` (`quickcheck/arbitrary_integer.mbt:33`) always draws from a fixed 4–5 element set regardless of `bits`. Measured over 1e6 draws, `Byte` now yields `0` ~7.1%, `1`/`2` ~7.9% each, `3` ~6.7% — about **30% of all bytes land in {0,1,2,3}** — while an ordinary byte like `77` appears ~0.156%. All 256 stay reachable, so this is a calibration choice rather than a bug. But `Arbitrary for Bytes` (`quickcheck/arbitrary.mbt:35`) draws each element through `Byte::arbitrary`, so every byte-oriented suite in the repo (`encoding/base64`, `encoding/hex`, `encoding/utf8`, `bytes`, `buffer`) now samples a materially narrower domain — a 10-byte `Bytes` averages ~3 bytes drawn from {0,1,2,3}. Was that intended? Scaling the bucket size with `bits` would keep the intent for 32/64-bit without flattening `Byte`.

**2. Keyed collections lost their duplicate-key coverage.** `Map`/`HashMap`/`SortedMap`/`Set`/`HashSet`/`SortedSet::arbitrary` pass the element *index* as the size hint. Under the old `rs.next_int() % size` that guaranteed `|key| < index`, so generated collections routinely contained repeated keys and exercised the overwrite / dedup / hash-collision paths. With full-width keys the birthday probability for a <=20-element collection is ~1e-7, so those paths are now effectively never generated. The updated snapshots show it directly: `<Set: [0, 2, 3, 1]>` became all-distinct 9–10 digit keys. A dedicated small-key generator for those impls would keep the collision paths covered.

**3. `BigInt` is now the odd one out.** It kept `if size == 0 { 0 }` (`quickcheck/arbitrary.mbt:151`) while every fixed-width type stopped consulting `size`. Since `gen` defaults `size` to `0`, `@quickcheck.gen[BigInt]()` is now deterministically `0` — and the updated README output shows exactly that (a trailing `0` after six full-width values). I realize `BigInt` is explicitly out of scope for the rewrite, but to a user following the documented `gen()` pattern this reads as a bug.

## Verified sound — no need to re-litigate these

- Bucket weights are exactly 40/20/20/10/10, and every value stays reachable via the full-width branch.
- Binary / decimal / extrema construction stays in range for all 8/16/32/64 signed and unsigned combinations; each `decimal_power_count` bound checks out against the corresponding type max.
- `next_uint(limit=)` rejection sampling is unbiased and terminates — correct mask for unit, power-of-two and arbitrary limits, zero aborts, and SplitMix's odd gamma gives full-period traversal so rejection can't get trapped.
- Exhaustive `Int16`/`UInt16` sweeps: every shrink ladder terminates and preserves "strictly closer to zero", signed minima included. Replacing the old `(x - z).abs() < x.abs()` guard with `z != 0` is a genuine fix — the old form overflowed at `Int::MIN` and silently produced an empty ladder. Worth a short comment noting the new guard relies on division truncating toward zero.
- The `.compare(0)` changes in `immut/sorted_map` and `immut/sorted_set` are **correct**, not a weakening. `Compare` specifies sign only (`builtin/traits.mbt:29-33`), so the original assertion of exact numeric equality between two independent `Compare` impls was over-specified.

`moon check --target all` and the full `moon test` (7546 passing) are green on this branch, across wasm/js/native; `moon fmt` and `pkg.generated.mbti` show no drift.

quickcheck/shrink/shrink.mbt
```diff
@@ -46,22 +46,12 @@ impl Shrink with fn shrink(_a) {
...
 
 ///|
 pub impl Shrink for Int64 with fn shrink(x) {
-  [|
-    ..[ for z = x / 2; (x - z).abs() < x.abs(); z = z / 2 => x - z ].rev_iter(),
-    ..if x != 0 {
-      [(0 : Int64)]
-    },
-  |]
+  [|..[ for z = x / 2; z != 0; z = z / 2 => x - z ], ..if x != 0 { [0L] }|]
```
bobzhang, line 54: **Blocking: preserve the default shrink budget for compound integers**

Putting zero after the full halving ladder fixes the single-scalar budget, but makes the greedy driver accept every intermediate value whenever another field already keeps the property failing. With the default `max_shrinks=100`, this reproduction:

```moonbit
@quickcheck.report(
  (_ : (Int64, Int64)) => false,
  count=1,
  seed=2,
)
```

stops at `(0, -93051221)` rather than `(0, 0)`; seed 7 similarly stops at `(0, -4096)`. Each full-width field can consume roughly 64 accepted steps, so this becomes common now that integer generation ignores size.

The straightforward fix is to restore zero-first ordering and raise the default budget to about 130 so a scalar `Int64` whose zero candidate is rejected can still finish. Please add regressions for both a scalar `Int64` property and an always-failing `(Int64, Int64)` property so the two cases stay covered.

CAIMEOX: There is an inherent tradeoff in candidate ordering with the current greedy, stateless shrink interface:

- With **zero first**, an always-failing compound value can clear each field immediately. However, if zero satisfies the property, the driver retries and rejects the same zero after every accepted shrink step. A full-width `Int64` then takes 127 attempts.
- With **zero last**, that scalar case takes only 64 attempts because zero is tested once at the end. However, compound values must walk each field’s halving ladder: two `Int64` fields can require 128 attempts.

Changing the ordering therefore moves the extra shrink work from one case to the other; neither ordering dominates universally. The driver cannot generically remember that zero was already rejected because `Shrink` is stateless and the input is not required to implement `Eq`.

Therefore, a recommended, feasible approach is to slightly increase the number of shrink iterations to around 130.

PR #4157, opened 2026-08-26
perf(json): index large replacer key sets

bobzhang:
Supersedes #4146 — the diagnosis and the original benchmark are @mizchi's; this changes where the index lives so no usage pattern regresses.

## Problem

`Replacer::keep` and `Replacer::exclude` closed over the key array and ran a linear `contains` for every object property, so stringifying an object with many properties against a large key list cost O(properties × keys).

## Approach

The replacer now carries its key set as *data* rather than as a closure, which is what makes indexing possible at all. On top of that:

- Key sets below a small threshold keep scanning linearly — cheaper than hashing at that size.
- Larger sets also scan linearly at first, and only build a `Map[StringView, Unit]` once the replacer has actually been consulted enough times to amortize it. The index then lives on the `Replacer` and is reused for its lifetime.

Deferring the build bounds the wasted work to `threshold × keys` and keeps one-shot uses allocation-free; caching it on the replacer means a reused replacer pays for the index once rather than once per call. `Map[StringView, Unit]` avoids copying every key with `to_owned()`.

## Results

Native, `Json::stringify` with 2,500 keep keys:

| object fields | main | this PR | |
| ---: | ---: | ---: | --- |
| 5 | 7.42 µs | 402.54 ns | 18× faster |
| 50 | 92.91 µs | 3.19 µs | 29× faster |
| 5,000 | 24.31 ms | 391.73 µs | 62× faster |

And the two cases that keep the design honest:

| case | main | this PR | |
| --- | ---: | ---: | --- |
| fresh replacer built per call, 2500 keys / 5 fields | 7.51 µs | 7.56 µs | unchanged |
| 3 keys / 5,000 fields | 91.4 µs | 95.1 µs | **~4% slower** |

That last row is a real regression and it reproduces: matching on the replacer enum costs a little more than the old direct closure call on the small-key hot path. I tried reordering the variants to recover it and it made no difference, so it looks like the indirection itself. It seems a fair trade for the rows above, but flagging it rather than burying it.

## Behaviour

Two things changed that are worth calling out:

**The key array is no longer re-read on every call.** Previously the replacer closed over the `ArrayView`, so mutating it between `stringify` calls was observed. Once a large key set has been indexed that no longer holds. This was an accident of closing over the view rather than an intended contract, so `keep`/`exclude` now document the key list as being read while the replacer is in use, with mutation afterwards not guaranteed to be observed.

**`Replacer`'s `@debug.Debug` output** is now `{ kind: Custom(<function: ...>) }` instead of `{ f: <function: ...> }`, reflecting the new private representation. `derive` is kept, so `pkg.generated.mbti` is byte-identical to main — no public API change.

## Also here

While documenting the above I noticed `keep`/`exclude` filter at *every* nesting level — `keep(["a"])` turns `{"a": {"b": 1}, "c": 2}` into `{"a": {}}`. That matches JavaScript's `JSON.stringify(value, keys)` and is not a change, but every doc comment and README example showed a flat object, so it was easy to be surprised by. Now documented and pinned by a test.

Tests: the new ones check that the indexed path agrees with an equivalent custom replacer for both `keep` and `exclude` across the threshold, that a replacer stays correct when reused after its index is built, and that `transform` agrees with `stringify`.

Validation: 218 tests pass on wasm, wasm-gc, JS and native; `moon check --target all`, `moon info` and `moon fmt` are clean; `pkg.generated.mbti` matches main exactly.

https://claude.ai/code/session_01MWKp8X7A5VmoV1PxkT8zaS

### DIFF
--- a/json/json.mbt
+++ b/json/json.mbt
@@ -116,14 +116,48 @@ priv enum WriteFrame {
 ///|
 /// A Replacer provides a way to filter and transform JSON object properties during stringification.
 /// 
-/// Replacers contain a function that takes a property key and value, and returns:
+/// A replacer takes a property key and value, and returns:
 /// - `Some(value)` to include the property in the output (possibly transformed)
 /// - `None` to exclude the property from the output
 /// 
 /// Only applies to object properties, not array elements.
+///
+/// The replacer is consulted at every nesting level, including inside objects
+/// it has just kept, matching JavaScript's `JSON.stringify`. So a replacer that
+/// keeps only `"a"` turns `{"a": {"b": 1}}` into `{"a": {}}` — the nested
+/// `"b"` is filtered by the same rule.
 pub struct Replacer {
-  priv f : (String, Json) -> Json?
+  priv kind : ReplacerKind
 } derive(@debug.Debug)
+
+///|
+priv enum ReplacerKind {
+  Custom((String, Json) -> Json?)
+  Keep(ArrayView[StringView])
+  Exclude(ArrayView[StringView])
+  AdaptiveKeep(ReplacerKeyLookup)
+  AdaptiveExclude(ReplacerKeyLookup)
+} derive(@debug.Debug)
+
+///|
+/// Lookup state for a key set large enough to be worth indexing. The index is
+/// built on first heavy use and then reused for the lifetime of the `Replacer`.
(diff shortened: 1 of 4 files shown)


bobzhang:

PR #4163, opened 2026-08-27
refactor(encoding): use v128 bitstring patterns

Yu-zh:
## Summary

- replace manual `Bytes`/`FixedArray` conversions and indexed v128 loads with `v128le` bitstring patterns in ASCII, Base64, hex, and UTF-16 codecs
- express scalar tails in the same pattern matches while preserving malformed-input and UTF-16 surrogate behavior
- remove four private `%identity` conversion helpers

The updated compiler scalar-replaces the tail-consuming view loops, so the native hot loops no longer perform per-block reference counting. The Base64 loop retains its source once because its 16-byte load advances by 12 bytes.

## Native spot checks

- ASCII decode: 11.98 us for 100K, 90.08 us for 1M
- Base64 encode: 14.14 us for 64K
- hex encode: 14.11 us for 64K
- UTF-16 ASCII decode: 8.60 us LE, 15.51 us BE for 100K code units

I also tried converting `BytesView::lexical_compare`, but left its indexed loop unchanged because the pattern version was slower.

## Verification

- `moon check --target all` (0 errors; the updated toolchain reports 117 existing qualification warnings in unchanged Buffer docs/tests)
- `moon info --target wasm,wasm-gc,js,native` (no `.mbti` changes)
- `moon test --target all`
  - wasm: 7563 passed
  - wasm-gc: 7563 passed
  - JS: 7504 passed
  - native: 7478 passed
- `moon test --target native --release` (7478 passed)

### DIFF
--- a/encoding/ascii/decode.mbt
+++ b/encoding/ascii/decode.mbt
@@ -29,10 +29,6 @@ fn suppress_unused_v128_import_on_js() -> Unit {
 }
 
 ///|
-#cfg(not(target="js"))
-fn unsafe_fixedarray_from_bytes(bytes : Bytes) -> FixedArray[Byte] = "%identity"
-
-///|
 fn finish_string(buffer : FixedArray[UInt16], len : Int) -> String {
   if len == buffer.length() {
     unsafe_fixedarray_uint16_to_string(buffer)
@@ -66,40 +62,33 @@ fn decode_scalar(bytes : BytesView) -> String raise Malformed {
 fn decode_v128(bytes : BytesView) -> String raise Malformed {
   let length = bytes.length()
   let result = FixedArray::make(length * 2, b'\x00')
-  let source = unsafe_fixedarray_from_bytes(bytes.data())
-  let source_start = bytes.start_offset()
-  let source_end = source_start + length
-  let mut source_offset = source_start
-  let mut result_offset = 0
-  while source_offset + 16 <= source_end {
-    let block = @v128.v128_load(source, source_offset)
-    let non_ascii = @v128.i8x16_bitmask(block)
-    if non_ascii != 0 {
-      let malformed_offset = source_offset - source_start + non_ascii.ctz()
-      raise Malformed(bytes[malformed_offset:])
+  for rest = bytes, result_offset = 0 {
+    match rest {
+      [v128le(block), .. tail] => {
+        let non_ascii = @v128.i8x16_bitmask(block)
+        if non_ascii != 0 {
+          raise Malformed(bytes[result_offset / 2 + non_ascii.ctz():])
+        }
+        @v128.v128_store(
+          result,
+          result_offset,
+          @v128.i16x8_extend_low_i8x16_u(block),
+        )
+        @v128.v128_store(
+          result,
+          result_offset + 16,
+          @v128.i16x8_extend_high_i8x16_u(block),
+        )
+        continue tail, result_offset + 32
+      }
+      [0..=0x7F as byte, .. tail] => {
+        result.unsafe_set(result_offset, byte)
+        result.unsafe_set(result_offset + 1, b'\x00')
+        continue tail, result_offset + 2
+      }
+      [] => break
+      _ => raise Malformed(bytes[result_offset / 2:])
     }
-    @v128.v128_store(
-      result,
-      result_offset,
-      @v128.i16x8_extend_low_i8x16_u(block),
-    )
-    @v128.v128_store(
-      result,
-      result_offset + 16,
-      @v128.i16x8_extend_high_i8x16_u(block),
-    )
-    source_offset = source_offset + 16
-    result_offset = result_offset + 32
-  }
-  while source_offset < source_end {
-    let byte = source.unsafe_get(source_offset)
-    if byte > b'\x7F' {
-      raise Malformed(bytes[source_offset - source_start:])
-    }
-    result.unsafe_set(result_offset, byte)
-    result.unsafe_set(result_offset + 1, b'\x00')
-    source_offset = source_offset + 1
-    result_offset = result_offset + 2
   }
   result.unsafe_reinterpret_as_bytes().to_unchecked_string()
 }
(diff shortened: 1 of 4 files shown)


bobzhang:
Reviewed and benchmarked this on all five backends. Correctness looks good; the perf picture differs from the native-only spot checks in the description, so posting the numbers.

## Method

Worktrees at `pr4163^` (3e57e09c) and `312c2662`, the existing `encoding/*` bench suites, A/B interleaved, `--release`, min of 5 reps (3 for llvm). Untouched benches in the same binaries — base64 `decode v128`, hex `decode`, all `*_scalar` — act as the noise control.

## Results

**native** — neutral, every touched path ≤1%:

| bench | base | head | Δ |
|---|---|---|---|
| ascii decode 1M | 78.7 µs | 79.3 µs | +0.8% |
| base64 encode v128 64k | 15.57 µs | 15.67 µs | +0.6% |
| hex encode 64k | 14.15 µs | 14.27 µs | +0.8% |
| utf16 decode ASCII BE | 13.50 µs | 13.43 µs | −0.5% |

(Native `*_scalar` benches swing ±30% between builds — untouched code, pure code-layout drift. Native is the noisiest target here, which matters since it's the one the description measured.)

**llvm** — neutral (±5%). Note `moon bench --target llvm` without `--release` is unoptimized and shows a spurious −15%; in release it disappears.

**js** — no code change at all. All four touched functions are `#cfg(not(target="js"))` or `#cfg(any(target="native", target="wasm"))`, so JS compiles none of them.

**wasm** — consistent regression on every touched path; controls within ±1%:

| bench | base | head | Δ |
|---|---|---|---|
| utf16 decode ASCII BE 100k | 21.07 µs | 26.03 µs | **+23.5%** |
| utf16 decode mixed BE 100k | 21.14 µs | 26.17 µs | **+23.8%** |
| utf16 decode ASCII LE 100k | 12.68 µs | 13.77 µs | +8.6% |
| ascii decode 1M | 98.7 µs | 108.4 µs | +9.8% |
| base64 encode v128 64k | 17.54 µs | 19.04 µs | +8.6% |
| hex encode 64k | 13.42 µs | 14.23 µs | +6.0% |
| *base64 decode v128 64k (control)* | *17.14 µs* | *17.15 µs* | *+0.1%* |

**wasm-gc** — ASCII decode only (the other three packages are cfg'd off there):

| bench | base | head | Δ |
|---|---|---|---|
| ascii decode 1M | 1.45 ms | 1.81 ms | **+24.8%** |
| ascii decode 100k | 154.1 µs | 188.5 µs | **+22.4%** |
| ascii decode n=64 | 96.3 ns | 122.4 ns | **+27.1%** |

## Cause

`v128le(block)` in a bitstring pattern lowers to an **out-of-line helper call** rather than an inline load. In `ascii.blackbox_test.wasm` the base loop has

```wat
local.get 5
local.get 38
i32.add
v128.load align=1
```

inline, while head has `local.get 5; local.get 8; call 155`, where func 155 is nothing but that same `i32.add; v128.load`. Same in `utf16_decode_be_no_surrogate_v128`: base's function body contains one `v128.load`, head's contains zero. That is one non-inlined call per 16-byte block in every hot loop.

On wasm-gc it's worse in kind — `v128_load` there is already emulated as 16 `array.get_u` + shifts, and the pattern adds *another* call layer on top (`108 → 169 → 170`, vs base's `90 → 93`).

## On wasm-opt

Worth noting, since it came up: `wasm-opt --all-features -O3` **does** inline the generated helper. Post-`-O3` the head loop contains the same inline `v128.load align=1` as base, so this specific call is erasable downstream. Two caveats though:

1. No wasm-opt step appears anywhere in `moon bench` / `moon test` / `moon build --target wasm --release --dry-run` for this module, so the numbers above are what the standard toolchain actually produces and runs. (`moon-wasm-opt` ships in `~/.moon/bin` but isn't in this build graph.)
2. I was not able to get a trustworthy end-to-end A/B timing of the `-O3`'d modules — moon relinks the `.wasm` on every bench run, and it resolves `moonc` from its own bin directory rather than `PATH`/`MOON_HOME`, so the optimizer can't be spliced into the pipeline without modifying the toolchain install. Driving the optimized modules through `moonrun` directly bypasses moon's warmup/batching and gave control benchmarks swinging ±250%, so I'm not quoting those numbers. The structural inlining is verified; the resulting timings are not.

The wasm-gc case is not covered by the wasm-opt argument either — different mechanism, and it's the largest regression of the set.

## Correctness

No issues found. `moon check --target all` clean; `moon test` on the four packages passes everywhere (wasm 84, wasm-gc 83, js 80, native 84, llvm 84). The risky parts are already well covered by tests this PR leaves green:

- `encoding/ascii/decode_v128_wbtest.mbt` quickchecks `decode_v128` against `decode_scalar` including the **error view** on unaligned view offsets — exactly the `result_offset / 2 + ctz` arithmetic that was rewritten.
- `encoding/hex` and `encoding/base64` v128-vs-scalar tests across all tail residues and view offsets.
- `encoding/utf16/quickcheck_test.mbt` oracle tests, including "a surrogate at every position of a vectorized block".

I also checked the tail-arm restructuring by hand: base64's `[b0,b1,b2,..]` / `[b0,b1]` / `[b0]` arms reproduce the old `>= 3` loop plus 1/2-byte padding exactly, and the `v128le` arm still requires 16 bytes while advancing 12.

Two small readability points, neither a bug, which I'll send as separate PRs:

1. `encoding/ascii/decode.mbt` — `raise Malformed(bytes[result_offset / 2:])` recovers the input offset by dividing the *output* offset. In that arm `rest` already is the view starting at the bad byte, so `_ as bad => raise Malformed(bad)` is simpler and matches `decode_scalar`. The v128 arm can likewise bind `as current` and use `current[non_ascii.ctz():]`.
2. `encoding/utf16/decode.mbt` — the leading `if bytes is [u16le(0xD800..=0xDFFF), ..] { return true }` (and its BE twin) is dead: the loop's own surrogate arm already covers the first code unit. It was dead on main too, but this is a natural time to drop it.

## Unrelated, pre-existing

On wasm-gc, `@ascii.decode` of 100 KB costs 154 µs vs 12 µs on wasm1, because `v128_load` is emulated byte-by-byte there. The `decode_v128` path is very likely a pessimization on wasm-gc regardless of this PR — probably worth its own issue.

PR #4169, opened 2026-08-31
perf: build Array::copy and the to_array paths without a redundant fill

Yu-zh:
Builds `Array::copy`, `Deque::to_array`, `SortedSet::to_array` and `HashMap::to_array` by allocating the result at its final length and writing each slot once, instead of filling the array and then overwriting it.

`Array::make(len, elem)` is `Array::unsafe_make_uninit(len)` followed by `len` `unsafe_set` calls. For a reference element type each of those also increfs `elem` and decrefs the slot's previous contents. Where the array is immediately overwritten, all of that is dead.

## Commits

1. **`test:` benchmarks** for the four functions, each with an Int and a Ref (String) variant. The element type is not incidental — it decides whether the per-element store carries reference-counting traffic, and the spread is large enough that an Int-only benchmark would hide it.
2. **`refactor(builtin):`** rename `Array::make_uninit` to `Array::unsafe_make_uninit` and export it `#doc(hidden)`, so packages outside `builtin` can build an array without going through `push`. `pkg.generated.mbti` is unchanged. No behavior change.
3. **`perf:`** the four conversions.

## Numbers

Native, release. Every A/B varies **one file only**, with all other files pinned, and rebuilds each round.

**`Array::copy`** — 4 rounds, ranges disjoint. `FixedArray::copy` is a control: its code is identical in both arms.

| | main-form | branch | min Δ | wins |
|---|---|---|---|---|
| `Array::copy` Int n=1000000 | 104.7 / 111.5 / 111.1 / 112.9 µs | 53.8 / 55.6 / 51.6 / 53.5 µs | **−50.7%** | 4/4 |
| `Array::copy` Ref n=1000000 | 2240 / 2200 / 2260 / 2250 µs | 1620 / 1700 / 1640 / 1780 µs | **−26.4%** | 4/4 |
| `FixedArray::copy` Ref *(control)* | 1110 / 1060 / 1110 / 1080 µs | 1060 / 1110 / 1080 / 1120 µs | 0.0% | 2/4 |

**`to_array`** — 3 rounds, `builtin` identical in both arms so only the three consumer files vary. Min of rounds.

| | push | unsafe_set | Δ |
|---|---|---|---|
| `Deque::to_array` Int n=50000 | 35.2 µs | 27.5 µs | **−21.9%** (3/3) |
| `Deque::to_array` Ref n=50000 | 99.9 µs | 91.7 µs | **−8.2%** (3/3) |
| `HashMap::to_array` Int n=50000 | 496.0 µs | 462.4 µs | **−6.8%** (3/3) |
| `HashMap::to_array` Ref n=50000 | 783.1 µs | 845.0 µs | no signal — ranges overlap |
| `SortedSet::to_array` Int n=50000 | 205.8 µs | 204.4 µs | −0.7% — no signal |
| `SortedSet::to_array` Ref n=50000 | 240.5 µs | 236.5 µs | −1.7% — no signal |

The gain tracks how little else the loop does per element. `Deque` only indexes and stores, so the store path is most of the work. `HashMap` allocates a tuple per pair and increfs both halves when they are references, which dominates. `SortedSet` is dominated by its `dfs` recursion. `SortedSet` and `HashMap` keep the conversion despite showing no measurable gain, so all four paths build their result the same way.

`HashMap::to_array` Ref is reported as no signal rather than a regression: its own push arm spans 14% across rounds, wider than the difference between the arms.

## Two things reviewers should know

**Bind the value to a local before `unsafe_set`.** Written inline as `xs.unsafe_set(i, self[i])`, the call sits between the buffer field load and the store, so the reference-counting pass brackets the buffer borrow *inside* the loop and emits an `incref`/`decref` pair per element — costing more than the bounds check it saves. Generated C, before and after:

```c
buf = xs->$0;  incref(buf);  v = at(self,i);  buf[i] = v;  decref(buf);   /* inline  */
v = at(self,i);  buf = xs->$0;  buf[i] = v;                               /* hoisted */
```

This is why every converted loop binds first, and it's recorded on `unsafe_make_uninit`'s doc comment.

**`unsafe_make_uninit` is only genuinely uninitialized for primitive element types.**

```
Array[Int]     moonbit_make_int32_array_raw    -- raw
Array[Double]  moonbit_make_double_array_raw   -- raw
Array[String]  moonbit_make_ref_array(n, ...)  -- filled with a placeholder
```

For reference elements the runtime fills the store with a valid placeholder, so slots are type-safe but meaningless, and part of the initializing pass still happens inside the function. It still beats `Array::make` there — `Array::make` adds `len` increfs and `len` decrefs on top of the same fill — but the margin is smaller, which matches the Int-vs-Ref split in both tables.

## Verification

- `moon check --target all` clean
- `moon test` — wasm-gc 7578, js 7518, wasm 7548, native 7463; all passing, matching the `main` baseline on the same tree
- `moon fmt`, `moon bundle`, `moon info` clean; `pkg.generated.mbti` unchanged

Empty-input paths keep their `len == 0` fastpath; dropping it costs 6–31% on empty input because `unsafe_make_uninit(0)` has no `[]` shortcut and `SortedSet` would still allocate its `dfs` closure. Covered by existing tests (`builtin/array_test.mbt:1250`, `deque/deque_test.mbt:1262`, `sorted_set/set_test.mbt:156`).

### DIFF
--- a/builtin/array.mbt
+++ b/builtin/array.mbt
@@ -74,7 +74,7 @@ pub fn[T] Array::from_fixed_array(arr : FixedArray[T]) -> Array[T] {
 /// One should use makei() instead which creates an object for each index.
 #owned(elem)
 pub fn[T] Array::make(len : Int, elem : T) -> Array[T] {
-  let arr = Array::make_uninit(len)
+  let arr = Array::unsafe_make_uninit(len)
   for i in 0..<len {
     arr.unsafe_set(i, elem)
   }
@@ -109,7 +109,7 @@ pub fn[T] Array::makei(length : Int, f : (Int) -> T raise?) -> Array[T] raise? {
   if length <= 0 {
     []
   } else {
-    let array = Array::make_uninit(length)
+    let array = Array::unsafe_make_uninit(length)
     for i in 0..<length {
       array.unsafe_set(i, f(i))
     }
@@ -588,7 +588,7 @@ pub fn[T, U] Array::map(
   self : Array[T],
   f : (T) -> U raise?,
 ) -> Array[U] raise? {
-  let arr = Array::make_uninit(self.length())
+  let arr = Array::unsafe_make_uninit(self.length())
   for i, v in self {
     arr.unsafe_set(i, f(v))
   }
@@ -636,7 +636,7 @@ pub fn[T, U] Array::mapi(
   if self.is_empty() {
     return []
   }
-  let arr = Array::make_uninit(self.length())
+  let arr = Array::unsafe_make_uninit(self.length())
   for i, v in self {
     arr.unsafe_set(i, f(i, v))
   }
@@ -775,7 +775,7 @@ pub fn[T] Array::rev_in_place(self : Array[T]) -> Unit {
 /// ```
 pub fn[T] Array::rev(self : Array[T]) -> Array[T] {
   let len = self.length()
-  let arr = Array::make_uninit(len)
+  let arr = Array::unsafe_make_uninit(len)
   for i in 0..<len {
     arr.unsafe_set(i, self.unsafe_get(len - i - 1))
   }
@@ -818,7 +818,7 @@ pub fn[T] Array::split_at(self : Array[T], index : Int) -> (Array[T], Array[T]) 
       len=len2,
     )
   } else {
-    Array::make_uninit(0)
+    Array::unsafe_make_uninit(0)
   }
   (v1, v2)
 }
@@ -1271,7 +1271,7 @@ pub fn[T] Array::flatten(self : Array[Array[T]]) -> Array[T] {
   } nobreak {
     len
   }
-  let res = Array::make_uninit(len)
+  let res = Array::unsafe_make_uninit(len)
   for xs in self; i = 0 {
     res.unsafe_blit(i, xs, 0, xs.length())
     continue i + xs.length()
--- a/builtin/array_make_blit_bench_test.mbt
+++ b/builtin/array_make_blit_bench_test.mbt
@@ -116,3 +116,23 @@ test "bench Array::push Ref n=1000000 resize" (it : @bench.T) {
     it.keep(data)
   })
 }
+
+///|
+fn make_blit_int_array(len : Int) -> Array[Int] {
+  Array::makei(len, i => i)
+}
+
+///|
+/// `Array::copy` allocates and fills in one shot. The Int and Ref variants are
+/// both kept because the element type decides whether the per-element store
+/// carries reference-counting traffic.
+test "bench Array::copy Int n=1000000" (it : @bench.T) {
+  let data = make_blit_int_array(make_blit_ref_bench_size)
+  it.bench(fn() { it.keep(data.copy()) })
+}
+
+///|
+test "bench Array::copy Ref n=1000000" (it : @bench.T) {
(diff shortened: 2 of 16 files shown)


bobzhang (commented):
## Review

LGTM. I verified both the code and the perf claims; the change is correct and the win is real — and it extends well beyond the native-only numbers in the PR description.

### Correctness

All four conversion sites write every slot exactly once before the array escapes, so `unsafe_make_uninit` is used safely:

- `Array::copy` (nonjs) → `unsafe_blit` fills all `len` slots.
- `Deque::to_array` → fills all `len` slots.
- `HashMap::to_array` → fills exactly `self.size` slots, with an early `break` once `n == self.size` (sound given the size invariant).
- `SortedSet::to_array` → in-order `dfs` fills exactly `self.size` slots.
- `Map::to_array` is a rename only — it already used the uninit + `unsafe_set` path.

The "bind before `unsafe_set`" discipline is followed in the three consumer loops, and every empty-input fastpath is preserved. `#doc(hidden)` correctly keeps the new `pub` fn out of `pkg.generated.mbti` (I ran `moon info builtin` — no diff), so "mbti unchanged" is accurate.

### Perf (independent A/B, `moon test --release`, one round of 10 winsorized samples per cell; main `3082191` vs branch `7bdf8a7`)

| benchmark | native | wasm-gc | js |
|---|---|---|---|
| `Array::copy` Int | 102.0→56.1 µs (**−45%**) | 862→478 µs (**−45%**) | 398→421 µs (+6%†) |
| `Array::copy` Ref | 2583→1698 µs (**−34%**) | 1091→707 µs (**−35%**) | 402→417 µs (+4%†) |
| `Deque::to_array` Int | 29.4→21.9 µs (**−26%**) | 108→81 µs (**−25%**) | 226→225 µs (−1%) |
| `Deque::to_array` Ref | 118.4→80.3 µs (**−32%**) | 122→98 µs (**−20%**) | 248→227 µs (**−8%**) |
| `HashMap::to_array` Int | 459→447 µs (−3%) | 522→472 µs (**−10%**) | 600→523 µs (**−13%**) |
| `HashMap::to_array` Ref | 688→645 µs (−6%) | 588→540 µs (**−8%**) | 602→570 µs (−5%) |
| `SortedSet::to_array` Int | 208→172 µs (**−17%**) | 160→141 µs (**−12%**) | 364→302 µs (**−17%**) |
| `SortedSet::to_array` Ref | 231→188 µs (**−19%**) | 190→166 µs (**−13%**) | 366→303 µs (**−17%**) |

† js `Array::copy` is the unchanged native `JSArray::copy`, so the +4–6% is run noise, as expected.

Two notes vs. the PR description:

1. **The win is not native-only.** It holds on wasm-gc (uniformly −8%…−45%) and, for the three consumer `to_array` paths, on js too. The description only claims native numbers, so this is strictly more good news.
2. **`SortedSet::to_array` shows a real improvement here** (−12…−19%), where the description reports "no signal". The new `dfs` also swaps the bounds-checked `arr[i] = root.value` for `arr.unsafe_set(n, v)`, which the "dominated by `dfs` recursion" analysis appears to understate. Worth reflecting in the final numbers table.

Caveat: these are single-round runs (the PR did 3–4). The large effects (`copy`, `Deque`) are unambiguous; the small `HashMap` effects sit near run-to-run noise but are directionally consistent.

No blockers from me.

PR #4170, opened 2026-08-31
docs(builtin): correct 15 inaccurate doc comments

bobzhang:
Corrects **15 inaccurate documentation comments** across 6 files. Documentation only — **no code, signature, or `.mbti` interface is changed**.

### What was wrong

| Count | Class | |
|---:|---|---|
| 7 | `contradicts-impl` | prose stated behavior the code does not have |
| 3 | `wrong-panic-contract` | documented panic/error behavior was wrong |
| 2 | `stale-reference` | reference to a renamed/removed item, or a dangling doc block |

### Representative fixes

**`Char::is_printable`** — `builtin/char.mbt` (`contradicts-impl`)

> **Was:** - Unassigned (Cn)
>
> **Now:** - Noncharacters (U+FDD0-U+FDEF and the U+nFFFE/U+nFFFF pairs)

**`Double::is_close`** — `builtin/double.mbt` (`contradicts-impl`)

> **Was:** relative or absolute tolerance. Returns `false` if either number is infinite.
>
> **Now:** relative or absolute tolerance. Returns `false` if the numbers are not exactly equal and either of them is infinite.

**`Double::abs`** — `builtin/double.mbt` (`contradicts-impl`)

> **Was:** Returns the absolute value of the input number. For any input `x`, the result is equivalent to `if x < 0.0 { -x } else { x }`.
>
> **Now:** Returns the absolute value of the input number by clearing its sign bit, so the result is never negative. In particular, `(-0.0).abs()` is `0.0`.

**`Int::abs`** — `builtin/int.mbt` (`contradicts-impl`)

> **Was:** Returns the absolute value of the integer.
>
> **Now:** Returns the absolute value of the integer. When the input is `@int.min_value` (-2147483648), returns `@int.min_value` itself, since its absolute value is not representable as an `Int`.

### Files

- `builtin/char.mbt`
- `builtin/double.mbt`
- `builtin/double_to_int64_js_wasm.mbt`
- `builtin/int.mbt`
- `builtin/int64.mbt`
- `builtin/intrinsics.mbt`

### Validation

| Check | Result |
|---|---|

### How this was produced

### DIFF
--- a/builtin/char.mbt
+++ b/builtin/char.mbt
@@ -348,7 +348,7 @@ pub fn Char::is_numeric(self : Self) -> Bool {
 /// - Format characters (Cf)
 /// - Line/paragraph separators (Zl, Zp)
 /// - Private use (Co)
-/// - Unassigned (Cn)
+/// - Noncharacters (U+FDD0-U+FDEF and the U+nFFFE/U+nFFFF pairs)
 /// - Surrogates (Cs)
 pub fn Char::is_printable(self : Self) -> Bool {
   // Check for control characters (Cc)
--- a/builtin/double.mbt
+++ b/builtin/double.mbt
@@ -55,8 +55,8 @@ pub fn Double::from_int(i : Int) -> Double {
 /// * `value` : The double-precision floating-point number to compute the
 /// absolute value of.
 ///
-/// Returns the absolute value of the input number. For any input `x`, the result
-/// is equivalent to `if x < 0.0 { -x } else { x }`.
+/// Returns the absolute value of the input number by clearing its sign bit, so
+/// the result is never negative. In particular, `(-0.0).abs()` is `0.0`.
 ///
 /// Example:
 ///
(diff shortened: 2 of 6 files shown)


bobzhang:

---

`builtin/int64.mbt:55-57` newly refers to `Int64::min_value()`, but no such method exists. The minimum is exposed by the `int64` package (`int64/int64.mbt:24-31`):

```mbt
pub const MIN_VALUE = -9223372036854775808L

#deprecated("Use `MIN_VALUE` instead")
pub let min_value = -9223372036854775808L
```

The edge-case behavior itself follows from `builtin/int64.mbt:68-73`:

```mbt
pub fn Int64::abs(self : Int64) -> Int64 {
  if self < 0L {
    -self
  } else {
    self
  }
}
```

Please replace both occurrences with `@int64.MIN_VALUE`. I verified the remaining changes against their bodies/intrinsics; in particular, the popcount is 14 and `0x4059000000000000` is the IEEE-754 encoding of `100.0`.

---

bobzhang:

---

Approved. The previously flagged `Int64::abs` issue is resolved: `@int64.MIN_VALUE` is the exported constant, and negating that value wraps to itself as documented. I also verified the remaining changes against their implementations, including the `Int64` bit examples, floating-point edge cases, clamp guards, Unicode noncharacters, byte construction, and zero-extension behavior; all are accurate improvements.

PR #4172, opened 2026-08-31
docs(builtin): correct 18 inaccurate doc comments

bobzhang:
Corrects **18 inaccurate documentation comments** across 9 files. Documentation only — **no code, signature, or `.mbti` interface is changed**.

### What was wrong

| Count | Class | |
|---:|---|---|
| 6 | `copy-paste-drift` | doc described a different function than the one it sits above |
| 4 | `stale-reference` | reference to a renamed/removed item, or a dangling doc block |
| 4 | `contradicts-impl` | prose stated behavior the code does not have |
| 1 | `wrong-example` | asserted value or comment in an example did not match the code |

### Representative fixes

**`FixedArray::blit_from_bytes`** — `builtin/bytes.mbt` (`contradicts-impl`)

> **Was:** Copy `length` chars from byte sequence `src`, starting at `src_offset`, into byte sequence `self`, starting at `bytes_offset`.
>
> **Now:** Copy `length` bytes from byte sequence `src`, starting at `src_offset`, into byte sequence `self`, starting at `bytes_offset`.

**`Iter::nth`** — `builtin/iterator.mbt` (`contradicts-impl`)

> **Was:** Returns the nth element of the iterator, or `None` if the iterator is shorter than `n` elements.
>
> **Now:** Returns the `n`-th element of the iterator, counting from zero, or `None` if `n` is negative or the iterator has fewer than `n + 1` elements. [

**`String::to_lower (and sibling StringView::to_lower / js variants at lines 2403, 2417, 2451)`** — `builtin/string_methods.mbt` (`contradicts-impl`)

> **Was:** Converts this string to lowercase.
>
> **Now:** Converts the ASCII uppercase letters (`'A'` to `'Z'`) in this string to lowercase. All other characters, including non-ASCII letters, are left unchanged. [applied to all three occurrences of the identical one-liner; range written as `'A'` to `'Z'` rather than `'A'..'Z'` to avoid reading as MoonBit r…

**`String::to_upper (and sibling StringView::to_upper at line 2515)`** — `builtin/string_methods.mbt` (`contradicts-impl`)

> **Was:** Converts this string to uppercase.
>
> **Now:** Converts the ASCII lowercase letters (`'a'` to `'z'`) in this string to uppercase. All other characters, including non-ASCII letters, are left unchanged. [applied to both occurrences of the identical one-liner]

### Files

- `builtin/bitstring.mbt`
- `builtin/bytes.mbt`
- `builtin/bytesview.mbt`
- `builtin/iterator.mbt`
- `builtin/string.mbt`
- `builtin/string_methods.mbt`
- `builtin/stringbuilder_buffer.mbt`
- `builtin/stringbuilder_concat.mbt`
- `builtin/stringview.mbt`

### Validation

| Check | Result |
|---|---|

### How this was produced

### DIFF
--- a/builtin/bitstring.mbt
+++ b/builtin/bitstring.mbt
@@ -218,7 +218,7 @@ fn ArrayView::unsafe_extract_uint_be_slow(
 /// # Invariants
 /// - It's guaranteed to have at least 5 bytes available for extraction
 /// - Only reads the necessary number of bytes based on the bit length (5-8 bytes)
-/// - For bit lengths < 33, use unsafe_extract_uint_le instead
+/// - For bit lengths [9..32], use unsafe_extract_uint_le instead
 ///
 #internal(experimental, "subject to breaking change without notice")
 #doc(hidden)
@@ -293,7 +293,7 @@ fn ArrayView::unsafe_extract_uint64_le_slow(
 /// # Invariants
 /// - It's guaranteed to have at least 5 bytes available for extraction
 /// - Only reads the necessary number of bytes based on the bit length (5-8 bytes)
-/// - For bit lengths < 33, use unsafe_extract_uint_be instead
+/// - For bit lengths [9..32], use unsafe_extract_uint_be instead
 ///
 #internal(experimental, "subject to breaking change without notice")
 #doc(hidden)
@@ -1156,7 +1156,7 @@ fn BytesView::unsafe_extract_uint_be_slow(
 /// # Invariants
 /// - It's guaranteed to have at least 5 bytes available for extraction
 /// - Only reads the necessary number of bytes based on the bit length (5-8 bytes)
-/// - For bit lengths < 33, use unsafe_extract_uint_le instead
+/// - For bit lengths [9..32], use unsafe_extract_uint_le instead
 ///
 #internal(experimental, "subject to breaking change without notice")
 #doc(hidden)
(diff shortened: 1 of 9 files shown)


bobzhang:

---

The four changes at `builtin/bitstring.mbt:221`, `:296`, `:1159`, and `:1234` still overstate the supported range. They say bit lengths `< 33` should use `unsafe_extract_uint_{le,be}`, but those functions only accept 9–32 bits.

For example, the ArrayView slow paths at `builtin/bitstring.mbt:125-149` and `:181-211` compute:

```mbt
let bytes_needed = (len + 7) / 8
```

Their match arms handle only 2, 3, or 4 bytes, ending with:

```mbt
_ => abort("Invalid byte count for int32 extraction")
```

The BytesView paths at `:1063-1087` and `:1119-1149` do the same. For lengths 1–8, `bytes_needed` is 1 and these helpers abort; one bit uses `unsafe_extract_bit`, while 2–8 bits use `unsafe_extract_byte`. These four comments should say “For bit lengths 9–32.”

I traced the remaining changed claims through their implementations and recomputed all affected example outputs; I found no other factual problems.

---

bobzhang:

---

Approved. The previously flagged `builtin/bitstring.mbt` issue is resolved: all four hints now correctly route only 9–32-bit extraction to `unsafe_extract_uint_{le,be}`; 1-bit and 2–8-bit extraction use `unsafe_extract_bit` and `unsafe_extract_byte`, respectively. I also rechecked the remaining bytes, view bounds, iterator, UTF-16/string case-conversion, and StringBuilder documentation against their implementations; all changed claims are accurate.

PR #4173, opened 2026-08-31
docs(builtin): correct 7 inaccurate doc comments

bobzhang:
Corrects **7 inaccurate documentation comments** across 5 files. Documentation only — **no code, signature, or `.mbti` interface is changed**.

### What was wrong

| Count | Class | |
|---:|---|---|
| 1 | `wrong-example` | asserted value or comment in an example did not match the code |

### Representative fixes

**`SourceLoc::repr`** — `builtin/autoloc.mbt` (`contradicts-impl`)

> **Was:** Returns a string representation of the source location, typically in the format "@package:file:start_line:start_column-end_line:end_column".
>
> **Now:** Returns a string representation of the source location, in the format "file:start_line:start_column-end_line:end_column@module".

**`ArgsLoc::to_json`** — `builtin/autoloc.mbt` (`contradicts-impl`)

> **Was:** representation. Each location in the array is either represented as a string if present, or "null" if absent. ... Returns a JSON array string where each element is either a string representation of a source location or "null".
>
> **Now:** representation. Each location in the array is either represented as a JSON object with `filename`, `start_line`, `start_column`, `end_line` and `end_column` fields if present, or "null" if absent. ... Returns a JSON array string where each element is either a JSON object describing a source location…

**`Hasher::Hasher`** — `builtin/hasher.mbt` (`contradicts-impl`)

> **Was:** * `seed` : An integer value used to initialize the hasher's internal state. Defaults to 0.
>
> **Now:** * `seed` : An integer value used to initialize the hasher's internal state. Defaults to 0 on every target except JavaScript, where the default is a randomly chosen per-process value.

**`InspectError`** — `builtin/console.mbt` (`wrong-example`)

> **Was:** inspect(x, content="42") // Raises InspectError with detailed failure message
>
> **Now:** inspect(x, content="42") // A mismatch would raise InspectError with a detailed message

### Files

- `builtin/LinkedHashMap.mbt.md`
- `builtin/autoloc.mbt`
- `builtin/console.mbt`
- `builtin/deprecated.mbt`
- `builtin/hasher.mbt`

### Validation

| Check | Result |
|---|---|

### How this was produced

### DIFF
--- a/builtin/LinkedHashMap.mbt.md
+++ b/builtin/LinkedHashMap.mbt.md
@@ -91,7 +91,7 @@ test {
 
 ### Size & Capacity
 
-You can use `size()` to get the number of key-value pairs in the map, or `capacity()` to get the current capacity.
+You can use `length()` to get the number of key-value pairs in the map, or `capacity()` to get the current capacity.
 
 ```mbt check
 ///|
--- a/builtin/autoloc.mbt
+++ b/builtin/autoloc.mbt
@@ -31,8 +31,8 @@ pub(all) type SourceLoc
 /// * `source_location` : A source code location containing information about the
 /// file path, line number, and column number.
 ///
-/// Returns a string representation of the source location, typically in the
-/// format "@package:file:start_line:start_column-end_line:end_column".
+/// Returns a string representation of the source location, in the format
+/// "file:start_line:start_column-end_line:end_column@module".
 ///
 /// Note: This function is primarily used internally by the compiler for error
 /// reporting and debugging purposes. Source locations are automatically created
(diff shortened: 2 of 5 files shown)


bobzhang:
## Codex review — ✅ approved

---

Approved. I verified the map length/capacity calculation by hand (three entries produce length 3 and capacity 4), traced `SourceLoc` parsing and `ArgsLoc` JSON serialization, checked the `inspect` mismatch branch, confirmed both `Byte` shift implementations and parameter names, and followed the target-gated hasher seed initialization. I also checked the affected examples: `inspect(42, content="42")` matches, and seeds 0 and 42 necessarily yield distinct hashes through the invertible rotate/odd-multiply/avalanche operations.

PR #4174, opened 2026-08-31
docs(numerics): correct 31 inaccurate doc comments

bobzhang:
Corrects **31 inaccurate documentation comments** across 15 files. Documentation only — **no code, signature, or `.mbti` interface is changed**.

### What was wrong

| Count | Class | |
|---:|---|---|
| 9 | `copy-paste-drift` | doc described a different function than the one it sits above |
| 6 | `contradicts-impl` | prose stated behavior the code does not have |
| 6 | `stale-readme` | README prose no longer matched the package |
| 5 | `language-error` | typo or broken markdown that changed meaning |

### Representative fixes

**`struct BigInt`** — `bigint/bigint_default.mbt` (`contradicts-impl`)

> **Was:** A big integer represented as an array of Int.
>
> **Now:** A big integer represented as an array of UInt limbs. (proposed fix applied unchanged; `limbs : FixedArray[UInt]` at line 39. Note the adjacent non-doc `//` design block still says "FixedArray of Int" — left alone because only `` lines may be edited)

**`const DECIMAL_RATIO`** — `bigint/bigint_default.mbt` (`contradicts-impl`)

> **Was:** The ratio of the number of decimal digits to the number of radix digits.
>
> **Now:** The ratio of the number of decimal digits to the number of bits. (proposed fix applied unchanged after independent verification: DECIMAL_RATIO = 0.302 = log10(2) = decimal digits per bit. In this file a radix digit is a limb — RADIX = 1UL << RADIX_BIT_LEN with RADIX_BIT_LEN = 32 — so decimal-per-lim…

**`Float::from_int`** — `float/methods.mbt` (`contradicts-impl`)

> **Was:** // Convert back to double for comparison since Float doesn't implement Show
>
> **Now:** // `Float` implements `Show`, so `f` could also be inspected directly [

**`Float::from_byte`** — `float/methods.mbt` (`contradicts-impl`)

> **Was:** // Convert to double for comparison since Float doesn't implement Show
>
> **Now:** // `Float` implements `Show`, so `f` could also be inspected directly [same substitution as Float::from_int]

### Files

- `bigint/bigint_default.mbt`
- `bigint/bigint_js.mbt`
- `double/README.mbt.md`
- `float/README.mbt.md`
- `float/methods.mbt`
- `int/README.mbt.md`
- `int16/int16.mbt`
- `int64/README.mbt.md`
- `math/log.mbt`
- `math/prime.mbt`
- `math/scalbn.mbt`
- `math/trig.mbt`
- `math/trig_double_js.mbt`
- `math/trig_double_nonjs.mbt`
- `uint/README.mbt.md`

### Validation

| Check | Result |
|---|---|

### How this was produced

### DIFF
--- a/bigint/bigint_default.mbt
+++ b/bigint/bigint_default.mbt
@@ -17,7 +17,7 @@
 // performance regression.
 
 ///|
-/// A big integer represented as an array of Int.
+/// A big integer represented as an array of UInt limbs.
 //
 // Design explained:
 // - Why use an FixedArray of Int with a len field instead of an Array[Int]?
@@ -93,7 +93,7 @@ const RADIX : UInt64 = 1UL << RADIX_BIT_LEN // TODO: This can be generalized onc
 const RADIX_MASK : UInt64 = RADIX - 1
 
 ///|
-/// The ratio of the number of decimal digits to the number of radix digits.
+/// The ratio of the number of decimal digits to the number of bits.
 const DECIMAL_RATIO : Double = 0.302 // log10(2)
 
 ///|
@@ -115,7 +115,7 @@ let one : BigInt = 1N
 ///
 /// Parameters:
 ///
-/// * `value` : The 32-bit signed integer (`Int`) to be converted.
+/// * `n` : The 32-bit signed integer (`Int`) to be converted.
 ///
 /// Returns a `BigInt` equivalent to the input integer.
 ///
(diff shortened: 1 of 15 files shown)


bobzhang:
## Codex review — ✅ approved

---

Approved. I reviewed the exact 15-file merge-base diff, traced every changed claim through the relevant implementations, helpers, and target-specific variants, and independently recomputed the numeric examples—including BigInt conversions/remainders, IEEE-754 results, and endian byte sequences. I found no factual errors or regressions.

PR #4175, opened 2026-08-31
docs(mutable collections): correct 21 inaccurate doc comments

bobzhang:
Corrects **21 inaccurate documentation comments** across 11 files. Documentation only — **no code, signature, or `.mbti` interface is changed**.

### What was wrong

| Count | Class | |
|---:|---|---|
| 7 | `stale-reference` | reference to a renamed/removed item, or a dangling doc block |
| 6 | `contradicts-impl` | prose stated behavior the code does not have |
| 5 | `stale-readme` | README prose no longer matched the package |
| 3 | `copy-paste-drift` | doc described a different function than the one it sits above |

### Representative fixes

**`Deque::flatten`** — `deque/deque.mbt` (`contradicts-impl`)

> **Was:** - Uses the first inner deque as base and appends subsequent deques.
>
> **Now:** - Allocates a new buffer of the combined length and copies every inner deque into it; none of the input deques are modified.

**`HashMap::capacity`** — `hashmap/utils.mbt` (`contradicts-impl`)

> **Was:** Returns the current capacity of the hash map. The capacity is the number of key-value pairs the hash map can hold before it needs to reallocate its internal storage.
>
> **Now:** Returns the current capacity of the hash map, that is, the length of its internal storage array. The map grows once it is half full, so it can hold at most `capacity / 2` key-value pairs before it reallocates.

**`PriorityQueue::copy`** — `priority_queue/README.mbt.md` (`contradicts-impl`)

> **Was:** `copy()` creates a shallow clone.
>
> **Now:** `copy()` returns a deep copy of the queue: the copy shares no heap nodes with the original, so modifying one does not affect the other. [

**`SortedMap::range`** — `sorted_map/map.mbt` (`contradicts-impl`)

> **Was:** Returns a new array of key-value pairs that are within the specified range [low, high].
>
> **Now:** Returns an iterator over the key-value pairs whose keys are within the specified range [low, high] (both bounds inclusive).

### Files

- `deque/deque.mbt`
- `hashmap/README.mbt.md`
- `hashmap/hashmap.mbt`
- `hashmap/utils.mbt`
- `hashset/README.mbt.md`
- `priority_queue/README.mbt.md`
- `queue/README.mbt.md`
- `set/README.mbt.md`
- `sorted_map/README.mbt.md`
- `sorted_map/map.mbt`
- `sorted_set/types.mbt`

### Validation

| Check | Result |
|---|---|

### How this was produced

### DIFF
--- a/deque/deque.mbt
+++ b/deque/deque.mbt
@@ -522,11 +522,6 @@ pub fn[A] Deque::remove(self : Deque[A], index : Int) -> A {
 }
 
 ///|
-/// Creates a new deque from a fixed array, preserving the order of elements.
-///
-/// Parameters:
-
-///|
 /// Returns the number of elements in the deque.
 ///
 /// Parameters:
@@ -702,25 +697,6 @@ test "unsafe_pop_front after many push_front" {
 }
 
 ///|
-/// Removes and discards the first element from the deque. This function is a
-/// deprecated version of `unsafe_pop_front`.
-///
-/// Parameters:
-///
-/// * `self` : The deque to remove the first element from.
-///
-/// Throws a runtime error if the deque is empty.
-///
-/// Example:
-///
-/// ```mbt test
-///   let dq = @deque.from_array([1, 2, 3])
-///   dq.unsafe_pop_front()
-///   inspect(dq, content="@deque.from_array([2, 3])")
-/// ```
-///
-
-///|
 /// Removes a back element from a deque.
 ///
 /// # Example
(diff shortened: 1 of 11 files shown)


bobzhang:

---

Blocking: `priority_queue/README.mbt.md:150` calls `copy()` a deep copy and claims modifying one queue cannot affect the other, but `copy_node` directly reuses every stored value (`priority_queue/priority_queue.mbt:57-78`):

```mbt
content: root.content
// ...
content: child.content
// ...
content: sib.content
```

`peek()` likewise returns that same `content` (`priority_queue/priority_queue.mbt:281-285`). Since `A` can be mutable—for example, `Array[Int]` implements `Compare` at `builtin/array.mbt:343`—both queues can retain the same mutable element, and mutation through one is observable through the other. The heap nodes are independent, but the elements are shallow-copied, so the replaced “shallow clone” was more accurate.

---

bobzhang:

---

Approved. The previously flagged `priority_queue` issue is resolved: `copy()` recreates every heap node, making push/pop operations independent, while retaining the same stored element values/references, so mutable elements remain shared.

I also verified the constructor/API names, HashMap capacity threshold, deque flattening, sorted-map range/copy behavior, builtin `Map` ordering, `Char`/`Bool` hashing, sorted-set mutation semantics, and orphan-doc removals. All changed documentation is factually correct.

PR #4176, opened 2026-08-31
docs(immutable collections): correct 34 inaccurate doc comments

bobzhang:
Corrects **34 inaccurate documentation comments** across 19 files. Documentation only — **no code, signature, or `.mbti` interface is changed**.

### What was wrong

| Count | Class | |
|---:|---|---|
| 10 | `contradicts-impl` | prose stated behavior the code does not have |
| 9 | `stale-reference` | reference to a renamed/removed item, or a dangling doc block |
| 6 | `stale-readme` | README prose no longer matched the package |
| 2 | `wrong-panic-contract` | documented panic/error behavior was wrong |
| 2 | `copy-paste-drift` | doc described a different function than the one it sits above |

### Representative fixes

**`Path::is_last`** — `immut/internal/path/path.mbt` (`contradicts-impl`)

> **Was:** Returns if the path contains only a single segment.
>
> **Now:** Returns if at most one index segment remains in the path. This also holds for the exhausted path, which has no segment left.

**`SparseArray::difference`** — `immut/internal/sparse_array/sparse_array.mbt` (`contradicts-impl`)

> **Was:** Keep indices and values only present in self but not in other.
>
> **Now:** Keep every index of self: an index absent from other keeps its value, an index present in both keeps f's result, or is dropped when f returns None.

**`SortedMap::fold / rev_fold (README prose)`** — `immut/sorted_map/README.mbt.md` (`contradicts-impl`)

> **Was:** Use `fold()` to fold over the key-value pairs of the map. The default order is Pre-order; use `rev_fold()` for a Post-order fold.
>
> **Now:** Use `fold()` to fold over the key-value pairs of the map. The default order is ascending by key; use `rev_fold()` to fold in descending key order.

**`SortedMap::fold`** — `immut/sorted_map/utils.mbt` (`contradicts-impl`)

> **Was:** Pre-order fold.
>
> **Now:** Fold over the key-value pairs in ascending key order.

### Files

- `immut/hashmap/HAMT.mbt`
- `immut/hashmap/types.mbt`
- `immut/hashset/README.mbt.md`
- `immut/hashset/types.mbt`
- `immut/internal/path/extends.mbt`
- `immut/internal/path/path.mbt`
- `immut/internal/sparse_array/sparse_array.mbt`
- `immut/priority_queue/README.mbt.md`
- `immut/priority_queue/priority_queue.mbt`
- `immut/sorted_map/README.mbt.md`
- `immut/sorted_map/utils.mbt`
- `immut/sorted_set/README.mbt.md`
- `immut/sorted_set/immutable_set.mbt`
- `immut/sorted_set/types.mbt`
- `immut/vector/deprecated.mbt`
- `immut/vector/tree.mbt`
- `lazy_list/README.mbt.md`
- `list/README.mbt.md`
- `list/list.mbt`

### Validation

| Check | Result |
|---|---|

### How this was produced

### DIFF
--- a/immut/hashmap/HAMT.mbt
+++ b/immut/hashmap/HAMT.mbt
@@ -79,6 +79,8 @@ pub fn[K : Eq + Hash, V] HashMap::get(self : HashMap[K, V], key : K) -> V? {
 
 ///|
 /// Get value with `at` access semantics.
+/// Aborts if the key is not present; use `get` for the `Option`-returning
+/// version.
 #alias("_[_]")
 pub fn[K : Eq + Hash, V] HashMap::at(self : HashMap[K, V], key : K) -> V {
   guard! self.data is Some(node)
--- a/immut/hashmap/types.mbt
+++ b/immut/hashmap/types.mbt
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 ///|
-/// An non-empty immutable hash set data structure
+/// A non-empty immutable hash map data structure
 #unsafe_cycle_free
 priv enum Node[K, V] {
   /// a subtree holding exactly one entry, at any depth; carries the
--- a/immut/hashset/README.mbt.md
+++ b/immut/hashset/README.mbt.md
@@ -18,7 +18,7 @@ test "creating immutable sets" {
   let from_array_result = @hashset.HashSet([1, 2, 3, 2, 1]) // Duplicates removed
   inspect(from_array_result.length(), content="3")
 
-  // From fixed array
+  // From an array of distinct values
   let from_fixed = @hashset.HashSet([10, 20, 30])
   inspect(from_fixed.length(), content="3")
 
--- a/immut/hashset/types.mbt
+++ b/immut/hashset/types.mbt
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 ///|
-/// An non-empty immutable hash set data structure
+/// A non-empty immutable hash set data structure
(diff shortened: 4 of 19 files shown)


bobzhang:

---

### Blocking

`immut/sorted_set/types.mbt:20` claims that one child’s size is always at most five times the other’s. This is false for any valid two-element set, whose child sizes are `1` and `0`.

The implementation explicitly permits that base case in `immut/sorted_set/immutable_set.mbt:1093`:

```mbt
let left_size = left.length()
let right_size = right.length()
if left_size + right_size < 2 {
  create(left, value, right)
} else if left_size > right_size * BALANCE_RATIO {
```

With `BALANCE_RATIO = 5` (`immutable_set.mbt:1018`), child sizes `(1, 0)` violate the documented claim because `1 > 5 * 0`. Please document the small-tree exception. I verified the remaining changed claims against their implementations and hand-evaluated the affected examples; they are correct.

---

bobzhang:

---

Approved. The previously flagged `immut/sorted_set/types.mbt` issue is resolved: `BALANCE_RATIO` is 5, and `balance` special-cases `left_size + right_size < 2`, so the documented greater-than-two-element rule, strict “exceeds 5 times” threshold, and two-element 1/0 exception match the implementation.

I also rechecked the remaining API names, failure semantics, traversal orders, vector invariants, list behavior, and README corrections against their implementations; all changed lines are accurate improvements.

PR #4179, opened 2026-08-31
docs(quickcheck, splitmix & random docs): correct 8 inaccurate doc comments

bobzhang:
Corrects **8 inaccurate documentation comments** across 4 files. Documentation only — **no code, signature, or `.mbti` interface is changed**.

### What was wrong

| Count | Class | |
|---:|---|---|
| 2 | `copy-paste-drift` | doc described a different function than the one it sits above |
| 1 | `stale-readme` | README prose no longer matched the package |

### Representative fixes

**`RandomState::next_float`** — `quickcheck/splitmix/random.mbt` (`contradicts-impl`)

> **Was:** Get the next random number as a float in [0, 1]
>
> **Now:** Get the next random number as a float in [0, 1)

**`RandomState::next_double`** — `quickcheck/splitmix/random.mbt` (`contradicts-impl`)

> **Was:** Get the next random number as a double in [0, 1]
>
> **Now:** Get the next random number as a double in [0, 1)

**`Rand::bigint (summary line)`** — `random/random.mbt` (`contradicts-impl`)

> **Was:** Generates a random positive `BigInt` with a specified number of bits.
>
> **Now:** Generates a random non-negative `BigInt` with a specified number of bits.

**`int_range`** — `quickcheck/generator.mbt` (`wrong-panic-contract`)

> **Was:** Generates an integer in the half-open interval `[lower, upper)`.
>
> **Now:** Generates an integer in the half-open interval `[lower, upper)`.

### Files

- `quickcheck/generator.mbt`
- `quickcheck/splitmix/README.mbt.md`
- `quickcheck/splitmix/random.mbt`
- `random/random.mbt`

### Validation

| Check | Result |
|---|---|

### How this was produced

### DIFF
--- a/quickcheck/generator.mbt
+++ b/quickcheck/generator.mbt
@@ -162,6 +162,9 @@ pub fn[T] Generator::resize(self : Generator[T], size : Int) -> Generator[T] {
 
 ///|
 /// Generates an integer in the half-open interval `[lower, upper)`.
+///
+/// The degenerate range where `lower == upper` always yields `lower`. Aborts
+/// if `lower` is greater than `upper`.
 pub fn int_range(lower : Int, upper : Int) -> Generator[Int] {
   guard lower < upper else {
     if lower == upper {
--- a/quickcheck/splitmix/README.mbt.md
+++ b/quickcheck/splitmix/README.mbt.md
@@ -282,7 +282,7 @@ SplitMix provides:
 ## Performance Characteristics
 
 - **Generation speed**: Very fast (few CPU cycles per number)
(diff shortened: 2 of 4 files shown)


bobzhang:
## Codex review — ✅ approved

---

Approved. I verified `int_range`’s equality and error branches; confirmed SplitMix stores `seed` and `gamma`, returns `UInt` halves, and computes maxima of `1 - 2^-24`/`1 - 2^-53`; and traced `Rand::new` to `&Source` plus `bigint` to unsigned octets. The 8-bit example is guaranteed because its single byte is in `0..255`, so `bit_length() <= 8`.

PR #4180, opened 2026-08-31
docs(tooling & small packages): correct 14 inaccurate doc comments

bobzhang:
Corrects **14 inaccurate documentation comments** across 10 files. Documentation only — **no code, signature, or `.mbti` interface is changed**.

### What was wrong

| Count | Class | |
|---:|---|---|
| 8 | `stale-reference` | reference to a renamed/removed item, or a dangling doc block |
| 1 | `copy-paste-drift` | doc described a different function than the one it sits above |

### Representative fixes

**`OptionAction::Set`** — `argparse/arg_spec.mbt` (`contradicts-impl`)

> **Was:** - `Set` keeps the last provided value.
>
> **Now:** - `Set` accepts the value once; a repeated occurrence is an error. (proposed fix accepted verbatim; verified there is no code path where a later occurrence overwrites an earlier one — `check_duplicate_set_occurrence` in parser_values.mbt:156-168 raises on the second argv occurrence for both long (pa…

**`FlagArg::FlagArg`** — `argparse/arg_spec.mbt` (`contradicts-impl`)

> **Was:** - If `env` is set, accepted boolean values are: `1`, `0`, `true`, `false`, `yes`, `no`, `on`, `off`.
>
> **Now:** - If `env` is set on a `SetTrue` / `SetFalse` flag, accepted boolean values are: `1`, `0`, `true`, `false`, `yes`, `no`, `on`, `off`. - If `env` is set on a `Count` flag, the value must be a non-negative integer.

**`cmp.minmax_by_key`** — `cmp/cmp.mbt` (`contradicts-impl`)

> **Was:** Returns the minimum and maximum of two values based on a comparison function.
>
> **Now:** Returns the minimum and maximum of two values based on a key extracted from each of them. (Applied auditor's fix verbatim.

**`Eq for V128 with fn not_equal`** — `v128/simd_basic.mbt` (`copy-paste-drift`)

> **Was:** Compares two `V128` values for bitwise equality across both 64-bit words.
>
> **Now:** Compares two `V128` values for bitwise inequality across both 64-bit words. (Applied the auditor's proposed fix verbatim; verified against the body `lo(self) != lo(other) || hi(self) != hi(other)`, which returns true when either 64-bit word differs.)

### Files

- `argparse/arg_spec.mbt`
- `cmp/README.mbt.md`
- `cmp/cmp.mbt`
- `env/README.mbt.md`
- `error/README.mbt.md`
- `option/README.mbt.md`
- `prelude/README.mbt.md`
- `range/range.mbt`
- `result/README.mbt.md`
- `v128/simd_basic.mbt`

### Validation

| Check | Result |
|---|---|

### How this was produced

### DIFF
--- a/argparse/arg_spec.mbt
+++ b/argparse/arg_spec.mbt
@@ -29,7 +29,7 @@ pub(all) enum FlagAction {
 ///|
 /// Behavior for option args.
 ///
-/// - `Set` keeps the last provided value.
+/// - `Set` accepts the value once; a repeated occurrence is an error.
 /// - `Append` keeps all provided values in order.
 pub(all) enum OptionAction {
   Set
@@ -89,8 +89,10 @@ pub struct FlagArg {
 /// - At least one of `short`, `long`, or `env` must be available.
 /// - `global=true` makes the flag available in subcommands.
 /// - `negatable=true` accepts `--no-<long>` for long flags.
-/// - If `env` is set, accepted boolean values are:
-///   `1`, `0`, `true`, `false`, `yes`, `no`, `on`, `off`.
+/// - If `env` is set on a `SetTrue` / `SetFalse` flag, accepted boolean
+///   values are: `1`, `0`, `true`, `false`, `yes`, `no`, `on`, `off`.
+/// - If `env` is set on a `Count` flag, the value must be a non-negative
+///   integer.
 #alias(new, deprecated="Use `FlagArg()` instead")
 pub fn FlagArg::FlagArg(
   name : StringView,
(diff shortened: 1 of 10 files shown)


bobzhang:
## Codex review — ✅ approved

---

Approved. I checked every changed claim against its implementation, including argparse duplicate handling and environment parsing, cmp key extraction, JavaScript `Date.now()`, error cleanup guidance, deprecated Option/Result/prelude APIs, range iteration, and V128 inequality. I also manually recomputed the range outputs and the Int/Int64 minimum-value shrink sequences; all counts and asserted values match.

PR #4182, opened 2026-09-01
perf(json): format integral numbers without Ryu (2.25x on numeric documents)

bobzhang:
`Json::stringify` routed every number through `Show for Double` (Ryu). Integral values in `Int` range — the overwhelming majority of numbers in real JSON — are now formatted as integers instead.

```mbt
if n >= -2147483648.0 && n <= 2147483647.0 && n.to_int().to_double() == n {
  buf.write_object(n.to_int())
} else {
  buf.write_object(n)
}
```

### Why this path

Profiling `stringify` by payload type showed where the time actually goes — same 10,000 elements, native release:

| 10,000 elements of | Time |
|---|---|
| numbers | 345.22 µs |
| strings | 104.02 µs |
| booleans | 55.02 µs |

**~84% of the cost on a numeric document was double→string formatting**, not structure. Ryu's `d2d_small_int` path helps, but integral values still go through generic decimal emission in `to_chars`.

### Results

native, `--release`:

| Benchmark | before | after | Δ |
|---|---|---|---|
| numbers n=10000 | 345.22 ± 8.86 µs | **153.45 ± 2.77 µs** | **−55.6% (2.25×)** |
| `stringify` compact n=10000 | 345.86 ± 9.09 µs | **159.65 ± 2.21 µs** | −53.8% |
| `stringify` deep wide indent=2 n=10000 | 438.86 ± 7.01 µs | **257.47 ± 4.23 µs** | −41.3% |
| shallow wide indent=2 n=500 | 165.93 ± 1.35 µs | 156.50 ± 2.09 µs | −5.7% |

Strings (104.02 → 106.02 µs) and booleans (55.02 → 56.02 µs) are unchanged within noise — no regression on non-numeric payloads.

### Output is unchanged

The rendering of a JSON number is by definition the shortest round-tripping decimal form of the double, i.e. `Double::to_string`. `Int` formatting agrees with it on exactly the values the guard admits.

`json/stringify_number_test.mbt` asserts that equality directly — the oracle is `d.to_string()`, taken from the definition rather than copied from the implementation:

- **Exhaustive** over `-2000..=2000`, where the fast path is always taken
- **Both sides of the range boundary**: 2147483646/7/8/9, −2147483647/8/9/50, 2³², 2⁵³, 1e15/1e16/1e17
- **Non-integral values that must not take the path**, including `2147483646.5` whose truncation *is* in range, plus subnormals and `Double::max_value`
- **Inside a document**, so the path is exercised through the real stringify loop rather than only at the root

`-0.0` takes the integer path and prints `0`, which is what `Double::to_string(-0.0)` already produced.

**Mutation-checked:** removing the integrality guard breaks 10 tests (`123.45` renders as `123`).

### Verification

|---|---|---|---|---|
| `moon test` | 7555 ✓ | 7585 ✓ | 7525 ✓ | 7470 ✓ |

`moon check --deny-warn --target all` clean. `moon info` shows no `.mbti` diff — no public API change.

Contained entirely to the `json` package; `builtin`'s `Double` formatting is untouched.

### DIFF
--- a/json/json.mbt
+++ b/json/json.mbt
@@ -312,7 +312,16 @@ pub fn Json::stringify(
           }
           Number(n, repr~) =>
             match repr {
-              None => buf.write_object(n)
+              None =>
+                // Integral values in Int range are the overwhelming majority of
+                // JSON numbers; formatting them as integers skips Ryu entirely.
+                if n >= -2147483648.0 &&
+                  n <= 2147483647.0 &&
+                  n.to_int().to_double() == n {
+                  buf.write_object(n.to_int())
+                } else {
+                  buf.write_object(n)
+                }
               Some(r) => buf.write_string(r)
             }
           True => buf.write_string("true")
--- a/json/stringify_number_test.mbt
+++ b/json/stringify_number_test.mbt
@@ -1,0 +1,86 @@
+// Copyright 2026 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+///|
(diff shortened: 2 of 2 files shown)


bobzhang:
this belongs to ryu

bobzhang:
Superseded by the `builtin`-level version of the same fast path, which delivers the identical JSON win (152.22 vs 153.45 µs) while also speeding up every other `Double::to_string` caller 3.46×. Closing in its favour.

PR #4183, opened 2026-09-01
perf(json): fill indent cache incrementally, plus oracle test and common-path benchmarks

bobzhang:
Follow-up to #4145, which added the depth-indexed indentation cache. Three small things on top of it.

### 1. Fill the cache incrementally

```diff
- cache.push("\n" + " ".repeat(indent * cache.length()))
+ match cache.last() {
+   None => cache.push("\n")
+   Some(prev) => cache.push(prev + unit)
+ }
```

Each level is the previous level plus one indentation unit. Building a level is **still O(indent × level)** — the concatenation copies `last` — but it replaces a fresh `repeat(indent * level)` with a `repeat(indent)`, roughly halving the bytes written per level. (An earlier revision of this description claimed an asymptotic improvement; that was wrong, as Codex pointed out in review.)

The unit is built inside `write_indent` rather than once per `stringify` call, so a document that emits no indentation never allocates it.

It also removes the `indent * level` product. That matters because `String::repeat` **aborts on a negative count** (`builtin/string_methods.mbt:1606`), so an overflowing product yields `abort("negative repeat count")` — or, if it wraps to a small positive, silently emits the wrong indentation width. Reaching this needs an extreme `indent × depth`, so it is a robustness cleanup rather than a live bug.

**I want to be straight about the performance: it is not a measurable win.**

| Benchmark (native, release) | before | after | Δ |
|---|---|---|---|
| deep wide indent=2 n=10000 | 438.86 ± 7.01 µs | 428.88 ± 7.08 µs | −2.3% |
| compact n=10000 | 345.86 ± 9.09 µs | 341.59 ± 5.58 µs | −1.2% |
| shallow indent=2 n=500 | 165.93 ± 1.35 µs | 164.28 ± 4.02 µs | −1.0% |
| shallow compact n=500 | 140.55 ± 1.72 µs | 137.87 ± 1.85 µs | −1.9% |

Directionally consistent across all four, but every delta is within ~1–2σ. The reason is structural: that workload is ~84% double→string formatting (see the companion PR), so indentation changes cannot move it much. **If you'd rather drop this hunk and take only the tests and benchmarks below, that is a reasonable call.**

### 2. An oracle-driven regression test

`json/stringify_indent_test.mbt` checks indentation against `indent * level` computed straight from the definition, rather than against anything the implementation produces.

Exhaustive over indent widths 0..=6 crossed with nesting depths 0..=12 — spanning **both sides of the 8-space lookup cutoff** that the pre-cache implementation special-cased. The existing snapshots cover indent 1/2/4/5/7 at shallow depth and never reach past it, which is exactly where the cache changed behavior. A second test covers a document that returns to shallower levels after going deep, guarding the lazy fill against stale entries.

Mutation-checked: an off-by-one in the fill breaks 7 tests.

### 3. Benchmarks for the paths that were unguarded

The benchmark added in #4145 wraps a 10,000-element array in six singleton arrays, placing every separator at depth 7 — past the old fast path. That measures the best case and nothing else. Added:

- **compact path** (`indent=0`) — the default, and by far the most common
- **realistic shallow document** — 500 small objects nested four deep, like ordinary API/config output

so a future regression on the common paths is visible rather than silent.

### Verification

|---|---|---|---|---|
| `moon test` | 7553 ✓ | 7583 ✓ | 7523 ✓ | 7468 ✓ |

`moon check --deny-warn --target all` clean. `moon info` shows no `.mbti` diff.

Independent of the companion number-formatting PR — either can merge first.

### DIFF
--- a/json/json.mbt
+++ b/json/json.mbt
@@ -91,15 +91,19 @@ fn write_indent(
   buf : StringBuilder,
   cache : Array[String],
   level : Int,
-  unit : String,
+  indent : Int,
 ) -> Unit {
   // Each level is the previous one plus a single indentation unit, so filling
   // the cache costs O(indent) per new level rather than O(indent * level), and
   // never computes `indent * level` — a product that overflows `Int` to a
   // negative value for a large `indent`, which `String::repeat` rejects.
+  //
+  // The unit is built here rather than once per `stringify` call so that a
+  // document emitting no indentation at all never allocates it: callers reach
+  // this only when `indent > 0`, so `repeat` never sees a negative count.
   while cache.length() <= level {
     match cache {
-      [.., last] => cache.push(last + unit)
+      [.., last] => cache.push(last + " ".repeat(indent))
       [] => cache.push("\n")
     }
   }
@@ -281,7 +285,6 @@ pub fn Json::stringify(
 ) -> String {
   let buf = StringBuilder(size_hint=0)
   let indent_cache : Array[String] = []
-  let indent_unit = " ".repeat(indent)
 
   // Explicit stack to replace recursive calls
   let stack : Array[WriteFrame] = []
@@ -297,7 +300,7 @@ pub fn Json::stringify(
               depth += 1
               buf.write_char('{')
               if indent > 0 {
-                write_indent(buf, indent_cache, depth, indent_unit)
+                write_indent(buf, indent_cache, depth, indent)
               }
               // After child value printed, we resume from this frame
               stack.push(Object(members.iter(), first=true))
@@ -309,7 +312,7 @@ pub fn Json::stringify(
               depth += 1
               buf.write_char('[')
               if indent > 0 {
-                write_indent(buf, indent_cache, depth, indent_unit)
+                write_indent(buf, indent_cache, depth, indent)
               }
               stack.push(Array(arr, i=0))
             }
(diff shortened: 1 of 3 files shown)


bobzhang:
## Codex review — bug hunt at `xhigh`

*Automated review by Codex CLI (`gpt-5.6-sol`, `xhigh` reasoning), run read-only.*

**Round 1 — one blocking bug, confirmed and fixed.**

> `json/json.mbt:284` constructs `indent_unit` before any `indent > 0` check. On `origin/main`, negative indentation takes the compact path because every `write_indent` call is guarded; now `Json::stringify(indent=-1)` aborts because `String::repeat` rejects negative counts.

Reproduced before fixing: on this branch `stringify(indent=-1)` aborted, on `main` it did not. Fixed in `bbb81ddd` by building the unit inside `write_indent` (reached only when `indent > 0`), with regression tests for non-positive indent and for a huge indent on documents that emit none. Both tests were mutation-checked — they fail against the eager version.

**Round 2 — no bugs.**

> The previously flagged runtime bug is genuinely fixed, and I found no remaining functional safety issue. The runtime behavior is otherwise correct, including non-positive indentation, level zero, array-pattern matching, and cache indexing.

Codex's one remaining note was that the code comment overstated the complexity improvement — concatenation copies `last`, so a level is still O(indent × level). Correct, and fixed in `223e6b65`; the PR description was corrected too.

Verified on all four backends: wasm 7555, wasm-gc 7585, js 7525, native 7470 — all passing. `moon check --deny-warn --target all` clean, no `.mbti` drift.

PR #4184, opened 2026-09-01
perf(builtin): format integral doubles without the Ryu search (3.46x)

bobzhang:
Supersedes #4182 by moving the integer fast path to where the cost actually is: `ryu_to_string` itself, so every caller of `Double::to_string` benefits, not just `Json::stringify`.

```mbt
const MAX_EXACTLY_REPRESENTABLE_INT : Double = 9_007_199_254_740_992 // 2^53

if val >= -MAX_EXACTLY_REPRESENTABLE_INT && val <= MAX_EXACTLY_REPRESENTABLE_INT {
  if val >= -2147483648.0 && val <= 2147483647.0 {
    let i = val.to_int()
    if i.to_double() == val { return i.to_string() }
  } else {
    let i = val.to_int64()
    if i.to_double() == val { return i.to_string() }
  }
}
```

Placed after the existing `val == 0.0` check in `builtin/double_ryu_nonjs.mbt`. Two tiers: the dominant small-integer band takes the cheaper `Int` formatting; the band up to 2⁵³ — millisecond timestamps, large IDs — takes `Int64`.

**Why the bound is 2⁵³ and cannot be wider:** below 2⁵³ every integer is exactly representable, so no shorter decimal can round-trip to it and the shortest form IS the exact digits. Above it, doubles are spaced 2+ apart and Ryu legitimately shortens: 2⁶² renders as `4611686018427388000`, not its exact `...87904`. The wbtest pins 2⁶² and 2⁶⁰ so a widened bound fails tests rather than silently changing output. (The name deliberately isn't `MAX_SAFE_INTEGER` — JS defines that as 2⁵³ − 1, and this bound includes 2⁵³ itself.)

### Results (native, `--release`)

| | before | after | Δ |
|---|---|---|---|
| `Double::to_string`, 10k small integral values | 311.14 µs | **93.20 µs** | **3.34×** |
| `Double::to_string`, 10k ms-timestamps (~1.7e12) | 350.81 µs | **158.17 µs** | **2.22×** |
| `Json::stringify`, 10k-number document | 345.22 µs | **152.22 µs** | 2.27× |
| `Double::to_string`, 10k fractional values | 575.59 µs | 486.32 µs | not regressed¹ |

¹ Fractional values pay only the extra range checks; the measured improvement is likely code-layout noise and is not claimed.

This matches the JSON-local prototype in #4182 (153.45 µs) while also reaching `Show for Double`, string interpolation, `@debug`, `Buffer` — every double rendering on wasm/wasm-gc/native.

**JS is unaffected**: `ryu_to_string` on js is `number.toString()`, already integer-fast.

### Edge cases, explicitly

| Value | Path | Why |
|---|---|---|
| `NaN` | full algorithm | fails both range comparisons (all comparisons with NaN are false) |
| `±Infinity` | full algorithm | fails the range comparisons |
| `-0.0` | never reached | the existing `val == 0.0` check above returns `"0"` first |
| `2147483647.0` / `-2147483648.0` | Int tier | exactly representable, in range |
| `2147483648.0` / `-2147483649.0` | Int64 tier | between the tiers |
| `±9007199254740992.0` (2⁵³) | Int64 tier | inclusive bound, exactly representable |
| `±9007199254740994.0` (2⁵³+2) | full algorithm | outside the guard |
| `2147483646.5`, `1700000000000.5` | full algorithm | truncation in range, but `i.to_double() == val` fails |
| `0.9999999999999999`, `1.0000000000000002` | full algorithm | nearest doubles to 1.0; integrality fails |
| subnormals, `5e-324` | full algorithm | integrality fails |
| 2⁶⁰, 2⁶² | full algorithm | shortest form ≠ exact digits — pinned so the bound cannot widen |

### Correctness evidence

**Differential checksum** over ~330,000 values — dense band ±100k, power-of-two neighbourhoods ±2 through 2⁵⁶ in both signs, powers of ten (trailing-zero mantissas), 30k pseudo-random 53-bit integral/fractional/quotient values:

```
main:     LEN=2913311 HASH=-1997157211
patched:  LEN=2913311 HASH=-1997157211
```

Byte-identical rendering.

**New wbtest** (`builtin/double_ryu_int_wbtest.mbt`): oracle agreement with `Int::to_string` over a dense band, pinned boundaries on both sides of each tier and of 2⁵³, special values, nearest-to-integral doubles, tier-2 fractional values, above-bound values whose shortest form is shortened, and power-of-two neighbourhoods for p in 0..=53 in both signs.

**Mutation-checked three ways**: removing the tier-1 integrality test breaks 41 tests; removing the tier-2 integrality test breaks 1; widening the bound to 2⁶² breaks 2.

### Verification

|---|---|---|---|---|
| `moon test` | 7562 ✓ | 7592 ✓ | 7532 ✓ | 7477 ✓ |

`moon check --deny-warn --target all` clean; no `.mbti` drift (no public API change).

One design note: `Double::to_string` carries `#intrinsic("%f64.to_string")`, so on backends where the compiler substitutes the intrinsic this body is bypassed entirely — the change simply doesn't apply there, it cannot diverge.

### DIFF
--- a/builtin/double_ryu_int_wbtest.mbt
+++ b/builtin/double_ryu_int_wbtest.mbt
@@ -30,7 +30,7 @@ test "ryu integral fast path at the Int range boundary" {
   inspect(ryu_to_string(2147483647.0), content="2147483647")
   inspect(ryu_to_string(-2147483647.0), content="-2147483647")
   inspect(ryu_to_string(-2147483648.0), content="-2147483648")
-  // Just outside: must fall back to the full algorithm, same rendering.
+  // Just outside Int range: now served by the Int64 tier, same rendering.
   inspect(ryu_to_string(2147483648.0), content="2147483648")
   inspect(ryu_to_string(2147483649.0), content="2147483649")
   inspect(ryu_to_string(-2147483649.0), content="-2147483649")
--- a/builtin/double_ryu_int_wbtest.mbt
+++ b/builtin/double_ryu_int_wbtest.mbt
@@ -1,0 +1,72 @@
+// Copyright 2026 International Digital Economy Academy
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+///|
+/// The integral fast path must be indistinguishable from the full algorithm.
+/// The oracle is `Int::to_string`, which is the definition of the shortest
+/// decimal form for an integer; for values outside the fast path the expected
+/// strings are pinned so a widened guard cannot silently change them.
+test "ryu integral fast path agrees with Int::to_string over a dense band" {
+  for i in -3000..<=3000 {
+    assert_true(ryu_to_string(i.to_double()) == i.to_string())
+  }
+}
+
+///|
+test "ryu integral fast path at the Int range boundary" {
+  // Inside the guard, both signs.
+  inspect(ryu_to_string(2147483646.0), content="2147483646")
+  inspect(ryu_to_string(2147483647.0), content="2147483647")
+  inspect(ryu_to_string(-2147483647.0), content="-2147483647")
+  inspect(ryu_to_string(-2147483648.0), content="-2147483648")
+  // Just outside: must fall back to the full algorithm, same rendering.
+  inspect(ryu_to_string(2147483648.0), content="2147483648")
+  inspect(ryu_to_string(2147483649.0), content="2147483649")
+  inspect(ryu_to_string(-2147483649.0), content="-2147483649")
+  inspect(ryu_to_string(-2147483650.0), content="-2147483650")
+}
+
+///|
+test "ryu integral fast path never captures special or fractional values" {
+  // -0.0 is handled by the existing zero check before the fast path.
+  inspect(ryu_to_string(-0.0), content="0")
+  inspect(ryu_to_string(0.0 / 0.0), content="NaN")
+  inspect(ryu_to_string(1.0 / 0.0), content="Infinity")
+  inspect(ryu_to_string(-1.0 / 0.0), content="-Infinity")
+  // Fractional values whose truncation is in range must not take the path.
+  inspect(ryu_to_string(0.5), content="0.5")
+  inspect(ryu_to_string(-0.5), content="-0.5")
+  inspect(ryu_to_string(2147483646.5), content="2147483646.5")
+  inspect(ryu_to_string(-2147483647.5), content="-2147483647.5")
+  // The next representable double below 1.0 and above 1.0: integrality test
+  // must reject both.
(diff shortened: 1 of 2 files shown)


bobzhang:
## Codex review — ✅ approved (edge-case hunt at `xhigh`)

*Automated review by Codex CLI (`gpt-5.6-sol`, `xhigh` reasoning), run read-only, prompted specifically to attack the floating-point edge cases.*

---

Approved — no correctness defects found.

- NaN is rejected by the direct `%f64.ge`/`%f64.le` comparisons, and each infinity fails one range bound.
- Wasm/Wasm-GC use saturating `i32.trunc_sat_f64_s`; native/LLVM guard before their unchecked conversion. The exact inclusive bounds make conversion safe.
- `-0.0` is caught by the existing zero check. Even if reordered, it would convert to `0`, compare equal to `0.0`, and still render `"0"`.
- Fractional neighbors and subnormals truncate but fail the exact `Int → Double` round trip, so they fall through. Existing coverage pins `5e-324`.
- Scientific notation starts at decimal exponent 21, while `Int` values reach only exponent 9, making every fast-path rendering identical to the full algorithm.
- `ryu_to_string` is private and pure; no caller relies on executing `d2d_small_int`.
- I checked all 6,369 JavaScript cases represented by the new tests against `Number.prototype.toString`; none differed.

`git diff --check` passed. The Moon test command itself could not run because the read-only sandbox rejected build-output writes.

---

*Note: Codex could not run `moon test` inside its read-only sandbox; the test runs reported in the PR body (all four backends green) were executed separately.*

bobzhang:
## Codex re-review of the widened (2⁵³) fast path — ✅ approved

*Automated review by Codex CLI (`gpt-5.6-sol`, `xhigh` reasoning), read-only, prompted to attack the widening's edge cases: the 2⁵³ rounding-interval math, tier boundaries, per-backend conversion semantics, and every JS-visible test pin.*

### Round 1 — code sound, two comments wrong

### Blocking: correct the widening’s boundary documentation

- `builtin/double_ryu_nonjs.mbt:658` calls 2^53 the largest integral Double that round-trips through Int64, but `builtin/double_ryu_int_wbtest.mbt:101-102` correctly observes that 2^62 also does. The relevant property is that every integer through 2^53 is exactly representable as binary64.
- `builtin/double_ryu_int_wbtest.mbt:33` says values immediately outside the Int range fall back to the full algorithm. After this widening, those values use the Int64 tier.

The code itself is sound. For N = 2^53, the rounding interval is `[N − 0.5, N + 1]`; the nearest decimals with at most 15 significant digits are N − 2 and N + 8, both outside it, while exact N wins over the same-length N + 1 by closeness. Thus the inclusive bound is valid. `9007199254740993.0` is already N and formats as `9007199254740992`; `9007199254740994.0` is exact and correctly falls through. The JavaScript pins also match `Number.prototype.toString()`.

Both comments were corrected in `45c5ece6` (comment-only change).

### Round 2 — approved

Approved.

- `builtin/double_ryu_nonjs.mbt:658` now correctly identifies 2^53 as the inclusive boundary through which every integer is exactly representable, while acknowledging that particular larger values such as 2^62 can still round-trip through `Int64` but require the full shortest-decimal algorithm.
- `builtin/double_ryu_int_wbtest.mbt:33` correctly states that values just outside the `Int` range use the `Int64` tier, matching the branch at `builtin/double_ryu_nonjs.mbt:687`.
- Commit `45c5ece6` changes only these comment blocks; no implementation, assertions, or expected outputs changed.

No remaining defects.

---

*Codex could not run `moon test` in its read-only sandbox; the runs in the PR body (all four backends green, 7562/7592/7532/7477) were executed separately.*

PR #4189, opened 2026-09-02
fix(deque): stop nulling vacated slots so views cannot read invalid memory

bobzhang:
Split out of #4135 (the `Deque` half; `Array` is in its own PR) so that each container can be reviewed on its own. The two are independent and can land in either order; `%fixedarray.set_null` loses its last user in core once both are in.

`Deque::as_views` hands out `ArrayView`s that share the deque's buffer, and every shrinking operation cleared the slots it vacated with `%fixedarray.set_null`. A view created before such a mutation then read a null slot. `Deque::as_views` is plain `pub`, so this is undefined behaviour reachable from entirely safe code:

```moonbit
let dq = @deque.from_array([Box::{ v: 11, pad: "a" }, Box::{ v: 22, pad: "b" }, Box::{ v: 33, pad: "c" }])
let (front, _) = dq.as_views()
let _ = dq.pop_back()
println(front[2].v)   // reads a nulled slot: SIGSEGV on native
```

## Approach

Vacated slots are no longer cleared, so a view can only ever observe valid values of `A`. Mutating a deque while a view of it is alive stays a program error, but what the view yields is now merely unspecified rather than invalid.

The cost is that a removal retains what it removes. That applies uniformly -- `clear` empties a deque the same way `pop_back` shortens it, and neither writes to the slots it gives up -- so there is no operation-by-operation rule to learn and no existing signature changes. The elements are released when a later push reuses the slot, when the buffer grows, or when the deque is dropped, which means the clear-and-refill pattern self-heals: each push releases one old occupant, so repeated fill/drain cycles on one deque hold flat.

| | cost | capacity |
| --- | --- | --- |
| `release_unused(placeholder~)` — new | one pass over the unused capacity, no allocation | kept |

`release_unused` overwrites every slot not currently holding an element -- the complement of the occupied run, which may wrap around the end of the buffer -- which is exactly the region any removal leaves behind. `shrink_to_fit` was already releasing those elements by letting the old buffer go; that is now documented rather than incidental.

```
Deque::release_unused(Self[A], placeholder~ : A)
```

One new method, nothing else in the public surface moves. `Deque::clear` and `Deque::truncate` both collapse to length assignments, and `Deque::truncate(0)` no longer routes through `Deque::clear`. `Deque::drain` keeps every blit and loses only the nulling passes; the branch that existed solely to null an emptied back run is folded away. `Deque::extract_if` no longer needs the final write index, so the loop's `nobreak` result goes with it.

Writing into the unused region is safe without tracking a high-water mark: it is always either NULL or a live reference, never garbage. `UninitializedArray::make` (`%fixedarray.make_uninit`) NULL-fills for reference element types — it has to, since `moonbit_drop_object` walks a REF_ARRAY's full capacity and skips slots with `if (!obj) continue` — growing blits into a fresh NULL-filled buffer, and `push_front`/`push_back` already store into exactly such slots.

## Changes relative to #4135

Code is identical to the `Deque` half of #4135 (rebased onto current `main`). An independent review pass found the prose imprecise in two places, fixed here: after a shifting removal (`remove`, `extract_if`, `retain`, a middle `drain`) the vacated slots hold duplicate references to shifted survivors, not "the removed elements"; and the changelog gets an "Added" entry for the new method. Unlike `Array`, every deque growth path already copies only the live elements, so "released once the buffer grows" holds on every path here.

## Review notes

Consequences the review surfaced that are inherent to the retain-instead-of-null design and worth weighing explicitly:

- A long-lived deque of reference types keeps its high-water-mark contents alive after `clear()` until overwritten; `queue` is the in-core user of this pattern. Any FFI finalizers on those elements run later than before.
- A popped or removed value is no longer uniquely referenced until its slot is reused, so unique-owner fast paths in the runtime are not taken for it.
- `release_unused` needs a placeholder value of `A`; element types without a cheap one can only release via `shrink_to_fit`.
- Pre-existing and untouched: a tail drain of an unwrapped deque (`drain(start=n-1)`) still shifts the `n-1` survivors right by one before adjusting `head`, which is wasted work the back-only branch of the same function avoids.

## Testing

- `moon test` on `--target native` (7486), `wasm-gc` (7601), `js` (7541) and `wasm` (7571), all passing
- 8 new regression tests covering: views after `clear`/`truncate`/`pop_back`/`drain`; `release_unused` after each of them, including the wrapped case both for a rotated deque and for one built with `push_front`; and that neither `truncate(0)` nor `clear` replaces the buffer, pinned by pushing after emptying and observing through a view taken beforehand
- `moon info`, `moon fmt` — `.mbti` churn is the one added method and nothing else


bobzhang:
Renamed the new method: `Deque::fill_unused(value)` is now `Deque::release_unused(placeholder~ : A)`, matching the `Array` rename in #4190. Same body, same docs; the PR description has been updated.

PR #4190, opened 2026-09-02
fix(builtin): stop nulling vacated Array slots so views cannot read invalid memory

bobzhang:
Split out of #4135 (the `Array` half; `Deque` is in its own PR) so that each container can be reviewed on its own. The two are independent and can land in either order; `%fixedarray.set_null` loses its last user in core once both are in.

Taking a view from an `Array` shares the underlying buffer, and every shrinking operation cleared the slots it vacated with `%fixedarray.set_null`. A view created before such a mutation then read a null slot. `Array::view` is plain `pub`, so this is undefined behaviour reachable from entirely safe code:

```moonbit
let arr = [Box::{ v: 11, pad: "a" }, Box::{ v: 22, pad: "b" }, Box::{ v: 33, pad: "c" }]
let view = arr[0:3]
let _ = arr.pop()
println(view[2].v)   // SIGSEGV on native (exit 139) with the current core
```

## Approach

Vacated slots are no longer cleared, so a view can only ever observe valid values of `T`. Mutating an array while a view of it is alive stays a program error, but what the view yields is now merely unspecified rather than invalid.

The cost is that a removal retains what it removes. That applies uniformly -- `clear` empties an array the same way `pop` shortens it, and neither writes to the slots it gives up -- so there is no operation-by-operation rule to learn and no existing signature changes. The elements are released when a later push reuses the slot, when the buffer grows, or when the array is dropped, which means the clear-and-refill pattern self-heals: each push releases one old occupant, so repeated fill/drain cycles on one array hold flat.

| | cost | capacity |
| --- | --- | --- |
| `release_unused(placeholder~)` — new | one pass over the unused capacity, no allocation | kept |

`release_unused` overwrites every slot from `length()` to `capacity()`, which is exactly the region any removal leaves behind. `shrink_to_fit` was already releasing those elements by letting the old buffer go; that is now documented rather than incidental.

```
Array::release_unused(Self[T], placeholder~ : T)
```

One new method, nothing else in the public surface moves. `Array::resize` shrinks like `truncate` and no longer releases on that branch. On the JavaScript backend `release_unused` is a documented no-op: shrinking a JS array already releases the removed elements, and a view reaching past the current length observes `undefined` there as before, which the `ArrayView` documentation now states.

Writing into `[length(), capacity())` is safe without tracking a high-water mark: that region is always either NULL or a live reference, never garbage. `%fixedarray.make_uninit` NULL-fills for reference element types — it has to, since `moonbit_drop_object` walks a REF_ARRAY's full capacity and skips slots with `if (!obj) continue` — and `moonbit_make_ref_array_with_blit` NULL-fills everything outside the blitted range when a buffer grows. It is the same region `Array::push` writes into on every push past the previous high-water mark.

## Changes relative to #4135

An independent review pass over the split found two places where the code and the prose disagreed, both fixed here:

- `Array::resize_buffer` now copies only the live prefix into the new buffer. It used to copy the whole old capacity, which was a memcpy of NULLs before and would have carried the retained elements into the new buffer on the `reserve_capacity` path, contradicting the "released once the buffer grows" wording; every other growth path already copied only `length()` elements.
- `Array::release_unused` uses the `%fixedarray.fill` intrinsic with an `#owned` placeholder, exactly as `Array::fill` and `Array::resize` do for this same region, instead of a hand-written `unsafe_set` loop.

Docs were tightened where they were imprecise: after a shifting removal (`remove`, `drain`, `retain`, ...) the vacated tail holds duplicate references to shifted survivors, not "the removed elements"; the shared `clear`/`truncate` docs now say the JS backend releases immediately; the changelog gets an "Added" entry for the new method. The buffer-reuse test gained a `remove` that actually shifts and a `drain` leg, which its comment already claimed.

## Review notes

Consequences the review surfaced that are inherent to the retain-instead-of-null design and worth weighing explicitly:

- Long-lived reusable buffers of reference types keep their high-water-mark contents alive after `clear()` until overwritten. The in-core instance is `lexbuf/storage.mbt`, where `discard_before` calls `chunks.clear()` on an `Array[String]` of consumed input chunks, so a streaming lexer now keeps up to `capacity()` consumed chunks alive until later pushes overwrite them. Any FFI finalizers on those elements run later than before.
- A popped or removed value is no longer uniquely referenced until its slot is reused, so the runtime's unique-source fast paths (`moonbit_make_ref_array_with_blit`'s memcpy branch, in-place reuse) are not taken for it.
- `release_unused` needs a placeholder value of `T`; element types without a cheap one (closures, handles) can only release via `shrink_to_fit`.

## Testing

- `moon test` on `--target native` (7485), `wasm-gc` (7600), `js` (7533) and `wasm` (7570), all passing
- 7 new regression tests (`not js` targets) covering: views after `clear`/`truncate`/`pop`/`remove`/`drain`; `release_unused` after each of them and a run of removals settled by one call; and that no shrinking operation replaces the buffer, pinned by pushing after emptying and observing through a view taken beforehand
- `moon info`, `moon fmt` — `.mbti` churn is the one added method and nothing else
- The reproducer above was run against the installed core on native and exits with SIGSEGV

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,9 +13,11 @@ changelog should follow [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 - Core commit: `bd827dc85`
 - Added `Debug` trait with `derive(Debug)` support, including `ignore=[..]` configuration for non-debuggable nested types
 - Added new `moonbitlang/async` APIs including `@process.spawn`, advisory file locking, `@fs.tmpdir`, `@async.all`, and `@async.any`
+- Added `Array::release_unused(placeholder~)`, which overwrites the unused capacity of an array in place with a placeholder, releasing the elements that earlier removals left there
 
 #### Changed
 
+- `Array` shrinking operations no longer null out the slots they vacate, so an `ArrayView` created before the mutation can no longer read uninitialized memory (previously a segfault on native and `null` on wasm-gc; the JavaScript backend is unchanged, and such a view still observes `undefined` past the array's current length there). Mutating an array while a view of it is alive remains a program error, but it now yields unspecified *valid* values rather than undefined behavior. The removed elements stay reachable from the buffer until later pushes reuse those slots, the buffer grows, or the array is dropped -- uniformly, `clear` included, which no longer releases them. `Array::release_unused(placeholder~)` overwrites the unused capacity in place with a placeholder to release them on demand, while `shrink_to_fit`, which already reallocated, lets them go with the old buffer. No existing signature changes
 - **BREAKING**: `@json.parse` now rejects unpaired `\uXXXX` surrogate escapes in strings, raising `ParseError::InvalidChar` at the backslash that opens the offending escape; an escaped leading surrogate must be followed immediately by an escaped trailing surrogate, and the pair decodes to the character it denotes. Previously such escapes were decoded unchecked and produced a `String` that was not well-formed Unicode. `@json.valid` reports the same documents as invalid. JSON emitted by `JSON.stringify` in JavaScript can contain these escapes, so input that JavaScript and Python accept may now be rejected — as it is by Rust's serde_json
 - `@json.inspect` has been migrated to `json_inspect`
 - `String::sub` and `StringView::sub` now panic on invalid indices instead of raising `CreatingViewError`. The `CreatingViewError` type has been removed.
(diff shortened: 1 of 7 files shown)


bobzhang:
Renamed the new method: `Array::fill_unused(value)` is now `Array::release_unused(placeholder~ : T)`. Same body, same docs; the name says what it is for and the labelled argument says what the value is. The PR description has been updated to match. #4189 gets the same rename for `Deque`.

PR #4191, opened 2026-09-03
fix async lexbuf test, adapt new cancellation semantic

Guest0x0:
We are adding first class support for cancellation in the compiler. A new `nocancel` effect is added to represent `async` function that are not cancellable. As a consequence, signature of certain async primitive should now be marked with `nocancel`. In particular, `%async.suspend` with `noraise` type should add `+ nocancel` to its signature, otherwise the code will trigger ICE. Test for async lexbuf in `moonbitlang/core` has binding for `%async.suspend` with `noraise` signature, so should be fixed.

Notice that raw async primitives are *not* stable, public API intended for end users. In general only `moonbitlang/async` should invoke them, and there is no guarantee at all for users binding `%async.suspend`/`%async.run` directly.

### DIFF
--- a/lexbuf/lexbuf_test.mbt
+++ b/lexbuf/lexbuf_test.mbt
@@ -132,10 +132,12 @@ test "Lexbuf returns an empty retained view" {
 }
 
 ///|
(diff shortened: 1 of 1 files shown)


bobzhang:
cc @hackwaly

PR #4192, opened 2026-09-03
refactor(deque): skip the survivor shift on a tail drain, tighten two comments

bobzhang:
Follow-up to #4189 (merged as d9930b60), carrying the items from its review that were not worth holding that PR for.

## One behaviour change: no survivor shift on a tail drain

`Deque::drain` handled a drain that ends at the end of the front run with a single `else` arm: shift the front survivors right by the drained length and advance `head`. That is required when something survives *after* the drained range (the rest of the front, or a non-empty back run) so the survivors stay contiguous with it. When the deque is unwrapped and the drain takes its tail, nothing survives after the range, the survivors already sit at `[head, head + start)`, and the shift was O(`start`) work that moved every survivor one slot for no reason. The arm is now guarded on `len < front_max_drain || back.length() != 0`; the tail case falls through with `head` untouched.

The new test `deque tail drain leaves the survivors in place` pins this through a view taken before the drain: the survivors stay at their original offsets and the drained slot keeps its occupant, whereas the old shift made the view read the first survivor twice. Reverting the guard fails exactly that test.

## Comment fixes

`clear` and `truncate` called "an emptied deque restarts at offset zero" an *invariant* that `clear`, `truncate` and `drain` maintain. It is not one: `pop_front` and `extract_if` can empty a deque without resetting `head`, and nothing depends on it. The comments now say those three functions arrange it.

## Extra test

`deque release_unused covers both sides of an unwrapped run` exercises a deque whose unused capacity sits before *and* after the occupied run, so each of the two `fill_slots` calls in the unwrapped branch is pinned by a test on its own.

## Testing

- `moon test -p moonbitlang/core/deque` on native (248), wasm-gc, js and wasm (258 each), all passing
- `moon fmt`, `moon info`: no drift, no interface change

### DIFF
--- a/deque/deque.mbt
+++ b/deque/deque.mbt
@@ -1052,8 +1052,8 @@ pub fn[A] Deque::rev_eachi(
 /// ```
 pub fn[A] Deque::clear(self : Deque[A]) -> Unit {
   // The surviving elements are none, so nothing moves: an emptied deque
-  // restarts at offset zero, the invariant `Deque::truncate` and
-  // `Deque::drain` also maintain.
+  // restarts at offset zero, as `Deque::truncate` and `Deque::drain` also
+  // arrange.
   self.len = 0
   self.head = 0
 }
@@ -1410,8 +1410,8 @@ pub fn[A] Deque::release_unused(self : Deque[A], placeholder~ : A) -> Unit {
 pub fn[A] Deque::truncate(self : Deque[A], len : Int) -> Unit {
   guard len >= 0 && len < self.len else { return }
   // The surviving elements keep their slots, so there is nothing to move: only
-  // the length changes. An emptied deque restarts at offset zero, matching the
-  // invariant `Deque::clear` and `Deque::drain` maintain.
+  // the length changes. An emptied deque restarts at offset zero, as
+  // `Deque::clear` and `Deque::drain` also arrange.
   self.len = len
   if len == 0 {
     self.head = 0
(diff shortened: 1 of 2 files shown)


bobzhang (commented):
Reviewed the diff in `deque/deque.mbt` and `deque/deque_test.mbt`.

## `Deque::drain` logic — correct

In the "draining front only" arm the old `else` always shifted the `start` front
survivors right by `len` and advanced `head`. That shift is only needed when
something survives *after* the drained range: the rest of the front
(`len < front_max_drain`) or a non-empty back run (`back.length() != 0`). The new
guard `len < front_max_drain || back.length() != 0` covers exactly those cases, and
the fall-through (tail drain of an unwrapped deque) correctly leaves `head` and the
survivors at `[head, head + start)` untouched. I walked every combination of
`start`/`len`/wrapped-or-not and the arm dispatches correctly in each.

## Comment fixes — accurate

`pop_front` advances `head` without resetting it when the deque empties, so "an
emptied deque restarts at offset zero" was never an invariant; nothing relies on
`head == 0` when `len == 0`. The "arrange" wording is precise.

## Tests — meaningful, pin real behaviour

- `deque release_unused covers both sides of an unwrapped run` exercises both
  `fill_slots` calls (unused capacity before and after the occupied run).
- `deque tail drain leaves the survivors in place` observes a pre-drain
  `as_views()` view; reverting the guard makes that view read the first survivor
  twice, so it genuinely fails on the old behaviour.

## Verification

- `moon test -p moonbitlang/core/deque` → 258/258 pass (native)
- `moon check --target all deque` clean
- `moon fmt --check deque` clean

No issues found.

PR #4199, opened 2026-09-03
refactor(hashmap, set): remove the deprecated Show impls

bobzhang:
Removes the deprecated `Show` impls for `HashMap` and linked-hash `Set`
(both were `#deprecated` in favor of `@debug.Debug`), together with their
hidden `Show::{output, to_string}` extends and the hashmap test that only
exercised the deprecated `Show` rendering. Nothing in the tree renders
these collections through `Show` anymore, so the impls are simply gone
(-85 lines).

Rewritten from the earlier `<+>`-refactor version of this PR after review
feedback: polishing deprecated impls is wasted effort, removing them is
cleaner.

Verified: `moon check` 0 warnings/0 errors; hashmap 139/139, set 70/70.

Generated with SeekMoon

### DIFF
--- a/hashmap/utils.mbt
+++ b/hashmap/utils.mbt
@@ -324,34 +324,6 @@ pub fn[K, V] HashMap::eachi(
 }
 
 ///|
-/// Provides string representation for hash maps.
-/// 
-/// Notice that the order of key-value pairs in the output string is not guaranteed
-///
-/// Parameters:
-///
-/// * `self` : The hash map to be converted to string.
-/// * `logger` : The buffer to write the string representation to.
-#deprecated("Use @debug.Debug instead of Show for debugging purposes. See https://github.com/moonbitlang/core/blob/main/debug/README.mbt.md")
-pub impl[K : Show, V : Show] Show for HashMap[K, V]
-
-///|
-pub impl[K : Show, V : Show] Show for HashMap[K, V] with fn output(self, logger) {
-  logger.write_string("HashMap::from_array([")
-  self.eachi((i, k, v) => {
-    if i > 0 {
-      logger.write_string(", ")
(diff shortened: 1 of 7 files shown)


hashmap/utils.mbt
```diff
@@ -337,18 +337,14 @@ pub impl[K : Show, V : Show] Show for HashMap[K, V]
 
 ///|
 pub impl[K : Show, V : Show] Show for HashMap[K, V] with fn output(self, logger) {
```
bobzhang: should `impl Show for HashMap` be deprecated?
if it is maybe remove it now

set/linked_hash_set.mbt
```diff
@@ -468,13 +468,13 @@ pub impl[K : Show] Show for Set[K]
 
 ///|
 pub impl[K : Show] Show for Set[K] with fn output(self, logger) {
```
bobzhang: same

hashmap/utils.mbt
bobzhang: You're right - this `Show` impl is `#deprecated` (Debug migration). Per your suggestion I've turned the PR into a removal instead: `Show for HashMap` is deleted together with the hidden `Show::{output, to_string}` extends and the test that only exercised the deprecated Show. `moon check` is clean; hashmap 139/139 and set 70/70 pass.

set/linked_hash_set.mbt
bobzhang: Same here - `Show for Set` was deprecated in favor of `@debug.Debug`; the impl, its hidden `Show` extends, and any test only exercising the deprecated Show are now removed. hashmap 139/139 and set 70/70 pass.

PR #4195, opened 2026-09-03
refactor: use `<+` template writing in place of write_string sequences

bobzhang:
Replaces literal and sequence `StringBuilder`/`&Logger` writes with the `<+`
template writer across `Show` implementations and streaming helpers, part of
the ongoing project-wide simplification from `write_string` sequences to
`<+` template writing.

Changes:

- **builtin**: `BytesView` `Show` and `ToJson` now share one
  `BytesView::escape_to` streaming helper; `Char`/`StringView` writes of
  non-printable `\u{...}` escapes collapse into a single template; `Iter`
  and `Map` `Show` stream through `<+`; a base64 `==` tail and a
  `FixedArray` iterator-test builder simplify.
- **diff**: `Range`/`HunkHeader` `Show` stream via inline-writer holes.
- **hashmap**, **set**: `HashMap`/`Set` `Show` collapse punctuation writes
  into `<+` templates.
- **json**: `ParseError` and the deprecated `Json` `Show` impls collapse
  write sequences into templates.
- **json_path**: JSON Pointer rendering streams the recursive path through
  one template per step instead of interleaved `write_*` calls.
- **v128**: zero-padded hex formatting streams through a new
  `u64_hex_to` helper used by the `Show` template.

`\{l => ...}` inline-writer holes preserve monomorphic `write_object`
dispatch, so output is unchanged throughout (full test suite: 7499/7499).

This re-applies the still-relevant content of the closed
#4110 (simplify-template-writing) on top of current `main`; the JSON
stringify/`escape_to` part of that PR was superseded by the merged
SIMD `need_escape`/`write_escaped` work, so it is intentionally not
re-applied.

Generated with SeekMoon

PR #4200, opened 2026-09-03
refactor(diff, v128): stream Show output through `<+>` and writer helpers

bobzhang:
Part of the `<+>` template-writing simplification, split out of the closed
#4195 for reviewability.

- `Range`/`HunkHeader` `Show` write through `<+>` templates; a lone range
  is written via `write_object`
- `V128` hex formatting streams through a new `u64_hex_to` writer helper
  (the string `u64_hex` wrapper is kept for the Debug impl), and `Show`
  renders `V128(...)` through a `<+>` template with `\{cb => ...}` holes

Output is unchanged.

Verified: `moon check` 0 warnings/0 errors; diff 127/127, v128 50/50.

Generated with SeekMoon

### DIFF
--- a/diff/hunk.mbt
+++ b/diff/hunk.mbt
@@ -26,8 +26,7 @@ impl Show for Range with fn output(self, logger) {
       // empty ranges begin at line just before the range
       beginning -= 1
     }
-    logger <+
-      "\{cb => cb.write_object(beginning)},\{cb => cb.write_object(len)}"
+    logger <+ "\{beginning},\{len}"
   }
 }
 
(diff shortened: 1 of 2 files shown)


diff/hunk.mbt
```diff
@@ -20,15 +20,14 @@ impl Show for Range with fn output(self, logger) {
...
   } else {
     if len == 0 {
       // empty ranges begin at line just before the range
       beginning -= 1
     }
-    logger.write_string(beginning.to_string())
-    logger.write_char(',')
-    logger.write_string(len.to_string())
+    logger <+
+      "\{cb => cb.write_object(beginning)},\{cb => cb.write_object(len)}"
```
bobzhang: logger <+ "\{beginning},\{len}"

bobzhang: Done - the else branch now writes plain interpolation `logger <+ "\{beginning},\{len}"` since both members are Ints; no writer holes needed. Pushed as 64dfbbad (diff tests 127/127).

PR #4212, opened 2026-09-08
fix(json): encode non-finite numbers as strings in stringify

bobzhang:
## Problem

`Json::stringify` wrote a `Number` holding NaN or an infinity as a bare token:

```moonbit
Json::number(@double.not_a_number).stringify()  // NaN
Json::number(@double.infinity).stringify()      // Infinity
Json::number(@double.neg_infinity).stringify()  // -Infinity
```

None of those are JSON — no RFC 8259 parser reads them back, `@json.parse` included.

## Fix

Encode the three non-finite values the way `Double::to_json` already does — as the strings `"NaN"`, `"Infinity"` and `"-Infinity"` — which is exactly the set `Double::from_json` accepts, so the value survives a round trip:

```moonbit
let inf = Json::number(@double.infinity)
inspect(inf.stringify(), content=(#|"Infinity"))
let back : Double = @json.from_json(@json.parse(inf.stringify()))
inspect(back == @double.infinity, content="true")
```

Only a directly constructed `Number` reaches this path. The parser records the source text in `repr` for every literal that overflows to an infinity (`1e400` → `Number(Infinity, repr=Some("1e400"))`), and that branch writes `repr` verbatim without consulting the double — so parsed documents are unaffected, and `repr` still takes precedence where it is present.

The guard is a tiny `#inline` entry with the non-finite tail outlined, so inlining it into the `stringify` loop stays cheap for ordinary numbers.

## Docs

The fix leans on the `repr` contract, which was undocumented (a bare `// 1.0000000000000000000e100` comment on the constructor). Now stated on both `pub enum Json` and `Json::number`:

- when present, `repr` must be valid JSON number syntax denoting the accompanying `Double` — `stringify` writes it in place of the double, so text that is not a JSON number makes the output invalid JSON or gives it a different JSON type, and text denoting another value silently changes what the document says;
- which literals the parser attaches it to: every integer literal past the exact-integer range (whether or not the value lands on one — `9007199254740992` keeps its text too) plus everything overflowing to an infinity, and nothing else (`0.1` and `1e-400` get `None`);
- equality compares only the numeric value and ignores `repr`, so two equal `Json` values can stringify differently.

## Verification

- Mutation-checked: restoring `None => buf.write_object(n)` fails exactly the two tests added here — the `Json::number` doc test and `stringify non-finite number` (3227 passed / 2 failed).
- `moon test`: 7529 passed, 0 failed.
- `moon test --target all -p moonbitlang/core/json -p moonbitlang/core/builtin`: wasm 3229, wasm-gc 3229, js 3199, native 3188 — all passed.
- `moon fmt`, `moon check`, `moon info` clean; no `.mbti` change (both new helpers are private).

Reviewed and signed off by Codex CLI (three rounds; it blocked twice on doc claims that overstated when `repr` is attached and what a bad `repr` produces).

https://claude.ai/code/session_01VubGDsJHzgC6ykiq7t4hrY

### DIFF
--- a/json/json.mbt
+++ b/json/json.mbt
@@ -113,6 +113,38 @@ fn write_indent(
 }
 
 ///|
+#inline
+fn write_number(buf : StringBuilder, number : Double) -> Unit {
+  // The guard is kept tiny and the non-finite tail outlined so that inlining
+  // this into the `stringify` loop stays cheap on every ordinary number.
+  if number.is_nan() || number.is_inf() {
+    write_non_finite_number(buf, number)
+  } else {
+    buf.write_object(number)
+  }
+}
+
+///|
+fn write_non_finite_number(buf : StringBuilder, number : Double) -> Unit {
+  // JSON has no syntax for NaN or the infinities, so a `Number` holding one
+  // cannot be written as a bare literal. Encode them the way `Double::to_json`
+  // does — as the strings "NaN", "Infinity" and "-Infinity" — since
+  // `Double::from_json` accepts exactly those three, so the value still
+  // survives a round trip.
+  //
+  // Only a directly constructed `Number` reaches this: the parser stores the
+  // source text in `repr` for every literal that overflows to an infinity, and
+  // that branch writes `repr` verbatim without consulting the double.
+  if number.is_nan() {
+    buf.write_string("\"NaN\"")
+  } else if number.is_pos_inf() {
+    buf.write_string("\"Infinity\"")
+  } else {
+    buf.write_string("\"-Infinity\"")
+  }
+}
+
+///|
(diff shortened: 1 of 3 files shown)


json/json.mbt
```diff
@@ -112,6 +112,38 @@ fn write_indent(
...
+  // this into the `stringify` loop stays cheap on every ordinary number.
+  if number.is_nan() || number.is_inf() {
+    write_non_finite_number(buf, number)
+  } else {
+    buf.write_object(number)
+  }
+}
+
+///|
+fn write_non_finite_number(buf : StringBuilder, number : Double) -> Unit {
```
bobzhang, line 128: inine this function?

PR #4219, opened 2026-09-10
math: return 1 for pow(1, NaN) and pow(±1, ±inf) per IEEE 754

### LINKED ISSUE #4059: math: pow special cases pow(1, NaN) and pow(±1, ±inf) return NaN, diverging from IEEE 754/C99 pow
## Behavior

On every backend (wasm-gc, wasm, js, native), the following all return `NaN`:

```moonbit
@math.pow(1.0, @double.not_a_number)   // NaN, IEEE 754 / C99 Annex F say 1.0
@math.pow(1.0, @double.infinity)       // NaN, IEEE 754 / C99 Annex F say 1.0
@math.pow(1.0, @double.neg_infinity)   // NaN, IEEE 754 / C99 Annex F say 1.0
@math.pow(-1.0, @double.infinity)      // NaN, IEEE 754 / C99 Annex F say 1.0
@math.pow(-1.0, @double.neg_infinity)  // NaN, IEEE 754 / C99 Annex F say 1.0
```

## Discussion

- IEEE 754-2019 (§9.2.1) and C99 Annex F (F.9.4.4) specify `pow(1, y) = 1` for **any** y, including NaN, and `pow(±1, ±∞) = 1` — the rationale being "1 to any power is 1".
- The current behavior follows the classic fdlibm ordering (NaN check first) and matches ECMAScript's `Math.pow` / `**`, so it is at least internally consistent and consistent across backends.
- However, the function's doc comment claims IEEE 754 semantics, so today either the docs or the behavior is wrong.

## Possible resolutions

1. Align behavior with IEEE 754/C99: special-case `x == 1.0` (and `|x| == 1.0` with infinite y) before the NaN check. This is what glibc/musl/Rust/Go do.
2. Keep JS-`Math.pow` behavior and fix the doc comment to say so explicitly.

Either way the divergence should stop being implicit. Found while adding QuickCheck special-value coverage in #4055 (the property suite currently pins the *actual* behavior, and can be flipped to the IEEE behavior if option 1 is chosen).

superbigcup325:
Fixes #4059.

## Problem and resulting behavior

`@math.pow(1.0, NaN)`, `@math.pow(1.0, ±inf)`, and `@math.pow(-1.0, ±inf)` previously returned `NaN` on every backend. They now return `1.0`, matching IEEE 754-2019 §9.2.1, C99 Annex F.9.4.4, and the existing `powf` behavior.

The inconsistency came from different implementations. `powf` is a MoonBit port of FreeBSD's `e_powf.c`, shared by all backends, including JS; its unit-base special cases have returned `1` since the implementation was introduced. Double `pow` instead called `Math.pow` on JS and used an older fdlibm implementation on other backends, both producing `NaN` for these inputs. The [ECMAScript specification](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-numeric-types-number-exponentiate) explicitly retains these differing results for historical compatibility.

Aligning the Double implementation makes Float and Double agree on these special cases and fulfills `pow`'s documented IEEE 754 semantics. This changes the five special-value results above; `pow(-1, NaN)` and negative finite bases with non-integer finite exponents still return `NaN`. Public function signatures are unchanged.

## Changes

- Handle unit-base special cases before the non-JS NaN check, while preserving the existing zero-exponent handling.
- Wrap the JS `Math.pow` call with the same special-case guards.
- Remove the now-unreachable non-JS branch that returned `y - y` for unit bases with infinite exponents, along with its stale NaN comment.
- Update the previous expectation and add coverage for all five changed cases, NaN propagation for base `-1`, and ordinary finite exponents.

## Validation

After the cleanup and synchronization with `main` in `0844b77d1`:

- `moon test -p moonbitlang/core/math --target <backend>`: 196/196 passed on each of wasm-gc, wasm, js, and native.
- `moon check --target all`: passed.
- `moon info` followed by `moon fmt`: passed; no `.mbti` changes.
- `git diff --check`: passed.

### DIFF
--- a/math/pow_double_nonjs.mbt
+++ b/math/pow_double_nonjs.mbt
@@ -205,6 +205,13 @@ pub fn pow(x : Double, y : Double) -> Double {
   if (iy.reinterpret_as_uint() | ly) == 0 {
     return ONE
   }
+  // |x|==1: pow(1,y) = 1 even if y is NaN, pow(-1,+-inf) = 1
+  // (IEEE 754-2019 section 9.2.1, same as powf)
+  if ix == 0x3ff00000 && lx == 0 {
+    if hx > 0 || (iy == 0x7ff00000 && ly == 0) {
+      return ONE
+    }
+  }
   // +-NaN return x+y
   if ix > 0x7FF00000 ||
     (ix == 0x7FF00000 && lx != 0) ||
--- a/math/pow_double_nonjs.mbt
+++ b/math/pow_double_nonjs.mbt
@@ -244,9 +244,7 @@ pub fn pow(x : Double, y : Double) -> Double {
   // special value of y
   if ly == 0 {
     if iy == 0x7ff00000 { // y is +-inf
-      if ((ix.reinterpret_as_uint() - 0x3ff00000) | lx) == 0 {
-        return y - y // inf**+-1 is NaN
-      } else if ix >= 0x3ff00000 { // (|x|>1)**+-inf = inf,0
+      if ix >= 0x3ff00000 { // (|x|>1)**+-inf = inf,0
         return if hy >= 0 { y } else { ZERO }
       } else { // (|x|<1)**-,+inf = inf,0
         return if hy < 0 { -y } else { ZERO }
--- a/math/pow_double_js.mbt
+++ b/math/pow_double_js.mbt
@@ -37,4 +37,18 @@
 ///   inspect(@math.pow(@double.infinity, -1.0), content="0")
 /// }
 /// ```
-pub fn pow(base : Double, exponent : Double) -> Double = "Math" "pow"
+pub fn pow(base : Double, exponent : Double) -> Double {
+  // ECMAScript's Math.pow returns NaN for pow(1, NaN) and pow(+-1, +-inf);
+  // align with IEEE 754-2019 section 9.2.1 as the other backends do.
+  if base == 1.0 {
+    return 1.0
+  }
+  if base == -1.0 &&
+    (exponent == @double.infinity || exponent == @double.neg_infinity) {
+    return 1.0
+  }
+  math_pow_js(base, exponent)
+}
+
+///|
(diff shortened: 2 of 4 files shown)


math/pow_double_nonjs.mbt
```diff
@@ -205,6 +205,13 @@ pub fn pow(x : Double, y : Double) -> Double {
   if (iy.reinterpret_as_uint() | ly) == 0 {
     return ONE
   }
+  // |x|==1: pow(1,y) = 1 even if y is NaN, pow(-1,+-inf) = 1
+  // (IEEE 754-2019 section 9.2.1, same as powf)
+  if ix == 0x3ff00000 && lx == 0 {
```
bobzhang, line 210: Non-blocking cleanup: this early return makes the existing `|x| == 1` branch inside the infinite-exponent handling below (currently lines 247–248, `return y - y`) unreachable. Please consider removing that branch and its stale NaN comment so the remaining implementation does not describe the behavior this PR is replacing.

bobzhang (commented):
Reviewed 1b96df0. This is a worthwhile correctness fix, and I found no blocking issues.

The strongest reason for the change is consistency with the existing `powf` implementation: switching between Float and Double should not change these special-value results. It also brings `pow` into line with its documented IEEE 754 semantics. The [ECMAScript specification](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-numeric-types-number-exponentiate) explicitly describes its differing NaN results as a historical compatibility choice.

The new guards are narrowly scoped, and the tests cover all five changed cases as well as `pow(-1, NaN)` and finite negative-base cases that must retain their behavior. I would be comfortable merging this, with one optional dead-code cleanup noted inline.

Local validation on this commit:
- Math package tests: 196/196 passed on each of wasm-gc, wasm, js, and native.
- `moon check --target all`: passed.

PR #4226, opened 2026-09-12
fix(math): enforce probable-prime bit length

bobzhang:
`probable_prime(0, rand)` and `probable_prime(1, rand)` cannot terminate, while valid requests can return a prime shorter than the requested bit length.

Reject bit lengths below two before generation and set each candidate's highest requested bit. Add invalid-input and exact-bit-length regressions, and clarify the Miller–Rabin false-positive probability in the documentation.

Validation:

- `moon check --deny-warn`
- `moon test math --target all`


bobzhang (commented):
Reviewed commit 38d5c3d869799dd4499450e5d3a320427d82f5cb. No blocking findings.

The input guard prevents nontermination for zero- and one-bit requests and rejects negative bit lengths before generation. Setting the highest requested bit correctly turns the random generator's at-most-bits guarantee into an exact-bit-length guarantee, without exceeding the requested range or excluding the prime 2 for two-bit requests. Uniform candidate sampling remains uniform after setting the high bit.

The documentation correctly distinguishes the probability of accepting a fixed composite under independent uniform Miller–Rabin bases from the posterior probability that a returned value is prime.

Independent validation in a detached checkout:
- `moon check --deny-warn` passed.
- `moon test math --target all`: 200/200 tests passed on each of wasm, wasm-gc, js, and native.
- `moon info` produced no interface changes; `moon fmt --check` passed.
- Running the new exact-bit-length regression against the original implementation failed with an actual bit length of 2 for a requested length of 3, confirming that it detects the original bug.

Non-blocking suggestion: extend the exact-bit-length cases with 63, 64, 65, and 128 to cover larger BigInt storage boundaries. The current implementation is suitable to merge as-is.

PR #4227, opened 2026-09-12
fix(quickcheck): exclude unchanged Float shrink candidates

bobzhang:
A smaller Double shrink candidate can round back to the original Float during conversion. For values such as `3.14F`, this returns an unchanged failing input and prevents shrinking from progressing.

Filter out candidates equal to the original value after narrowing to Float. Add a regression covering values that previously reproduced this behavior.

Validation:

- `moon check --deny-warn`
- `moon test quickcheck/shrink --target all`


bobzhang:
cc @CAIMEOX

PR #4229, opened 2026-09-12
fix(argparse): omit the Node script path from default argv

bobzhang:
On Node, `process.argv` contains both the runtime executable and the script path. Default argument parsing removed only the first item, causing the script path to be parsed as a positional argument and potentially rejecting otherwise valid commands.

Remove both prefix items on JavaScript and add a test that supplies a synthetic Node command line and restores the process arguments afterward. Document explicit argv and environment-map behavior.

Validation:

- `moon check --deny-warn`
- `moon test argparse --target all`


bobzhang:
Cross-backend argument conventions, for reference:

I ran `fn main { println(@env.args()) }` with `--verbose hello` using Moon 0.1.20260904. With build paths abbreviated, the results were:

```text
wasm / moonrun:    […/single.wasm, --verbose, hello]
wasm-gc / moonrun: […/single.wasm, --verbose, hello]
native:           […/single.exe, --verbose, hello]
js / Node:        [/usr/local/bin/node, …/single.js, --verbose, hello]
```

Thus this PR fixes a real mismatch: removing only one entry on Node leaves the script path to be parsed as a positional argument. Removing two entries for ordinary Node script execution makes the default parser input consistent with native and moonrun. Explicit `parse(argv=...)` already expects user arguments only.

Validation: applying the PR diff to a temporary copy of the local core checkout, `moon test argparse --target all` passed all 613 tests (153 each on wasm, wasm-gc, and native; 154 on JS). A separate executable probe on all four backends successfully parsed `--verbose` through both default arguments and explicit `argv`.

Other language conventions:

- **OCaml native and bytecode:** both expose `[program, user arguments…]` through `Sys.argv`. For example, `ocamlrun -b ./demo.byte a b` produces `["./demo.byte", "a", "b"]`; neither `ocamlrun` nor its `-b` option is included. The bytecode runtime strips the interpreter prefix before initializing `Sys.argv`. See the [OCaml runtime manual](https://ocaml.org/manual/5.5/runtime.html), [bytecode startup](https://github.com/ocaml/ocaml/blob/5.5/runtime/startup_byt.c), and [native startup](https://github.com/ocaml/ocaml/blob/5.5/runtime/startup_nat.c).
- **Python:** `python app.py a b` gives `sys.argv == ["app.py", "a", "b"]`. Interpreter identity and original interpreter arguments are available separately through `sys.executable` and `sys.orig_argv`. See the [Python sys documentation](https://docs.python.org/3/library/sys.html#sys.argv).
- **Java:** both `java Main a b` and `java -jar app.jar a b` pass only `["a", "b"]` to `main(String[] args)`, excluding the runtime and application name. See the [Java launcher documentation](https://docs.oracle.com/en/java/javase/24/docs/specs/man/java.html).
- **Node:** ordinary script execution exposes `[runtime, script, user arguments…]`. This is a documented API convention, not a technical requirement: Node also normalizes `argv[0]` to `process.execPath` and separates runtime options into `process.execArgv`. I did not find an authoritative statement of the original design motivation. See [process.argv](https://nodejs.org/api/process.html#processargv) and [process.execArgv](https://nodejs.org/api/process.html#processexecargv).

Design consideration: a future change could normalize `@env.args()` itself to `[program, user arguments…]` on every supported CLI backend, with JS dropping the Node executable but retaining the script path. This follows Python/OCaml and prevents every consumer from needing backend-specific slicing. It would change the existing JS API behavior, so compatibility and the corresponding removal of argparse's JS-specific extra skip would need to be handled together. The current PR is a narrower fix for argparse.

Scope: the Wasm results above are for moonrun; custom Wasm hosts supply their own arguments. Browser JS without `process.argv` returns an empty array. The two-entry Node prefix assumes ordinary script execution, not every possible embedding or launch mode.

PR #4230, opened 2026-09-12
fix(env): propagate WASI entropy failures

bobzhang:
The linear Wasm entropy adapter discarded the return code from WASI `random_get` and always reported success. A host error therefore exposed a buffer as successful random output.

Report success only for errno 0, allowing `env.rand` to return `None` on host failure, including availability probes. Clarify the runtime argument and entropy contracts, remove the unconditional-WASI-availability claim from the test name, and correct the timestamp-label guidance.

Validation:

- `moon check --deny-warn`
- `moon test env --target all`
- Compiled-Wasm host injection with errno 0/5 and lengths -1/0/32: all six cases pass.


peter-jerry-ye (approved):
The errno check correctly propagates WASI entropy failures as None for both byte requests and availability probes. Looks good as-is.

PR #4238, opened 2026-09-14
refactor(argparse): share argument value policy

bobzhang:
Argument parsing repeats value-source policy in global merging and environment assignment. Share the global precedence/copy tail (return immediately after both-argv append), reuse the identical conflict-label helper, and route env flags through assign_value while retaining the Help/Version no-op. Remove two unused private index parameters from option-value boundary checks. Append order, copies, sources, errors, hyphen-value policy and public interfaces stay unchanged.

Validation on the final three-file diff: `moon check --target all --deny-warn`, `moon test argparse --target all` (153 per target), full `moon test --target all` (wasm 7609, wasm-gc 7636, JS 7569, native 7524), `moon bundle --all`, `moon info`, and `moon fmt` pass. No interface or snapshot changes. Final source/diff review found no actionable findings.

GitHub check snapshot for 7d8e888f8: GitGuardian pending; Devin Review reports pass with the review skipped.


bobzhang (commented):
## Sign-off: approve

Automated review of the two-commit refactor (base `2a34e7a35`, head `7d8e888f8`; 3 files, +26/-89).

**Verdict: behavior-preserving refactor, no blocking findings.**

Why it is safe:
- The removed precedence/copy tail was textually identical to the shared one; `if ... { if both_argv { ...; return } } else if ... { raise }` + tail is equivalent to the previous control flow for every (append/set, both-argv) combination. Append order, copies, sources, and the `Set` conflict raise are unchanged.
- `assign_value`'s `SetTrue`/`SetFalse`/`Count` arms are the exact statements `apply_env` inlined, with the same `Env` source; the retained Help/Version no-op keeps `assign_value`'s `InvalidArgument` unreachable.
- `global_option_conflict_label` and `option_conflict_label` are character-for-character identical.

Verification performed:
- Differential probe on both trees (44 scenarios): env -> Set/Append/positional options, env -> SetTrue/SetFalse/Count (incl. `0`, `03`, falsy) with sources, invalid env values with full error text, argv-over-env, hyphen boundary policy (strict / `allow_hyphen_values` / bare `-`), and 13 global-merge scenarios including the cross-subcommand `Set` conflict label; output is identical base vs head.
- `moon check --deny-warn --target all` clean; `moon test argparse --target all` 153/153 on wasm, wasm-gc, js, native; `moon info` leaves no diff (no interface/snapshot change); `moon fmt --check` clean; tree byte-identical to `origin/pr-4238`.

Non-blocking notes:
1. `argparse/parser_values.mbt:207` (`FlagInfo(action=Help | Version, ..) => ()`) is unreachable: `parser_validate.mbt:289` rejects `env` on help/version args. Safe to keep (dropping it would let `assign_value` raise if validation is ever relaxed); a one-line comment would explain the arm, which is also the line coveralls reports as a coverage regression.
2. Coveralls' "1 uncovered change" in the merge file is a dedup artifact: the tail lines now exist once instead of twice.

Not verified here: the full-suite counts in the description (only `argparse` was rerun); the rest of the suite is untouched by the diff and CI is green on `7d8e888f8`.

PR #4242, opened 2026-09-14
refactor(builtin): reuse scalar count and whitespace operations

bobzhang:
Byte::popcnt duplicates bit-counting machinery already provided by the builtin UInt intrinsic. Delegate after zero extension and validate all 256 inputs with an independent arithmetic bit count. Char::is_whitespace now reuses the identical ASCII predicate while keeping its Unicode branch. Public interfaces remain unchanged.

Validation: `moon check --target all --deny-warn`, focused all-target tests, full `moon test --target all`, `moon bundle --all`, `moon info`, and `moon fmt` pass. Focused tests also pass in release mode on all four targets. No generated-interface or snapshot changes. Final source/diff review found no actionable findings.

GitHub check snapshot at publication: no checks reported yet.


bobzhang:
Performance review of `Byte::popcnt` at ea1e0c45ba7ed4602b7aea4ae09461d03a9b0516: **no performance blocker found; reasonable to merge based on these measurements.** This is a throughput/latency tradeoff, not a universal speedup.

Measured on Apple M3 Max (ARM64), release builds, moon 0.1.20260904 / moonc v0.10.12+1634b282e, Node v24.10.0, moonrun 0.1.20260904, Apple Clang 16.0.0. I toggled the actual `Byte::popcnt` body between the original SWAR and this PR, keeping callers identical. Two before/after rounds, sequential backend runs, bench calibration/warmup and ten timed batches per benchmark. Each invocation performs 65,536 operations; input construction is outside timing and the result is retained with `Bench::keep`.

Times below are the two runs' reported means, in microseconds (lower is better):

| Backend | Scan: original | Scan: PR | Dependent loop: original | Dependent loop: PR |
|---|---:|---:|---:|---:|
| JS | 94.48 / 95.97 | 94.28 / 97.25 | 216.67 / 217.19 | 285.55 / 313.51 |
| Native (C/Clang) | 54.51 / 57.67 | 25.65 / 26.04 | 198.17 / 201.40 | 319.01 / 303.13 |
| Wasm | 127.75 / 127.89 | 57.06 / 57.03 | 261.33 / 238.09 | 299.26 / 300.59 |
| Wasm-GC | 103.27 / 102.08 | 28.30 / 28.42 | 240.35 / 237.03 | 299.05 / 322.30 |

The scan sums `data[i].popcnt()` over 65,536 bytes, initialized with `(i * 73 + 19).to_byte()` so all 256 byte values occur equally often. It shows substantial throughput improvements on Native/Wasm/Wasm-GC and effectively unchanged performance on JS.

The dependent loop is an intentionally synthetic latency benchmark:

```moonbit
let mut x : Byte = 173
for i in 0..<65536 {
  x = (x.popcnt() + i).to_byte()
}
```

Each input depends on the previous result. All four backends are slower here, but this is not an identified application hotspot and does not establish a real-world regression. It is not sufficient by itself to block this refactoring or require retaining the old JS implementation.

Generated-code inspection:

- JS expands the intrinsic to a 32-bit software popcount, including `>>> 8` and `>>> 16` reduction stages that a byte-specific implementation does not need.
- Wasm and Wasm-GC emit `i32.popcnt`.
- Native maps to `__builtin_popcount`; on this machine the function and dependent loop use `fmov → cnt.8b → uaddlv.8b → fmov`, consistent with the observed latency tradeoff.

Validation: `moon test byte popcnt_review --target all --release` passed 13 tests on each of the four backends, including the PR's exhaustive arithmetic check over all 256 byte values and an old/new equivalence check. No x86 measurements were taken. The experimental standalone LLVM target could not be tested because this toolchain reports `LLVM backend is disabled`.

Scope: this is a focused popcnt performance review, not a claim that every workload or platform becomes faster.

PR #4247, opened 2026-09-18
refactor(json)!: consolidate Json methods in builtin

peter-jerry-ye:
`Json` is defined in `builtin`, but its stringification, transformation, and serialization implementation were spread across `json`. Move `Replacer`, `Json::stringify`, `Json::transform`, and `ToJson for Json` into `builtin` so they live alongside the type. Parsing and `FromJson` remain in `json`.

Re-export `Replacer` from `json` to preserve `@json.Replacer` construction. Keep the serialization algorithms intact, reuse builtin SIMD operations with the equivalent unsigned `<= 0x1F` control-character check, and implement `Debug` in `debug` with the same representation. Escaping tests retain independent model, scalar, and parser checks; a builtin white-box test also exercises `Json(value)` without depending on `json`.

Remove the deprecated `Json::as_*`, `item`, and `value` methods, `Show for Json`, its `to_string`/`output` promotions, and `Replacer.to_repr()`. Use pattern matching, `stringify`, or `debug_inspect` as appropriate.

The hidden, deprecated `ToJson` promotion moves to `builtin`; the equivalent `FromJson` promotion stays in `json`. The current compiler requires these explicit declarations to avoid reintroducing implicit public methods. Callers should use `Json(value)` and `@json.from_json`.


peter-jerry-ye:
This is for eliminating the special case in the core where the type and methods are defined in different packages. This is supposed to be the last one of them. @Guest0x0 would you confirm for the others please?

PR #4249, opened 2026-09-20
fix(json): qualify FromJson in blackbox tests

peter-jerry-ye:
`test_unqualified_package` causes warning-denying checks to fail on JSON's blackbox test derives. `FromJson` is normally available unqualified through the prelude, but JSON's blackbox tests implicitly import the package under test, whose `FromJson` shadows the prelude re-export.

Qualify the three derives as `@json.FromJson` to make that resolution explicit. This changes only JSON's test fixtures: other packages continue to use the prelude's unqualified `FromJson`, and runtime behavior and public APIs are unchanged.


myfreess (commented):
lgtm

